PR #35, opened 2017-03-21
Fix Echo example and improve instructions

acchou:
Added header search path to the Echo example to allow it to build. Also
added instructions to run the script in the third party directory to
clone the swift-protobuf repo locally, which is required by the Echo
example but not mentioned. Also update the toplevel readme to make the
required make command explicit, and fix the name `swift-proto` to
`swift-protobuf`


acchou:
Signed the CLA.

timburks:
Looks good, thanks!

Issue #38, opened 2017-03-27
Inconsistency in proto message naming convention, too many underscores


mickeyreiss:
Hi @timburks,

While trying to incorporate grpc-swift into my project, I noticed a minor inconsistency between grpc-swift and swift-protobuf in the naming conventions for messages. This caused my generated code to fail to compile.

You can apply this patch, recompile `echo.proto` and then attempt to build the Echo app to see what I'm talking about:

```diff
diff --git a/Examples/Echo/echo.proto b/Examples/Echo/echo.proto
index 1328b5f..ce418b6 100644
--- a/Examples/Echo/echo.proto
+++ b/Examples/Echo/echo.proto
@@ -14,7 +14,7 @@

 syntax = "proto3";

-package echo;
+package echo_package;

 service Echo {
   // Immediately returns an echo of a request.
```

`swift-protobuf` creates messages with names like `EchoPackage_EchoRequest` while `grpc-swift` assumes `Echo_Package_EchoRequest`.

MrMage:
The CgRPC implementation is now officially deprecated; closing this for now.

PR #39, opened 2017-03-27
Fix inconsistency in proto message naming convention

mickeyreiss:
Here's a simple fix for #38 that is working for me as a workaround.

I acknowledge that my proposed patch here is a bit hacky, and I think it could use some unit testing. Moreover, I'm wondering if there's a way this library could pull in [some code](https://github.com/apple/swift-protobuf/blob/0.9.28/Sources/protoc-gen-swift/MessageGenerator.swift) from `swift-protobuf` and ensure that the exact same logic is used for handling Swifty naming conventions. Unfortunately, I don't have enough context in the projects to make such a change.

Thanks!

### DIFF
--- a/Plugin/Sources/protoc-gen-swiftgrpc/main.swift
+++ b/Plugin/Sources/protoc-gen-swiftgrpc/main.swift
@@ -52,18 +52,25 @@ extension String {
   }
 }
 
+// Transforms .some.package_name.FooBarRequest -> Some_PackageName_FooBarRequest
 func protoMessageName(_ name :String?) -> String {
   guard let name = name else {
     return ""
   }
-  let parts = name.undotted.components(separatedBy:"_")
-  var capitalizedParts : [String] = []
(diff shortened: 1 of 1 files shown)


mickeyreiss:
I'm also noticing that this patch does not fix the naming conventions in the gRPC-only classes, such as `Echo_Package_EchoService`.

jcanizales:
+1 for reusing swift-protobuf's logic to generate names. I ended up doing the same in the Objective-C client to keep me sane.

timburks:
We have a long-unused offer from @tbkka to provide a name-generation interface into swift-protobuf's code.

timburks:
I cherry-picked this commit into https://github.com/grpc/grpc-swift/tree/swift3.1.

Let's keep https://github.com/grpc/grpc-swift/issues/38 open until we have a test case that verifies that the names are what we want them to be.

PR #43, opened 2017-04-17
Swift3.1

timburks:
Bring Swift 3.1 compatibility to master; Xcode builds currently have problems.


timburks:
@mickeyreiss Could you please reply with your ok that your commit is in this PR? (see the note from @googlebot above.)

mickeyreiss:
🆗

Examples/Echo/RUNME
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+#
+# Use this to run the swift-proto generator
+#
+protoc echo.proto --swift_out=. --swift_opt=Visibility=Public --plugin=/Users/tim/bin/protoc-gen-swift
```
mickeyreiss: Neat! Is this Visibility flag new?

timburks: It's [not that new](https://github.com/apple/swift-protobuf/blame/d5ccb9558c0083f7fbe686db0889030560d2c5fe/Documentation/PLUGIN.md), but it wasn't needed for 0.9.24, so grpc-swift wasn't using it.

timburks: I haven't checked the podspec for this branch, so it's likely not to work without fixes. If you would advise on that, it would be great.

mickeyreiss: > It's not that new, but it wasn't needed for 0.9.24, so grpc-swift wasn't using it.

Got it. Thanks for explaining!

> I haven't checked the podspec for this branch, so it's likely not to work without fixes. If you would advise on that, it would be great.

I will try and report back.

Issue #55, opened 2017-05-25
Why is public visibility required for the source generated from the protos?


thomasvl:
The _README.md_ documents the need for `--swift_opt=Visibility=Public` as:

> The `--swift_opt` flag passes an option to `protoc-gen-swift` that makes message classes public, which is necessary to use them with gRPC.

But that explication doesn't make sense.  If the _.pb.swift_ files and grpc files are compiled in the same module, then the default internal visibility would be fine.

The real reason one might need public visibility is if the _.pb.swift_ files (and grpc ones) will be placed in a different module from the code that uses them.

timburks:
The generated gRPC methods are public, and since they use structs generated by swift-proto, those structs need to also be public.

Since the generated gRPC methods might be used in a separate module, "public" access seemed like a safer choice. 

Would you recommend that gRPC also support --swift_opt and make that user-controllable?

thomasvl:
It all depends on how developers want to use it, if the intent was to have grpc as an impl detail within some other library, forcing public can seem debatable.  It also could get in the way of WMO since it might force the compiler/optimizer to keep things exposed incase an external caller is added.

timburks:
Thanks - a simpler alternative would be to completely remove the "public" access specifiers and wait for someone to request a Visibility option.

thomasvl:
>Thanks - a simpler alternative would be to completely remove the "public" access specifiers and wait for someone to request a Visibility option.

Yup, because of the Well Known Types (and exposing the PluginLibrary), we had to add the option to SwiftProtobuf.

timburks:
OK, out they go.

PR #56, opened 2017-05-30
Change access level for generated classes and methods to "internal"

timburks:
"public" is replaced by a template parameter which is currently hard-coded to "internal".


mickeyreiss:
Is this related to #55? I was planning to compile my generated files in a framework, which is included in my app. Is this setup no longer supported?

timburks:
Yes, but if you need it, I'll add support for swift-proto's Visibility option. In the meantime, you can set the access level to public with a one-line change [here](https://github.com/grpc/grpc-swift/blob/master/Plugin/Sources/protoc-gen-swiftgrpc/main.swift#L108).

PR #62, opened 2017-07-11
Allows passing \'--swiftgrpc_opt=Visibility=Public\' to protoc to cha…

deatondg:
…nge the visibility of the generated code for use in projects using multiple modules.

Hello! I'm new to open-source, but hopefully I did this whole pull request thing properly. Let me know if there's anything I should do differently next time.

**What does this do?**

This PR adds the ability to change the access controls of the generated code when calling 'protoc' by passing options with '--swiftgrpc_opt='. The only option that this PR adds is 'Visibility'. Setting 'Visibility=Internal' or not specifying an option at all will produce code identical to before this PR. Setting 'Visibility=Public' will change 'access' in the context of the Stencil rendering to 'public' in order to change the access of generated code to 'public', which allows compiling and using the generated code in separate modules.

**Possible regressions**

If an invalid option is specified, code generation through the grpc-swift plugin will return an error. Previously, the option was simply ignored. Any workflow which relied on passing junk options will no longer function; however, I do not anticipate that being a meaningful regression. 

**Another gotcha**

As the functionality of this PR is intended to mirror the visibility options of the swift-protobuf plugin, the logic used to implement this feature in grpc-swift is basically identical to the logic from swift-protobuf. The 'options.swift' file is basically just a feature-stripped and slightly refactored version of the [GeneratorOptions.swift](https://github.com/apple/swift-protobuf/blob/master/Sources/protoc-gen-swift/GeneratorOptions.swift) file from protobuf-swift. The logic for actually changing the access controls was already in the Stencil templates, so the only code needed for this PR just allowed setting options. I'm not really sure if this is okay or not. Both projects appear to be licensed under Apache 2.0, but I don't understand code licensing very well.

### DIFF
--- a/Plugin/Sources/protoc-gen-swiftgrpc/main.swift
+++ b/Plugin/Sources/protoc-gen-swiftgrpc/main.swift
@@ -56,7 +56,9 @@ func main() throws {
   // read plugin input
   let rawRequest = try Stdin.readall()
   let request = try Google_Protobuf_Compiler_CodeGeneratorRequest(serializedData: rawRequest)
-
+  
+  let options = try GeneratorOptions(parameter: request.parameter)
+  
   var generatedFileNames = Set<String>()
 
   // process each .proto file separately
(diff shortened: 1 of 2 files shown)


timburks:
Thanks for doing this! I've been on vacation and am at a conference this week, but will review this soon so that we can get this in.

deatondg:
Sounds great, thanks!

PR #76, opened 2017-09-29
Added some changes to create a channel for a GRPC server which takes 3 certs as input..

ujjwalap:


timburks:
@ujjwalap Thanks for raising this! - We consider this covered by #244.

PR #81, opened 2017-10-25
Add Cancel to Call

KyoheiG3:
It should be able to cancel. However, I do not know if it is necessary to cancel about streaming.


timburks:
Thank you!

PR #83, opened 2017-10-27
Fix typo in README

walkingbrad:
When describing how to `make` the binary the "Plugins" directory is referenced, however the actual directory name is "Plugin".


timburks:
Fixed.

Issue #93, opened 2017-12-14
*.client.pb.swift missing services


bootstraponline:
I have a Kotlin gRPC server:

```kotlin
private val server: Server = ServerBuilder.forPort(port)
            .addService(GeneralSeedyImpl())
            .addService(SeedyAssignmentsImpl())
            .addService(SeedyConversationsImpl())
            .addService(SeedyCoursesImpl())
            .addService(SeedyDiscussionsImpl())
            .addService(SeedyEnrollmentsImpl())
            .addService(SeedyFilesImpl())
            .addService(SeedyGradingPeriodsImpl())
            .addService(SeedyGroupsImpl())
            .addService(SeedyPagesImpl())
            .addService(SeedyQuizzesImpl())
            .addService(SeedySectionsImpl())
            .addService(SeedyUsersImpl())
            .build()
```

I'm generating Swift:

```
/code/ios/scripts/grpc/protoc-3.5.0-osx-x86_64/bin/protoc \
  --plugin=protoc-gen-swift=/code/ios/scripts/grpc/protoc-3.5.0-osx-x86_64/bin/protoc-gen-swift \
  --plugin=protoc-gen-swiftgrpc=/code/ios/scripts/grpc/protoc-3.5.0-osx-x86_64/bin/protoc-gen-swiftgrpc \
  -I /code/android-uno/dataseedingapi/src/main/proto \
  "groups.proto" \
  "health_check.proto" \
  "courses.proto" \
  "quizzes.proto" \
  "sections.proto" \
  "assignments.proto" \
  "enrollments.proto" \
  "grading_periods.proto" \
  "discussion_topics.proto" \
  "conversations.proto" \
  "pages.proto" \
  "files.proto" \
  "user.proto" \
  --swift_out="/code/ios/scripts/grpc/../../rn/Teacher/ios/TeacherUITests/soseedy" \
  --swiftgrpc_out="/code/ios/scripts/grpc/../../rn/Teacher/ios/TeacherUITests/soseedy"
protoc-gen-swift: WARNING: untested version of protoc (3.5.0).
```

`soseedy.client.pb.swift` only contains for `groups.proto`

```swift
/*
 * DO NOT EDIT.
 *
 * Generated by the protocol buffer compiler.
 * Source: groups.proto
 *
 */
```

[...]

PR #94, opened 2017-12-14
Fix gRPC Swift generation when files have same package

bootstraponline:
Fix https://github.com/grpc/grpc-swift/issues/93


timburks:
Thanks for posting this. I'm going to make one small change to remove the "0" for the first file generated; that will avoid breaking other users of generated code (except for you, unfortunately) and later I'd like to revisit this to match the behavior of protoc-gen-swift: if that plugin generates only one file for these multi-file service definitions, then I think protoc-gen-swiftgrpc should too.

bootstraponline:
Thanks! We're using this for a project at work. This is the only issue I've encountered so far.

PR #107, opened 2018-01-31
call completion when GRPC_CALL_OK in Call.swift

MihoSasaki:


MihoSasaki:
Sorry, my fix is wrong.
I will take a look.

Issue #92, opened 2017-12-13
Unary callback never executed


pinda:
Trying to build a simple app with this but in discovery I ran into the issue that the callback is not executed. In the example it shows:
`_ = try service.get(requestMessage) {responseMessage, callResult in ...`

If I try it with this piece of code it never goes to the completionHandler, also no error is thrown.

```
let service = Service(address: "127.0.0.1:50051")
let userRequest = RegisterRequest.with {
   $0.name = "foo"
 }

do {
    _ = try service.register(userRequest) {responseMessage, callResult in
      if let responseMessage = responseMessage {
           print(responseMessage.success)
       } else  {
           print("No message received. \(callResult)")
         }
      }
} catch {
    print(error)
}
```

when changing it to the following it does get executed:

```
let service = Service(address: "127.0.0.1:50051")
let userRequest = RegisterRequest.with {
   $0.name = "foo"
 }

do {
    let call = try service.register(userRequest) {responseMessage, callResult in
      if let responseMessage = responseMessage {
           print(responseMessage.success)
       } else  {
           print("No message received. \(callResult)")
         }
      }
    print(call) // Can imagine it will be evaluated at that point, but the example works otherwise.
} catch {
    print(error)
}
```

pinda:
Another test I ran if I change it to the following, the sync will get executed every time but the callback not every time but most of the time.

```
let service = Service(address: "127.0.0.1:50051")
let userRequest = RegisterRequest.with {
   $0.name = "foo"
 }

do {
    let resp = try service.register(userRequest)
    print("sync: \(resp.success)")

    _ = try service.register(userRequest) {responseMessage, callResult in
      if let responseMessage = responseMessage {
           print(responseMessage.success)
       } else  {
           print("No message received. \(callResult)")
         }
      }
} catch {
    print(error)
}
```

bilalyaylak:
@pinda Does second one work without printing 'call', but with assignment? 
I also couldn't achieve running async calls with completion closures. 

```swift
let service = Service(address: "127.0.0.1:50051")
let userRequest = RegisterRequest.with {
   $0.name = "foo"
 }

do {
    let call = try service.register(userRequest) {responseMessage, callResult in
      if let responseMessage = responseMessage {
           print(responseMessage.success)
       } else  {
           print("No message received. \(callResult)")
         }
      }
    print(call) // Can imagine it will be evaluated at that point, but the example works otherwise.
} catch {
    print(error)
}
```

[...]

PR #108, opened 2018-02-01
make return completion when event is GRPC_QUEUE_SHUTDOWN at Completio…

MihoSasaki:
## Description
when we call method with completion and it fails, completion does not return anything
because completion does not call in case of GRPC_QUEUE_SHUTDOWN.

## Related Issue
This fix may be related to #92


timburks:
Thank you!

Issue #103, opened 2018-01-23
Service initializer interface compliance


sgammon:
Using `grpc-swift` latest (as of this writing, `0.3.2`) with _SwiftProtobuf_ latest (`1.0.2`) produces services with the following initializers:

```swift
  /// Create a client that makes insecure connections.
  public required init(address: String) {
    [...]
  }

  /// Create a client that makes secure connections.
  public required init(address: String, certificates: String?, host: String?) {
    [...]
  }
```

In libraries that use multiple services, one might be tempted to create a generic-factory to setup service objects and register them centrally, or otherwise prepare them for use by higher-order code.  Perhaps you want to prepare endpoints in a uniform way, or certs, or what not.
Consider the following example:

```swift
/**
 * Main protocol for a remotely-supported gRPC service.
 */
internal protocol RPCService {
  init(address: String)
  init(address: String, certificates: String?, host: String?)
}

/**
 * Generic RPC service factory.
 */
internal final class RPCServiceFactory<Service: RPCService> {
  static func factory(endpoint: RPCEndpoint) -> Service {
    if let secure = endpoint as? SecureRPCEndpoint {
      // connect over TLS
      return Service.init(
        address: "\(secure.hostname ?? secure.host):\(secure.port)",
        certificates: secure.chain,
        host: secure.hostname)
    } else {
      // connect over plaintext
      return Service.init(
        address: "\(endpoint.host):\(endpoint.port)")
    }
  }
}
```

Whoops! This doesn't work because the initializers in the compiled gRPC code are not designated as `required`. Additionally, they aren't marked as `final`, implying they could be extended, but also aren't marked `open` for cross-module extension, implying they should be extended only inside other compiled code.

I'm working in a Swift package context here, and extending the class from the same module not only seems excessive just to enforce initializer interface compliance, but hard to pull off in a repeatable way if I'm copying Swift into a folder containing custom extension code.

**tl;dr**
Can `grpc-swift` safely add `required` to the initializers for those compiled services, or is there a better approach to generic initialization? I also don't see a compiled/embedded protocol that does this already, perhaps there is and I could use that.

Happy to contribute code if it helps

timburks:
Hi Sam, I need to learn more about Swift initializers!

When you say this:
> This doesn't work because the initializers in the compiled gRPC code are not designated as required.  

Do you mean that your calling code fails to compile? 

[...]

PR #114, opened 2018-02-11
Finalize classes

takasek:
Without special intention,  classes are better to be final.
I want a service class to make abstract using protocol in my project, but non-final class cannot do so.
It fulfills a part of a point of #103


timburks:
Thank you!

Issue #99, opened 2017-12-30
Upgrading vendored CgRPC


ishkawa:
Hi.

The latest release of gRPC is 1.8.2, but gRPC vendored in grpc-swift seems to be 1.6.0.
It would be great if vendored gRPC is upgraded.

bootstraponline:
I think updating all the vendored dependencies would be nice. Currently it's difficult to adopt Cocoapods as the version used are so old.

MrMage:
**tl;dr: Upgrading to CgRPC 1.7.3 appears to be easy, upgrading to CgRPC 1.8.0 and later (current is 1.9.1) would require the Swift 4.1 toolchain, which is not available yet.**

I have been trying my luck at upgrading to CgRPC 1.9.1, the latest version. Upgrading up to 1.7.3 seems to be easy, but it appears that gRPC-Core does include C++ files from version 1.8.0 on, and those files need to be compiled with support for C++11.

Enabling C++11 can be realized by passing `-Xcc -std=c++11` to the swift compiler or, better, by setting `cxxLanguageStandard: .cxx11` in the Package.swift manifest (requires setting the swift tools version to 4 with the line `// swift-tools-version:4.0`). The problem with that flag, however, is that it gets passed to _all_ files in the compilation unit, even the C files. This in turn yields an `error: invalid argument '-std=c++11' not allowed with 'C/ObjC'` message when compiling e.g. `third_party/nanopb/pb_common.c`. The solution would be to pass different C/C++ standard flags to individual files instead of setting the standard per target. That is actually implemented in https://github.com/apple/swift-package-manager/commit/90abb3c4c9415afee34291e66e655c58a1902784 (tracking bug: https://bugs.swift.org/browse/SR-6430), but is only available from Swift 4.1 on, which should be released in a few weeks with Xcode 9.3.

SebastianThiebaud:
Are we still waiting a little bit longer before to upgrade `CgRPC`? We've been using newer version of `CgRPC` in production for awhile in our fork, without any issues. I even upgraded to `1.12` 2 weeks ago. Thoughts?

bootstraponline:
Maybe submit a pull request and see if the CI passes?

MrMage:
[...]

PR #118, opened 2018-02-19
Upgrade the vendored version of gRPC to 1.7.3.

MrMage:
Upgrading to gRPC 1.8.0 and later appears to be a bit more problematic, see #99 for details.

Process used:

    cd third_party
    ./RUNME.sh
    cd grpc
    git checkout v1.7.3
    cp ../../swift-vendoring.sh.template templates
    time tools/buildgen/generate_projects.sh
    cd ../..
    ./vendor-grpc.sh


timburks:
Thanks!

Issue #120, opened 2018-02-20
What's the best practice for testing SwiftGRPC clients/servers?


MrMage:
Are there any best practices on how to test SwiftGRPC clients and servers, and mocking the other side, so that we can write hermetic tests? Let's take the `Echo.Update` bidirectional streaming call from https://github.com/grpc/grpc-swift/tree/master/Examples/Echo, for example.

As far as I can tell, my client consumes an object of type `Echo_EchoUpdateCall` and calls that object's methods to send/receive messages. Therefore, it would be nice to mock the `Echo_EchoUpdateCall` to return custom reply messages and record the sent messages, to ensure the correct ones are sent. However, as of #114 (cc @takasek), that class is actually `final`. So, how am I supposed to test that the client sends the correct messages on my RPC? Would it maybe make sense to declare `Echo_EchoUpdateCall` as a protocol instead, and only make the actual implementation a class?

The same applies if I'd like to test the bigger picture, namely that the right RPCs are called on `Echo_EchoService`. That's also a final class — should it maybe be a protocol as well, letting me mock the whole service? (In fact, that class _needs_ to be mock-able if I e.g. wanted to fake that class's `func get(_ request: Echo_EchoRequest) throws -> Echo_EchoResponse` call.)

Same for server-side testing: My server class needs to implement `Echo_EchoProvider`, whose `update` methods gets passed a `Echo_EchoUpdateSession`:

    func update(session : Echo_EchoUpdateSession) throws

Again, it would be nice to be able to fake that class, letting me provide fake requests and ensure the right responses are returned. So, should `Echo_EchoUpdateSession` be a protocol for better testability as well?

By the way, testing the `get` method would in theory be slightly easier, given that it simplay takes in an `Echo_EchoRequest` and returns a `Echo_EchoResponse`:

    func get(request : Echo_EchoRequest, session : Echo_EchoGetSession) throws -> Echo_EchoResponse

However, **even that can't be tested, because it's impossible to fake `Echo_EchoGetSession`**, with that class's initializer being `fileprivate`. So, as far as I can tell, to test these things, at the moment I'd need to wrap everything in another layer of protocols and test fake implementations of those protocols, resulting in a ton of boilerplate. (The "real" implementation of those testable protocols would simply call through to the gRPC methods and wrap the results of those methods in other protocols.)

[...]

PR #124, opened 2018-02-22
Extract the interface of generated code into protocols for better testability

MrMage:
...and make the implementing classes conform to those protocols. This lets us create custom implementations of those protocols, e.g. mocks/stubs/fakes/spies for testing purposes. In particular, this commit also includes an option to have simple fake implementations generated for the above protocols in addition to the "real" implementations (see #120 for a detailed discussion).

The only breaking change in this code is that the client-side RPC server interface now needs to be instantiated as e.g. `Echo_EchoServiceImpl(address:secure:)` instead of `Echo_EchoService(address:secure:)`. Anywhere else, the protocols have the same name and expose the same interface as the previously-used `final class` implementations, with classes still being used under the hood.


timburks:
I was unsure of the "generate_mock_code" option name, but it seems to be used by the C++ plugin. Are there any other plugins that use this same option name?

MrMage:
Yeah, I used that option name because I found it for the C++ plugin. It appears that [the Objective-C plugin](https://github.com/grpc/grpc/blob/master/src/compiler/objective_c_plugin.cc) doesn't have an option to generate such mocks. The Go client libraries [can apparently be mocked with Go's `mockgen` tool](https://github.com/grpc/grpc-go/blob/master/Documentation/gomock-example.md). For Java, I guess custom mocks aren't necessary thanks to the presence of great reflection-based mocking frameworks such as Mockito. In theory, one could also use a tool like [Cuckoo](https://github.com/Brightify/Cuckoo) to generate Swift-based mocks _after running the SwiftGRPC codegen_, but I feel that having an option to have the compiler generate primitive mocks without introducing another dependency wouldn't hurt.

I don't care much about the option's name, happy to change it to whatever you propose :-)

timburks:
OK, sorry to nitpick on naming, but here goes...

I'd like to use something more concise than "generate_mock_code", either "mocks" or "stubs", perhaps using [this discussion](https://martinfowler.com/articles/mocksArentStubs.html)) as a guide. On a cursory read, I think that "stubs" is probably the right word, and it matches your usage in the code.

Also the "Impl" suffix doesn't feel very Swifty to me. Could we remove it and instead rename the protocol by adding the "Protocol" suffix? (as in [this style discussion](https://github.com/raywenderlich/swift-style-guide/issues/123#issuecomment-276894689)). Thus  "Echo_EchoSession" and "Echo_EchoSessionStub" would both conform to "Echo_EchoSessionProtocol". This would keep the common case simple.

MrMage:
Have a look at the comment right after that one: https://github.com/raywenderlich/swift-style-guide/issues/123#issuecomment-277068418

I feel that `...Protocol` is redundant — the name itself should describe what this is about, and consumers should only care about the interface exposed by the protocol, not whether they are dealing with a protocol or a class or the concrete implementation or a stub. So library consumers should simply consume the protocols, anyway. (Also, as far as I know, the Swift standard library has moved away from `...Protocol` suffixes.)

Also, this change specifically makes almost `Impl` classes `fileprivate`, so the actual implementation name shouldn't matter, except for the case of `Echo_EchoServiceImpl`. How about renaming that one to e.g. `Echo_EchoServiceClient`, though? Also happy to replace `Impl` in the other cases with something else, if you have a suggestion.

In terms of codegen options, I have renamed the option to `TestStubs` now, and renamed the classes to `...TestStubs` to avoid confusion with gRPC's client-side RPC shims (i.e. the things clients use to call RPC functions on), which are also called stubs by e.g. the C++ implementation. (According to the linked article, these indeed are stubs, not mocks.)

MrMage:
In case this ends up being blocked only by having to replace `...Impl` in the generated `fileprivate` classes (no externally visible symbols are affected), would it be possible to merge this and change the `...Impl` names in a follow-up pull request? I’d like to try moving most of the generated code into a few generic classes in the runtime library instead (to help make reasoning about the generated code easier), but this pull request needs to be merged first and is blocking me from starting work on that.

timburks:
Replacing "Impl" with "Client" works for me. We can make the other Impl changes in a followup PR. When things stabilize, l'll tag and release an 0.4.0 version since there are some breaking API changes already in master.

PR #127, opened 2018-02-22
Fix a similar set of warnings in the vendored gRPC code

MrMage:
See https://github.com/grpc/grpc-swift/pull/126 for details. If you'd rather not have manual changes like this to the vendored code, no objections. In that case, I'd just suggest leaving them in there until we are able to get rid of the vendored code completely, then fix them upstream in gRPC if still necessary.


timburks:
To me, diverging from the source gRPC code seems worse than having the warnings. Let's see if these can be fixed upstream. @muxi, could you check on that for us?

MrMage:
Agreed, especially as e.g. CocoaPods has an option to suppress warnings from included pods anyway. Closing this.

PR #129, opened 2018-02-22
Remove "Plugin/Sources/protoc-gen-swiftgrpc/templates.swift" from source control

MrMage:
This file actually gets regenerated by running `make` in the `Plugin` directory, but it is easy to forget to check in changes to it when making changes to the underlying templates. So to avoid inconsistencies, I think it is better to only rely on the user's freshly-generated local copy.


timburks:
Thanks, this is great! I wasn't aware of the --product option for "swift build". That makes the plugin build much cleaner.

PR #137, opened 2018-02-23
Make more variables `let` and use `Mutex.synchronize` instead of `Mutex.lock()+unlock()` in a few more cases for consistency

MrMage:


timburks:
Looks great. Thanks!

PR #138, opened 2018-02-23
Make most completion blocks optional, so that clients don't need to pass them

MrMage:
I've also changed several completion blocks to accept a boolean argument whether the corresponding operation was successful, instead of just never calling the completion block in case of errors.
In general, I think we should soon take a hard look at how errors are handled currently and ensure that no errors can lead to unexpected behavior (e.g. some blocks never getting called etc.), both in the client and server use-cases.


timburks:
@MrMage would you mind resolving these merge conflicts?

MrMage:
Seems like I tried too much at once here ^^ Updated.

Issue #111, opened 2018-02-08
Cocoapods Support?


AttilaTheFun:
I'm interested in using this library for a prototype Swift iOS application. It's nice that https://github.com/apple/swift-protobuf allows for installation. I was wondering if it would be possible to offer the SwiftGRPC.xcodeproj through Cocoapods, though I suppose it would require BoringSSL, CgRPC, gRPC, and CzLib all be available through Cocoapods as well.

muxi:
We have that in the list.

gRPC and BoringSSL are in Cocoapods already.

timburks:
@AttilaTheFun I started the podspec and ran into problems, so I renamed it to avoid over-promising (since it didn't work). I'm not a Cocoapods user, so am hoping @muxi can help from his experience with Objective-C gRPC.

AttilaTheFun:
@muxi @timburks I tried forking the library and linting the podspec. I got an error about a double-import from the boringSSL module map. I googled around about the error and I think it has since been fixed, but the version of gRPC-Core depended on by this library is significantly out of date. (1.0.1 vs 1.9.1 which is current). When I tried to bump the gRPC-Core version to 1.9.1, I got a bunch of compile errors because I think this library used some non-backwards compatible APIs. Is it possible to get grpc-swift to be compatible with the latest version of grpc-core?

AttilaTheFun:
Btw here is the discussion about the issue I see currently:
https://github.com/GoogleCloudPlatform/ios-docs-samples/issues/39

MrMage:
I have created a Podspec depending on the latest `gRPC` version (1.9.1) that seems to compile for my project. You should be able to try it out via the following `pod` includes:

	pod 'BoringSSL', '~> 9.1'
	pod 'gRPC-Core', :git => 'https://github.com/mrmage/grpc', :branch => 'fix-cocoapods-swift'
	pod 'SwiftGRPC', :git => 'https://github.com/mrmage/grpc-swift/tree/podspec', :branch => 'podspec'

Not sure if this is production ready yet (I might have missed something), but I would appreciate if you could try it out and give feedback, so I can send a pull request sometime soon.

AttilaTheFun:
@MrMage I tried adding your branches as dependencies in my pod file but was unable to install them because I don't believe they're published to the master specs repo. I then attempted to clone your forks and push your pod specs to my private repo, and the gRPC-Core spec didn't validate. This was the output of the failed validation:
```
Validating spec
[...]

PR #139, opened 2018-02-23
Create a first version of a Podspec that seems to be working

MrMage:
At least for me, this Podspec is working. Also, as long as the podspec is called `Podspec-incomplete.swift`, running `pod install` fails right away, so I think using the proper name makes sense already, for those who'd like to try it out. See https://github.com/grpc/grpc-swift/issues/111#issuecomment-366733622 on how to use the Podspec for now.

What still needs to be done, though, is changing the resulting module name, while currently the generated code imports `gRPC` (also see https://github.com/grpc/grpc-swift/issues/111#issuecomment-366733622). I would actually suggest always using `SwiftGRPC` as the name for this project's module, instead of `gRPC`, to avoid confusion with `gRPC-Core` (which gets built as `gRPC` as well). WDYT?

### DIFF
--- a/SwiftGRPC.podspec
+++ b/SwiftGRPC.podspec
@@ -29,7 +29,7 @@
 
 Pod::Spec.new do |s|
   s.name = 'SwiftGRPC'
-  s.version = '0.1.0'
+  s.version = '0.3.3'
   s.license  = 'New BSD'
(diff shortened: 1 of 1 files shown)


timburks:
Thanks, this is huge! I think using SwiftGRPC for the module name is appropriate - it's also consistent with SwiftProtobuf.

PR #141, opened 2018-02-23
Add some free methods as static members to a `gRPC` class instead

MrMage:
Add the free methods `initialize`, `shutdown`, `version` and `gStandsFor` as static members to a `gRPC` class, to avoid namespace collision.

I figured this might help avoid confusion what a call to e.g. `initialize()` stands for, but YMMV. Feel free to reject if you think this is not a good idea.


timburks:
This is good. I'm assuming we're also renaming the package from "gRPC" to "SwiftGRPC", which will hopefully also reduce ambiguity.

PR #151, opened 2018-02-28
Fix code generator and samples for latest gRPC API.

timburks:


timburks:
@MrMage I merged a bunch of PRs without waiting for tests to run - then noticed the plugin was generating code for the old APIs. I'll iterate here until the tests pass.

timburks:
@MrMage it looks like we're passing again. Thanks for adding the example tests!

PR #154, opened 2018-02-28
Add functional testing of the Echo example to the top-level "make test".

timburks:


MrMage:
According to CI, `Echo collect >> test.out` never terminates. I could imagine that there's still a "hanging" bug somewhere inside the client or server code causing that.

timburks:
I tested on Linux and found synchronization problems in the Echo demo that allowed the Linux results to differ from what I got on my Mac. Prior to the fix, "make test" didn't block on Linux, but it may still be related to the CI failure.

timburks:
Looks like there still was a race:
```
diff test.out test.gold
27a28
> update closed
31d31
< update closed
```
I removed those "closed" messages because their ordering appears to depend on gRPC library internals.

MrMage:
Alternatively, if you only care that things happen, but not about their order, you could sort the "gold" example and the output of the binary before comparing the two.

(I also hope to add some tests with expectations soon that lore elegantly check that callbacks are called soon.)

timburks:
@MrMage agreed - let's think about doing things like that in more sophisticated (possibly hand-coded) tests.

PR #155, opened 2018-02-28
Extract most generated code into generic runtime support classes

MrMage:
Note: this depends on pull request #138, and should become somewhat simpler once that has been merged.

This removes most code from the codegen templates and puts it into dedicated runtime library classes in the new `GenCodeSupport` directory instead. This makes it easier to reason about the generated code and make modifications to the client/server code, as we no longer have to mess with hard-to-read gencode snippets most of the time.

Note that this PR does *not* change the _interface_ of the generated code — you use it exactly as before this PR (but after #138, as that PR's changes are included here). I plan to make some more changes to the generated code soon, and this PR should make that _much_ easier.

(I've also changed most occurrences of `...Impl` to `...Base`. Not sure if that is the most adequate, but I figured it would be "good enough" for now. Some of those classes are now publicly exposed by the runtime library (e.g. `ClientCallUnaryBase`), but still only used by the generated code, so the name shouldn't have too significant effects.)


timburks:
This looks great. Would you also run "make test" in the Plugin directory and then, if you're satisfied with the generated code, run "make deploy" there? That will update Examples/Echo/Generated/echo.grpc.swift to the new code generation output.

(I was thinking of taking that generated code out of the repo but it seems useful to illustrate the generation output.)

MrMage:
I agree that keeping echo.grpc.swift for illustration purposes is a good idea. But I think I did already update that file with this PR (that's why the tests pass) — could you please double check?

timburks:
@MrMage Got it - I missed it earlier, maybe because it was collapsed in the GitHub diff view. Looks great!

timburks:
I just hand-merged changes in the Echo main.swift and plan to merge this as soon as the tests pass.

timburks:
New error:
```
starting insecure server
ERROR: ServiceServer has been asked to handle a request even though it has already been deallocated
Unknown error occurred.
```

MrMage:
I think that's kind of expected, given that the server leaves scope right after it is started. Maybe assign the server to a global variable, or at least one that stays around until the test finishes?

BTW, I've seen a `Group` statement in the examples `main.swift`. Where is that declared, and what exactly does it do? Just curious.

MrMage:
I have added another commit that fixes the test for me, makes them runnable on a case-sensitive file system (by executing `Echo` with a `./` prefix), and ensures that their output don't show up in the git index (in the future, we may want to simply write all outputs as well as the `Echo` binary into a separate subdirectory rather than adding all these files individually to the `.gitignore`).

timburks:
OK, I'll merge this and work on the broken test.

For the most part, I don't mind testing with debug. The Google samples have a dependency (https://github.com/google/auth-library-swift) that does RSA signatures in Swift; that is 60x slower in debug builds than release builds!

MrMage:
Thanks! I suspect the culprit with the test are difference in the Linux implementation of libdispatch. The tests might be fixable by introducing short delays to introduce ordering and avoid race conditions, but of course that is not optimal. I’m planning to look through the whole execution flow soon anyway, so maybe I’ll find a way to make this always work as expected in a consistent fashion.

Issue #130, opened 2018-02-22
User-Agent in Metadata


SebastianThiebaud:
Hello,

I've been looking through the entire project to see what is responsible in overriding the `User-Agent` I'm providing in the service metadata. 

I can see that `fill_common_header()` is setting the user agent, so I forked this repository to remove it for testing, but the user agent sent to the server is still `grpc-c/4.0.0 (ios; chttp2; garcia)`.

Any idea what's going on here? I'd like to be able to change the user agent. Thank you.

SebastianThiebaud:
I've been investigating this "issue" for the last hour and came up with a solution. I forked this repository, updated the `channel.c`, `Channel.swift` and the protoc plugin template to let customize the user agent when initializing a service. It works fine now, but I'd love to see the official support of this.

timburks:
Thanks for raising this. We'll investigate and propose something.

SebastianThiebaud:
This is solved with channel arguments. Closing this issue.

MrMage:
FYI, I’m using Endpoints Service Proxy, which does not forward the user agent field to the backend. My workaround was to simply write the user agent field into a different metadata header and check that one in the request handler.

PR #158, opened 2018-03-01
Add support for customized user agent

SebastianThiebaud:
This is a proposal for https://github.com/grpc/grpc-swift/issues/130


Sources/CgRPC/shim/channel.c
```diff
@@ -23,19 +23,29 @@
...
-  grpc_channel_args channel_args;
-  channel_args.num_args = 0;
-  c->channel = grpc_insecure_channel_create(address, &channel_args, NULL);
+
+  grpc_channel_args *channelArgs = gpr_malloc(sizeof(grpc_channel_args));
+  channelArgs->args = gpr_malloc(1 * sizeof(grpc_arg));
+
+  grpc_arg *arg = &channelArgs->args[0];
+  arg->type = GRPC_ARG_STRING;
+  arg->key = gpr_strdup("grpc.primary_user_agent");
```
MrMage: While at it, could you use `GRPC_ARG_PRIMARY_USER_AGENT_STRING` for the argument to `grp_strdup`? (Also below.)

Sources/CgRPC/src/core/ext/filters/http/client/http_client_filter.c
```diff
@@ -505,9 +505,7 @@ static grpc_slice user_agent_from_args(const grpc_channel_args *args,
     }
   }
 
-  gpr_asprintf(&tmp, "%sgrpc-c/%s (%s; %s; %s)", is_first ? "" : " ",
-               grpc_version_string(), GPR_PLATFORM_STRING, transport_name,
-               grpc_g_stands_for());
+  gpr_asprintf(&tmp, "");
```
MrMage: Modifications to the gRPC-Core code will be overwritten by any updates to the vendored code, and are not included in the CocoaPods distribution of the library (because that uses the public gRPC-Core repo directly). Is there another way to accomplish this?

Sources/gRPC/Channel.swift
```diff
@@ -36,12 +36,14 @@ public class Channel {
   ///
   /// - Parameter address: the address of the server to be called
   /// - Parameter secure: if true, use TLS
-  public init(address: String, secure: Bool = true) {
+  /// - Parameter userAgent: an optional user agent used for request
+  public init(address: String, secure: Bool = true, userAgent: String?) {
```
MrMage: How about `userAgent: String = defaultUserAgent`? (Also three more times below below.)

Sources/gRPC/Channel.swift
```diff
@@ -77,3 +80,5 @@ public class Channel {
     return Call(underlyingCall: underlyingCall, owned: true, completionQueue: completionQueue)
   }
 }
+
+private let defaultUserAgent = "\(grpc_g_stands_for()):\(grpc_version_string())"
```
MrMage: I would make this more prominent. Maybe move it to the top of the file or into the `gRPC` class. Also, would you mind using the `gRPC.gStandsFor` and `gRPC.version` variables instead? (I will file a separate PR to make them non-optional again.)

Sources/CgRPC/src/core/ext/filters/http/client/http_client_filter.c
SebastianThiebaud: Yeah, I agree, we shouldn't overwrite `CgRPC`. My short term goal is to make `grpc-swift` available through Carthage and we need to use libraries like `gRPC-Core`, `CZlib`, `BoringSSL` as dependencies. I'll look into it and see if there is another way but all my attempts so far weren't concluding.

PR #161, opened 2018-03-01
Add a first batch of XCTest-based tests covering the generated code

MrMage:
Also includes:

- removing a few sleep() calls that will hopefully no longer be necessary,
- extending `RootsEncoder` with an option to generate a Swift file with variables for the test certificates
- passing certificate strings directly to the various server initializers


MrMage:
FYI, this might also be fixing the test flakes we've seen under Linux...

MrMage:
Seems like Travis died downloading Swift. Retrying...

MrMage:
Seems like Travis got stalled, starting another one...

MrMage:
I can't believe it, the tests passed twice in a row! Let's merge this before Travis changes its mind.

(The run for `c77521d` only failed because I had broken building the plugin.)

MrMage:
Re-running this one more time, just to see if the tests still pass. Not 100% sure what fixed it, so I'm still not 100% sure whether they stopped being flaky now.

MrMage:
Go figure, it flakes again 😕

MrMage:
This will be made obsolete by #170.

PR #165, opened 2018-03-04
Runtime simplifications.

timburks:


Sources/gRPC/Runtime/ClientCallBidirectionalStreaming.swift
```diff
@@ -18,12 +18,7 @@ import Dispatch
 import Foundation
 import SwiftProtobuf
 
-public protocol ClientCallBidirectionalStreaming: ClientCall {
```
MrMage: I had kept the empty protocols because I felt that even though they don't add methods, they make it more clear what things are implementing. For example, something that conformed to `ClientCallBidirectionalStreaming` wouldn't conform to `ClientCallServerStreaming`, even though both protocols are empty. This would let a method explicitly require something that semantically is a bidirectional call. So I'd suggest keeping these empty protocols.

WDYT?

(This also applies to all other removed protocols.)

timburks: As we discussed, I've gotten some feedback to avoid using empty protocols for documentation purposes. There's a small runtime cost and also an element of [YAGNI](https://martinfowler.com/bliki/Yagni.html). But I see that your latest commits add a method to some of these protocols. I'll file a separate PR for the directory reorg and we can treat the protocol question separately. Thanks!

MrMage: Sounds good! No further objections to deleting the empty protocols where applicable, though; I don't feel strongly about them.

PR #170, opened 2018-03-06
Add a first set of XCTest-based tests

MrMage:
Obsoletes #161, but is based on #168, so that PR needs to be merged first (which should also make the diff cleaner).

Also includes support for directly initializing servers with certificate strings instead of URLs to certificate files, as well as codegen support (and generated code) to have the certificate strings available from the tests.


MrMage:
Closing and re-opening to re-run CI, I might have seen another test flake...

PR #173, opened 2018-03-06
Make `Handler.sendResponse` and `.receiveMessage` call through to the corresponding methods on `Call`

MrMage:
(Based on #170.)

This lets us re-use the message queue already implemented on `Call` when sending replies from the server.

While at it:
- unify naming by replacing a few instances of `callback` with `completion`.
- make `Call.messageQueueMaxLength` an optional integer, and ignore its value if it is nil.


MrMage:
Weird, this one works for me.

Issue #156, opened 2018-02-28
We should run CI tests both on Linux and macOS environments


MrMage:
See https://docs.travis-ci.com/user/reference/osx/. No idea whether macOS environments are available on the free plan for open source projects, though.

timburks:
It's available. I've used it for [another project](https://github.com/nulang/nu/blob/master/.travis.yml).

timburks:
Merged in PR #178

PR #178, opened 2018-03-08
Add support for running CI on macOS as well

MrMage:
(Based on #170.)


MrMage:
This will also fix #156.

PR #182, opened 2018-03-09
move examples under Sources/Examples

timburks:


MrMage:
👍 Well spotted — much cleaner than putting each into a separate directory under `Sources`. Should we move the remaining examples there as well?

PR #147, opened 2018-02-27
Send status on client for streaming calls

SebastianThiebaud:
This PR is a proposal for https://github.com/grpc/grpc-swift/issues/144

Issue #144, opened 2018-02-25
Calls always return OK for streaming RPCs


cvanderschuere:
There appears to be a significant bug with the client implementation of streaming RPCs where the status code is never processed. I have been able to verify that this is a swift-specific problem through a comparison with other clients (python & go).

With some investigation, the problem appears to be with what operations are performed for streaming RPCs:

Server Streaming (Swift):
https://github.com/grpc/grpc-swift/blob/c7a788526044077deb584de05e4dde6dbe8005b3/Sources/gRPC/Call.swift#L232-L237

Server Streaming (Python):
```python
_UNARY_STREAM_INITIAL_DUE = (
    cygrpc.OperationType.send_initial_metadata,
    cygrpc.OperationType.send_message,
    cygrpc.OperationType.send_close_from_client,
    cygrpc.OperationType.receive_initial_metadata,
    cygrpc.OperationType.receive_status_on_client,
)
```

The same appears to be true for client and bi-directional streaming as well, although I haven't explicitly verified those.

Issues:
- Streaming RPCs should process status
- Default status code should not be `OK`

timburks:
@cvanderschuere could you confirm that PR #147 resolves this for you?

cvanderschuere:
Tests confirm this was fixed

PR #183, opened 2018-03-11
Added tests for server & bi-directional streaming

cvanderschuere:
Following up on #144, these tests verify that #147 fixed the problem.

### DIFF
--- a/Sources/protoc-gen-swiftgrpc/options.swift
+++ b/Sources/protoc-gen-swiftgrpc/options.swift
@@ -85,7 +85,7 @@ class GeneratorOptions {
   }
 
   static func parseParameter(string: String?) -> [(key: String, value: String)] {
-    guard let string = string, string.characters.count > 0 else {
+    guard let string = string, string.isEmpty else {
       return []
     }
     let parts = string.components(separatedBy: ",")
--- a/Sources/protoc-gen-swiftgrpc/options.swift
+++ b/Sources/protoc-gen-swiftgrpc/options.swift
@@ -85,7 +85,7 @@ class GeneratorOptions {
   }
 
   static func parseParameter(string: String?) -> [(key: String, value: String)] {
-    guard let string = string, string.isEmpty else {
+    guard let string = string, !string.isEmpty else {
       return []
     }
     let parts = string.components(separatedBy: ",")
--- a/Sources/SwiftGRPC/Core/Call.swift
+++ b/Sources/SwiftGRPC/Core/Call.swift
@@ -108,7 +108,7 @@ public struct CallResult: CustomStringConvertible {
       initialMetadata = op.receivedInitialMetadata()
       trailingMetadata = op.receivedTrailingMetadata()
     } else {
-      statusCode = .ok
+      statusCode = .unknown
       statusMessage = nil
       resultData = nil
       initialMetadata = nil
--- a/Tests/SwiftGRPCTests/GRPCTests.swift
+++ b/Tests/SwiftGRPCTests/GRPCTests.swift
@@ -44,13 +44,13 @@ let initialClientMetadata =
     "x": "xylophone",
     "y": "yu",
     "z": "zither"
-  ]
+]
 let initialServerMetadata =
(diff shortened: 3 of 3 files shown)


MrMage:
Thanks a lot for this, it's great to see more cases of the API's usage covered with tests. I'll leave a full review to @timburks (he's the maintainer and knows this portion of the code better).

I think that it would be nicer to just have one or two extra test cases in [`EchoTests`](https://github.com/grpc/grpc-swift/blob/master/Tests/SwiftGRPCTests/EchoTests.swift) (possibly with a slightly customized version of `EchoProvider`), but I think the current generated server-side code would just "swallow" errors like that instead of returning them, so I guess this can't be tested in `EchoTests` right now.

Also, would you mind adapting your indentation to match the project's (2-spaces, no tabs)? Unfortunately, `swift package generate-xcodeproj` always resets that setting to the system defaults, maybe I can create a script to automatically change the defaults for the generated project.

Sources/protoc-gen-swiftgrpc/options.swift
```diff
@@ -85,7 +85,7 @@ class GeneratorOptions {
   }
 
   static func parseParameter(string: String?) -> [(key: String, value: String)] {
-    guard let string = string, string.characters.count > 0 else {
+    guard let string = string, string.isEmpty else {
```
timburks: This isEmpty condition should be inverted, i.e. it should be !string.isEmpty instead of string.isEmpty. Fortunately, our plugin test (`make test-plugin`) caught it!

If you don't mind, please update that in your PR. When the tests pass, I'll merge it.

Thank you!

timburks:
With the inverted guard condition fixed, I'm still seeing a [test failure on Linux](https://travis-ci.org/grpc/grpc-swift/jobs/352506500): 
```
/home/travis/build/grpc/grpc-swift/Tests/SwiftGRPCTests/GRPCTests.swift:253: error: gRPCTests.testConnectivitySecure : XCTAssertEqual failed: ("ok") is not equal to ("resourceExhausted") - 
Test Case 'gRPCTests.testConnectivitySecure' failed (0.045 seconds)
```

cvanderschuere:
@timburks I cannot figure out what's going on here. There seems to be some sort of race condition, but nothing is standing out to me. I also cannot reproduce this error on OSX.

timburks:
@cvanderschuere Have you tested on Linux? I'll try to reproduce this in a Linux VM.

timburks:
@cvanderschuere I can confirm that the test fails on Ubuntu 16.04. The "ok" result is being set on [this line](https://github.com/grpc/grpc-swift/blob/975822c1dc693c36df1a8f02e768bc76c1fd9adc/Sources/SwiftGRPC/Core/Call.swift#L111) which suggests that the operation group failed (and .ok should be something else, possibly .unknown). I'll continue investigating. Thanks for creating this test!

timburks:
@cvanderschuere I don't think [this line](https://github.com/postmates/grpc-swift/blob/e4914ebd4fda472388b233d1d533708002b020e4/Tests/SwiftGRPCTests/GRPCTests.swift#L263) does what your comment suggests that it will do. @MrMage recently added that queue and from reading [Call.swift](https://github.com/grpc/grpc-swift/blob/975822c1dc693c36df1a8f02e768bc76c1fd9adc/Sources/SwiftGRPC/Core/Call.swift#L162), I think it only synchronizes message sends and not call starts. 

Also, I think we need to add `sem.wait()` [here](https://github.com/postmates/grpc-swift/blob/e4914ebd4fda472388b233d1d533708002b020e4/Tests/SwiftGRPCTests/GRPCTests.swift#L288) to keep us from ending the call before the server has closed the connection.

cvanderschuere:
@timburks Thanks for the help! I certainly misunderstood what `messageQueueEmpty` was doing and was stuck in not being able to reproduce the issues.

I made the changes you suggested and all looks to be well.

timburks:
@cvanderschuere Thanks for working through this!

PR #184, opened 2018-03-12
This week's improvements

MrMage:
As discussed via Hangouts with @timburks, I'm going to gather all my changes for this week in this PR, then have them merged at once.

Notable changes:
- Add a script to fix the project's indentation settings. Requires the CocodaPods 'xcodeproj' gem.
- Use the system-default `zlib` module instead of a vendored one.
- Make sure that the caller gets notified of errors in client calls (and add tests for that).
- Fix the SSL-enabled tests.
- Several fixes to `gRPCTests` to make them less flaky (before, they would still flake out from time to time on my machine).
- Make the syntax of calling `send` and `receive` homogenous across all stream-sending/receiving client and server calls.
- Add tests to ensure that timeouts are handled as expected, both client- and server-side. I'd also like to introduce a server-enforced timeout at some point, but that has more time.
- Make sure that the client is notified about server errors (and add tests for that).
- In line with the above, homogenize the interface of closing server sessions and sending results. Unfortunately, that means that you now have to call e.g. `session.sendAndClose(response: response, status: .ok, completion: .nil)` instead of just `session.sendAndClose(response)`, because methods in protocols can't have default arguments. If you prefer, I could add another method to the protocols for the "default" case (status OK, no completion handler) in lieu of default arguments.

FYI, the increase in lines is almost exclusively due to tests.  
Also, I tried to keep the number of commits reasonable, but there is just a lot of stuff in this PR.


SebastianThiebaud:
Nice! I've been struggling to switch indentation settings every time I switch projects. I really hope Xcode projects could embed/store a custom indentation configuration.

MrMage:
They can, but `swift generate-xcodeproj` overrides that when regenerating the project :-/

MrMage:
This should be ready to review now. Will look into the CI failures soon.

MrMage:
Closing and re-opening to measure the effects of the dependency cache...

Makefile
```diff
@@ -8,6 +8,8 @@ all:
 	
 project:
 	swift package generate-xcodeproj
+	# Optional: set the generated project's indentation settings.
```
timburks: Let's add a comment here to explain how to install the gem. I think it's safe to assume people will have ruby installed, but `sudo gem install xcodeproj` might not be obvious to non-rubyists.

Package.swift
```diff
@@ -23,15 +23,15 @@ let package = Package(
     .library(name: "SwiftGRPC", targets: ["SwiftGRPC"]),
   ],
   dependencies: [
-    .package(url: "https://github.com/Zewo/zlib.git", from: "0.4.0"),
+    .package(url: "https://github.com/apple/swift-nio-zlib-support.git", from: "1.0.0"),
```
timburks: I'm not hugely-confident that this project won't move, but since the zewo version was deprecated, I'm ok with this.

timburks:
Thanks for all these improvements! See a few minor comments above, but overall it looks great.

PR #188, opened 2018-03-19
Two important fixes (and possibly more improvements)

MrMage:
Fixed two critical memory leaks in the C shim, as well as another flake in GRPCTests.

### DIFF
--- a/Tests/SwiftGRPCTests/GRPCTests.swift
+++ b/Tests/SwiftGRPCTests/GRPCTests.swift
@@ -19,12 +19,21 @@ import Foundation
 import XCTest
 
 class gRPCTests: XCTestCase {
+  // We have seen this test flake out in rare cases fairly often due to race conditions.
+  // To detect such rare errors, we run the tests several times.
+  // (By now, all known errors should have been fixed, but we'd still like to detect new ones.)
+  let testRepetitions = 10
+  
   func testConnectivity() {
-    runTest(useSSL: false)
+    for _ in 0..<testRepetitions {
+      runTest(useSSL: false)
+    }
   }
 
   func testConnectivitySecure() {
-    runTest(useSSL: true)
+    for _ in 0..<testRepetitions {
+      runTest(useSSL: true)
+    }
   }
 
   static var allTests: [(String, (gRPCTests) -> () throws -> Void)] {
@@ -75,11 +84,8 @@ let helloServerStream = "/hello.server-stream"
 let helloBiDiStream = "/hello.bidi-stream"
 
 // Return code/message for unary test
-let oddStatusCode = StatusCode.ok
 let oddStatusMessage = "OK"
-
-let evenStatusCode = StatusCode.notFound
-let eventStatusMessage = "Not Found"
+let evenStatusMessage = "some other status message"
 
 func runTest(useSSL: Bool) {
   gRPC.initialize()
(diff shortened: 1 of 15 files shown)


MrMage:
Trying to re-run CI...

MrMage:
Please don't merge this yet. I might have introduced new errors.

MrMage:
Actually, I think it's fine ;-)

Tests/SwiftGRPCTests/GRPCTests.swift
```diff
@@ -313,21 +336,28 @@ func handleUnary(requestHandler: Handler, requestCount: Int) throws {
...
+  try requestHandler.receiveMessage(initialMetadata: initialMetadataToSend) {
+    if let messageData = $0 {
+      let messageString = String(data: messageData, encoding: .utf8)
+      XCTAssertEqual(messageString, clientText)
+    } else {
+      XCTFail("handleUnary message missing")
+    }
   }
 
+  // We need to return status OK in both cases, as it seems like the server might never send out the last few messages
```
cvanderschuere: This seems like a huge problem. We need to assert that the server sends an appropriate status code and if it's not currently then tests should be failing

Tests/SwiftGRPCTests/GRPCTests.swift
```diff
@@ -157,23 +166,32 @@ func callUnary(channel: Channel) throws {
     try call.start(.unary, metadata: metadata, message: message) {
       response in
       // verify the basic response from the server
-      XCTAssertEqual(response.statusCode, (i % 2  == 0) ? evenStatusCode : oddStatusCode)
```
cvanderschuere: This means we never check for non `.ok` status codes which was the original bug filed in #144.

cvanderschuere:
@timburks @MrMage Why was this merged? It invalidates work done for #144 in #147 & #183

MrMage:
The tests you provided regularly had "unexpected response" messages (which I converted into `XCTFail` calls) and even occasional crashes (due to forcibly unwrapped optionals in the tests). It appears that if code returns a non-OK status, one can not rely on the gRPC library to still send all sent messages to the client. It took me several hours to figure this out. If you want tests that verify that a non-OK status code is sent, make sure to not send any messages back, as those might get dropped.

Also notice that we still have tests for this — it is now covered (and I'd say in a much cleaner way) by `ServerThrowingTests`.

MrMage:
P.S.: I agree that we might want to raise the issue of the server not sending back all remaining messages when it is given a non-OK status to return with the gRPC-Core team. However, do keep in mind that this is now tested (with much less wrapping code that can cause unrelated flakes) in `ServerThrowingTests`, and I specifically left the part in that tests for custom status messages, given that status codes and messages are always sent together.

cvanderschuere:
> It appears that if code returns a non-OK status, one can not rely on the gRPC library to still send all sent messages to the client. It took me several hours to figure this out. If you want tests that verify that a non-OK status code is sent, make sure to not send any messages back, as those might get dropped.

This doesn't make any sense. Servers regularly return non-OK statuses and continue to operate. I think this is a problem specific to the swift implementation.

My concern is the haste in which these changes were made and the lack of debate over them. It doesn't inspire much faith in the project and quality users can expect from it.

timburks:
@cvanderschuere Thanks for submitting tests and raising your concerns. Please consider everything since our last tagged version tentative.

MrMage:
FYI, I have filed https://github.com/grpc/grpc/issues/14842 with the gRPC project to inquire about the issue with non-OK status codes.

MrMage:
Received an update on https://github.com/grpc/grpc/issues/14842#issuecomment-377953950, with a link to https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/grpc-io/OwZZq27awUQ/uNRL0_7ICQAJ:

> It is guaranteed if the status code of the RPC is OK. If the RPC terminates with any non-OK status, it is not guaranteed that the client will see all messages sent by the server.

So I think the server will need to keep returning `.ok`, or the tests need to be rewritten. We still have [`ServerThrowingTests`](https://github.com/grpc/grpc-swift/blob/master/Tests/SwiftGRPCTests/ServerThrowingTests.swift) to check that non-OK status codes are sent, though.

Issue #186, opened 2018-03-14
Thoughts on adding support for connectivity state


SebastianThiebaud:
This Swift library doesn't expose connectivity states and doesn't include an API to watch state changes: https://github.com/grpc/grpc/blob/master/doc/connectivity-semantics-and-api.md

dcow:
+1. We would like to inform a user/client about the state of a channel at any given time (basically build a channel state UI, nothing more).

MrMage:
Agreed. I am also working on improving the library's error reporting. For example, if you try to connect to an endpoint that simply does not exist, the library will let you send data just fine in streaming requests and simply throw an `.endOfStream` error when trying to read responses. It should instead inform the user earlier that the connection failed.

MrMage:
I'm adding a simple API for this in #188 (e68cec0).

SebastianThiebaud:
I've been looking at your PR @MrMage and thanks for being on top of this. I've been working on a different implementation in my fork matching a little bit more what gRPC Python is offering with observers. What are your thoughts on updating the API to use observers?

MrMage:
My concern with adding an observer is that it adds _another_ polling thread. So running that extra thread should at least be optional I think.

dcow:
FWIW `grpc-java` punts on an observation mechanism too for exactly the same rationale as @MrMage. You'd need to "lock the world" to deliver connectivity state information in a way that's guaranteed to be accurate. From what I've gathered the compromise was to have a way to get a callback when a provided state is no longer the current state. You "watch" the provided state and get a callback when it changes. If the channel state has already changed from the provided state, the callback fires immediately, see: https://github.com/grpc/grpc-java/issues/3763

MrMage:
Can you share that code so we can look at it? :-)

timburks:
What would be a good way to verify this (in tests)?

SebastianThiebaud:
Closing this issue since #196 and https://github.com/MrMage/grpc-swift/commit/e68cec0db8b1cd90f319ff99c9a8417a711a0c0a are adding support for this.

Issue #189, opened 2018-03-19
Further improvements


MrMage:
I might not have time to do all of these, but I figured it would make sense to list them here:

Implementation:
- [ ] **Migrate to Swift 4**
- [ ] **[Improve iOS compatibility](https://github.com/grpc/grpc-swift/issues/218#issuecomment-385353307)**
- [ ] **`ServerCancellingTests.testServerCancelsUnary` and `ServerTimeoutTests.testTimeoutUnary` still emit `ServerSessionUnaryBase.run error sending response: unknown`; we should avoid sending responses when the call has been cancelled anyway. (For that, we would need to know when the call has been cancelled, probably by listening for a `receiveCloseOnServer` operation.)**
- [x] **Check for memory leaks caused by calling `grp_strdup` without freeing the result. — might be fixed by #217** (see also https://github.com/grpc/grpc-swift/pull/199#issuecomment-379598537)
- [ ] **Fix a potential memory leak of `grpc_channel_credentials` in `cgrpc_channel_create_secure`.**
- [ ] Improve concurrency and handling of closing calls (https://github.com/grpc/grpc-swift/issues/19).
- [ ] Look into deallocating (and thus draining) completion queues on a background thread (see https://github.com/grpc/grpc-swift/issues/96)
- [ ] Consider providing the status code with a unary call to fix https://github.com/grpc/grpc-swift/issues/88.
- [x] Check whether we are leaking server address/certificate or metadata string slices etc.
- [ ] Avoid throwing in `ServerSession....run()` — instead catch and log errors and send an error status to the client _(might be obsolete)_
- [x] Unify all the `ServerSession....run()` calls as much as possible; at least the error reporting is identical every time.
- [x] Convert `ServiceServer` to a plug-in architecture: Initialize it with any number of `...Server` implementations that can announce what methods they'd like to respond to. (This is easier than it sounds — essentially, we'd just stop e.g. `Echo_EchoServer` from subclassing `ServiceServer` and instead have it provide a list of methods it responds to. This would let us serve multiple services on the same port.
- [ ] Ensure that the server is going to tell us once it is pointless to send responses because the call has timed out (or been cancelled for other reasons). We don't want the server to keep sending stuff because it doesn't know that the connection has been terminated already. (Also needs tests.)
- [ ] Move the CFlags specified in the `Makefile` into `Package.swift`.
- [ ] Ensure better completion queue shutdowns:
- [ ] - Shut down even when `handler.close...()` throws
[...]

PR #196, opened 2018-03-26
Implement connectivity state property and observers

cvanderschuere:
Implementation for #186 

The goal was to copy the interface that the python client [exposes](https://github.com/grpc/grpc/blob/bd6bdf93279a39a8cd92978fd7c9d14eccd98fc2/src/python/grpcio/grpc/_channel.py#L855).

```python
    def subscribe(self, callback, try_to_connect=None)
```


Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -81,4 +81,145 @@ public class Channel {
     let underlyingCall = cgrpc_channel_create_call(underlyingChannel, method, host, timeout)!
     return Call(underlyingCall: underlyingCall, owned: true, completionQueue: completionQueue)
   }
+
+  public func connectivityState(tryToConnect: Bool = false) -> ConnectivityState {
+    return ConnectivityState.connectivityState(cgrpc_channel_check_connectivity_state(underlyingChannel, tryToConnect ? 1 : 0))
+  }
+
+  public func subscribe(sourceState: ConnectivityState, tryToConnect: Bool = false, callback: @escaping (ConnectivityState) -> ()) {
```
MrMage: Would you mind adding a comment to this method, especially to explain what `sourceState` is for? It looks to me like the callback is only fired if the channel's state changes to `sourceState`, is that correct? Why not subscribe to _all_ state changes and have the caller exclude unwanted ones themselves in the callback (could be done with a simple `if state != sourceState { return }` in the callback)? That would also avoid spinning up a separate thread for each source state.

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -81,4 +81,145 @@ public class Channel {
...
+        while self.polling {
+          guard !self.callbacks.isEmpty && !self.tryToConnect else {
+            self.polling = false
+            break
+          }
+
+          defer { self.tryToConnect = false }
+
+          let deadline: TimeInterval = 0.2
+          cgrpc_channel_watch_connectivity_state(self.underlyingChannel, self.underlyingCompletionQueue, underlyingState, deadline, nil)
```
MrMage: Shouldn't `underlyingState` be the last observed state? At the moment this is only updated once, at the very beginning of the spinloop, so this will fire for any states that are different from the state when the loop was started, instead of simply firing every time we detect any change.

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -81,4 +81,145 @@ public class Channel {
...
+      self.state = state
+      self.underlyingChannel = underlyingChannel
+      self.tryToConnect = tryToConnect
+      self.underlyingCompletionQueue = cgrpc_completion_queue_create_for_next()
+      self.queue = CompletionQueue(underlyingCompletionQueue: self.underlyingCompletionQueue, name: "Connectivity State")
+      self.lastState = ConnectivityState.connectivityState(cgrpc_channel_check_connectivity_state(self.underlyingChannel, 0))
+    }
+
+    deinit {
+      queue.shutdown()
```
MrMage: It looks to me like this will never get deallocated, as you are acquiring a strong reference to `self` right before the spinloop starts and never let go of it until the end of the spinloop. I would suggest shutting down all observers in `Channel.deinit` as well. Let me know if I'm missing something.

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -81,4 +81,145 @@ public class Channel {
...
+    }
+
+    static func == (lhs: ConnectivityObserver, rhs: ConnectivityObserver) -> Bool {
+      return lhs.state == rhs.state
+    }
+  }
+}
+
+extension Channel {
+  public enum ConnectivityState {
```
MrMage: Thanks! This is much better than what I had added before.

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -81,4 +81,145 @@ public class Channel {
...
+}
+
+private extension Channel {
+  class ConnectivityObserver: Equatable {
+    let state: ConnectivityState
+    let queue: CompletionQueue
+    let underlyingChannel: UnsafeMutableRawPointer
+    let underlyingCompletionQueue: UnsafeMutableRawPointer
+    private(set) var tryToConnect: Bool
+    var callbacks: [(ConnectivityState) -> ()] = []
```
MrMage: Access to `callbacks` definitely needs to be synchronized.

cvanderschuere:
Update: haven't had a chance to get to these fixes. I plan on coming back to it by the end of the week.

MrMage:
No worries! Most of these are just nits, anyway; but the possible thread leak and synchronization issues should definitely be investigated — maybe there actually no problems, but it's still worth checking.

timburks:
@cvanderschuere Looks great! We'll merge after those requested changes. Thanks for this!

timburks:
@cvanderschuere @MrMage Is there more coming in this PR? 

If not, I'll merge it and we can address anything remaining in future PRs.

cvanderschuere:
I don’t think this should be merged until the comments have been addressed.

I have not had chance get back to this, but do plan on it eventually. I would also not be opposed to someone taking over the PR and making the necessary change so.

SebastianThiebaud:
I changed a few things. Let me know @MrMage if it's more inline with what you have in mind.

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -86,83 +85,77 @@ public class Channel {
...
-    private(set) var tryToConnect: Bool
-    var callbacks: [(ConnectivityState) -> ()] = []
+  class ConnectivityObserver {
+    private let completionQueue: CompletionQueue
+    private let underlyingChannel: UnsafeMutableRawPointer
+    private let underlyingCompletionQueue: UnsafeMutableRawPointer
+    private let callback: (ConnectivityState) -> ()
     private var lastState: ConnectivityState
-
+    private let queue: OperationQueue
```
MrMage: Please just use `SwiftGRPC.Mutex` to guard **all** accesses to `polling` (including the loop below and `connectivityObservers.forEach { $0.polling = false }` above), an operation queue is overkill here. (Can be disregarded if we eliminate `polling`, but please still remove the `queue` variable in that case.)

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -86,83 +85,77 @@ public class Channel {
...
+      queue.qualityOfService = .background
     }
 
     deinit {
-      queue.shutdown()
+      shutdown()
     }
 
     private func run() {
       DispatchQueue.global().async { [weak self] in
```
MrMage: Please use the thread-spawning pattern now used in e.g. https://github.com/grpc/grpc-swift/blob/8941a852aaec9cacf07c6cacc6a6dbbe65c42619/Sources/SwiftGRPC/Core/CompletionQueue.swift#L118 here now instead of dispatching to the global queue.

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -86,83 +85,77 @@ public class Channel {
...
-        guard let `self` = self, let underlyingState = self.lastState.underlyingState else { return }
+        guard let `self` = self else { return }
 
         while self.polling {
```
MrMage: Replace `self.polling` with `(lock.synchronize { self.polling })`. (Can be disregarded if we get rid of `polling`.)

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -86,83 +85,77 @@ public class Channel {
...
+            let newState = ConnectivityState.connectivityState(cgrpc_channel_check_connectivity_state(self.underlyingChannel, 1))
 
             guard newState != self.lastState else { continue }
             defer { self.lastState = newState }
 
-            if self.lastState == self.state {
-              self.callbacks.forEach({ $0(newState) })
-            }
+            self.callback(newState)
           }
```
MrMage: I'd suggest to handle queue shutdown at this point as well and exit the loop in that case. FYI, this loop retains `self`, which means that `deinit` (and thus `shutdown`) will never get called by itself — but shutdown will be indirectly called by `Channel.deinit`, so we should be fine here.

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -86,83 +85,77 @@ public class Channel {
...
+          
           let deadline: TimeInterval = 0.2
           cgrpc_channel_watch_connectivity_state(self.underlyingChannel, self.underlyingCompletionQueue, underlyingState, deadline, nil)
-          let event = self.queue.wait(timeout: deadline)
+          let event = self.completionQueue.wait(timeout: deadline)
 
-          if event.success == 1 || self.tryToConnect {
-            let newState = ConnectivityState.connectivityState(cgrpc_channel_check_connectivity_state(self.underlyingChannel, self.tryToConnect ? 1 : 0))
+          if event.success == 1 {
+            let newState = ConnectivityState.connectivityState(cgrpc_channel_check_connectivity_state(self.underlyingChannel, 1))
```
MrMage: Are you sure that this should always try to connect? This would make the connectivity observer reconnect all the time, even though the actual channel is currently not used by the client code.

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -86,83 +85,77 @@ public class Channel {
...
-    let queue: CompletionQueue
-    let underlyingChannel: UnsafeMutableRawPointer
-    let underlyingCompletionQueue: UnsafeMutableRawPointer
-    private(set) var tryToConnect: Bool
-    var callbacks: [(ConnectivityState) -> ()] = []
+  class ConnectivityObserver {
+    private let completionQueue: CompletionQueue
+    private let underlyingChannel: UnsafeMutableRawPointer
+    private let underlyingCompletionQueue: UnsafeMutableRawPointer
+    private let callback: (ConnectivityState) -> ()
```
MrMage: super-nit: how about `-> Void`?

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -86,83 +85,77 @@ public class Channel {
...
-
-    if observer == nil {
-      let newObserver = ConnectivityObserver(state: sourceState, underlyingChannel: underlyingChannel, tryToConnect: tryToConnect)
-      observers.append(newObserver)
-      observer = newObserver
-    }
-
-    observer?.callbacks.append(callback)
-    observer?.polling = true
+  public func subscribe(callback: @escaping (ConnectivityState) -> ()) {
```
MrMage: nit: `-> Void`?

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -86,83 +85,77 @@ public class Channel {
...
 
         while self.polling {
-          guard !self.callbacks.isEmpty && !self.tryToConnect else {
-            self.polling = false
-            break
-          }
-
-          defer { self.tryToConnect = false }
-
+          guard let underlyingState = self.lastState.underlyingState else { return }
```
MrMage: Log an error in this case, as this would be very unexpected?

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -86,83 +85,77 @@ public class Channel {
...
-    init(state: ConnectivityState, underlyingChannel: UnsafeMutableRawPointer, tryToConnect: Bool) {
-      self.state = state
+    init(underlyingChannel: UnsafeMutableRawPointer, callback: @escaping (ConnectivityState) -> ()) {
       self.underlyingChannel = underlyingChannel
-      self.tryToConnect = tryToConnect
       self.underlyingCompletionQueue = cgrpc_completion_queue_create_for_next()
-      self.queue = CompletionQueue(underlyingCompletionQueue: self.underlyingCompletionQueue, name: "Connectivity State")
+      self.completionQueue = CompletionQueue(underlyingCompletionQueue: self.underlyingCompletionQueue, name: "Connectivity State")
+      self.callback = callback
       self.lastState = ConnectivityState.connectivityState(cgrpc_channel_check_connectivity_state(self.underlyingChannel, 0))
```
MrMage: nit: I would add a `currentState` argument to `init` and use the result of that here. The initializer call in `Channel` would become

    let observer = ConnectivityObserver(underlyingChannel: underlyingChannel, currentState: self.connectivityState(), callback: callback)

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -86,83 +85,77 @@ public class Channel {
...
+    private let callback: (ConnectivityState) -> ()
     private var lastState: ConnectivityState
-
+    private let queue: OperationQueue
+    
     var polling: Bool = false {
       didSet {
-        if polling == true && oldValue == false {
-          run()
+        queue.addOperation { [weak self] in
```
MrMage: It looks to me like `polling` will only set to `true` once, right after initializing. So how about calling `self.run()` at the end of `init()`, and make `shutdown` available to `Channel.deinit` instead? In that case, we might even get by without a `polling` variable altogether if you add support for handling shutdown events in the spinloop.

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -59,13 +58,13 @@ public class Channel {
...
-    underlyingChannel = cgrpc_channel_create_secure(address, certificates, host)
-    completionQueue = CompletionQueue(
-      underlyingCompletionQueue: cgrpc_channel_completion_queue(underlyingChannel), name: "Client")
+    underlyingChannel = cgrpc_channel_create_secure(address, certificates, &argumentValues, Int32(arguments.count))
+    completionQueue = CompletionQueue(underlyingCompletionQueue: cgrpc_channel_completion_queue(underlyingChannel), name: "Client")
     completionQueue.run() // start a loop that watches the channel's completion queue
   }
 
   deinit {
+    connectivityObservers.forEach { $0.polling = false }
```
MrMage: Either use a lock for setting `polling`, or use `$0.shutdown()` (see the other comments).

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -86,83 +85,77 @@ public class Channel {
...
-      observers.append(newObserver)
-      observer = newObserver
-    }
-
-    observer?.callbacks.append(callback)
-    observer?.polling = true
+  public func subscribe(callback: @escaping (ConnectivityState) -> ()) {
+    let observer = ConnectivityObserver(underlyingChannel: underlyingChannel, callback: callback)
+    observer.polling = true
+    connectivityObservers.append(observer)
```
MrMage: In theory there could be race conditions here of `subscribe` were to be called from different threads, but I think we can ignore that possibility here.

Sources/SwiftGRPC/Core/Channel.swift
SebastianThiebaud: I might miss something here, but I don't think this loop retains `self`. It's been weakened.

Sources/SwiftGRPC/Core/Channel.swift
SebastianThiebaud: You're right, I missed that one.

Sources/SwiftGRPC/Core/Channel.swift
SebastianThiebaud: Done

Sources/SwiftGRPC/Core/Channel.swift
SebastianThiebaud: Done

SebastianThiebaud:
You're right and I got rid of `polling` variable, which makes things way easier to handle and read.

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -73,7 +73,6 @@ public class Channel {
   }
 
   deinit {
-    connectivityObservers.forEach { $0.polling = false }
```
MrMage: Please keep this line (see below).

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -106,66 +104,57 @@ private extension Channel {
...
-            self.run()
-          } else if self.polling == false && oldValue == true {
-            self.shutdown()
-          }
-        }
-      }
-    }
-
-    init(underlyingChannel: UnsafeMutableRawPointer, callback: @escaping (ConnectivityState) -> ()) {
+    init(underlyingChannel: UnsafeMutableRawPointer, currentState: ConnectivityState, callback: @escaping (ConnectivityState) -> ()) {
```
MrMage: nit: `(ConnectivityState) -> Void`

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -106,66 +104,57 @@ private extension Channel {
...
+            if newState != self.lastState {
+              self.callback(newState)
+            }
+            
+            self.lastState = newState
+          case .queueShutdown:
+            break spinloop
+          case .queueTimeout:
+            continue spinloop
+          default:
```
MrMage: print an error here as well and/or `continue` (see above)?

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -106,66 +104,57 @@ private extension Channel {
...
+      let spinloopThreadQueue = DispatchQueue(label: "SwiftGRPC.ConnectivityObserver.run.spinloopThread")
+
+      spinloopThreadQueue.async { [weak self] in
         guard let `self` = self else { return }
 
-        while self.polling {
-          guard let underlyingState = self.lastState.underlyingState else { return }
+        spinloop: while true  {
+          guard let underlyingState = self.lastState.underlyingState else {
+            print("Couldn't retrieve `underlyingState`")
```
MrMage: nit: I prefer to also include which class printed the error message. How about

    print("ERROR: `Channel.ConnectivityObserver.lastState.underlyingState` nil. This should never happen!")

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -106,66 +104,57 @@ private extension Channel {
...
+            
+            if newState != self.lastState {
+              self.callback(newState)
+            }
+            
+            self.lastState = newState
+          case .queueShutdown:
+            break spinloop
+          case .queueTimeout:
+            continue spinloop
```
MrMage: Could have just `break` without `spinloop` here to exit the switch; up to you.

You might also get rid of the `spinloop: ` loop label by replacing `break spinloop` above with return.

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -106,66 +104,57 @@ private extension Channel {
...
+          case .complete:
+            let newState = ConnectivityState.connectivityState(cgrpc_channel_check_connectivity_state(self.underlyingChannel, 0))
+            
+            if newState != self.lastState {
+              self.callback(newState)
+            }
+            
+            self.lastState = newState
+          case .queueShutdown:
+            break spinloop
```
MrMage: Should we call the observer callback with a `nil` argument once we exit the loop, to indicate that no further calls are expected?

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -106,66 +104,57 @@ private extension Channel {
...
-      shutdown()
+      completionQueue.shutdown()
     }
 
     private func run() {
-      DispatchQueue.global().async { [weak self] in
+      let spinloopThreadQueue = DispatchQueue(label: "SwiftGRPC.ConnectivityObserver.run.spinloopThread")
+
+      spinloopThreadQueue.async { [weak self] in
         guard let `self` = self else { return }
```
MrMage: This is the line that retains `self` by acquiring a strong reference to it. It does not make sense to try to get rid of it; instead avoid the `[weak self] in guard let self = self` altogether and keep the `forEach { $0.shutdown() }` line above. So I think at the moment your code still leaks threads.

To be honest, I'd like to have a test that spins up 100 observers to ensure all their spinloop threads get spun down once the channel closes (i.e. ensure no leaks are happening), but I can understand if you don't want to add one (should be fairly easy, though).

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -94,9 +93,8 @@ public class Channel {
     return ConnectivityState.connectivityState(cgrpc_channel_check_connectivity_state(underlyingChannel, tryToConnect ? 1 : 0))
   }
 
-  public func subscribe(callback: @escaping (ConnectivityState) -> ()) {
-    let observer = ConnectivityObserver(underlyingChannel: underlyingChannel, callback: callback)
-    observer.polling = true
+  public func subscribe(callback: @escaping (ConnectivityState) -> Void) {
+    let observer = ConnectivityObserver(underlyingChannel: underlyingChannel, currentState: connectivityState(), callback: callback)
```
MrMage: ultra-nit: `observer` could be inlined now.

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -106,66 +104,57 @@ private extension Channel {
...
+          case .queueTimeout:
+            continue spinloop
+          default:
+            break spinloop
           }
         }
       }
     }
-    
-    private func shutdown() {
```
MrMage: I would have kept this to call in `Channel.shutdown` (see above)

Sources/SwiftGRPC/Core/Channel.swift
SebastianThiebaud: I'd say no. If I subscribe to get connectivity state changes, I'd assume I will get a callback only when there is a connectivity state change. If the queue gets shut down, that means, somehow the channel has been deallocated and at that point, I would not expect to get a connectivity state change, especially since it's most likely I'm the one initiating this event.

SebastianThiebaud: I also dislike introducing optionals where they're not really needed.

Sources/SwiftGRPC/Core/Channel.swift
SebastianThiebaud: Done.

Sources/SwiftGRPC/Core/Channel.swift
SebastianThiebaud: Done.

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -152,12 +152,13 @@ private extension Channel {
...
-          case .queueShutdown:
-            break spinloop
           case .queueTimeout:
-            continue spinloop
+            continue
+          case .queueShutdown:
+            return
           default:
-            break spinloop
+            print("Event's completion type is `unknown`")
```
MrMage: Please also improve the formulation of this error as above.

SebastianThiebaud: To be honest, I'm not fan of having `print` statements all over the place. Especially in libraries. I'd rather not have these in here. I'm going to remove them and feel free to put them back in another PR if you really want them.

MrMage: Fair enough. For what it's worth, I have a task in #189 to add a proper logging class. At that point, it should become much easier to properly detect and notice such errors (and possibly investigate why they happen).

Sources/SwiftGRPC/Core/Channel.swift
MrMage: FYI, I think this has not been resolved yet (1).

Sources/SwiftGRPC/Core/Channel.swift
MrMage: FYI, I think this has not been resolved yet (2).

Sources/SwiftGRPC/Core/Channel.swift
MrMage: FYI, I think this has not been resolved yet (3).

SebastianThiebaud:
@MrMage Just in case you missed the last commit.

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -124,41 +124,43 @@ private extension Channel {
...
 
     deinit {
-      completionQueue.shutdown()
+      shutdown()
     }
 
     private func run() {
       let spinloopThreadQueue = DispatchQueue(label: "SwiftGRPC.ConnectivityObserver.run.spinloopThread")
 
       spinloopThreadQueue.async { [weak self] in
```
MrMage: No need to use `[weak self]` here, as mentioned before. Shutdowns are initiated by `Channel`.

SebastianThiebaud:
Done :)

PR #199, opened 2018-04-02
Add channel arguments

SebastianThiebaud:
This PR exposes channel arguments through an enum `Arg`. A `Channel` can be created with a list of arguments passed in the initializer as follows:
```swift
let args: [Arg] = [.primaryUserAgent("myApp (iPhone, iOS 11.1)"),
                   .keepAliveTime(10.seconds),
                   .keepAliveTimeout(10.seconds),
                   .reconnectBackoffMin(5.seconds),
                   .reconnectBackoffMax(10.seconds),
                   .http2MinSentPingInterval(10.seconds),
                   .http2MaxPingsWithoutData(0)]

let client = ServiceClient(address: "example.com", certificates: certificates, args: args)
```

I didn't expose all arguments available in the C library at this time.


MrMage:
Thanks for this! I had been wanting to add a setting for backoff soon, and this is much nicer than anything I would have come up with. If you don’t mind I’ll add a few smaller comments tomorrow (holiday here today). Also, do you think it would be possible to try testing this, e.g. at least check that the configured user agent is sent to the server as desired?

SebastianThiebaud:
I've been only testing this against our servers but didn't write any tests for this. I'll think about it.

SebastianThiebaud:
@MrMage Makes total sense, and I agree with that. I've been looking at `Server` and can't find a way to access the metadata (I assume the `user agent` is sent through metadata). Any idea?

MrMage:
I would suppose so; however the generated-code server should log the metadata of each request IIRC, and I haven't seen the user-agent in there. I would suggest experimenting a bit with that and see if there's a good way to get that metadata. You could also try testing that other configuration arguments are working, but I would suppose that checking for the user-agent is still the most straightforward.

Sources/CgRPC/shim/cgrpc.h
```diff
@@ -118,6 +118,31 @@ typedef struct grpc_event {
...
 } grpc_event;
 
+typedef enum grpc_arg_type {
+  GRPC_ARG_STRING,
+  GRPC_ARG_INTEGER,
+  GRPC_ARG_POINTER
+} grpc_arg_type;
+
+typedef struct grpc_arg_pointer_vtable {
+  void* (*copy)(void* p);
```
MrMage: Nit: Would you mind having the asterisk after the space? I.e. ` void *`, not `void* `. (Also below.)

Sources/CgRPC/shim/cgrpc.h
```diff
@@ -126,17 +151,22 @@ void grpc_shutdown(void);
...
 void grpc_completion_queue_destroy(cgrpc_completion_queue *cq);
 
 // helper
 void cgrpc_free_copied_string(char *string);
 
 // channel support
-cgrpc_channel *cgrpc_channel_create(const char *address);
+cgrpc_channel *cgrpc_channel_create(const char *address, 
+                                    grpc_arg *args,
+                                    int number_args);
```
MrMage: nit: `num_args`?

Sources/CgRPC/shim/channel.c
```diff
@@ -23,44 +23,34 @@
 #include <stdlib.h>
 #include <string.h>
 
-cgrpc_channel *cgrpc_channel_create(const char *address) {
+cgrpc_channel *cgrpc_channel_create(const char *address,
+                                    grpc_arg *args,
+                                    int number_args) {
   cgrpc_channel *c = (cgrpc_channel *) malloc(sizeof (cgrpc_channel));
+
+  grpc_channel_args *channel_args = gpr_malloc(sizeof(grpc_channel_args));
```
MrMage: Do we need to deallocate this structure after use, or is ownership transferred to the channel? The existing code only allocates this on the stack, which would indicate that ownership is not transferred.

Sources/CgRPC/shim/channel.c
```diff
@@ -23,44 +23,34 @@
...
-  if (host) {
-    argCount++;
-    arg = &channelArgs->args[1];
-    arg->type = GRPC_ARG_STRING;
-    arg->key = gpr_strdup("grpc.ssl_target_name_override");
-    arg->value.string = gpr_strdup(host);
-  }
-
-  channelArgs->num_args = argCount;
+  grpc_channel_args *channel_args = gpr_malloc(sizeof(grpc_channel_args));
```
MrMage: Same as above. Given that the old version of this one allocates arguments on the stack, I suppose that one of the two either leaked memory or caused undefined behavior.

Sources/SwiftGRPC/Core/Arg.swift
```diff
@@ -0,0 +1,138 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#if SWIFT_PACKAGE
+  import CgRPC
+#endif
+import Foundation // for String.Encoding
+
+public enum Arg {
```
MrMage: Nit: have a more descriptive name, e.g. `ChannelArgument`?

Sources/SwiftGRPC/Core/Arg.swift
```diff
@@ -0,0 +1,138 @@
...
+    case let .http2MaxPingsWithoutData(value):
+      return arg("grpc.http2.max_pings_without_data", value: value)
+    case let .sslTargetNameOverride(value):
+      return arg("grpc.ssl_target_name_override", value: value)
+    }
+  }
+
+  private func arg(_ key: String, value: String) -> grpc_arg {
+    var arg = grpc_arg()
+    arg.key = gpr_strdup(key)
```
MrMage: Who is responsible for releasing the references allocated with gpr_strdup? If it's us, does this leak memory?

Sources/CgRPC/shim/cgrpc.h
```diff
@@ -126,17 +151,22 @@ void grpc_shutdown(void);
...
 void cgrpc_completion_queue_drain(cgrpc_completion_queue *cq);
 void grpc_completion_queue_destroy(cgrpc_completion_queue *cq);
 
 // helper
 void cgrpc_free_copied_string(char *string);
 
 // channel support
-cgrpc_channel *cgrpc_channel_create(const char *address);
+cgrpc_channel *cgrpc_channel_create(const char *address, 
+                                    grpc_arg *args,
```
MrMage: Can this be made `const`? (Might not be the case, but worth checking.)

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -40,12 +40,15 @@ public class Channel {
   ///
   /// - Parameter address: the address of the server to be called
   /// - Parameter secure: if true, use TLS
-  public init(address: String, secure: Bool = true) {
+  /// - Parameter args: list of arguments
+  public init(address: String, secure: Bool = true, args: [Arg] = []) {
     host = address
+    var cargs = args.map({ $0.toCArg() })
```
MrMage: In case the corresponding argument to the C function can be made `const` (see above), we might be albe to change this into `let`.

Also, would you mind using the trailing closure without parentheses?

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -40,12 +40,15 @@ public class Channel {
   ///
   /// - Parameter address: the address of the server to be called
   /// - Parameter secure: if true, use TLS
-  public init(address: String, secure: Bool = true) {
+  /// - Parameter args: list of arguments
```
MrMage: A bit non-descriptive, how about e.g. "list of channel configuration options"? Up to you. (Also consider renaming the parameter.)

Tests/SwiftGRPCTests/BasicEchoTestCase.swift
```diff
@@ -54,7 +54,7 @@ class BasicEchoTestCase: XCTestCase {
                                keyString: String(data: keyForTests, encoding: .utf8)!,
                                provider: provider)
       server.start(queue: DispatchQueue.global())
-      client = Echo_EchoServiceClient(address: address, certificates: certificateString, host: "example.com")
+      client = Echo_EchoServiceClient(address: address, certificates: certificateString, args: [Arg.sslTargetNameOverride("example.com")])
```
MrMage: micro-nit: I think you can replace the `Arg.` with `.` (also below).

Sources/SwiftGRPC/Core/Arg.swift
```diff
@@ -0,0 +1,138 @@
...
+    case let .http2MinSentPingInterval(value):
+      return arg("grpc.http2.min_time_between_pings_ms", value: value * 1_000)
+    case let .http2MaxPingsWithoutData(value):
+      return arg("grpc.http2.max_pings_without_data", value: value)
+    case let .sslTargetNameOverride(value):
+      return arg("grpc.ssl_target_name_override", value: value)
+    }
+  }
+
+  private func arg(_ key: String, value: String) -> grpc_arg {
```
MrMage: Nit: call these e.g. `makeArg` or `makeArgument` instead and maybe make them free functions (still in the same file) instead, given that they don't explicitly refer to `self`? (Up to you.)

Sources/SwiftGRPC/Core/Arg.swift
```diff
@@ -0,0 +1,138 @@
+/*
+ * Copyright 2016, gRPC Authors All rights reserved.
```
MrMage: 2018?

Sources/CgRPC/shim/cgrpc.h
SebastianThiebaud: No it can't: 
```
typedef struct {
  size_t num_args;
  grpc_arg *args;
} grpc_channel_args;
```

MrMage: I see, thanks! Was worth a check ;-)

MrMage:
Thank you for the changes, much appreciated! Apart from the question whether we leak memory by using `gpr_strdup` in `ChannelArgument.swift`, everything looks good.

Given that `channel_args` now seems to only be used temporarily, you could probably also replace

    grpc_channel_args *channel_args = gpr_malloc(sizeof(grpc_channel_args));

with

    grpc_channel_args channel_args;

and remove the corresponding `gpr_free` call to save a `malloc`, but I would leave that up to you :-)

timburks:
@SebastianThiebaud Thanks for this! This looks like a great use of enums. I notice that the Echo test has a build failure related to the new Arg type. Could you address that before we merge this?

@MrMage, would you LGTM when you think this is ready to merge? Thanks!!

SebastianThiebaud:
@MrMage I don't think we're having a leak, but I haven't been writing C code for over a decade now, so I feel a little bit rusty :) 
@timburks My last push should solve that.

MrMage:
@SebastianThiebaud `grp_strdup` simply allocates a `char *` buffer on the heap and copies the current string to that buffer. As far as I can tell, `grpc_channel_create` copies those strings again, without freeing the input, so we would be responsible for that. So please make sure to `grp_free` all string buffers created with `grp_strdup` after use (and document this in `Channel.Argument.toCArg()`.

Also, were you able to add tests for this? Those would e.g. help us detect if the procedure above would actually end up over-releasing the strings, causing a crash.

SebastianThiebaud:
@MrMage When freeing those strings, I get random failures in my tests that I just wrote. Something tells me that we shouldn't free these strings. Thoughts?

SebastianThiebaud:
Also not 100% sure why `ServiceClientTests` fails.

Tests/SwiftGRPCTests/ChannelArgumentTests.swift
```diff
@@ -0,0 +1,48 @@
...
+import CgRPC
+#endif
+import Foundation
+@testable import SwiftGRPC
+import XCTest
+
+class ChannelArgumentTests: XCTestCase {
+  func testArgumentKey() {
+    let argument: Channel.Argument = .defaultAuthority("default")
+    XCTAssertEqual(String(cString: argument.toCArg().key), "grpc.default_authority")
```
MrMage: Pretty sure that these (and below) leak memory, though, but OTOH they are only one-of tests. @timburks, what do you think?

MrMage:
Travis provides these error messages:

	/Users/travis/build/grpc/grpc-swift/Tests/SwiftGRPCTests/ServiceClientTests.swift:33:43: error: incorrect argument label in call (have 'channel:', expected 'address:')
	      let client1 = Echo_EchoServiceClient(channel: sharedChannel)
	                                          ^~~~~~~~
	                                           address
	/Users/travis/build/grpc/grpc-swift/Tests/SwiftGRPCTests/ServiceClientTests.swift:39:43: error: incorrect argument label in call (have 'channel:', expected 'address:')
	      let client2 = Echo_EchoServiceClient(channel: sharedChannel)
	                                          ^~~~~~~~
	                                           address
	/Users/travis/build/grpc/grpc-swift/Tests/SwiftGRPCTests/ServiceClientTests.swift:53:43: error: incorrect argument label in call (have 'channel:', expected 'address:')
	      let client1 = Echo_EchoServiceClient(channel: sharedChannel)
	                                          ^~~~~~~~
	                                           address
	/Users/travis/build/grpc/grpc-swift/Tests/SwiftGRPCTests/ServiceClientTests.swift:61:43: error: incorrect argument label in call (have 'channel:', expected 'address:')
	      let client2 = Echo_EchoServiceClient(channel: sharedChannel)
	                                          ^~~~~~~~
	                                           address
	/Users/travis/build/grpc/grpc-swift/Tests/SwiftGRPCTests/ServiceClientTests.swift:54:23: warning: no calls to throwing functions occur within 'try' expression
	      let response1 = try client1.get(Echo_EchoRequest(text: "foo")).text
	                      ^
	/Users/travis/build/grpc/grpc-swift/Tests/SwiftGRPCTests/ServiceClientTests.swift:62:23: warning: no calls to throwing functions occur within 'try' expression
	      let response2 = try client2.get(Echo_EchoRequest(text: "foo")).text
	                      ^
	error: terminated(1): /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swift-build-tool -f /Users/travis/build/grpc/grpc-swift/.build/debug.yaml test -v

Doesn't seem related to your changes, but if you don't mind, could you fix those so we can get a full CI run for your code?

I would have preferred to also have tests that verify that passing at least some of these arguments to a channel works functionally, but I'm fine with getting this merged as-is once CI passes.

However, I would appreciate if you could provide _how_ the tests fail if you try to release the variables, and provide details on _how_ you released them.

SebastianThiebaud:
@MrMage I've looked at the failing tests, but can't figure out what changed / broke and how to fix them.
I agree for the tests for channel, but I'm not aware of any way to get the arguments after the channel creation (the C lib isn't exposing this afaik?)

MrMage:
@SebastianThiebaud did you merge this with the current master branch?

In terms of tests, I was thinking more of "integration tests", eg. checking that passing a user agent argument to the call does affect the user agent that gets sent to the server. You should be able to read the user agent from `handler.caller` if I'm not mistaken.

SebastianThiebaud:
Yes, this branch has been rebased with the current master branch and is up to date. I'm struggling to pinpoint the root cause of these failing tests.

Looking at the integration tests tonight and will update you.

MrMage:
@SebastianThiebaud what are your plans to update this? Or should I try my luck at the remaining issues?

SebastianThiebaud:
@MrMage Please do. I've been trying to come up with useful tests but didn't succeed.

Issue #200, opened 2018-04-04
Support for Channel specification on ServiceClient.init


rebello95:
Python's gRPC implementation allows for consumers to [specify a channel](https://github.com/grpc/grpc/blob/7542ba6e0caee740f124dc25706287fd62f546d8/examples/python/multiplex/multiplex_client.py#L110) when creating clients, like so:
```
def run():
    channel = grpc.insecure_channel('localhost:50051')
    my_stub = helloworld_pb2_grpc.FirstStub(channel)
    my_other_stub = helloworld_pb2_grpc.SomeOtherStub(channel)
    ...
````

The Objective-C implementation uses a [singleton channel](https://github.com/grpc/grpc/blob/master/src/objective-c/GRPCClient/private/GRPCHost.m#L115), where the current Swift implementation [creates a new one](https://github.com/grpc/grpc-swift/blob/master/Sources/SwiftGRPC/Runtime/ServiceClient.swift#L54) every time.

Ideally, consumers would be able to specify their own `Channel` between multiple clients, which would alleviate the need for creating a new one each time:
```
/// Create a client with a pre-defined channel.
public init(channel: Channel) {
    gRPC.initialize()
    self.channel = channel
    self.metadata = Metadata()
}
```

rebello95:
Makes sense. Do you foresee any complications with the underlying C channel management if we were to implement an initializer like the above (i.e., regarding managing connectivity on a long-living/reused channel)? I have some local changes that seem to work as long as `gRPC` is initialized by the time the `Channel` is created.

MrMage:
I think this should be fine; we could just have `Channel.init()` call `gRPC.initialize()` as well.

rebello95:
Yep, sounds good! Opened https://github.com/grpc/grpc-swift/pull/202

PR #202, opened 2018-04-05
Support for Channel specification on ServiceClient

rebello95:
Adds support to allow consumers to specify a `Channel` when creating `ServiceClient` instances.

This is something already supported by other gRPC language flavors (like [Python](https://github.com/grpc/grpc/blob/7542ba6e0caee740f124dc25706287fd62f546d8/examples/python/multiplex/multiplex_client.py#L110)), and alleviates the need to create a new channel every time a call is made.

Usage:
```swift
let sharedChannel = Channel(address: "https://api.myservice.com", secure: true)
HelloWorldClient(channel: sharedChannel).getHelloWorld(...)
OtherClient(channel: sharedChannel).getOther(...)
```

Resolves https://github.com/grpc/grpc-swift/issues/200

### DIFF
--- a/Sources/SwiftGRPC/Runtime/ServiceClient.swift
+++ b/Sources/SwiftGRPC/Runtime/ServiceClient.swift
@@ -55,6 +55,13 @@ open class ServiceClientBase: ServiceClient {
     metadata = Metadata()
   }
 
+  /// Create a client using a pre-defined channel.
+  public init(channel: Channel) {
+    gRPC.initialize()
+    self.channel = channel
+    self.metadata = Metadata()
+  }
+
   /// Create a client that makes secure connections with a custom certificate and (optional) hostname.
   public init(address: String, certificates: String, host: String?) {
     gRPC.initialize()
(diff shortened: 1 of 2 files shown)


rebello95:
@MrMage if you'd like to take a look 😀 

Thought about adding additional tests for this as well, but it looks like reusing channels is already covered [here](https://github.com/grpc/grpc-swift/blob/master/Tests/SwiftGRPCTests/GRPCTests.swift#L156). Let me know if you think there's something I should add!

Sources/SwiftGRPC/Runtime/ServiceClient.swift
```diff
@@ -55,6 +55,13 @@ open class ServiceClientBase: ServiceClient {
     metadata = Metadata()
   }
 
+  /// Create a client using a pre-defined channel.
+  public init(channel: Channel) {
+    gRPC.initialize()
```
MrMage: I think we can remove this `gRPC.initialize()` call, given that this is now done by `Channel.init`. Also in the other constructors above and below.

Also, I'd suggest moving this constructor below or above all the other ones?

timburks:
LGTM.

Thanks @rebello95 for adding this, and @MrMage for reviewing! I agree that future tests would be important for this, but will merge this now.

rebello95:
Thank you for the feedback @MrMage and @timburks! As this is already merged, I went ahead and wrote some tests in https://github.com/grpc/grpc-swift/pull/203 that should cover it

PR #203, opened 2018-04-06
Add ServiceClient tests for shared channels

rebello95:
Adds tests for support added in https://github.com/grpc/grpc-swift/pull/202 that allows `ServiceClient` instances to share a single `Channel`.


timburks:
Thanks!

PR #207, opened 2018-04-12
Fix SwiftGRPC.podspec

### LINKED ISSUE #11: Cocoapods build configuration (Podspec is broken)
I've put an initial podspec in `SwiftGRPC.podspec` and put a reference to it in `Examples/Speech/Podfile`. Attempting to build it has a few problems:

1. `#import "internal.h"` needs to be removed from the automatically-generated `SwiftGRPC-umbrella.h`. 
2. The CgRPC code refers to `grpc_slice` and the latest tagged version of gRPC-Core uses the older `gpr_slice`.

### LINKED ISSUE #45: Support for pod.
When grpc-swift support for pod? 
master branch podspec name have "incomplete".

### LINKED ISSUE #111: Cocoapods Support?

bootstraponline:
The existing podspec is broken with `inttypes.h` errors. Updating the dependencies resolves the issue.

Fix #111
Fix #11 
Fix #45

### DIFF
--- a/SwiftGRPC.podspec
+++ b/SwiftGRPC.podspec
@@ -45,7 +45,7 @@ Pod::Spec.new do |s|
   s.source_files = 'Sources/SwiftGRPC/*.swift', 'Sources/SwiftGRPC/**/*.swift', 'Sources/CgRPC/shim/*.[ch]'
   s.public_header_files = 'Sources/CgRPC/shim/cgrpc.h'
 
-  s.dependency 'gRPC-Core', '~> 1.9.1'
-  s.dependency 'BoringSSL', '~> 9.1'
-  s.dependency 'SwiftProtobuf', '~> 1.0.2'
+  s.dependency 'gRPC-Core', '~> 1.11.0-pre2'
+  s.dependency 'BoringSSL', '~> 10.0'
+  s.dependency 'SwiftProtobuf', '~> 1.0.3'
 end


muxi:
Would it work to change the version string to increase with gRPC core version? Something like `~> 1.11`?

bootstraponline:
> Would it work to change the version string to increase with gRPC core version? Something like ~> 1.11?

I think we should update it once 1.11 is officially released.

muxi:
My point is to let it sync with newest minor version of gRPC core if there is not a particular reason that prevents grpc-swift from doing this.

MrMage:
Makes sense to wait for gRPC-Core 1.11 before tagging. I would not tie SwiftGRPC releases to gRPC-Core ones, though, as development of SwiftGRPC is mostly independent of gRPC-Core.

MrMage:
Allowing minor gRPC updates sounds sensible, as long as those don't break anything (which so gar they don't seem to).

bootstraponline:
> My point is to let it sync with newest minor version of gRPC core if there is not a particular reason that prevents grpc-swift from doing this.

We need to be shipping a podspec that compiles. Here's what happens when `~> 1.11` is used:

```
[!] CocoaPods could not find compatible versions for pod "gRPC-Core":
  In Podfile:
    SwiftGRPC (from `https://github.com/bootstraponline/grpc-swift.git`) was resolved to 0.3.3, which depends on
      gRPC-Core (~> 1.11)

None of your spec sources contain a spec satisfying the dependency: `gRPC-Core (~> 1.11)`.

You have either:
 * out-of-date source repos which you can update with `pod repo update` or with `pod install --repo-update`.
 * mistyped the name or version.
 * not added the source repo that hosts the Podspec to your Podfile.

Note: as of CocoaPods 1.0, `pod repo update` does not happen on `pod install` by default.
```

As I mentioned, we'd have to wait for the official 1.11 release to use that version range.

MrMage:
Agreed. I'd be fine with committing this as-is now, with a follow-up PR once 1.11 final is out, and then tagging SwiftGRPC 0.4.1.

muxi:
Yea that sounds good.

bootstraponline:
Agreed.

PR #217, opened 2018-04-21
Fix a few memory leaks related to unreleased slices in observers and metadata arrays

MrMage:


Sources/SwiftGRPC/Core/Metadata.swift
```diff
@@ -20,25 +20,42 @@ import Foundation // for String.Encoding
...
+    self.ownsFields = ownsFields
   }
 
   public init() {
     underlyingArray = cgrpc_metadata_array_create()
+    ownsFields = true
   }
 
-  public init(_ pairs: [String: String]) {
+  public init(_ pairs: [String: String]) throws {
```
timburks: If I read this correctly, it seems that init throws only because it calls add(), which throws an error that will never occur in practice (because ownsFields is true). Did I miss another throw or do you expect add() to throw other errors in the future? If not, it seems better to use try! add and not throw from init.

MrMage: I had this before (see https://github.com/grpc/grpc-swift/pull/217/commits/7b4f47d91dd2980f13c146979760badf43045e37), but decided that it is safer not to rely on `add` not failing in that context. It is guaranteed to not fail _now_, but that might change if the implementation of `add` gets changes. And I think most clients won't have a problem adding the `try` statement.

timburks: SGTM.

Issue #220, opened 2018-04-26
`swift-nio-zlib-support` pulls in incompatible version of `ncurses`


tsabirgaliev:
When building on a Mac with Macports, `pkgConfig: "zlib"` in `swift-nio-zlib-support` emits `-I/opt/local/include` flag, because **Macports comes with its own `pkg-config`**.  It potentially breaks the build, because `/opt/local/include` may contain incompatible versions of libraries referred from PlatformSDK.  For example: https://github.com/grpc/grpc-swift/issues/205#issuecomment-384274112

@MrMage probably this commit: https://github.com/grpc/grpc-swift/commit/263e66bad2e6db7853e719d2f271fed0952461ab broke the build

MrMage:
Given that the current implementation does build on clean systems, what would you suggest to fix this on systems with MacPorts installed?

moldabekov:
@MrMage I've opened issue in Apple's repo https://github.com/apple/swift-nio-zlib-support/issues/2.

There are three ways for fellow devs to fix it by themselves described here: https://github.com/grpc/grpc-swift/issues/205#issuecomment-384666911. All of them aren't elegant.

While Apple thinks about the fix, it'd be nice to vendor `swift-nio-zlib-support` by ownself. Still it's a simple wrapper with a few lines `Package.swift` file. This seems to me the most convenient way to fix conflict with MacPorts.

MrMage:
I'm not particularly excited about vendoring that `Package.swift`, but I wouldn't object to it either, especially if it's only temporary.

From my side feel free to prepare a pull request, but the final call will have to be made by @timburks of course.

tsabirgaliev:
@MrMage what about something like this:

```swift
#if os(macOS)
let dependencies: [Package.Dependency] = [
    .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.0.2"),
    .package(url: "https://github.com/kylef/Commander.git", from: "0.8.0")
  ]
#else
let dependencies: [Package.Dependency] = [
    .package(url: "https://github.com/apple/swift-nio-zlib-support.git", from: "1.0.0"),
    .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.0.2"),
    .package(url: "https://github.com/kylef/Commander.git", from: "0.8.0")
  ]
#endif

let package = Package(
  name: "SwiftGRPC",
  ...
  dependencies: dependencies,
```

tsabirgaliev:
[...]

PR #221, opened 2018-04-27
Leave `zlib` dependency only for non-Apple platforms

tsabirgaliev:
because Apple already ships with its own.

This somewhat fragile fix allows building grpc-swift when Macports is present on a system and ships its own `zlib`.
Without this hack pkgConfig emits `-I/opt/local/include` and brings various header version incompatibilities,
because Macports (unlike Homebrew) doesn't offer scoped header includes. Further here:
https://github.com/apple/swift-nio-zlib-support/issues/2#issuecomment-384681975

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -17,16 +17,21 @@
  */
 import PackageDescription
 
+var dependencies: [Package.Dependency] = [
+  .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.0.2"),
+  .package(url: "https://github.com/kylef/Commander.git", from: "0.8.0")
+]
+
+#if !os(macOS)
+dependencies.insert(.package(url: "https://github.com/apple/swift-nio-zlib-support.git", from: "1.0.0"), at: 0)
+#endif
+
 let package = Package(
   name: "SwiftGRPC",
   products: [
     .library(name: "SwiftGRPC", targets: ["SwiftGRPC"]),
   ],
-  dependencies: [
-    .package(url: "https://github.com/apple/swift-nio-zlib-support.git", from: "1.0.0"),
-    .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.0.2"),
-    .package(url: "https://github.com/kylef/Commander.git", from: "0.8.0")
-  ],
+  dependencies: dependencies,
   targets: [
     .target(name: "SwiftGRPC",
             dependencies: ["CgRPC", "SwiftProtobuf"]),


tsabirgaliev:
This solves issue https://github.com/grpc/grpc-swift/issues/220

tsabirgaliev:
Someone without Macports and/or Homebrew please test/review this.

Package.swift
```diff
@@ -17,16 +17,21 @@
  */
 import PackageDescription
 
+var dependencies: [Package.Dependency] = [
+  .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.0.2"),
+  .package(url: "https://github.com/kylef/Commander.git", from: "0.8.0")
+]
+
+#if !os(macOS)
+dependencies.insert(.package(url: "https://github.com/apple/swift-nio-zlib-support.git", from: "1.0.0"), at: 0)
```
MrMage: How about using .append instead?

MrMage: Also, please add a brief comment explaining why this is necessary, possibly with links to the corresponding issues.

timburks:
I just noticed that builds fail on Ubuntu 16.04 (using directions [here](https://github.com/grpc/grpc-swift/wiki/Try-gRPC-Swift-in-a-Google-Compute-Engine-VM)).

```
/home/timburks/grpc-swift/Sources/CgRPC/src/core/lib/compression/message_compress.c:123: error: undefined reference to 'inflateInit2_'
/home/timburks/grpc-swift/Sources/CgRPC/src/core/lib/compression/message_compress.c:124: error: undefined reference to 'inflate'
/home/timburks/grpc-swift/Sources/CgRPC/src/core/lib/compression/message_compress.c:133: error: undefined reference to 'inflateEnd'
/home/timburks/grpc-swift/Sources/CgRPC/src/core/lib/compression/message_compress.c:97: error: undefined reference to 'deflateInit2_'
/home/timburks/grpc-swift/Sources/CgRPC/src/core/lib/compression/message_compress.c:99: error: undefined reference to 'deflate'
/home/timburks/grpc-swift/Sources/CgRPC/src/core/lib/compression/message_compress.c:109: error: undefined reference to 'deflateEnd'
/home/timburks/grpc-swift/Sources/CgRPC/src/core/lib/compression/stream_compression_gzip.c:143: error: undefined reference to 'deflate'
/home/timburks/grpc-swift/Sources/CgRPC/src/core/lib/compression/stream_compression_gzip.c:173: error: undefined reference to 'inflate'
/home/timburks/grpc-swift/Sources/CgRPC/src/core/lib/compression/stream_compression_gzip.c:193: error: undefined reference to 'inflateInit2_'
/home/timburks/grpc-swift/Sources/CgRPC/src/core/lib/compression/stream_compression_gzip.c:193: error: undefined reference to 'inflate'
/home/timburks/grpc-swift/Sources/CgRPC/src/core/lib/compression/stream_compression_gzip.c:196: error: undefined reference to 'deflateInit2_'
/home/timburks/grpc-swift/Sources/CgRPC/src/core/lib/compression/stream_compression_gzip.c:196: error: undefined reference to 'deflate'
/home/timburks/grpc-swift/Sources/CgRPC/src/core/lib/compression/stream_compression_gzip.c:212: error: undefined reference to 'inflate'
/home/timburks/grpc-swift/Sources/CgRPC/src/core/lib/compression/stream_compression_gzip.c:217: error: undefined reference to 'inflateEnd'
/home/timburks/grpc-swift/Sources/CgRPC/src/core/lib/compression/stream_compression_gzip.c:219: error: undefined reference to 'deflateEnd'
```

Adding `-Xlinker -lz` to `swift build` in the Makefile fixes the build. I guess that's related to this change. I'll investigate more before suggesting any changes.

MrMage:
FYI, the build also fails for me using the generated project unless I comment out the `#if !os(macOS)` condition. That might be my personal setup (a gazillion Homebrew packages), though, given that it works on a "clean" Mac.

Alternatively, maybe there's a way to add that linker option to Package.swift directly?

tsabirgaliev:
I admit, this patch is very fragile. If there is a way to externalize the zlib dependency configuration, that would be more robust IMO

tsabirgaliev:
My main concern is that `swift build` is not a mature enough tool to manage that

MrMage:
I checked, and I did indeed not find a way to pass extra arguments via Package.swift. Damn.

Maybe we could augment [`fix-indentation-settings.rb`](https://github.com/grpc/grpc-swift/blob/master/fix-indentation-settings.rb) to also add these linker flags?

PR #224, opened 2018-05-02
Update to cocoapods example

jbolinger:
A minor update to the example to eliminate an extra step (also added a .gitignore). 

Note that I did this to see what it would take to generate the client as a new pod, but that may not be possible right now due to the visibility of the generated code. It might be useful to consider if you want to use Cocoapods as a means to distribute the tools/scripts for generating things.


timburks:
Thanks for this! The plugin takes a Visibility option that can be used to make generated interfaces public... but now that I look for a description of that option in the READMEs, it seems we lost it in a reorganization. It's the same option that's used by [swift-protobuf](https://github.com/apple/swift-protobuf/blob/master/Documentation/PLUGIN.md).

PR #225, opened 2018-05-03
Finalize adding channel arguments

MrMage:
This is heavily based on #199 by @SebastianThiebaud. I have just cleaned up a few memory leaks and added an "integration test" that ensures that the primary and secondary user agent arguments are respected.


timburks:
@SebastianThiebaud @MrMage Thanks for adding this! Channel.Arguments looks like a great use of enums.

PR #228, opened 2018-05-05
Streamline the interface of the server-side response handling code

MrMage:
Based on #227.

In the default (successful) case, most request handlers can now simply return `.ok` and have `ServiceServer` handle sending back the status for them. This also streamlines the `run()` implementations of the `ServerSession` subclasses.


MrMage:
Looks like a build system flake to me. Rerunning CI.

PR #230, opened 2018-05-12
Add server/client codegen options to protoc

rebello95:
Implements support for specifying additional options for `protoc` gRPC codegen to exclude client and/or server code.

Usage:
```
protoc TestEvent.proto --swiftgrpc_out=Client=true,Server=false:.
```

Also updated the readme with examples of all supported parameters.


rebello95:
For sure! I had considered some alternatives - personally I think it'd be really nice to be able to do `protoc MyProto.proto --swiftgrpc_out=. --outputs=client,server --visibility=internal` but that's outside the realm of `protoc`'s `Google_Protobuf_Compiler_CodeGeneratorRequest` as far as I know 🤷‍♂️

PR #232, opened 2018-05-16
Bump podspec to 0.4.3

rebello95:
A new release was tagged last night, so updating the podspec accordingly.


rebello95:
Spec has been bumped on master.

PR #233, opened 2018-05-22
Add timeout argument to makeCall

KyoheiG3:
Should be able to set a timeout when making a call as written in the comment.
```swift
/// - Parameter timeout: a timeout value in seconds
```
Thanks.


KyoheiG3:
Thank you for your reply. I have already made it.

MrMage:
Yes, but this is not for use with the generated code, right?

KyoheiG3:
Yes, right.

timburks:
Thanks!

PR #239, opened 2018-05-29
Update CgRPC to v1.12.0 and BoringSSL to 10.0.4 (I guess?)

MrMage:
This change requires Swift 4.1 and Ruby (with the 'xcodeproj' gem) installed to compile if building via SPM.

To cut through the noise, here are the non-vendored files with changes:

- `Makefile` (added a `-DPB_FIELD_16BIT=1` flag)
- `Package.swift` (added C/C++ language versions)
- `SwiftGRPC.podspec` (updated the Podspec to use gRPC 1.12 as well)
- `fix-project-settings.rb` (added a `-DPB_FIELD_16BIT=1` flag)
- `vendor-boringssl.sh` (updated some paths, compiled some new files)
- `.travis-install.sh` (updated to Swift 4.1)

For unknown reasons, the `Makefile` needed an extra `-DPB_FIELD_16BIT=1` flag, while `SwiftGRPC.podspec` didn't.

While at it, I have also re-run RootsEncoder.

Thanks to @SebastianThiebaud for https://github.com/postmates/grpc-swift/blob/carthage_structure/c%2B%2B-compile-settings.rb, which I blatantly copied.


MrMage:
Note: This also reverts #221, which causes the build to fail with Xcode 9.3, both locally and on CI. @tsabirgaliev, I think we'll need to find a different solution for you there, possibly by manually patching the Xcode project as seen in `fix-project-settings.rb` (in this PR), or by having you maintain a custom branch.

MrMage:
@timburks as an alternative to adding all these compiler flags (`-DTSI_OPENSSL_ALPN_SUPPORT=0 -DPB_FIELD_16BIT=1`) to the `Makefile` and having all SPM-based consumers of the library add them to their `swift build` command, what would you think of having `vendor-grpc.sh` automatically insert them as macros near the top of https://github.com/grpc/grpc-swift/blob/master/Sources/CgRPC/include/grpc/grpc.h? That would save us from having a `swift build` line that is getting longer and longer.

timburks:
@MrMage Yes, I think having `vendor-grpc.sh` "fix up" the header is a good idea here; at least it's consistent with what `vendor-boringssl.sh` does to fix BoringSSL build problems.

tsabirgaliev:
@MrMage 
> This also reverts #221, which causes the build to fail with Xcode 9.3, both locally and on CI.

Yes, my patch is very disturbing even for me, so I ended up dedicating a clean macos VM to work with grpc

MrMage:
> @MrMage Yes, I think having vendor-grpc.sh "fix up" the header is a good idea here; at least it's consistent with what vendor-boringssl.sh does to fix BoringSSL build problems.

@timburks done — please take another look.

timburks:
@MrMage It looks like I un-reverted #221 when merging. If you could take a look and and fix the OS X build, I'll merge this first thing Monday (Pacific Time).

MrMage:
@timburks done — PTAL.

tsabirgaliev:
@MrMage BTW, when I `make project` and build the project from within Xcode, it fails with zlib link errors. _This is on a clean Mac without Macports or Homebrew_. I added `-Xlinker -lz` to `CFLAGS` and now it builds in CLI and Xcode.

Is there a particular reason against this?

MrMage:
@tsabirgaliev did you building try from this branch? I think this branch should be working.

tsabirgaliev:
@MrMage this branch fails with the same linker error:

error

I managed to fix this multiple ways:
1. add `libz` to CgRPC in project settings: 
fix1

2. add `-lz` to linker flags here:
fix2

3. add `-Xlinker -lz` to `CFLAGS` and make `make project` respect these flags:
fix3


The 1 and 2 require manual work, 3 is better with a cons that the linker flag is set globally

MrMage:
@tsabirgaliev please check if https://github.com/grpc/grpc-swift/pull/246 compiles out of the box for you :-)

PR #244, opened 2018-06-01
Add support for TLS mutual auth

### LINKED ISSUE #212: Unable to configure tls mutual auth with current API
I have three artifacts that our gRPC java client uses

- ca.pem      -> our ca
- client.pem -> client cert
- client.key   -> key

However, the current API doesn't really communicate how to use these entities in concert to make a secure connection.

Using only the client cert, I receive:

> OPENSSL_internal:CERTIFICATE_VERIFY_FAILED

However, I can run:

> openssl s_client -CAfile /certs/ca.pem -cert /certs/client.pem -key /certs/client.key -debug -connect localhost:1408

and properly establish a connection to the remote endpoint.

So, what am I missing in order to get this to properly work with grpc-swift?

n8armstrong:
This fixes #212.


Sources/CgRPC/shim/channel.c
```diff
@@ -47,7 +49,14 @@ cgrpc_channel *cgrpc_channel_create_secure(const char *address,
   channel_args.args = args;
   channel_args.num_args = num_args;
 
-  grpc_channel_credentials *creds = grpc_ssl_credentials_create(pem_root_certs, NULL, NULL);
+  grpc_ssl_pem_key_cert_pair *client_credentials = malloc(sizeof (struct grpc_ssl_pem_key_cert_pair));
+  client_credentials->cert_chain = client_certs;
+  client_credentials->private_key = client_private_key;
+  if (client_certs == NULL || client_private_key == NULL) {
+      client_credentials = NULL;
+  }
```
n8armstrong: Is there a more elegant way to do this? Also, do I need to `free(client_credentials)` after setting it to `NULL`?

bootstraponline:
This is the error we're getting:

> Thread 1: Fatal error: 'try!' expression unexpectedly raised an error: SwiftGRPC.RPCError.callError(successful, status unavailable: Connect Failed
initialMetadata: [:]
trailingMetadata: [:])

```swift
let bundle = Bundle(identifier: "com.instructure.SoSeedySwift")!
let trustCertCollectionFilePath = bundle.path(forResource: "ca", ofType: "crt")!
let trustCertCollectionString = try! String(contentsOfFile: trustCertCollectionFilePath)

let clientCertChainFilePath = bundle.path(forResource: "client", ofType: "crt")!
let clientCertChainString = try! String(contentsOfFile: clientCertChainFilePath)

let clientPrivateKeyFilePath = bundle.path(forResource: "client", ofType: "pem")!
let clientPrivateKeyString = try! String(contentsOfFile: clientPrivateKeyFilePath)
let address = "localhost:50051"

let generalClient = Soseedy_SeedyGeneralServiceClient(
  address: address,
  certificates: trustCertCollectionString,
  clientCertificates: clientCertChainString,
  clientKey: clientPrivateKeyString)
```

The same certs work fine on the Java client.


```kotlin
val host = "localhost"
val port = 50051
val trustCertCollectionFilePath = Certs.trustCertCollectionFile.toString()
val clientCertChainFilePath = Certs.clientCertChainFile.toString()
val clientPrivateKeyFilePath = Certs.clientPrivateKeyFile.toString()

val client = MutualAuthClientTls(host, port,
            buildSslContext(trustCertCollectionFilePath, clientCertChainFilePath, clientPrivateKeyFilePath))
```

bootstraponline:
lldb output:

```
(lldb) expr -- generalClient.getHealthCheck(Soseedy_HealthCheckRequest())
2018-06-02 10:26:27.124518-0400 Teacher[14279:70976] Task <4FD4659C-8F87-4EBB-B0E1-2104342DBA10>.<17> finished with error - code: -1001
2018-06-02 10:26:27.124655-0400 Teacher[14279:70996] [BoringSSL] Function boringssl_session_errorlog: line 2881 [boringssl_session_read] SSL_ERROR_ZERO_RETURN(6): operation failed because the connection was cleanly shut down with a close_notify alert
2018-06-02 10:26:27.125019-0400 Teacher[14279:70996] [BoringSSL] Function boringssl_session_errorlog: line 2881 [boringssl_session_read] SSL_ERROR_ZERO_RETURN(6): operation failed because the connection was cleanly shut down with a close_notify alert
2018-06-02 10:26:27.126495-0400 Teacher[14279:70996] [BoringSSL] Function boringssl_session_errorlog: line 2881 [boringssl_session_read] SSL_ERROR_ZERO_RETURN(6): operation failed because the connection was cleanly shut down with a close_notify alert
2018-06-02 10:26:27.126976-0400 Teacher[14279:70996] [BoringSSL] Function boringssl_session_errorlog: line 2881 [boringssl_session_read] SSL_ERROR_ZERO_RETURN(6): operation failed because the connection was cleanly shut down with a close_notify alert
2018-06-02 10:26:27.127476-0400 Teacher[14279:70996] [BoringSSL] Function boringssl_session_errorlog: line 2881 [boringssl_session_read] SSL_ERROR_ZERO_RETURN(6): operation failed because the connection was cleanly shut down with a close_notify alert
E0602 10:26:27.160424000 123145504768000 ssl_transport_security.cc:1063] Handshake failed with fatal error SSL_ERROR_SSL: error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED.
(SwiftGRPC.RPCError) $E0 = callError {
  callError = {
    success = true
    statusCode = unavailable
    statusMessage = "Connect Failed"
    resultData = nil
    initialMetadata = 0x000060c00023ef80 {
      underlyingArray = (_rawValue = 0x000060c000424200)
      ownsFields = false
    }
    trailingMetadata = 0x000060c0004246c0 {
      underlyingArray = (_rawValue = 0x000060c0004246a0)
      ownsFields = false
    }
  }
}
```

> ssl_transport_security.cc:1063] Handshake failed with fatal error SSL_ERROR_SSL: error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED.

The same certs work with a Java client. I think they're not being passed through Swift/CgRPC properly.

MrMage:
Hmm, the general logic looks sound to me. Are the certificate files in PEM format?

You could also try running the server and/or the client with gRPC verbosity enabled (if you aren't already): https://github.com/grpc/grpc/blob/master/doc/environment_variables.md

Will add a few code review comments.

Also, do you think it will be possible to add a test for this once you get it running?

Sources/CgRPC/shim/channel.c
MrMage: This is indeed leaking memory. Please look at https://github.com/grpc/grpc-swift/blob/38b66c346c007cbf52b7304a9ceddf907217d9a2/Sources/CgRPC/shim/server.c#L42 on how this should be done. You could have something like the following:

```
grpc_ssl_pem_key_cert_pair client_credentials;
grpc_ssl_pem_key_cert_pair *client_credentials_pointer = NULL;
if (client_certs != NULL && client_private_key != NULL) {
  client_credentials.cert_chain = client_certs;
  client_credentials.private_key = client_private_key;
  client_credentials_pointer = &client_credentials;
}
grpc_channel_credentials *creds = grpc_ssl_credentials_create(pem_root_certs, client_credentials_pointer, NULL);
```

MrMage: Also, please use 2 spaces for indentation. (Run `make project` to change this in your xcodeproj.)

MrMage:
To me, it looks like what you are doing should work. It looks very similar to https://github.com/grpc/grpc/blob/17f682d8274ef0b7d1376eeee5e94839a0750e0e/src/objective-c/GRPCClient/private/GRPCHost.m#L129, actually.

What I would recommend would be to add several `printf` statements to your C code that print out the certificate chain and private keys passed to `cgrpc_channel_create_secure` and which branch is taken. Also, double-check that your certs and keys are in the PEM format expected by gRPC.

n8armstrong:
@MrMage Thanks for your help. I did add several `printf` statements to the c code as you suggested and everything looks good there but we are still getting an SSL error when trying to connect to the gRPC server.
```
---E0605 14:20:25.789013000 123145321992192 ssl_transport_security.cc:1063] Handshake failed with fatal error SSL_ERROR_SSL: error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED.
```

 We are generating the certificates using openSSL. Should we be using BoringSSL?

Here is the script we are running to generate the certificates:
```bash
#!/bin/bash

# https://github.com/grpc/grpc-java/tree/master/examples

# Changes these CN's to match your hosts in your environment if needed.
SERVER_CN=localhost
CLIENT_CN=localhost # Used when doing mutual TLS

echo Generate CA key:
openssl genrsa -passout pass:1111 -des3 -out ca.key 4096
echo Generate CA certificate:
# Generates ca.crt which is the trustCertCollectionFile
openssl req -passin pass:1111 -new -x509 -days 365 -key ca.key -out ca.crt -subj "/CN=${SERVER_CN}"
echo Generate server key:
openssl genrsa -passout pass:1111 -des3 -out server.key 4096
echo Generate server signing request:
openssl req -passin pass:1111 -new -key server.key -out server.csr -subj "/CN=${SERVER_CN}"
echo Self-signed server certificate:
# Generates server.crt which is the certChainFile for the server
openssl x509 -req -passin pass:1111 -days 365 -in server.csr -CA ca.crt -CAkey ca.key -set_serial 01 -out server.crt
echo Remove passphrase from server key:
openssl rsa -passin pass:1111 -in server.key -out server.key
echo Generate client key
openssl genrsa -passout pass:1111 -des3 -out client.key 4096
echo Generate client signing request:
openssl req -passin pass:1111 -new -key client.key -out client.csr -subj "/CN=${CLIENT_CN}"
echo Self-signed client certificate:
# Generates client.crt which is the clientCertChainFile for the client (need for mutual TLS only)
openssl x509 -passin pass:1111 -req -days 365 -in client.csr -CA ca.crt -CAkey ca.key -set_serial 01 -out client.crt
echo Remove passphrase from client key:
openssl rsa -passin pass:1111 -in client.key -out client.key
echo Converting the private keys to X.509:
# Generates client.pem which is the clientPrivateKeyFile for the Client (needed for mutual TLS only)
openssl pkcs8 -topk8 -nocrypt -in client.key -out client.pem
# Generates server.pem which is the privateKeyFile for the Server
openssl pkcs8 -topk8 -nocrypt -in server.key -out server.pem
```

bootstraponline:
I updated the script to generate AES keys. I'm still getting verification errors. Maybe something has to be done on the iOS side to support self signed certificates?

```
$ openssl s_client -CAfile ca.crt -cert client.crt -key client.pem -connect localhost:50051
CONNECTED(00000005)
depth=0 CN = localhost
verify error:num=18:self signed certificate
verify return:1
depth=0 CN = localhost
verify return:1
---
Certificate chain
 0 s:/CN=localhost
   i:/CN=localhost
---
Server certificate
-----BEGIN CERTIFICATE-----
<removed>
-----END CERTIFICATE-----
subject=/CN=localhost
issuer=/CN=localhost
---
Acceptable client certificate CA names
/CN=localhost
---
SSL handshake has read 2003 bytes and written 2168 bytes
---
New, TLSv1/SSLv3, Cipher is ECDHE-RSA-AES256-GCM-SHA384
Server public key is 4096 bit
Secure Renegotiation IS supported
Compression: NONE
Expansion: NONE
No ALPN negotiated
SSL-Session:
    Protocol  : TLSv1.2
    Cipher    : ECDHE-RSA-AES256-GCM-SHA384
    Session-ID: 90EC34E84A23C6476901EF6667224B97193E7C5F3CDCE387C845A50116453D33
    Session-ID-ctx:
    Master-Key: F2E1A1B2CE2DB717E70E80E9BF7460983CB458E60BB9559861CB443A91307D17632335F03BC256404BF9D8A513BBBF60
    Start Time: 1528249571
    Timeout   : 300 (sec)
    Verify return code: 18 (self signed certificate)
---
read:errno=0
```

rlubke:
Hey, sorry for not chiming in sooner.

We're using the java-grpc certs for testing our grpc projects.  I was getting similar errors on connection until I changed the host to 'foo.test.google.fr' (don't forget to add a reference to /etc/hosts to redirect to your local interface).

MrMage:
It feels like we are starting to mix client certs and server certs here.

**For client certs:** those should work with this PR (I guess? at least I didn't find any obvious mistakes), but you of course need to configure your gRPC server to accept them.

**For server certs:** the SwiftGRPC client will currently NOT connect to servers with self-signed certificates. The solution would be to add support for passing in a custom set of root certificates to the channel constructor. See https://github.com/grpc/grpc-swift/pull/247#issuecomment-394493819 for a short explanation of that. Note that that should be in a separate PR.

bootstraponline:
blocked on https://github.com/grpc/grpc-swift/pull/250

MrMage:
Update on this one: sorry for sending you down the wrong path with self-signed certificates not working, for some reason I had never noticed that method :-(

With connecting to a server that doesn’t require client certificates now working (if I read #250 correctly), could you please provide an update on the error messages encountered now? Also, could you provide more details into how the server is set up to require client certs (e.g. with a code snippet demonstrating how the server is set up, even if the server is written in a different language (Java?))?

n8armstrong:
This is the error:
```
Handshake failed with fatal error SSL_ERROR_SSL: error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED.
```
And this is the server: https://gist.github.com/n8armstrong/f0a1ef6106b583d882d7b0edc117d77a

bootstraponline:
They probably want a server that doesn't use our internal code. I'll make one real quick today.

MrMage:
Just an example that shows how certificate verification is supposed to work would help already.

@bootstraponline @n8armstrong are you guys on the same team? If so, which one?

bootstraponline:
> @bootstraponline @n8armstrong are you guys on the same team? If so, which one?

Yeah, we're on the Instructure mobile team. Nate is an iOS developer. You can see how we're using gRPC Swift. A gRPC server is defined in Kotlin which we're trying to call from Swift. Works great if SSL is disabled. However we need to deploy the server to enable Firebase Test Lab iOS physical device automation. Before deployment, SSL needs to be enabled along with authentication. The Swift client isn't able to communicate to the Java server with SSL enabled. A Java client connects without issue.

- [Instructure iOS](https://github.com/instructure/instructure-ios)
- [Instructure Android data seeding](https://github.com/instructure/instructure-android/blob/master/automation/soseedygrpc/src/main/kotlin/com/instructure/dataseeding/OutOfProcessServer.kt)

I'll work on creating a standalone proof of concept that can be executed to reproduce the issue.

n8armstrong:
@MrMage Here is an example project for you: https://github.com/bootstraponline/grpc_java_auth. The swift tests have a case for regular ssl and mutual auth against this pr. Both of the tests fail to connect to the java server due to an ssl connection error.

Let me know if you have any questions about the demo.

MrMage:
@n8armstrong thanks! I got the regular SSL auth working. I noticed that the client would accept the server if we passed the server cert for verification (`Certs.certChain`) instead of the CA cert (`Certs.trustCertCollection`). Changing the CA generation line to this:

```openssl req  -new -x509 -days 365 -key ca.key -out ca.crt -subj "/C=US/ST=Foo/L=Bar/O=Baz/CN=Xyz"```

fixed that. (See https://github.com/MrMage/grpc_java_auth/blob/mrmage/soseedygrpc/openssl/keys.sh.)

I also got client auth to work by adding `, arguments: [.sslTargetNameOverride("example.com")]` to that client's initializer arguments. See https://github.com/mrmage/grpc_java_auth/blob/mrmage/JavaAuthSwift/JavaAuthTests/JavaAuthTests.swift (note that this is custom code, but you should get the essentials).

**Comment by n8armstrong** [comment]
Awesome, @MrMage. It's working! Thank you so much for your help.

I'd like to add a test for this so that we can get it merged but I'm not sure how to go about editing the proto file so that the server will accept the `rootCerts`. I'm referring to this change that you made in our demo: https://github.com/MrMage/grpc_java_auth/commit/d193ba64f90bba0c859d92604a1112c3774257b4#diff-d5503c9eee3f880120fb53c5d0d0488aR265

```diff
-  internal init?(address: String, certificateString: String, keyString: String, provider: Echo_EchoProvider) {
+  internal init?(address: String, certificateString: String, keyString: String, rootCerts: String?, provider: Echo_EchoProvider) {
```

You edited the generated file directly where I would need to edit the proto file. Could you provide some guidance on this?

MrMage:
Sure, thanks for your willingness to test this!

Here are the methods you'd need to edit:

* https://github.com/grpc/grpc-swift/blob/8ce6ad0d3631cc53f92ea9f372be1a1d4cee994d/Sources/protoc-gen-swiftgrpc/Generator-Server.swift#L86, then regenerate `echo.grpc.swift` with `make test-plugin; cp /tmp/echo.grpc.swift Sources/Examples/Echo/Generated/echo.grpc.swift`
* https://github.com/grpc/grpc-swift/blob/8ce6ad0d3631cc53f92ea9f372be1a1d4cee994d/Sources/CgRPC/shim/cgrpc.h#L191
* https://github.com/grpc/grpc-swift/blob/8ce6ad0d3631cc53f92ea9f372be1a1d4cee994d/Sources/CgRPC/shim/server.c#L36 (need to pass the root certs instead of `NULL` in https://github.com/grpc/grpc-swift/blob/8ce6ad0d3631cc53f92ea9f372be1a1d4cee994d/Sources/CgRPC/shim/server.c#L47; while at it I'd suggest to also add an option to enable server-side cert checking (currently always 0, i.e. disabled) by adding a parameter for the argument in https://github.com/grpc/grpc-swift/blob/8ce6ad0d3631cc53f92ea9f372be1a1d4cee994d/Sources/CgRPC/shim/server.c#L50)
* https://github.com/grpc/grpc-swift/blob/8ce6ad0d3631cc53f92ea9f372be1a1d4cee994d/Sources/SwiftGRPC/Core/Server.swift#L53
* https://github.com/grpc/grpc-swift/blob/8ce6ad0d3631cc53f92ea9f372be1a1d4cee994d/Sources/SwiftGRPC/Runtime/ServiceServer.swift#L35

Also, note that currently this can't be tested because we don't have client certs and a root CA to test with. You could essentially use your `keys.sh` script to add these (and replace the current server certs in https://github.com/grpc/grpc-swift/tree/master/Tests and https://github.com/grpc/grpc-swift/blob/master/Tests/SwiftGRPCTests/TestKeys.swift).

So I am afraid this is one of the annoying cases where the tests are more work than the actual change. However, I would be okay with submitting this as-is first and adding tests in a follow-up PR.

bootstraponline:
I think it's best to have tests in the same PR as the feature.

bootstraponline:
## errors


> Caused by: javax.net.ssl.SSLHandshakeException: error:10000410:SSL routines:OPENSSL_internal:SSLV3_ALERT_HANDSHAKE_FAILURE

Java client hostname wrong. Fix by updating `CLIENT_CN="localhost"` to `CLIENT_CN="example.com"`


> SwiftGRPC.RPCError.callError(successful, status unavailable: Connect Failed

Swift client hostname wrong. Fix by updating `CLIENT_CN="example.com"` to `CLIENT_CN="localhost"`


---

I added Java tests to https://github.com/bootstraponline/grpc_java_auth

I can either get the Swift tests passing (`CLIENT_CN="localhost"`) or the Java tests passing (`CLIENT_CN="example.com"`) but not both.  I'm using the `SslTargetNameOverride` equivalent in Java `.overrideAuthority()` however the tests still fail when the keys are generated with `CLIENT_CN="localhost"`.

If you have any ideas about making the tests pass for both Java & Swift with the same certs, that'd be great.

bootstraponline:
I rewrote the script using certstrap. Now all the tests are passing (Java & Swift).

https://github.com/bootstraponline/grpc_java_auth/blob/master/soseedygrpc/openssl/keys.sh

MrMage:
In case we commit the script to the SwiftGRPC repo (which might make sense to reproduce how the keys are created) I would prefer if it hadn't the extra dependency on certstrap. Did you try replacing the "/DN=example.com" part with "/DN=example.com/CN=example.com"? (In either case, the Java tests wouldn't ne committed to SwiftGRPC so the old keys.sh would suffice for SwiftGRPC I think.)

bootstraponline:
> In case we commit the script to the SwiftGRPC repo (which might make sense to reproduce how the keys are created) I would prefer if it hadn't the extra dependency on certstrap.

I agree, I [replaced certstrap with cfssl](https://github.com/bootstraponline/grpc_java_auth/blob/27b762258a4c42df61aca071be8176ca9e7886d7/soseedygrpc/openssl/cfssl.sh).

For the PR we'll check in an `openssl` only script based on the old `keys.sh`. I debugged the issue and Java expects `CN` to match on [`ca`](https://github.com/bootstraponline/grpc_java_auth/blob/27b762258a4c42df61aca071be8176ca9e7886d7/soseedygrpc/openssl/ca-csr.json) and [`client`](https://github.com/bootstraponline/grpc_java_auth/blob/27b762258a4c42df61aca071be8176ca9e7886d7/soseedygrpc/openssl/client-csr.json). [`server`](https://github.com/bootstraponline/grpc_java_auth/blob/27b762258a4c42df61aca071be8176ca9e7886d7/soseedygrpc/openssl/server-csr.json) works with a different `CN`.

bootstraponline:
I [updated the `openssl` script](https://github.com/bootstraponline/grpc_java_auth/commit/f301932e9bcec44f0808894359ac39d613e95bc1) and verified the certs work in both Java & Swift.

MrMage:
Did you need to modify the Swift code as well to support that `grpc_auth` client CN?

bootstraponline:
> Did you need to modify the Swift code as well to support that grpc_auth client CN?

Nope, it all works out of the box. The initial confusion came from [the template script which is parameterized incorrectly.](https://github.com/grpc/grpc-java/tree/master/examples#generating-self-signed-certificates-for-use-with-grpc)

```
ca "/CN=${SERVER_CN}"
server "/CN=${SERVER_CN}"
client "/CN=${CLIENT_CN}"
```

when it needs to be:

```
ca "/CN=${CLIENT_CN}"
server "/CN=${SERVER_CN}"
client "/CN=${CLIENT_CN}"
```

The example uses the same `localhost` value for both so they missed the error. I'll submit a PR upstream which corrects the issue.

n8armstrong:
Okay, tests have been added. Thanks for the detailed instructions, @MrMage.

scripts/makecert
```diff
@@ -1,8 +1,34 @@
...
+openssl x509 -req -days 365 -in server.csr -CA ca.crt -CAkey ca.key -set_serial 01 -out server.crt
+
+# Client
+openssl genrsa -out client.key $SIZE
+openssl req -new -key client.key -out client.csr -subj "/CN=${CN_CLIENT}"
+openssl x509 -req -days 365 -in client.csr -CA ca.crt -CAkey ca.key -set_serial 01 -out client.crt
+
+# netty only supports PKCS8 keys. openssl is used to convert from PKCS1 to PKCS8
+# http://netty.io/wiki/sslcontextbuilder-and-private-key.html
+# Generates client.pem which is the clientPrivateKeyFile for the Client (needed for mutual TLS only)
```
bootstraponline: > # Generates client.pem which is the clientPrivateKeyFile for the Client (needed for mutual TLS only)

I think we can remove that comment.

Sources/SwiftGRPC/Core/Server.swift
```diff
@@ -50,8 +50,8 @@ public class Server {
   /// - Parameter address: the address where the server will listen
   /// - Parameter key: the private key for the server's certificates
   /// - Parameter certs: the server's certificates
-  public init(address: String, key: String, certs: String) {
-    underlyingServer = cgrpc_server_create_secure(address, key, certs)
+  public init(address: String, key: String, certs: String, rootCerts: String? = nil) {
```
MrMage: Please document the rootCerts argument, that it is for client validation and will enforce client validation when provided.

Tests/SwiftGRPCTests/BasicEchoTestCase.swift
```diff
@@ -31,36 +31,53 @@ extension Echo_EchoResponse {
...
   var address: String { return "localhost:5050" }
 
   override func setUp() {
     super.setUp()
     
     provider = makeProvider()
-    
-    if secure {
-      let certificateString = String(data: certificateForTests, encoding: .utf8)!
+
```
MrMage: Rename to serverCertificate and serverKey, respectively?

Tests/SwiftGRPCTests/EchoTests.swift
```diff
@@ -38,7 +38,11 @@ class EchoTests: BasicEchoTestCase {
 }
 
 class EchoTestsSecure: EchoTests {
-  override var secure: Bool { return true }
+  override var security: Security { return .ssl }
+}
+
```
MrMage: Please add this test class to LinuxMain.swift.

n8armstrong: LinuxMain.swift is another generated file. Am I in for a world of hurt?

MrMage: That file shouldn't be autogenerated?

n8armstrong: Oh you're right. I was looking at LinuxMain.swift in swift-protobuf. 😌

Sources/SwiftGRPC/Core/Server.swift
```diff
@@ -50,6 +50,7 @@ public class Server {
   /// - Parameter address: the address where the server will listen
   /// - Parameter key: the private key for the server's certificates
   /// - Parameter certs: the server's certificates
+  /// - Parameter rootCerts: for client validation and will enforce client validation when provided
```
MrMage: super-nit: "used to validate client certificates; will enable enforcing valid client certificates when provided"

n8armstrong: Fixed 😄

n8armstrong:
@MrMage this is ready for you to review again.

MrMage:
LGTM

MrMage:
FYI, once this is merged, #76 can be closed. (/cc @timburks)

n8armstrong:
@MrMage do you have merge permissions? I do not..

bootstraponline:
@timburks will merge once he has time.

bootstraponline:
If there's anything I can do to help get the PR merged, let me know. We're looking to use this for our iOS apps.

MrMage:
I think this is ready to be merged, just waiting on @timburks to actually do the merge.

bootstraponline:
image
image

Do you know why I'd be getting these compile errors? I rebuilt `protoc-gen-swift` and `protoc-gen-swiftgrpc` from this branch. I also updated cocoapods to point to `pod 'SwiftGRPC', git: 'https://github.com/n8armstrong/grpc-swift.git', branch: 'tls-mutual-auth'`

bootstraponline:
nevermind, that error is due to building the `protoc` plugins from master instead of the feature branch.

timburks:
Looks great! Thanks for adding the test.

bootstraponline:
Thanks for merging this!

idi888:
Is there a way to grab information from the client certificate on the server, IE CN name ?

PR #245, opened 2018-06-04
Update Google sample for recent API changes.

timburks:


MrMage:
Sorry for breaking this — should we try to add building these to the CI runs, possibly at least on the Mac?

PR #246, opened 2018-06-04
Fix building with a clean Xcode and add a CI run to ensure just that.

MrMage:
See https://github.com/grpc/grpc-swift/pull/239#issuecomment-394280855. Also needs that PR to be merged first.


tsabirgaliev:
@MrMage this builds without issues, both from CLI and from within Xcode

timburks:
Fantastic! Thanks @MrMage

PR #247, opened 2018-06-04
Expose Google Credentials

jseibert:
Exposes the built-in default Google credential support, via the GOOGLE_APPLICATION_CREDENTIALS environment variable.

### DIFF
--- a/Sources/CgRPC/include/grpc/grpc_security.h
+++ b/Sources/CgRPC/include/grpc/grpc_security.h
@@ -133,7 +133,7 @@ GRPCAPI void grpc_channel_credentials_release(grpc_channel_credentials* creds);
 /** Creates default credentials to connect to a google gRPC service.
    WARNING: Do NOT use this credentials to connect to a non-google service as
    this could result in an oauth2 token leak. */
-GRPCAPI grpc_channel_credentials* grpc_google_default_credentials_create(void);
+GRPCAPI grpc_channel_credentials* grpc_google_default_credentials_create(const char* pem_root_certs);
 
 /** Callback for getting the SSL roots override from the application.
    In case of success, *pem_roots_certs must be set to a NULL terminated string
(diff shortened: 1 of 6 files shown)


MrMage:
Any changes to the gRPC C library would be overwritten the next time we integrate the upstream gRPC library, so I'd rather avoid those.

Also, why do we need to pass in root certs into `grpc_google_default_credentials_create`? What exactly does this change do that we couldn't do before?

jseibert:
@MrMage ahh great point on the C changes. 

Would appreciate your advice--maybe there is a better way to patch the Google credentials without dipping into C? Basically, the Swift library provides its own hard-coded `roots_pem()` when creating its normal secure channel, but the Google credentials internally create a secure channel with `nullptr`, and this was the only way I could figure out how to pipe them through?

MrMage:
@jseibert no idea about that. However, I would suppose that the gRPC developers had intended that method to be useful somehow even with NULL roots. To be honest, I don’t know how that method works at all and what problems would come from using no roots there (or, on the other hand passing them now).

So I’d suggest asking about that on the regular gRPC mailing list. Alternatively, if you think this is a bug, you could file an issue on the main gRPC repo.

jseibert:
Here's a summary of the issue: https://github.com/grpc/grpc/issues/4834

It might be possible to expose this in the shim without having to touch the core library: https://github.com/grpc/grpc/pull/4934

MrMage:
@jseibert that option looks better to me! For what it’s worth, I‘ve also been thinking of adding a „roots“ argument to the existing channel constructors; but I guess that wouldn’t help in this case as you explicitly need to call the google credentials method. Excuse the stupid question, but what exactly does that method do, anyway?

jseibert:
Our forks have hopelessly diverged for the time being due to other conflicts with Vapor+NIO, so I will let this lie for now.

Issue #249, opened 2018-06-06
SwiftGRPC client doesn't to servers with self-signed cert


bootstraponline:
> For server certs: the SwiftGRPC client will currently NOT connect to servers with self-signed certificates. The solution would be to add support for passing in a custom set of root certificates to the channel constructor. See #247 (comment) for a short explanation of that. Note that that should be in a separate PR.
>
> https://github.com/grpc/grpc-swift/pull/247#issuecomment-394493819
> https://github.com/grpc/grpc-swift/pull/244#issuecomment-394960991

Connecting to self signed servers isn't possible with grpc-swift.

bootstraponline:
@MrMage we're struggling with figuring out how to make this happen. Is this possible with only Swift side changes?

MrMage:
I think that should be possible by making `roots_pem()` customizable in https://github.com/grpc/grpc-swift/blob/7279202d7fe84ec6705d4623b45ee1ba6235035f/Sources/SwiftGRPC/Core/Channel.swift#L51. I'm not sure about the best way to expose that, though — @timburks, do you have any preference?

MrMage:
Never mind — if you use the method in https://github.com/grpc/grpc-swift/blob/7279202d7fe84ec6705d4623b45ee1ba6235035f/Sources/SwiftGRPC/Core/Channel.swift#L64, that should work already. Feel free to try it and let me know if it works for you.

bootstraponline:
Thanks, we're working on submitting a failing test case that reproduces the problem. It's not working for us with self signed certificates.

PR #250, opened 2018-06-06
Add tests to demonstrate security issues

n8armstrong:
There are 2 issues that I am hoping to surface by these tests.

**1. Basic ssl with `roots_pem()` does not appear to be working**
SSL requests using the `roots_pem()` certificate via `ServiceClientBase.init(address:secure:)` where `secure` is `true` results in error. See `EchoTestsRootSecurity` test.

**2. Self-signed ssl certificates do not work** #249 
I have been unable to send requests using a self-signed cert/key pair. See `EchoTestsSelfSigned` test. I included self-signed credentials that I generated using openssl commands.


MrMage:
In the case of `.root`, you are setting up a non-SSL Server but trying to connect via SSL. That’s not going to work. Also, the case of the default root certs can’t really be tested, because that would require a server cert signed by one of the default root CAs, i.e. a „real“ web server certificate.

MrMage:
Also, which commands exactly did you use to generate the self-signed certs? The existing test certificate is self-signed as well, so in principle self-signed certs should work.

MrMage:
FYI, the tests assume the certificates to be created for the „example.com“ domain; yours need to be created for that domain as well. See the `sslTargetNameOverride` channel argument.

n8armstrong:
@MrMage Thank you for the prompt feedback. What you said about `.root` makes perfect sense to me now. That test can be ignored.

This is how we are generating the self-signed certs:

n8armstrong:
```
openssl req -passin pass:1111 -new -x509 -days 365 -key ca.key -out ca.crt -subj "/CN=${SERVER_CN}"
openssl genrsa -passout pass:1111 -des3 -out server.key 4096
```

n8armstrong:
Self-signing with the correct domain (`example.com`) fixed the issue. Still can't get tls mutual auth to work even with #244 changes but we'll keep that issue separate.

Thanks.

PR #252, opened 2018-06-07
Make ChannelArgument extensible with custom values.

MrMage:
Also includes a refactoring, replacing most enum cases with factories that create the actual underlying enum cases. Should be source-compatible with the old implementation.

The purpose of this is to allow passing any gRPC channel argument without having to wait until that argument has been added to this repo.

@SebastianThiebaud, please let me know what you think about this; happy to scrap this if you think that it e.g. makes readability worse.

### DIFF
--- a/Sources/SwiftGRPC/Core/ChannelArgument.swift
+++ b/Sources/SwiftGRPC/Core/ChannelArgument.swift
@@ -18,91 +18,97 @@
 #endif
 import Foundation // for String.Encoding
 
-public extension Channel {
-  enum Argument {
+extension Channel {
+  public enum Argument {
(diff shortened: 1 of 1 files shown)


SebastianThiebaud:
Less code and the grpc key is next to the declaration of the case, so that's better to me. Tiny comment: I'd rather not make it open to custom keys.

MrMage:
> I'd rather not make it open to custom keys (except if I'm missing a reason why this would be useful)

The purpose of this is to allow passing any gRPC channel argument without having to wait until that argument has been added to this repo. There's hundreds of gRPC options and we might not even want to support all of them in this repo.

PR #255, opened 2018-06-12
For service servers, replace inheritance with composition

MrMage:
This lets us spin up one server handling multiple gRPC services, where each service has its own `Provider` implementation. I.e. we can now serve multiple services from the same port.

(This is not 100% source compatible with the previous code, but the necessary changes are very small, and I think SwiftGRPC is seldomly used on the server at the moment, anyway.)


Sources/SwiftGRPC/Runtime/ServiceServer.swift
```diff
@@ -23,39 +23,36 @@ open class ServiceServer {
...
+  public init(address: String, services: [ServiceProvider]) {
     gRPC.initialize()
     self.address = address
     server = Server(address: address)
+    servicesByName = Dictionary(uniqueKeysWithValues: services.map { ($0.serviceName, $0) })
   }
 
   /// Create a server that accepts secure connections.
-  public init(address: String, certificateString: String, keyString: String) {
+  public init(address: String, certificateString: String, keyString: String, services: [ServiceProvider]) {
```
timburks: How would you feel about changing "services:" to "serviceProviders:"? I think our "server", "service", and "provider" terminology can get confusing; more consistency could help here. (Maybe I'm still thinking like an Objective-C developer)

timburks: (same comment for all three init methods)

PR #256, opened 2018-06-18
Fix trailing metadata corruption if accessed after the original has been deallocated

MrMage:
We used to not copy trailing metadata we receive from the gRPC library. If we try to access that after the original metadata object's lifetime has expired, this causes memory corruption errors. We fix this by copying the metadata right away.


MrMage:
Actually, I did find the bug I was looking for and fixed it in a second commit (also added a test for it).

Tests/SwiftGRPCTests/ServerThrowingTests.swift
```diff
@@ -53,15 +55,28 @@ class ServerThrowingTests: BasicEchoTestCase {
...
-        else { XCTFail("unexpected error \(error)"); return }
-      XCTAssertEqual(.permissionDenied, callResult.statusCode)
-      XCTAssertEqual("custom status message", callResult.statusMessage)
+      caughtError = error as! RPCError
     }
+    
+    guard case let .callError(callResult) = caughtError
+      else { XCTFail("unexpected error \(caughtError)"); return }
+    
+    // Send another RPC to cause the original metadata array to be deallocated.
```
rebello95: What are your thoughts on splitting this into another test since it's validating something different than the original test?

Tests/SwiftGRPCTests/ServerThrowingTests.swift
```diff
@@ -53,15 +55,28 @@ class ServerThrowingTests: BasicEchoTestCase {
...
     do {
       let result = try client.get(Echo_EchoRequest(text: "foo")).text
       XCTFail("should have thrown, received \(result) instead")
+      return
     } catch {
-      guard case let .callError(callResult) = error as! RPCError
-        else { XCTFail("unexpected error \(error)"); return }
-      XCTAssertEqual(.permissionDenied, callResult.statusCode)
-      XCTAssertEqual("custom status message", callResult.statusMessage)
+      caughtError = error as! RPCError
```
rebello95: Should we fail the test if this is not an `RPCError`? i.e., `catch let as RPCError { ... }`... `catch { XCTFail(...) }`

Tests/SwiftGRPCTests/ServerThrowingTests.swift
```diff
@@ -53,15 +55,28 @@ class ServerThrowingTests: BasicEchoTestCase {
...
     }
+    
+    guard case let .callError(callResult) = caughtError
+      else { XCTFail("unexpected error \(caughtError)"); return }
+    
+    // Send another RPC to cause the original metadata array to be deallocated.
+    // If we were not copying the metadata array correctly, this would cause the metadata of the call result to become
+    // corrupted.
+    _ = try? client.get(Echo_EchoRequest(text: "foo")).text
```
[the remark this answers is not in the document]
MrMage: That call *will* actually throw; look at how the provider is implemented in that case. We don't care whether it throws; we just need the call for gRPC to shuffle some memory around (forcing the error).

Tests/SwiftGRPCTests/MetadataTests.swift
```diff
@@ -0,0 +1,31 @@
...
+import XCTest
+
+class MetadataTests: XCTestCase {
+  static var allTests: [(String, (MetadataTests) -> () throws -> Void)] {
+    return [
+      ("testCopy", testCopy)
+    ]
+  }
+  
+  func testCopy() {
```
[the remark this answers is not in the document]
MrMage: Suggestions? It's a fairly trivial test anyway.

Tests/SwiftGRPCTests/ServerThrowingTests.swift
MrMage: Done.

Tests/SwiftGRPCTests/ServerThrowingTests.swift
MrMage: Good idea! Done.

MrMage:
PTAL; updated the description as well.

Tests/SwiftGRPCTests/MetadataTests.swift
```diff
@@ -0,0 +1,31 @@
...
+
+class MetadataTests: XCTestCase {
+  static var allTests: [(String, (MetadataTests) -> () throws -> Void)] {
+    return [
+      ("testCopyDoesNotCorruptData", testCopyDoesNotCorruptData)
+    ]
+  }
+  
+  func testCopyDoesNotCorruptData() {
+    let metadata = try! Metadata(["foo": "bar"])
```
rebello95: Same comment here regarding `try`/`throws`

Tests/SwiftGRPCTests/ServerThrowingTests.swift
```diff
@@ -52,15 +55,37 @@ class ServerThrowingTests: BasicEchoTestCase {
...
 
 extension ServerThrowingTests {
-  func testServerThrowsUnary() {
+  func testServerThrowsUnary() throws {
     do {
       let result = try client.get(Echo_EchoRequest(text: "foo")).text
       XCTFail("should have thrown, received \(result) instead")
-    } catch {
-      guard case let .callError(callResult) = error as! RPCError
+      return
```
rebello95: No need for this return since `XCTFail` should end the test

Tests/SwiftGRPCTests/ServerThrowingTests.swift
```diff
@@ -52,15 +55,37 @@ class ServerThrowingTests: BasicEchoTestCase {
...
       XCTAssertEqual("custom status message", callResult.statusMessage)
+      XCTAssertEqual(["some_long_key": "bar"], callResult.trailingMetadata?.dictionaryRepresentation)
+    }
+  }
+  
+  func testServerThrowsUnary_noTrailingMetadataCorruptionAfterOriginalTrailingMetadataGetsReleased() throws {
+    do {
+      let result = try client.get(Echo_EchoRequest(text: "foo")).text
+      XCTFail("should have thrown, received \(result) instead")
+      return
```
rebello95: Same here 🤷‍♂️

Tests/SwiftGRPCTests/MetadataTests.swift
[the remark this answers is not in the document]
rebello95: Yea, the change you made is fine 👍

Tests/SwiftGRPCTests/ServerThrowingTests.swift
MrMage: XCTFail generates a failure but does not exit. In fact, the compiler would have warned about those return statements otherwise.

Tests/SwiftGRPCTests/ServerThrowingTests.swift
MrMage: See anove

Tests/SwiftGRPCTests/MetadataTests.swift
MrMage: See below

Tests/SwiftGRPCTests/ServerThrowingTests.swift
rebello95: Aha, I was thinking about `continueAfterFailure` - for some reason I thought it defaulted to `true`, but turns out it [defaults to `false`](https://developer.apple.com/documentation/xctest/xctestcase/1496260-continueafterfailure). My bad 🙃

Tests/SwiftGRPCTests/MetadataTests.swift
rebello95: This one was referring to marking the function as `throws` instead of using `try!`

MrMage: FWIW I actually prefer `try!` in tests; it gives more descriptive error messages. And given that the default case should be a pass, anyway, this should have little practical impact.

PR #266, opened 2018-07-13
add "make project"

woodcrust:
"make" without "make project" does not create the SwiftGRPC.xcodeproj file


MrMage:
LGTM. Good spot, thanks!

Issue #273, opened 2018-07-20
building EchoXcode example


mark1010:
I'm trying to build EchoXcode following the steps in the doc, with the current head and Xcode Version 9.4.1 (9F2000). 

The project seems to be missing dependencies on czlib, I apologise for the newbie question.

This is the process I've followed:

```
git clone https://github.com/grpc/grpc-swift
cd grpc-swift/

cd third_party
sh RUNME.sh
```
y (return), (return), all installs ok
```
cd ..
make && make project
```
All built ok, (An Echo target appears to have complied in this phase).

Open ./Examples/EchoXcode/Echo. xcodeprj in Xcode
Left Hand Nav (LHN) - right click Echo project Add Files, SwiftGRPC.xcodeproject, from parent folder.

Active Scheme is Echo (Echo project) > My Mac

Initial build fails, missing CgRPC.

LHN - click Echo project, General tab opens.

(Signing, Change Team).

Linked Frameworks and Libraries, 4x greyed out (unresolved) entries, only SwiftProtobuf.framework is 
listed.

base on README.md and `Echo.xcodeproj/project.pbxproj`, `PBXBuildFile section` the missing frameworks are: BoringSSL.framework,  Czlib.framework, CgRPC.framework, SwiftGRPC.framework.

+ button allows addition of: BoringSSL, CgRPC, Commander, SwiftGRPC, SwiftProtobufPluginLibrary, and is missing Czlib.

Add BoringSSL, SwiftGRPC, CgRPC.

Build with Command-B

Build fails, no changes,  missing CgRPC.
<img width="269" alt="screen shot 2018-07-20 at 13 51 24" src="https://user-images.githubusercontent.com/10847477/43003324-1812000e-8c24-11e8-9c37-809c567e0a3c.png">

Under Build Phases, Target Dependencies:
Add: BoringSSL, CgRPC, SwiftGRPC

no change to errors.

It's not clear to me whether I'm supposed to have provide Czlib, or is that part of the package management(?), and / or whether it's: Build Phases, Target Dependencies or Linked Frameworks / Libs that's causing the CgRPC issue, as CgRPC is there.

MrMage:
Hi Mark, thanks for the report! I think `Czlib` is an old module that is no longer module. Could you please try simply removing the module from the Xcode project and see if building works then? If not, could you try adding "libz" instead to the target's dependencies? And if you get that to work, would you mind filing a PR with those changes so we can include them in the repo? :-)

[...]

Issue #272, opened 2018-07-19
missing required module 'CgRPC'


alqa-dash:
Hi!

I want to clarify CgRPC integration process for myself. I looked through threads here and didn't find anything useful. 

I get whole SwiftGRPC project into my workspace and configure headers and Linked Frameworks. This way everything is fine. 

But in my case I have to create framework which uses SwiftGRPC as a dependency framework and share it with another team. To use that custom framework they must add custom framework to Linked Framework as well as SwiftGRPC, BoringSSL, CgRPC, SwiftProtobuf. It would be ok but this is not the end of story. Additionally they should configure Header Search Paths for CgRPC include folder and copy frameworks in Build Phases. Huh.. Now it should work.  Without these steps they got "missing required module 'CgRPC'" and so on.

My question: is it possible to avoid configuring Header Search Paths and (perhaps) copying frameworks?

JonasVautherin:
I am wondering the same thing, I believe. So if I understand correctly, you are trying the "manual integration", right?

I personally build manually in command line, with something like:

1. Clone the git repo
2. Run `make && make project`
3. Run something like `xcodebuild -target BoringSSL -target SwiftGRPC -target SwiftProtobuf -target CgRPC ...`

I end up with all four frameworks, but when I include them, CgRPC is not recognized as a module (same as you, I believe).

My solution is to create a `Modules/` and `Headers/` directory inside the built `CgRPC.framework`, copy `cgrpc.h` into `Headers/` and create the following modulemap in `Modules/` (notice the `framework` keyword):

```
framework module CgRPC {
    header "cgrpc.h"
    export *
}
```

I believe that I am doing manually what you do from Xcode (i.e. configuring Header Search Paths etc).

@MrMage: Would you have a comment on that? I must say I am not familiar with Xcode projects, so I mainly solved my issue through trial-and-error. So I'd like to have your opinion about what you think is happening and if that could be solved upstream.

Note that integrating through Cocoapods works flawlessly.

MrMage:
[...]

Issue #88, opened 2017-11-15
Consider throwing an error on non-success status code in unary call


novi:
In unary call on a client, We have no way to handle a status code of response.
We could fix that with like [this](https://github.com/novi/grpc-swift/blob/7ea88d1e8fbfa60c8da3478ccaeb21965f3e6609/Plugin/Templates/client-call-unary.swift#L22).

Plus we could add a new case for client service error.
```swift
/// Type for errors thrown from generated client code.
{{ access }} enum {{ .|clienterror:file,service }} : Error {
  case endOfStream
  case invalidMessageReceived
  case error(c: CallResult)
  case statusError(code: Int) // <-
}
```

aleczadikian:
I ended up changing `ClientCallUnary.swift` in the current latest SwiftGRPC Cocoapods code like this, and it seems to work right:

```
@@ -33,7 +33,7 @@ open class ClientCallUnaryBase<InputType: Message, OutputType: Message>: ClientC
       sem.signal()
     }
     _ = sem.wait()
-    if let returnResponse = returnResponse {
+    if let returnResponse = returnResponse, returnCallResult.statusCode == .ok {
       return returnResponse
     } else {
       throw RPCError.callError(returnCallResult)
```

MrMage:
Yes, that should work, but it discards the sent data completely. Ideally, we'd want to provide both the returned data and the status code in this case, but this would complicate our model. OTOH, I guess the case where we get a response _and_ a non-OK status code is fairly rare, so we could just throw the result like you did.

@timburks, what do you think?

MrMage:
The CgRPC implementation is now officially deprecated; closing this for now.

PR #274, opened 2018-07-21
Remove CzLib from dependencies

tikidunpon:
related issue: https://github.com/grpc/grpc-swift/issues/273 and https://github.com/grpc/grpc-swift/issues/272

I removed some CzLib dependencies from sample project. 
It works with an additional header search fixes but it seems like large.
so I divided PR for easy to review.


MrMage:
BTW, is the related issue #88 or #273 or something else?

tikidunpon:
@MrMage oh.. yes, it is related #273.. I wanna fix „Simple“ on other PR.

tikidunpon:
To running „Echo“ I also fixed some compile error.

tikidunpon:
I will fix Simple too.
> FYI, it looks like the „Simple“ example would need the same changes. I’ll leave it to you if you want to add those as well; otherwise I’ll fix those in a separate PR.

PR #281, opened 2018-08-05
First experiments with a SwiftNIO-based gRPC server

MrMage:
**This might be ready to review (and merge after review) now.**

As a weekend project, I looked into actually building a functional serving stack based on SwiftNIO. The serving part is actually working, but before proceeding I'd like to collect some feedback.

1. The use-case is that the user calls `GRPCServer.start()`, providing a generated provider class (see `EchoCallHandlerProvider`) that in turn calls through to their business logic (`EchoGRPCProvider`). We translate HTTP2 structs into their HTTP1 equivalents using `HTTP2ToHTTP1ServerCodec` (to allow for easier gRPC-via-HTTP1 support) and then decode the [gRPC wire protocol](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) via `HTTP1ToRawGRPCServerCodec`. (De-)Serializing messages is done later; see below.
2. Each time a request comes in, a `GRPCChannelHandler` is created to handle that request's channel.
3. Once that handler has received the request's headers (in particular, the method name), it looks up the corresponding provider and has the provider return a `GRPCCallHandler` object.
4. Subsequent messages and stream-close events sent on that channel are then passed to (and handled by) that `GRPCCallHandler`. (We should probably also forward errors to that object.)
5. The `GRPCCallHandler` itself then (de-)serializes messages to their actual `SwiftProtobuf.Message`-derived object types and forwards them to the client's code as appropriate. There are are four concrete implementations of that handler, `UnaryCallHandler`, `ClientStreamingCallHandler`, `ServerStreamingCallHandler` and `BidirectionalStreamingCallHandler` — one for each kind of gRPC request. (We can't do the deserialization any earlier because we don't know the expected types for the handler before the call's metadata has been sent. In particular, `GRPCServerCodec` is currently unused.)

FYI, `EchoServerTests.swift` also contains an example for roughly what the generated server code would look like (`EchoCallHandlerProvider`), and how the user would need to implement their business logic (`EchoGRPCProvider`).

So, here are a ton remarks. All feedback would be really appreciated, to any of these points (or others):

- **Converting a gRPC status to trailers and other channel handling should be separated (https://github.com/grpc/grpc-swift/pull/281#issuecomment-413554008)**
- The tests in `EchoServerTests` (which uses a regular SwiftGRPC client to connect to the NIO-based server) all pass, except for `testBidirectionalStreamingPingPong` and `testBidirectionalStreamingLotsOfMessagesPingPong` (for those, messages seem to be sent correctly, but the final status doesn't arrive).
- The method signatures that the library's consumer would need to implement (see `EchoGRPCProvider`) still look a bit funky. E.g. instead of passing the `headers` argument to all `EchoGRPCProvider` methods, we could always provide the corresponding `CallHandler` to all those methods (it is already being passed to server-streaming/bidirectional ones), and store the headers inside the `CallHandler`.
- Error handling is essentially non-existent. Currently, we immediately end the stream with a `ServerStatus.processingError` when encountering message (de-)serialization errors, and if the user passes an error to the status/result promises, we will also forward those to the user, but other than that, we have no error handling.
- - **In particular, I don't know how to forward errors encountered in the NIO network layer on to the library consumer. If some of the NIO people could chime in on this, that would be great.**
- - We might also want to automatically return an error status when `{Server,Bidirectional}StreamHandler.sendMessage()` fails.
- In general, it would be nice to get a code review from the SwiftNIO team, to make sure I'm not doing anything wrong here.
- Naming is terrible. I have several classes ending on `CallHandler`, `CallHandlerProvider`, `Provider`, `Server` etc. that could use better names (for example, there is a `UnaryCallHandler` that services specific unary gRPC calls that inherits from a `UnaryResponseHandler` that also partially handles client-streaming calls). Suggestions would be very welcome!
- I'm using too much subclassing right now. In particular, all actual request handlers derive from `StatusSendingHandler`, and sometimes `UnaryResponseHandler` as well, and those classes probably violate the single-responsibility principle. In general, the code shared between `UnaryCallHandler`, `ClientStreamingCallHandler`, `ServerStreamingCallHandler` and `BidirectionalStreamingCallHandler` could probably be cleaned up and made more consistent.
- I'm a bit concerned about having to pass `GRPCServerHandler` and `ChannelHandlerContext` arguments to the `StatusSendingHandler` initializer (and those of its subclasses). Feels like a leaky abstraction to me. (Those objects are at least not memory-leaked, though.)
- There are several assertions that I would want to replace with warnings in production.
- We still depend on SwiftGRPC (and in turn cGRPC), but luckily only in a few places. I think the long-term plan would be to move a small subset of SwiftGRPC that the NIO-based code also depends on into a shared library that would then be depended on by both `SwiftGRPC` and `SwiftGRPCNIO`.
- There is currently a crash in `NGHTTP2Session.purgeOldStreams` when sending ~ 1024 requests to one server (see `testUnaryLotsOfRequests`). This appears related to the `StreamManager.maxCachedStreamIDs` setting. Not sure if this is a bug in `NIOHTTP2` or whether we are leaking channels here.
- `{Server,Bidirectional}StreamHandler.sendMessage()` (and maybe `.sendStatus`) should probably return a future that more actions could be chained to (e.g. to handle backpressure and throttle automatically when trying to send too many responses at once on one channel).
- Not sure if the `statusPromise/responsePromise` approach in `StatusSendingHandler/UnaryResponseHandler` is beneficial.

/cc @timburks @bmhatfield


Lukasa:
> There is currently a crash in `NGHTTP2Session.purgeOldStreams` when sending ~ 1024 requests to one server (see testUnaryLotsOfRequests). This appears related to the StreamManager.maxCachedStreamIDs setting. Not sure if this is a bug in NIOHTTP2 or whether we are leaking channels here.

Relevant code: https://github.com/apple/swift-nio-http2/blob/80963ee509e2d82b68c6e576dde9fb126bc9bcb3/Sources/NIOHTTP2/NGHTTP2Session.swift#L285-L292

Well it's certainly *probable* that this crash is ours. I don't think you leaking channels could make this crash, the force-unwrap that is sitting there seems like the most likely culprit. If none of the streams there are "active", that would lead to exactly this issue. Should be easily reproduced: will do so when I get back from the dentist.

MrMage:
> Well it's certainly probable that this crash is ours. I don't think you leaking channels could make this crash, the force-unwrap that is sitting there seems like the most likely culprit. If none of the streams there are "active", that would lead to exactly this issue. Should be easily reproduced: will do so when I get back from the dentist.

The crash is indeed in the force-unwrap (sorry, forgot to mention), but I didn't pay closer attention to the underlying causes yet. Take your time!

Lukasa:
Ok, the crash is fixed by apple/swift-nio-http2#11.

Sources/SwiftGRPCNIO/GRPCServerCodec.swift
```diff
@@ -0,0 +1,63 @@
...
+
+public enum GRPCServerResponsePart<MessageType: Message> {
+  case headers(HTTPHeaders)
+  case message(MessageType)
+  case trailers(HTTPHeaders)
+}
+
+/// A simple channel handler that translates raw gRPC packets into decoded protobuf messages,
+/// and vice versa.
+/// **Currently unused, as we do not yet know the request's method name (and thus message types) when this object is instantiated.**
```
Lukasa: You can work around this by having this handler be a handler-factory: that is, this handler catches the initial inbound messages and, based on the request's method name, generates an appropriate handler and inserts it in the pipeline directly behind itself.

Does that make sense?

MrMage: That would definitely make sense; it's essentially what I had tried to in a much more clunky fashion by piping messages to the `GRPCChannelHandler`. Is there an example on how that insertion process would be done correctly?

Lukasa: Not directly, but there is some similar stuff in the `HTTPUpgradeHandler`: https://github.com/apple/swift-nio/blob/a7fee8d98ffb1c1fd695a92521c1f6e59a98c78b/Sources/NIOHTTP1/HTTPUpgradeHandler.swift#L212-L242

The general idea is that, in general, pipeline manipulating operations on the channel you're currently running on actually complete synchronously. You shouldn't rely on that fact (that is, you should be prepared to buffer data), but you can always assume any requirement to buffer data will be short-lived.

The only thing you're missing is addinghandlers. For that you probably want [`add(handler:after:)`](https://apple.github.io/swift-nio/docs/current/NIO/Classes/ChannelPipeline.html#/s:3NIO15ChannelPipelineC3addAA15EventLoopFutureCyytGSSSg4name_AA0B7Handler_p7handlerAaJ_p5aftertF), where `after` can be `self`.

MrMage: Thanks! I have changed the code to directly add the `ServerCodec` and `CallHandlers` now and remove the `ChannelHandler`.

MrMage:
@Lukasa @weissi do you have any other feedback on this PR? In particular, is the use of SwiftNIO (with e.g. `responsePromise` and `statusPromise`) idiomatic? Or am I doing something potentially dangerous, e.g. by keeping a (weak) reference to `ctx` in `StatusSendingHandler`?

weissi:
@MrMage sorry for the delayed response. I only had a brief look so apologies for my questions: The `responsePromise` seems to be fulfilled when the client is supposed to send a response. And usually the `responsePromise` is fulfilled when a `StatusSendingHandler` received `channelRead`(which then calls some subclass' `processMessage` when then in turns fulfils this promise). What I don't really understand is why isn't it firing a new event through the pipeline (`ctx.fireChannelRead(...)`)?

For example this function seems to me like a good candidate for a `ChannelInboundHandler`:

```swift
  func get(request: Echo_EchoRequest, headers: HTTPHeaders, eventLoop: EventLoop) -> EventLoopFuture<Echo_EchoResponse> {
    var response = Echo_EchoResponse()
    response.text = "Swift echo get: " + request.text
    return eventLoop.newSucceededFuture(result: response)
  }
```

couldn't that be something like

```swift
final class EchoGRPCProvider: ChannelInboundHandler {
    [...]
    func channelRead(...) {
        [...]
        switch receivedMessage {
            case .get(Echo_EchoRequest):
                var response = Echo_EchoResponse()
                response.text = "Swift echo get: " + request.text
                ctx.write(...(response)).whenFailure { /* error handling */ }
           [...]
       }
    }
}
```

so I maybe the pipeline could deliver requests inbound and would be delivered responses outbound.

Does that make some sense?

MrMage:
@weissi thanks for the feedback!

Sticking to the example of the "get" call, the way a _unary_ gRPC call works is that _one_ input message is passed in and _one_ output result (or an error) is returned. I think that can better be reflected by providing the service implementor with an interface that matches this mental model. If the user had to implement a `ChannelHandler` themselves, they'd need to add boilerplate for the request lifecycle and state handling (which should be abstracted away by the gRPC framework) and would be able to send more than reply, which would violate the request's semantics.

So, my goal is to:

- only provide the user with the information they need to service the request
- only have them implement the code to service the request, with little boilerplate
- only provide them the methods they need to service the request, without e.g. shooting themselves in the foot by returning more than one response where that would be inappropriate

This is similar to Vapor 3's model, where request handler are also passed in the request data and are asked to return a future of the corresponding response. The framework is then tasked with delivering that response to the client by writing it to the outbound channel.

Also, a more sophisticated example involving actual work on the server side (a database request offloaded to the event loop) could be this:

```
  func get(request: Echo_EchoRequest, headers: HTTPHeaders, eventLoop: EventLoop) -> EventLoopFuture<Echo_EchoResponse> {
    return database.fetchEchoObjects(withText: request.text).first() // this returns an `EventLoopFuture<EchoObject>`
      .then { echoObject in echoObject.title }
  }
```

Hope that makes sense — let me know if I'm missing something, or if this approach is problematic.

weissi:
@MrMage cool, that makes a lot of sense. In that case I think `StatusSendingHandler` for example does two tasks:

1. doing protocol work (like setting the `"grpc-status"` trailer)
2. taking the handling of the messages out of the `ChannelPipeline`

Both are totally totally valid things to do but I would suggest to separate the concerns and use two different channel handlers for that. One that does (1) and one that does (2). In that case if the user is a power-NIO user anyway, they could just not put (2) into the pipeline.

For ease of use, you might obviously provide a method that adds both of them together like we do in [`NIOHTTP1`'s `configureHTTPServerPipeline1](https://github.com/apple/swift-nio/blob/d53ee6dafc4d03a342aa32e79d5d282abc0c4f1a/Sources/NIOHTTP1/HTTPPipelineSetup.swift#L96)

MrMage:
**This might be ready to review (and merge after review) now.**

MrMage:
Closing and re-opening to trigger another CI run.

JonasVautherin:
What about trying to add `project-carthage` to build-carthage-debug (and probably build-carthage) ([here](https://github.com/grpc/grpc-swift/blob/master/Makefile#L48))?

JonasVautherin:
So I tried to run `make project-carthage`, and [it seems like it's going further](https://travis-ci.org/JonasVautherin/grpc-swift/jobs/441895928) (failing because log is too long).

I'll keep you posted, but until now I believe it confirms what I was expecting, so it doesn't sound too bad!

Sources/SwiftGRPCNIO/CallHandlers/BidirectionalStreamingCallHandler.swift
```diff
@@ -0,0 +1,42 @@
...
+public class BidirectionalStreamingCallHandler<RequestMessage: Message, ResponseMessage: Message>: BaseCallHandler<RequestMessage, ResponseMessage> {
+  public typealias EventObserver = (StreamEvent<RequestMessage>) -> Void
+  fileprivate var eventObserver: EventLoopFuture<EventObserver>?
+  
+  //! FIXME: Do we need to keep the context around at all here?
+  public private(set) var context: StreamingResponseCallContext<ResponseMessage>?
+
+  // We ask for a future of type `EventObserver` to allow the framework user to e.g. asynchronously authenticate a call.
+  // If authentication fails, they can simply fail the observer future, which causes the call to be terminated.
+  public init(channel: Channel, headers: HTTPRequestHead, eventObserverFactory: (StreamingResponseCallContext<ResponseMessage>) -> EventLoopFuture<EventObserver>) {
```
weissi: Quick question here: the parameter is `headers` but it hands you the whole request. Would recommend to either rename `request:` or to pass `HTTPHeaders` instead of a `HTTPRequestHead` here. You can get the headers using `request.headers`

MrMage: Well spotted! I was originally using `HTTPHeaders` there, then noticed that some other request information (e.g. `.uri`) could be useful to have (e.g. for middlewares). I then forgot to rename the field.

Renamed the field now — please take another look.

Package.swift
```diff
@@ -19,7 +19,10 @@ import PackageDescription
 var packageDependencies: [Package.Dependency] = [
   .package(url: "https://github.com/apple/swift-protobuf.git", .upToNextMinor(from: "1.1.1")),
   .package(url: "https://github.com/kylef/Commander.git", from: "0.8.0"),
-  .package(url: "https://github.com/apple/swift-nio-zlib-support.git", from: "1.0.0")
+  .package(url: "https://github.com/apple/swift-nio-zlib-support.git", from: "1.0.0"),
```
rebello95: We should version lock these more strictly I think. Maybe `upToNextMinor`?

Package.swift
```diff
@@ -19,7 +19,10 @@ import PackageDescription
...
+  .package(url: "https://github.com/apple/swift-nio.git", from: "1.9.0"),
+  .package(url: "https://github.com/apple/swift-nio-nghttp2-support.git", from: "1.0.0"),
+  .package(url: "https://github.com/lukasa/swift-nio-http2.git", .branch("cb-issue-24"))
```
rebello95: We shouldn't be pointing at a fork here

Package.swift
```diff
@@ -35,11 +38,14 @@ let package = Package(
...
   products: [
     .library(name: "SwiftGRPC", targets: ["SwiftGRPC"]),
+    .library(name: "SwiftGRPCNIO", targets: ["SwiftGRPCNIO"]),
   ],
   dependencies: packageDependencies,
   targets: [
     .target(name: "SwiftGRPC",
             dependencies: ["CgRPC", "SwiftProtobuf"]),
+    .target(name: "SwiftGRPCNIO",
+            dependencies: ["SwiftProtobuf", "NIOHTTP1", "NIOFoundationCompat", "NIOHTTP2"]),
```
rebello95: nit: alphabetical order?

Sources/SwiftGRPCNIO/CallHandlers/BidirectionalStreamingCallHandler.swift
```diff
@@ -0,0 +1,42 @@
...
+import SwiftProtobuf
+import NIO
+import NIOHTTP1
+
+// Handles bidirectional streaming calls. Forwards incoming messages and end-of-stream events to the observer block.
+// The observer block is implemented by the framework user and calls `context.sendResponse` as needed.
+// To close the call and send the status, fulfill `context.statusPromise`.
+public class BidirectionalStreamingCallHandler<RequestMessage: Message, ResponseMessage: Message>: BaseCallHandler<RequestMessage, ResponseMessage> {
+  public typealias EventObserver = (StreamEvent<RequestMessage>) -> Void
+  fileprivate var eventObserver: EventLoopFuture<EventObserver>?
```
rebello95: Can be `private`

Sources/SwiftGRPCNIO/CallHandlers/BidirectionalStreamingCallHandler.swift
```diff
@@ -0,0 +1,42 @@
...
+  
+  //! FIXME: Do we need to keep the context around at all here?
```
rebello95: Doesn't look like we need this based on how it's used below

Sources/SwiftGRPCNIO/CallHandlers/BidirectionalStreamingCallHandler.swift
```diff
@@ -0,0 +1,42 @@
...
+
+  // We ask for a future of type `EventObserver` to allow the framework user to e.g. asynchronously authenticate a call.
+  // If authentication fails, they can simply fail the observer future, which causes the call to be terminated.
+  public init(channel: Channel, request: HTTPRequestHead, eventObserverFactory: (StreamingResponseCallContext<ResponseMessage>) -> EventLoopFuture<EventObserver>) {
+    super.init()
+    self.context = StreamingResponseCallContextImpl<ResponseMessage>(channel: channel, request: request)
+    self.eventObserver = eventObserverFactory(context!)
+    // Terminate the call if no observer is provided.
+    self.eventObserver?.cascadeFailure(promise: context!.statusPromise)
+    context!.statusPromise.futureResult.whenComplete {
```
rebello95: Can we avoid force-unwrapping here as a best practice even though it's guaranteed to exist above? We could potentially do something like this to handle it:

```
let context = StreamingResponseCallContextImpl<ResponseMessage>(channel: channel, request: request)
self.context = context
context.statusPromise...
...

File: Sources/SwiftGRPCNIO/CallHandlers/ClientStreamingCallHandler.swift [hunk 9]
```diff
@@ -0,0 +1,41 @@
+import Foundation
+import SwiftProtobuf
+import NIO
+import NIOHTTP1
+
+// Handles client-streaming calls. Forwards incoming messages and end-of-stream events to the observer block.
+// The observer block is implemented by the framework user and fulfills `context.responsePromise` when done.
+public class ClientStreamingCallHandler<RequestMessage: Message, ResponseMessage: Message>: BaseCallHandler<RequestMessage, ResponseMessage> {
+  public typealias EventObserver = (StreamEvent<RequestMessage>) -> Void
+  fileprivate var eventObserver: EventLoopFuture<EventObserver>?
```
rebello95: I'll avoid commenting on all these cases, but we should clean up the `fileprivate --> private` and force unwraps throughout this PR

Sources/SwiftGRPCNIO/CallHandlers/ClientStreamingCallHandler.swift
```diff
@@ -0,0 +1,41 @@
...
+  
+  //! FIXME: Do we need to keep the context around at all here?
```
rebello95: Should be able to remove these from all 4 handlers

Sources/SwiftGRPCNIO/CallHandlers/ServerStreamingCallHandler.swift
```diff
@@ -0,0 +1,43 @@
...
+  
+  //! FIXME: Do we need to keep the context around at all here?
+  public private(set) var context: StreamingResponseCallContext<ResponseMessage>?
+  
+  public init(channel: Channel, request: HTTPRequestHead, eventObserverFactory: (StreamingResponseCallContext<ResponseMessage>) -> EventObserver) {
+    super.init()
+    self.context = StreamingResponseCallContextImpl<ResponseMessage>(channel: channel, request: request)
+    self.eventObserver = eventObserverFactory(context!)
+    context!.statusPromise.futureResult.whenComplete {
+      // When done, reset references to avoid retain cycles.
```
rebello95: What if we made these references `weak` instead? Also technically not a retain cycle, just a memory leak 😄

Sources/SwiftGRPCNIO/CallHandlers/ServerStreamingCallHandler.swift
```diff
@@ -0,0 +1,43 @@
...
+      self.eventObserver = nil
+      self.context = nil
+    }
+  }
+  
+  
+  public override func processMessage(_ message: RequestMessage) {
+    guard !hasReceivedRequest else {
```
rebello95: It'd be nice if we could remove this state and instead rely on `self.eventObserver` being nil-ed out when it's no longer needed. Then we could potentially simply `assert(self.eventObserver == nil)` since it's a library error if this happens and remove `hasReceivedRequest`

Sources/SwiftGRPCNIO/CallHandlers/UnaryCallHandler.swift
```diff
@@ -0,0 +1,43 @@
+import Foundation
+import SwiftProtobuf
+import NIO
+import NIOHTTP1
+
+// Handles unary calls. Calls the observer block with the request message.
+// The observer block is implemented by the framework user and returns a future containing the call result.
+// To return a response to the client, the framework user should complete that framework
```
rebello95: > the framework user should complete that framework

Can we make this a little clearer?

Sources/SwiftGRPCNIO/GRPCChannelHandler.swift
```diff
@@ -0,0 +1,60 @@
...
+
+// Listens on a newly-opened HTTP2 subchannel and waits for the request headers to become available.
+// Once those are available, asks the `CallHandlerProvider` corresponding to the request's service name for an
+// `GRPCCallHandler` object. That object is then forwarded the individual gRPC messages.
+public final class GRPCChannelHandler: ChannelInboundHandler {
+  public typealias InboundIn = RawGRPCServerRequestPart
+
+  public typealias OutboundOut = RawGRPCServerResponsePart
+
+  fileprivate let servicesByName: [String: CallHandlerProvider]
```
rebello95: `private`

Sources/SwiftGRPCNIO/GRPCChannelHandler.swift
```diff
@@ -0,0 +1,60 @@
...
+
+  public init(servicesByName: [String: CallHandlerProvider]) {
+    self.servicesByName = servicesByName
+  }
+
+  public func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
```
rebello95: Might be nice to put the protocol conformance functions (throughout this PR) in separate extensions in the same file to make it clearer as to where they're coming from

Sources/SwiftGRPCNIO/GRPCChannelHandler.swift
```diff
@@ -0,0 +1,60 @@
...
+    let requestPart = self.unwrapInboundIn(data)
+    switch requestPart {
+    case .head(let requestHead):
+      let uriComponents = requestHead.uri.components(separatedBy: "/")
+      guard uriComponents.count >= 3 && uriComponents[0].isEmpty,
```
rebello95: Might be nice to name these or add docs to explain what `0, 1, 2` represent

Sources/SwiftGRPCNIO/GRPCChannelHandler.swift
```diff
@@ -0,0 +1,60 @@
...
+
+      var responseHeaders = HTTPHeaders()
+      responseHeaders.add(name: "content-type", value: "application/grpc")
+      ctx.write(self.wrapOutboundOut(.headers(responseHeaders)), promise: nil)
+
+      let codec = callHandler.makeGRPCServerCodec()
+      ctx.pipeline.add(handler: codec, after: self)
+        .then { ctx.pipeline.add(handler: callHandler, after: codec) }
+        //! FIXME(lukasa): Fix the ordering of this with NIO 1.12 and replace with `remove(, promise:)`.
+        .whenComplete { _ = ctx.pipeline.remove(handler: self) }
```
rebello95: Do we need `[weak self]` here?

Sources/SwiftGRPCNIO/GRPCServer.swift
```diff
@@ -0,0 +1,56 @@
+import Foundation
+import NIO
+import NIOHTTP1
+import NIOHTTP2
+
+// Wrapper object to manage the lifecycle of a gRPC server.
+public final class GRPCServer {
+  public static func start(
```
rebello95: Can we add headerdocs for these and other public interfaces? I think that'd greatly help with documentation/clarity

Sources/SwiftGRPCNIO/GRPCServer.swift
```diff
@@ -0,0 +1,56 @@
...
+    hostname: String,
+    port: Int,
+    eventLoopGroup: EventLoopGroup,
+    serviceProviders: [CallHandlerProvider]) -> EventLoopFuture<GRPCServer> {
+    let servicesByName = Dictionary(uniqueKeysWithValues: serviceProviders.map { ($0.serviceName, $0) })
+    let bootstrap = ServerBootstrap(group: eventLoopGroup)
+      // Specify backlog and enable SO_REUSEADDR for the server itself
```
rebello95: Can we explain why?

Sources/SwiftGRPCNIO/GRPCServer.swift
```diff
@@ -0,0 +1,56 @@
...
+      .childChannelOption(ChannelOptions.socket(IPPROTO_TCP, TCP_NODELAY), value: 1)
+      .childChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)
+      .childChannelOption(ChannelOptions.maxMessagesPerRead, value: 1)
+      .childChannelOption(ChannelOptions.allowRemoteHalfClosure, value: true)
+
+    return bootstrap.bind(host: hostname, port: port)
+      .map { GRPCServer(channel: $0) }
+  }
+
+  fileprivate let channel: Channel
```
rebello95: `private`

Sources/SwiftGRPCNIO/GRPCServerCodec.swift
```diff
@@ -0,0 +1,66 @@
+import Foundation
+import SwiftProtobuf
+import NIO
+import NIOFoundationCompat
+import NIOHTTP1
+
+public enum GRPCServerRequestPart<MessageType: Message> {
```
rebello95: Would be awesome if we could add docstrings for these

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -0,0 +1,122 @@
...
+}
+
+public enum RawGRPCServerResponsePart {
+  case headers(HTTPHeaders)
+  case message(ByteBuffer)
+  case status(GRPCStatus)
+}
+
+/// A simple channel handler that translates HTTP/1 data types into gRPC packets,
+/// and vice versa.
```
rebello95: Would be useful to explain why this is needed

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -0,0 +1,122 @@
...
+
+    case .body(var body):
+      guard var buffer = buffer
+        else { preconditionFailure("buffer not initialized") }
+      assert(state.expectingBody, "received body while in state \(state)")
+      buffer.write(buffer: &body)
+
+      // Iterate over all available incoming data, trying to read length-delimited messages.
+      // Each message has the following format:
+      // - 1 byte "compressed" flag (currently always zero, as we do not support for compression)
```
rebello95: Thoughts on writing tests for this function?

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -0,0 +1,122 @@
...
+      // - 4 byte signed-integer payload length (N)
+      // - N bytes payload (normally a valid wire-format protocol buffer)
+      requestProcessing: while true {
+        switch state {
+        case .expectingHeaders: preconditionFailure("unexpected state \(state)")
+        case .expectingCompressedFlag:
+          guard let compressionFlag: Int8 = buffer.readInteger() else { break requestProcessing }
+          precondition(compressionFlag == 0, "unexpected compression flag \(compressionFlag)")
```
rebello95: Might want to note that this is only unexpected because SwiftGRPC doesn't yet support compression flags

Sources/protoc-gen-swiftgrpc/Generator-Server.swift
```diff
@@ -63,44 +82,84 @@ extension Generator {
...
+        println("case \(methodPath):")
         indent()
-        println(".run()")
-        outdent()
+        switch streamingType(method) {
+        case .unary, .serverStreaming:
+          println("return try \(methodSessionName)Base(")
+          indent()
+          println("handler: handler,")
+          println("providerBlock: { try self.\(methodFunctionName)(request: $0, session: $1 as! \(methodSessionName)Base) })")
```
rebello95: Probably shouldn't be force casting like this in generated code

Sources/protoc-gen-swiftgrpc/Generator.swift
```diff
@@ -73,12 +73,30 @@ class Generator {
       //\n
       """)
     
-    for moduleName in ["Foundation", "Dispatch", "SwiftGRPC", "SwiftProtobuf"] {
+    let moduleNames: [String]
+    if options.generateNIOImplementation {
+      moduleNames = [
+        "Foundation",
+        "Dispatch",
```
rebello95: Do we need to import `Dispatch` since we're importing `Foundation`? If so, alphabetical order? 😄

Tests/SwiftGRPCNIOTests/NIOServerTestCase.swift
```diff
@@ -0,0 +1,57 @@
...
+}
+
+extension Echo_EchoResponse {
+  init(text: String) {
+    self.text = text
+  }
+}
+
+class NIOServerTestCase: XCTestCase {
+  func makeProvider() -> Echo_EchoProvider { return EchoProvider() }
```
rebello95: IMO not worth having a function for this

Tests/SwiftGRPCNIOTests/NIOServerTests.swift
```diff
@@ -0,0 +1,302 @@
...
+    XCTAssertEqual("Swift echo get: foo", try! client.get(Echo_EchoRequest(text: "foo")).text)
+  }
+
+  func testUnaryLotsOfRequests() {
+    // Sending that many requests at once can sometimes trip things up, it seems.
+    client.timeout = 5.0
+    let clockStart = clock()
+    let numberOfRequests = 1_000  //! FIXME: If we set this higher, it causes a crash related to `StreamManager.maxCachedStreamIDs`.
+    for i in 0..<numberOfRequests {
+      if i % 1_000 == 0 && i > 0 {
```
rebello95: Why the `% 1_000` test here if `numberOfRequest == 1_000`?

Tests/SwiftGRPCNIOTests/NIOServerTests.swift
```diff
@@ -0,0 +1,302 @@
...
+      XCTAssertEqual("Swift echo get: foo \(i)", try client.get(Echo_EchoRequest(text: "foo \(i)")).text)
+    }
+    print("total time for \(numberOfRequests) requests: \(Double(clock() - clockStart) / Double(CLOCKS_PER_SEC))")
+  }
+}
+
+extension NIOServerTests {
+  func testClientStreaming() {
+    let completionHandlerExpectation = expectation(description: "final completion handler called")
+    let call = try! client.collect { callResult in
```
rebello95: Instead of `try!`ing here, can we mark the functions as `throws` and let the tests fail with the thrown errors?

Tests/SwiftGRPCNIOTests/NIOServerTestCase.swift
MrMage: This test case is only a skeleton that gets subclassed by concrete tests; the method will be overridden by other test cases to provide other provider implementations. See the non-NIO tests.

Sources/protoc-gen-swiftgrpc/Generator-Server.swift
MrMage: This is actually the old, non-NIO implementation. The force cast is necessary but safe in this case (note that we are instantiating `\(methodSessionName)Base(` above).

Package.swift
MrMage: Done (except for `NIOHTTP2`).

Package.swift
MrMage: Done (the required patch had not yet been submitted).

Package.swift
MrMage: Done.

Sources/SwiftGRPCNIO/CallHandlers/BidirectionalStreamingCallHandler.swift
MrMage: Done.

Sources/SwiftGRPCNIO/GRPCChannelHandler.swift
MrMage: Agreed; changed that on all handlers..

Sources/SwiftGRPCNIO/GRPCChannelHandler.swift
MrMage: No; this is supposed to terminate almost instantly.

Sources/SwiftGRPCNIO/GRPCServer.swift
MrMage: Grat idea! Done.

Sources/SwiftGRPCNIO/GRPCServer.swift
MrMage: Done.

Sources/SwiftGRPCNIO/GRPCServerCodec.swift
MrMage: Done.

MrMage: Done.

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
MrMage: Done.

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
MrMage: Implicitly tested by our high-level tests already.

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
MrMage: It's also unexpected because we do not indicate that we accept compression. Extended the message and added a FIXME.

Sources/protoc-gen-swiftgrpc/Generator.swift
MrMage: We might need these on Linux for the old implementation, but probably not for the NIO one.

Tests/SwiftGRPCNIOTests/NIOServerTests.swift
MrMage: Good point. I have increased this to 2000 now, as the underlying error got fixed by Corey.

Sources/SwiftGRPCNIO/GRPCChannelHandler.swift
MrMage: Done.

Sources/SwiftGRPCNIO/CallHandlers/ClientStreamingCallHandler.swift
MrMage: Done. Let me know in case I missed any.

Tests/SwiftGRPCNIOTests/NIOServerTests.swift
MrMage: I prefer `try!`, as that causes the test to crash on the line that caused the error. Otherwise it is much harder to figure out _on which line exactly_ the test failed. I have now replaced this with `XCTAssertNoThrow(try ` where possible, though.

MrMage:
I have addressed all comments now; please at least go through the ones I haven't marked yet resolved and resolve them if you are satisfied with the changes.

I have also added a short section to `README.md`; sample projects will probably come at some point, but I can't guarantee that I will be the one who will implement them. Also, I think the library should be more production-ready before we add sample projects.

Package.swift
```diff
@@ -18,8 +18,11 @@ import PackageDescription
 
 var packageDependencies: [Package.Dependency] = [
   .package(url: "https://github.com/apple/swift-protobuf.git", .upToNextMinor(from: "1.1.1")),
-  .package(url: "https://github.com/kylef/Commander.git", from: "0.8.0"),
-  .package(url: "https://github.com/apple/swift-nio-zlib-support.git", from: "1.0.0")
+  .package(url: "https://github.com/kylef/Commander.git", .upToNextMinor(from: "0.8.0")),
+  .package(url: "https://github.com/apple/swift-nio-zlib-support.git", .upToNextMinor(from: "1.0.0")),
+  .package(url: "https://github.com/apple/swift-nio.git", .upToNextMinor(from: "1.11.0")),
+  .package(url: "https://github.com/apple/swift-nio-nghttp2-support.git", .upToNextMinor(from: "1.0.0")),
+  .package(url: "https://github.com/apple/swift-nio-http2.git", .branch("master"))
```
rebello95: Let's pin this to avoid it changing from under us 😄 (we may need to issue a new release to pull in https://github.com/apple/swift-nio-http2/pull/27 // @Lukasa)

Sources/SwiftGRPCNIO/CallHandlers/BidirectionalStreamingCallHandler.swift
```diff
@@ -0,0 +1,44 @@
...
+
+/// Handles bidirectional streaming calls. Forwards incoming messages and end-of-stream events to the observer block.
+///
+/// - The observer block is implemented by the framework user and calls `context.sendResponse` as needed.
+/// - To close the call and send the status, fulfill `context.statusPromise`.
+public class BidirectionalStreamingCallHandler<RequestMessage: Message, ResponseMessage: Message>: BaseCallHandler<RequestMessage, ResponseMessage> {
+  public typealias EventObserver = (StreamEvent<RequestMessage>) -> Void
+  private var eventObserver: EventLoopFuture<EventObserver>?
+  
+  private var context: StreamingResponseCallContext<ResponseMessage>?
```
rebello95: Can we remove this? Looks like it should be captured by the closure below when it's used

Sources/SwiftGRPCNIO/CallHandlers/ClientStreamingCallHandler.swift
```diff
@@ -0,0 +1,43 @@
...
+import NIOHTTP1
+
+/// Handles client-streaming calls. Forwards incoming messages and end-of-stream events to the observer block.
+///
+/// - The observer block is implemented by the framework user and fulfills `context.responsePromise` when done.
+public class ClientStreamingCallHandler<RequestMessage: Message, ResponseMessage: Message>: BaseCallHandler<RequestMessage, ResponseMessage> {
+  public typealias EventObserver = (StreamEvent<RequestMessage>) -> Void
+  private var eventObserver: EventLoopFuture<EventObserver>?
+  
+  private var context: UnaryResponseCallContext<ResponseMessage>?
```
rebello95: Same in these other files

Sources/SwiftGRPCNIO/GRPCChannelHandler.swift
```diff
@@ -0,0 +1,69 @@
+import Foundation
+import SwiftProtobuf
+import NIO
+import NIOHTTP1
+
+/// Processes individual gRPC messages and stream-close events on a HTTP2 channel.
+public protocol GRPCCallHandler: ChannelHandler {
+  func makeGRPCServerCodec() -> ChannelHandler
```
rebello95: Can we document all the public functions/getters for these protocols?

Sources/SwiftGRPCNIO/GRPCServer.swift
```diff
@@ -0,0 +1,58 @@
...
+      .map { GRPCServer(channel: $0) }
+  }
+
+  private let channel: Channel
+
+  private init(channel: Channel) {
+    self.channel = channel
+  }
+
+  public var onClose: EventLoopFuture<Void> {
```
rebello95: Doesn't look like this is used anywhere

Sources/SwiftGRPCNIO/GRPCServerCodec.swift
```diff
@@ -0,0 +1,71 @@
...
+  public typealias InboundIn = RawGRPCServerRequestPart
+  public typealias InboundOut = GRPCServerRequestPart<RequestMessage>
+
+  public func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
+    switch self.unwrapInboundIn(data) {
+    case .head(let requestHead):
+      ctx.fireChannelRead(self.wrapInboundOut(.head(requestHead)))
+
+    case .message(var message):
+      let messageAsData = message.readData(length: message.readableBytes)!
```
rebello95: Should we handle this instead of force unwrapping?

MrMage: No. When the buffer says it has this many bytes readable, then we are guaranteed to read that many.

Sources/SwiftGRPCNIO/CallHandlers/BidirectionalStreamingCallHandler.swift
MrMage: No; the context should live for as long as the handler does. The closure, however, has a much shorter lifetime.

Sources/SwiftGRPCNIO/CallHandlers/ClientStreamingCallHandler.swift
MrMage: No; the context should live for as long as the handler does. The closure, however, has a much shorter lifetime.

Tests/SwiftGRPCNIOTests/NIOServerTests.swift
```diff
@@ -145,9 +145,9 @@ extension NIOServerTests {
       completionHandlerExpectation.fulfill()
     }
 
-    try! call.send(Echo_EchoRequest(text: "foo"))
-    try! call.send(Echo_EchoRequest(text: "bar"))
-    try! call.send(Echo_EchoRequest(text: "baz"))
+    XCTAssertNoThrow(try call.send(Echo_EchoRequest(text: "foo")))
```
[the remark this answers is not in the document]
MrMage: Yes, **but it will no longer tell you which line the failure occurred on.**

Sources/SwiftGRPCNIO/GRPCServer.swift
MrMage: This is for users of the library to be notified when the server shuts down. I've added a comment.

Sources/SwiftGRPCNIO/GRPCChannelHandler.swift
MrMage: Not sure what you mean with that, but I've added headerdocs for this protocol's methods now.

Package.swift
MrMage: Good idea! Done.

MrMage:
@rebello95 I hope I have addressed all your comments; please take another look.

Package.swift
MrMage: FYI, I should have pinned to a commit that includes that fix.

MrMage:
Great, thanks! I've been planning to squash all commits into one, what do you think?

rebello95:
Up to you if you want to do one or a couple, no strong opinion. Could make sense to break it into `core updates, tests, codegen` 🤷‍♂️

timburks:
LGTM - I think squashing this first big commit would be good.

weissi:
dance

Tests/SwiftGRPCNIOTests/NIOServerTests.swift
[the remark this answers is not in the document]
weissi: exactly, we've added `XCTAssertNoThrow` basically everywhere in NIO now for that reason. It's still missing in lots of places which is always a pain when debugging something that happened in CI.

MrMage:
Yeah, I saw that in one of your PRs. Sadly that can't be used when the return value of a throwing function is relevant, though.

weissi:
> Yeah, I saw that in one of your PRs. Sadly that can't be used when the return value of a throwing function is relevant, though.

yeah, return values are annoying with this. In NIO we have https://github.com/apple/swift-nio/blob/master/Tests/NIOTests/TestUtils.swift#L186 

    func assertNoThrowWithValue<T>(_ body: @autoclosure () throws -> T, defaultValue: T? = nil, message: String? = nil, file: StaticString = #file, line: UInt = #line) throws -> T {

PR #287, opened 2018-08-14
Fix cgrpc xcodeproj

### LINKED ISSUE #44: Xcode project configuration issues - modules, headers, etc.
It appears that module maps are present for each target but are not referenced by the generated Xcode project targets, and can probably be removed.

In order to "import CgRPC" from my main application, I had to:

- Set "Defines Module" to Yes on CgRPC target
- Add a "Headers" phase to CgRPC target build phases & add CgRPC.h as a public header

I'm not familiar with the Xcode project generation or I'd submit a PR.

Thanks for the continued great work! Happy to see it working in Xcode 8.3.2.

JonasVautherin:
Fixes #44 and #272 by setting "DEFINES_MODULE" to "YES" and by adding a Header phase in CgRPC after the `xcodeproj` is being created by SwiftPM.

@MrMage As discussed in #272 and [on the swift forum](https://forums.swift.org/t/generated-xcodeproj-for-c-language-target/15192), the root problem is probably the structure of the CgRPC target, and this fix may be removed in the future when CgRPC is fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,4 +58,4 @@ script:
   - make test-plugin
   - make test-echo
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then make xcodebuild; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then make clean && travis_wait 20 make build-carthage; fi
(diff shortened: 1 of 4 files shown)


MrMage:
Thanks a lot for this, Jonas! I'm glad we have a workable solution now.

PR #289, opened 2018-08-15
fix: lock deps to exact versions

anjmao:
fixes https://github.com/grpc/grpc-swift/issues/288

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -17,7 +17,7 @@
 import PackageDescription
 
 var packageDependencies: [Package.Dependency] = [
-  .package(url: "https://github.com/apple/swift-protobuf.git", upToNextMinor(from: "1.0.3")),
+  .package(url: "https://github.com/apple/swift-protobuf.git", .upToNextMinor(from: "1.0.3")),
   .package(url: "https://github.com/kylef/Commander.git", from: "0.8.0"),
   .package(url: "https://github.com/apple/swift-nio-zlib-support.git", from: "1.0.0")
 ]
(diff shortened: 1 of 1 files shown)


MrMage:
I think I would prefer to only version-lock the packages that need it, and only as much as needed. So for now, I would only lock `SwiftProtobuf` to `. upToNextMinor(from: "1.0.2")` (see https://github.com/apple/swift-package-manager/blob/master/Documentation/PackageDescriptionV4.md).

Let me know if you think differently.

anjmao:
@MrMage I think the most important thing is to ensure that grpc protoc plugin is used with the same version of swift grpc and swift protobuf. Currently this is how it looks:

https://github.com/grpc/grpc-swift/blob/master/SwiftGRPC.podspec#L44 In podspec version is
`s.dependency 'SwiftProtobuf', '~> 1.0.3`

In https://github.com/grpc/grpc-swift/blob/master/Package.swift#L21 version is `from: 1.0.2`

Since apple swift protobuf do not follow semver even having `upToNextMinor(from: "1.0.2")` can break things.

MrMage:
Do you have any documentation on SwiftProtobuf‘s versioning? Haven’t been broken by revisions since 0.9, and i wouldn’t expect the 1.0 minor to receive any breaking changes in the future given that 1.1 is already out, but it might still receive bug fixes.

Also, if you desire consistency with the CocoaPods, we could use `.upToNextMinor(from: "1.0.3")’? The end result will be the same anyway.

MrMage:
upToNextMinor should always stay below 1.1, so it should work. Why not just try it?

anjmao:
Yes it works, I updated PR, thanks. Now swift protobuf is both 1.0.3 in code generation and pod library.

PR #308, opened 2018-09-10
Add Package.swift for Swift 4.2

keith:
In order to allow dependent packages to build SwiftGRPC with Swift 4.2,
we have to use the new Package.swift format to define
`swiftLanguageVersions` that include the new enum values. In order to
remain backwards compatible with Xcode < 10 and Swift compilers before
4.2 on Linux we have to duplicate the Package.swift file and make one
specifically for 4.2.


keith:
It's probably worth waiting for a new release with this change https://github.com/kylef/Commander/pull/74 so we can update Commander for 4.2 as well

MrMage:
Sorry, but what are the advantages of building this package with Swift 4.2? Shouldn't the Swift compiler allow mixed builds, i.e. the library consumer is built with 4.2, but SwiftGRPC with 4.1?

keith:
Mix builds do work now, so this update isn't required to build dependent packages right now. I figured this was worth submitting because at some point you're likely going to want to update toolchains and move to this format instead

MrMage:
I think at that point it would make more sense to move the entire library to Swift 4.2. What do you think?

MrMage:
I think for now we don't need any features from Swift 4.2. Once we do, we should be able to drop support for older Swift versions anyway.

@timburks, WDYT?

keith:
Updated the PR to go down the path of requiring the 4.2 toolchain

MrMage:
I think it will be several more months until we want to go Swift 4.2 only. I would suggest closing this PR for now and re-opening when the time comes, to avoid accidentally merging it prematurely.

Issue #327, opened 2018-11-04
Crash during Channel destroying


slavabulgakov:
### New Issue Checklist

- [x] Updated grpc-swift to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/grpc/grpc-swift/blob/master/CONTRIBUTING.md)
- [x] I searched for [existing GitHub issues](https://github.com/grpc/grpc-swift/issues)(especially, there are many [resolved issues related to SSL](https://github.com/grpc/grpc-swift/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+ssl))

### Issue Description
When iOS tries to remove service client which inherited from ServiceClientBase, I get crash EXC_BAD_ACCESS in channel_stack.cc:84

<img width="583" alt="screenshot 2018-11-05 at 01 13 29" src="https://user-images.githubusercontent.com/1135591/47968736-0b0f4100-e098-11e8-937e-dd5018b639e0.png">

![image](https://user-images.githubusercontent.com/1135591/47968676-73115780-e097-11e8-8e10-60ddefa5cd15.png)

Crash goes because `cgrpc_channel_check_connectivity_state(self.underlyingChannel, 0)` or `cgrpc_channel_watch_connectivity_state(self.underlyingChannel, self.underlyingCompletionQueue, underlyingState, deadline, nil)` are called when underlyingChannel is already destroyed.

<img width="1258" alt="screenshot 2018-11-05 at 01 20 51" src="https://user-images.githubusercontent.com/1135591/47968841-802f4600-e099-11e8-94c3-633712b3dace.png">

I suggest to add `hasBeenShutdown` property to `ConnectivityObserver` which will be `true` when `shutdown()` is called and put `guard`s with `hasBeenShutdown` before calling`cgrpc_channel_check_connectivity_state` and `cgrpc_channel_watch_connectivity_state` in `spinloopThreadQueue`

### Environment

<details>
  <pre>
  
| Key                           | Value                                                         |
| ----------------------------- | --------------------------------------------------------------|
| OS Version                    |   12.1                                                       |
| Swift Version                 |   4.1                                                       |
| Xcode Version                 |   10.1                                                       |
| gRPC-Swift Version            |   0.6.0                                                       |
| protoc Version                |   3.6.1                                                       |
  </pre>
</details>

PR #328, opened 2018-11-04
Fix crash during Channel with subscription is destroying

slavabulgakov:
Fix for #327

The goal was to fix crash during Channel (which has subscription to connectivity state changes) is destroying

### DIFF
--- a/Sources/SwiftGRPC/Core/Channel.swift
+++ b/Sources/SwiftGRPC/Core/Channel.swift
@@ -132,6 +132,7 @@ private extension Channel {
     private let underlyingCompletionQueue: UnsafeMutableRawPointer
     private let callback: (ConnectivityState) -> Void
     private var lastState: ConnectivityState
+    private var hasBeenShutdown = false
 
     init(underlyingChannel: UnsafeMutableRawPointer, currentState: ConnectivityState, callback: @escaping (ConnectivityState) -> ()) {
       self.underlyingChannel = underlyingChannel
@@ -151,11 +152,20 @@ private extension Channel {
 
       spinloopThreadQueue.async {
         while true  {
+            
+          guard !self.hasBeenShutdown else {
+            return
+          }
+            
           guard let underlyingState = self.lastState.underlyingState else { return }
 
           let deadline: TimeInterval = 0.2
           cgrpc_channel_watch_connectivity_state(self.underlyingChannel, self.underlyingCompletionQueue, underlyingState, deadline, nil)
           let event = self.completionQueue.wait(timeout: deadline)
+          
+          guard !self.hasBeenShutdown else {
+            return
+          }
 
           switch event.type {
           case .complete:
@@ -178,6 +188,7 @@ private extension Channel {
     }
 
     func shutdown() {
+      hasBeenShutdown = true
       completionQueue.shutdown()
     }
   }


MrMage:
Thank you! Would it be possible to add a test case for this as well that would fail without your fixes?

Tests/SwiftGRPCTests/ChannelCrash/ChannelCrashProvider.swift
```diff
@@ -0,0 +1,27 @@
...
+//  SwiftGRPCTests
+//
+//  Created by Sviatoslav Bulgakov on 06/11/2018.
+//
+
+import Foundation
+import SwiftGRPC
+
+class ChannelCrashProvider: ChannelCrash_ChannelCrashProvider {
+
```
MrMage: Nit: please remove this newline.

Tests/SwiftGRPCTests/ChannelCrash/ChannelCrashProvider.swift
```diff
@@ -0,0 +1,27 @@
...
+//  ChannelCrashProvider.swift
+//  SwiftGRPCTests
+//
+//  Created by Sviatoslav Bulgakov on 06/11/2018.
+//
+
+import Foundation
+import SwiftGRPC
+
+class ChannelCrashProvider: ChannelCrash_ChannelCrashProvider {
```
MrMage: Could you pull this entire class into `ChannelCrashTests.swift`?

Tests/SwiftGRPCTests/ChannelCrash/Generated/channelCrash.grpc.swift
```diff
@@ -0,0 +1,108 @@
+//
```
MrMage: Could you use the existing generated code for `EchoService` instead, and move `ChannelCrashTests.swift` into the `SwiftGRPCTests` folder?

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -151,11 +152,20 @@ private extension Channel {
 
       spinloopThreadQueue.async {
         while true  {
+
```
MrMage: nit: please remove this newline

Tests/SwiftGRPCTests/ChannelCrashTests.swift
```diff
@@ -0,0 +1,83 @@
...
+                    print("expand error: \(error)")
+                }
+            }
+        }
+        return .ok
+    }
+}
+
+class ChannelCrashTests: XCTestCase {
+
```
MrMage: Sorry for noticing this only now, but could you please make this a subclass of https://github.com/grpc/grpc-swift/blob/master/Tests/SwiftGRPCTests/BasicEchoTestCase.swift? That should spare you from all the client/server setup work. Also, please remove this extra newline.

Tests/SwiftGRPCTests/ChannelCrashTests.swift
```diff
@@ -0,0 +1,83 @@
...
+        
+        DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(1)) {
+            self.client = nil
+        }
+
+        waitForExpectations(timeout: 5)
+    }
+    
+    private func receive(call: Echo_EchoExpandCall) {
+        try? call.receive { result in
```
MrMage: Please use `try!` and move this into the test method itself.

Tests/SwiftGRPCTests/ChannelCrashTests.swift
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+class ChannelCrashTests: XCTestCase {
+    
+    var provider = ChannelCrashProvider()
+    var server: ServiceServer!
+    var client: Echo_EchoServiceClient?
+
+    func testChannelCrash() {
```
MrMage: nit: Rename to `testDanglingConnectivityObserversDontCrash`?

Tests/SwiftGRPCTests/ChannelCrashTests.swift
```diff
@@ -0,0 +1,83 @@
...
+    }
+    
+    func update(session: Echo_EchoUpdateSession) throws -> ServerStatus? {
+        return .ok
+    }
+    
+    func expand(request: Echo_EchoRequest, session: Echo_EchoExpandSession) throws -> ServerStatus? {
+        let parts = request.text.components(separatedBy: " ")
+        for (i, part) in parts.enumerated() {
+            usleep(500000)
```
MrMage: Use `Thread.sleep` instead?

Tests/SwiftGRPCTests/ChannelCrashTests.swift
```diff
@@ -0,0 +1,83 @@
...
+        let request = Echo_EchoRequest(text: "foo bar baz foo bar baz")
+        let call = try! client!.expand(request) { callResult in
+            print("callResult.statusCode: \(callResult.statusCode)")
+            completionHandlerExpectation.fulfill()
+        }
+        
+        receive(call: call)
+        
+        DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(1)) {
+            self.client = nil
```
MrMage: Please add a comment that this is to ensure that the client gets deallocated. Also, can all the timeouts be shortened quite a bit, to avoid increasing the test runtime so much?

Tests/SwiftGRPCTests/ChannelCrashTests.swift
```diff
@@ -0,0 +1,83 @@
+//
```
MrMage: Please remove this comment block.

Tests/SwiftGRPCTests/ChannelCrash/channelCrash.proto
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2015, Google Inc.
```
MrMage: Please delete this file.

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -178,6 +187,7 @@ private extension Channel {
     }
 
     func shutdown() {
+      hasBeenShutdown = true
```
MrMage: As mentioned before, please synchronize accesses to this variable.

Tests/SwiftGRPCTests/ChannelCrashTests.swift
slavabulgakov: Do you suggest to move call "try! call.receive {...}" inside "testChannelCrash" method?
If yes then I cannot recursively call method "private func receive(call: Echo_EchoExpandCall)" as I do it now

MrMage: Could you instead use the blocking, synchronous variant of `receive`? Or leave out the `receive` part completely, given that it should not be relevant to the crash, anyway?

slavabulgakov: Agreed, it is not relevant to the crash. I will remove it.

PR #331, opened 2018-11-09
Add support for RPC status

### LINKED ISSUE #258: Metadata error
I was wondering if we were planning on supporting https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto

We came across this talk: https://schd.ws/hosted_files/kccncna17/ad/2017%20CloudNativeCon%20-%20Mod%20gRPC%20Services.pdf

SebastianThiebaud:
This closes #258 

The new changes provide a way for consumer of `grpc-swift` to get access to the data stored in the `grpc-status-details-bin` of metadata. 

Here's how to use it in the consumer app:
```
// Register message types that could be returned in the details field of Status
Google_Protobuf_Any.register(messageType: CustomError.self)

_ = try service.doSomething(request) { response, result in
  result.trailingMetadata?.status?.details // will contain a list of messages that carry the error details
}
```

### DIFF
--- a/Sources/SwiftGRPC/Core/Metadata.swift
+++ b/Sources/SwiftGRPC/Core/Metadata.swift
@@ -80,7 +80,16 @@ public class Metadata {
     defer { cgrpc_free_copied_string(valueData) }
     return String(cString: valueData, encoding: String.Encoding.utf8)
   }
-  
+
+  public func data(forKey key: String) -> Data? {
+    for index in 0..<count() {
+      guard self.key(index) == key else { continue }
+      let byteBuffer = ByteBuffer(underlyingByteBuffer: cgrpc_metadata_array_copy_data_value_at_index(underlyingArray, index))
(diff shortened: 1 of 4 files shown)


SebastianThiebaud:
@MrMage I added a simple test.

Sources/SwiftGRPC/Core/Metadata.swift
```diff
@@ -80,7 +80,16 @@ public class Metadata {
     defer { cgrpc_free_copied_string(valueData) }
     return String(cString: valueData, encoding: String.Encoding.utf8)
   }
-  
+
+  public func data(forKey key: String) -> Data? {
```
MrMage: nit: I think it would make more sense to move this into the extension at the bottom of the file, next to the subscript method.

SebastianThiebaud: done

PR #341, opened 2018-12-09
Add environment flag for BoringSSL/OpenSSL on MacOS

jgh-:
Sometimes one wants to use dependencies that rely on OpenSSL, and for those times updating build scripts is easier than forking projects.

This change adjusts the behavior of the OpenSSL vs BoringSSL selection in Package.swift.  If you're not on Linux and the environment variable `GRPC_USE_OPENSSL`  is present, it will use OpenSSL. If not it will use BoringSSL as usual.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -26,16 +26,16 @@ var packageDependencies: [Package.Dependency] = [
 var cGRPCDependencies: [Target.Dependency] = []
 
 #if os(Linux)
-// On Linux, Foundation links with openssl, so we'll need to use that instead of BoringSSL.
-// See https://github.com/apple/swift-nio-ssl/issues/16#issuecomment-392705505 for details.
 let isLinux = true
 #else
 let isLinux = false
 #endif 
 
+// On Linux, Foundation links with openssl, so we'll need to use that instead of BoringSSL.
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -23,13 +24,22 @@ var packageDependencies: [Package.Dependency] = [
...
+let isLinux = true
 #else
-cGRPCDependencies.append("BoringSSL")
-#endif
+let isLinux = false
+#endif 
+
+// swift build doesn't pass -Xswiftc flags to dependencies, so here using an environment variable
+// is easiest.
+if ProcessInfo.processInfo.environment.keys.contains("GRPC_USE_OPENSSL") || isLinux {
```
MrMage: Super-nit: could you move `isLinux` before the other check, as it is the more important check, and move the comment from above down here?

```
// On Linux, Foundation links with openssl, so we'll need to use that instead of BoringSSL.
 // See https://github.com/apple/swift-nio-ssl/issues/16#issuecomment-392705505 for details.
```

jgh-: done, lmk if it's still not right.

MrMage:
Awesome, thanks!

PR #347, opened 2018-12-21
Support proto-gen-swift's proto-to-module mapping.

allevato:
This feature lets us support build systems like Bazel where `.proto` files are spread out among multiple `proto_library` targets, and each such target becomes its own Swift module. Downstream modules (and in our case, generated service code) must know which module has the proto definitions that they depend on so that it can import them. This information is conveyed using a [text-proto-formatted file](https://github.com/apple/swift-protobuf/blob/master/Documentation/PLUGIN.md#generation-option-protopathmodulemappings---swift-module-names-for-proto-paths) that Bazel will generate as it traverses the build graph.

This will let us create a `swift_grpc_library` rule that lives in [bazelbuild/rules_swift](https://github.com/bazelbuild/rules_swift) alongside [`swift_proto_library`](https://github.com/bazelbuild/rules_swift/blob/master/swift/internal/swift_proto_library.bzl).


allevato:
Thank you for the quick review, @MrMage! Is there a chance a new release could be cut soon that includes this change so that we can add it as a dependency of bazelbuild/rules_swift?

rebello95:
@allevato just tagged 0.7.0 including your change: https://github.com/grpc/grpc-swift/releases/tag/0.7.0

Podspec update here as well: https://github.com/grpc/grpc-swift/pull/349

allevato:
Thank you!

PR #350, opened 2018-12-30
Add gRPC Web support.

sergiocampama:
Initial implementation for supporting gRPC Web directly from
Swift gRPC services.


sergiocampama:
The error in Travis is because of https://github.com/apple/swift-nio-http2/pull/36.

sergiocampama:
Added CORS management, with the latest version I can get the gRPC [demo app](https://github.com/grpc/grpc-web/tree/master/net/grpc/gateway/examples/helloworld) to work without the Envoy proxy.

MrMage:
This is great, thank you so much! I was hoping that using NIO'S HTTP1-based primitives would make it fairly easy to add support for gRPC-Web/pPRC, and it looks like that is the case.

I also really appreciate the sample code for serving via NIO; this had been on my Todo list for a while.

I'll give this a thorough review once I return from the holidays. Until then, two questions:

1. Could you give me two sentences of instructions to test this myself with the demo app to make the review easier for me? I have no experience with the gRPC-Web codebase.
2. Given that gRPC-Web aims to support gRPC via vanilla HTTP1 requests (if I'm not mistaken), do you think it would be possible to add some simple test cases (using e.g. NSURLSession or other Foundation HTTP code to send the requests in the tests; if you encounter problems getting those tests to pass on Linux simply run them on macOS only) for all the request methods supported by our gRPC-Web implementation?

Hope this makes sense; let me know what you think. Again, thanks a lot and will review this properly soon!

sergiocampama:
Thanks! I wasn't expecting an expedite review around the holidays :) 

I added a new commit that adds the `EchoWeb` example. You should just need to install `npm` and then just run `make`.

Even with this PR, you'll still need to patch apple/swift-nio-http2#36 locally. What I did was to do a `swift build --product Echo` to populate dependencies, and then modify the `.build/checkouts/swift-nio-http2.git-8449925081683998237/Sources/NIOHTTP2/HTTP2Parser.swift` file manually to add the patch. With this, you should be able to run `swift run Echo serve_nio` and then open the `EchoWeb/index.html` file in a browser. 

RE: Tests, that should be fairly straight forward, I'll look at adding them over the next few days.

Happy holidays!

Sources/Examples/EchoNIO/EchoProviderNIO.swift
```diff
@@ -0,0 +1,71 @@
...
+        parts.append(message.text)
+
+      case .end:
+        var response = Echo_EchoResponse()
+        response.text = "Swift echo collect: " + parts.joined(separator: " ")
+        context.responsePromise.succeed(result: response)
+      }
+    })
+  }
+  func update(context: StreamingResponseCallContext<Echo_EchoResponse>) -> EventLoopFuture<(StreamEvent<Echo_EchoRequest>) -> Void> {
```
MrMage: nit: newline above this line

Sources/Examples/EchoNIO/EchoProviderNIO.swift
```diff
@@ -0,0 +1,71 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import Foundation
+import NIO
+import SwiftGRPCNIO
+
+class EchoProviderNIO: Echo_EchoProvider_NIO {
+
```
MrMage: nit: would you mind removing this newline?

Sources/Examples/EchoNIO/Generated/echo.grpc.swift
```diff
@@ -0,0 +1,73 @@
+//
```
MrMage: Could you use a symlink to the existing variant of this file (somewhere in `Tests/SwiftGRPCNIOTests`)? Or convert that file into a symlink pointing to this one (depending on how we've been handling this for the non-NIO case)?

Sources/Examples/EchoNIO/Generated/echo.pb.swift
```diff
@@ -0,0 +1,132 @@
+// DO NOT EDIT.
```
MrMage: Also symlink this, please (see the `.grpc.swift` file).

Sources/Examples/EchoNIO/README.md
```diff
@@ -0,0 +1,6 @@
+# EchoNIO, a gRPC NIO Sample App
+
+This directory contains a simple echo server that demonstrates
+all four gRPC API styles (Unary, Server Streaming, Client
+Streaming, and Bidirectional Streaming) using the NIO based
+Swift gRPC implementation
```
MrMage: super-nit: Please add a full stop at the end.

Sources/Examples/EchoNIO/echo.proto
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) 2015, Google Inc.
```
MrMage: Also symlink this, please.

Sources/Examples/EchoNIO/main.swift
```diff
@@ -0,0 +1,56 @@
...
+
+// Common flags and options
+func addressOption(_ address: String) -> Option<String> {
+  return Option("address", default: address, description: "address of server")
+}
+
+let portOption = Option("port",
+                        default: "8080",
+                        description: "port of server")
+let messageOption = Option("message",
```
MrMage: Is this being used anywhere?

Sources/Examples/EchoNIO/main.swift
```diff
@@ -0,0 +1,56 @@
...
+    let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+
+    print("starting insecure server")
+    let echoServer = try! GRPCServer.start(hostname: address,
+                                          port: Int(port)!,
+                                          eventLoopGroup: eventLoopGroup,
+                                          serviceProviders: [EchoProviderNIO()])
+      .wait()
+
+    // This blocks to keep the main thread from finishing while the server runs,
```
MrMage: Thanks for explaining these two lines with comments 👍

Sources/SwiftGRPCNIO/GRPCChannelHandler.swift
```diff
@@ -55,9 +55,8 @@ extension GRPCChannelHandler: ChannelInboundHandler {
           return
       }
 
-      var responseHeaders = HTTPHeaders()
-      responseHeaders.add(name: "content-type", value: "application/grpc")
-      ctx.write(self.wrapOutboundOut(.headers(responseHeaders)), promise: nil)
+      // Send empty headers to start the outbound header workflow.
```
MrMage: Would you mind explaining this change in more detail?

Sources/SwiftGRPCNIO/GRPCHTTPProtocolSwitcher.swift
```diff
@@ -0,0 +1,97 @@
...
+import NIO
+import NIOHTTP1
+import NIOHTTP2
+
+/// Channel handler that creates different processing pipelines depending on whether
+/// the incoming request is HTTP 1 or 2.
+class GRPCHTTPProtocolSwitcher {
+  private let handlersInitializer: ((Channel) -> EventLoopFuture<Void>)
+
+  public init(handlersInitializer: (@escaping (Channel) -> EventLoopFuture<Void>)) {
```
MrMage: Would you mind making the entire class `public` or `open` as well, so low-level users of this library can make use of it as well?

Sources/SwiftGRPCNIO/GRPCHTTPProtocolSwitcher.swift
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    ctx.fireChannelRead(data)
+    _ = ctx.pipeline.remove(ctx: ctx)
+  }
+
+  /// Peek into the first line of the packet to check which HTTP version is being used.
+  private func protocolVersion(_ preamble: String) -> HTTPProtocolVersion? {
+    let range = NSRange(location: 0, length: preamble.utf16.count)
```
MrMage: Using all these `NS` types and UTF16-based strings here is necessary because there is no regular expression support in the Swift standard library I assume?

Sources/SwiftGRPCNIO/GRPCHTTPProtocolSwitcher.swift
```diff
@@ -0,0 +1,97 @@
...
+class GRPCHTTPProtocolSwitcher {
+  private let handlersInitializer: ((Channel) -> EventLoopFuture<Void>)
+
+  public init(handlersInitializer: (@escaping (Channel) -> EventLoopFuture<Void>)) {
+    self.handlersInitializer = handlersInitializer
+  }
+}
+
+extension GRPCHTTPProtocolSwitcher: ChannelInboundHandler {
+  typealias InboundIn = ByteBuffer
```
MrMage: This switcher seems to operate on a fairly low level and could be useful for other services supporting both HTTP1 and HTTP2; should we maybe try to get this merged into http://github.com/apple/swift-nio-http2 so others could profit from it, or at least drop the `GRPC` from the class name?

Sources/SwiftGRPCNIO/GRPCWebCorsHandler.swift
```diff
@@ -0,0 +1,73 @@
+import NIO
+import NIOHTTP1
+
+/// Handler that manages the CORS protocol for requests incoming from the browser.
+class GRPCWebCorsHandler {
```
MrMage: Super-nit: Should we spell `CORS` all-uppercase? We've been doing that with e.g. `GRPC`, but not sure which way is idiomatic.

MrMage: Also, make this `public` or even `open`? Besides the `x-grpc-web` header, this looks fairly non-GRPC-specific as well; should we drop the `GRPC` from the name?

Sources/SwiftGRPCNIO/GRPCWebCorsHandler.swift
```diff
@@ -0,0 +1,73 @@
...
+        headers.add(name: "Access-Control-Allow-Headers",
+                    value: "content-type,x-grpc-web,x-user-agent")
+        headers.add(name: "Access-Control-Max-Age", value: "86400")
+        ctx.write(self.wrapOutboundOut(.head(HTTPResponseHead(version: requestHead.version,
+                                                              status: .ok,
+                                                              headers: headers))),
+                  promise: nil)
+        return
+      }
+    case .body(_):
```
MrMage: super-nit: newline above this line, and (optionally) omit `(_)`?

Sources/SwiftGRPCNIO/GRPCWebCorsHandler.swift
```diff
@@ -0,0 +1,73 @@
...
+  typealias OutboundIn = HTTPServerResponsePart
+
+  public func write(ctx: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
+    let responsePart = self.unwrapOutboundIn(data)
+    switch responsePart {
+    case .head(let responseHead):
+      var headers = responseHead.headers
+      // CORS requires all requests to have an Allow-Origin header.
+      headers.add(name: "Access-Control-Allow-Origin", value: "*")
+      //! FIXME: Check whether we can let browsers to keep connections alive. It's not possible
```
MrMage: nit: remove `to`?

Sources/SwiftGRPCNIO/GRPCHTTPProtocolSwitcher.swift
```diff
@@ -0,0 +1,97 @@
...
+    // Depending on whether it is HTTP1 or HTTP2, created different processing pipelines.
+    // Inbound handlers in handlersInitializer should expect HTTPServerRequestPart objects
+    // and outbound handlers should return HTTPServerResponsePart objects.
+    switch version {
+    case .http1:
+      // Upgrade connections are not handled since gRPC connections already arrive in HTTP2,
+      // while gRPC-Web does not support HTTP2 at all, so there are no compelling use cases
+      // to support this.
+      _ = ctx.pipeline.configureHTTPServerPipeline(withErrorHandling: true)
+        .then { ctx.pipeline.add(handler: GRPCWebCorsHandler()) }
```
MrMage: I assume we do not need to add the `CORS` handler to `HTTP2` requests because we always expect those to be from a "true" gRPC client and not via Javascript from within a browser? Does gRPC-Web guarantee to never send its requests via HTTP2 (in which case we might need CORS support on the HTTP2 path as well)?

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -58,7 +65,9 @@ extension HTTP1ToRawGRPCServerCodec: ChannelInboundHandler {
 
       state = .expectingCompressedFlag
       buffer = ctx.channel.allocator.buffer(capacity: 5)
-
+      if let contentTypeHeader = requestHead.headers["Content-Type"].first {
```
MrMage: Should `"Content-Type"` be lowercased? (I think at least in the HTTP2 spec all headers are lowercase.)

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -67,6 +76,16 @@ extension HTTP1ToRawGRPCServerCodec: ChannelInboundHandler {
       assert(state.expectingBody, "received body while in state \(state)")
       buffer.write(buffer: &body)
 
+      // If the contentType is text, then decode the incoming data as Base64, and continue
+      // processing normally after decoding.
+      if contentType == .text {
```
MrMage: I think this will break down if the request is chunked (and I think all HTTP clients will start chunking for sufficiently large HTTP requests; you might want to add a test for that). Example:

1. First chunk is 400 bytes Base64, gets decoded to 300 bytes binary.
2. Second chunk (also Base64) arrives, is appended to the 300 binary bytes. The result is half-binary (first chunk), half-base64 (second chunk).

Let me know if I am missing something.

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -113,19 +132,40 @@ extension HTTP1ToRawGRPCServerCodec: ChannelInboundHandler {
...
+      ctx.write(self.wrapOutboundOut(.head(HTTPResponseHead(version: version, status: .ok, headers: headers))), promise: promise)
     case .message(var messageBytes):
       // Write out a length-delimited message payload. See `channelRead` fpor the corresponding format.
       var responseBuffer = ctx.channel.allocator.buffer(capacity: messageBytes.readableBytes + 5)
       responseBuffer.write(integer: Int8(0))  // Compression flag: no compression
       responseBuffer.write(integer: UInt32(messageBytes.readableBytes))
       responseBuffer.write(buffer: &messageBytes)
+
+      if contentType == .text {
+        if let binaryData = responseBuffer.readData(length: responseBuffer.readableBytes) {
```
MrMage: Nit: I think there is a method called `readAllAvailable()` (or something similar) on `buffer` that you could use instead of `readData(length: .readableBytes)`.

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -113,19 +132,40 @@ extension HTTP1ToRawGRPCServerCodec: ChannelInboundHandler {
...
+          let encodedData = binaryData.base64EncodedString()
+          responseBuffer = ctx.channel.allocator.buffer(capacity: encodedData.utf8.count)
+          responseBuffer.write(string: encodedData)
```
MrMage: If the current `responseBuffer` length has a length not divisible by three, will this inject `==` signs in the middle of the stream? Could that cause any problems?

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -113,19 +132,40 @@ extension HTTP1ToRawGRPCServerCodec: ChannelInboundHandler {
...
       // Write out a length-delimited message payload. See `channelRead` fpor the corresponding format.
       var responseBuffer = ctx.channel.allocator.buffer(capacity: messageBytes.readableBytes + 5)
       responseBuffer.write(integer: Int8(0))  // Compression flag: no compression
       responseBuffer.write(integer: UInt32(messageBytes.readableBytes))
       responseBuffer.write(buffer: &messageBytes)
+
+      if contentType == .text {
+        if let binaryData = responseBuffer.readData(length: responseBuffer.readableBytes) {
+          let encodedData = binaryData.base64EncodedString()
+          responseBuffer = ctx.channel.allocator.buffer(capacity: encodedData.utf8.count)
```
MrMage: nit: Is there a more efficient way to directly generate a new buffer from `encodedData`? (If not, please disregard.)

Tests/SwiftGRPCNIOTests/NIOServerTests.swift
```diff
@@ -300,3 +300,89 @@ extension NIOServerTests {
     waitForExpectations(timeout: defaultTimeout)
   }
 }
+
+// Only test Unary and ServerStreaming, as ClientStreaming is not
+// supported in HTTP1.
+extension NIOServerTests {
```
MrMage: Would it make sense to put these tests into a separate file with a separate test case object, WDYT?

Tests/SwiftGRPCNIOTests/NIOServerTests.swift
```diff
@@ -300,3 +300,89 @@ extension NIOServerTests {
...
+  private func gRPCEncodedMessage(_ string: String) -> Data {
+    var data = string.data(using: .utf8)!
+    data.insert(contentsOf: [UInt8]([0x0a, UInt8(data.count)]), at: 0)
```
MrMage: nit: Please explain the `0xa` byte, and the `0x00` ones in the line below.

Tests/SwiftGRPCNIOTests/NIOServerTests.swift
```diff
@@ -300,3 +300,89 @@ extension NIOServerTests {
...
+    data.insert(contentsOf: [UInt8]([0x00, 0x00, 0x00, 0x00, UInt8(data.count)]), at: 0)
+    return data
+  }
+  
+  private func sendOverHTTP1(rpcMethod: String, message: String, handler: @escaping (Data?, Error?) -> Void) {
+    let serverURL = URL(string: "http://localhost:5050/echo.Echo/\(rpcMethod)")!
+    var request = URLRequest(url: serverURL)
+    request.httpMethod = "POST"
+    request.setValue("application/grpc-web-text", forHTTPHeaderField: "Content-Type")
```
MrMage: I assume `application/grpc-web-text` will be the more common use case than `application/grpc-web`?

MrMage: Should `Content-Type` be lowercased?

Sources/Examples/EchoNIO/EchoProviderNIO.swift
sergiocampama: Done

Sources/Examples/EchoNIO/EchoProviderNIO.swift
sergiocampama: Done

Sources/Examples/EchoNIO/Generated/echo.grpc.swift
sergiocampama: Done

Sources/Examples/EchoNIO/Generated/echo.pb.swift
sergiocampama: Done

Sources/Examples/EchoNIO/README.md
sergiocampama: Done

Sources/Examples/EchoNIO/echo.proto
sergiocampama: Done

Sources/Examples/EchoNIO/main.swift
sergiocampama: Removed

Sources/Examples/EchoNIO/main.swift
sergiocampama: Copied from the tests 😛

simplified as we don't need the `echoServer` variable for now

Sources/SwiftGRPCNIO/GRPCChannelHandler.swift
sergiocampama: Added more, please let me know if it makes sense. I tried modifying the architecture so that the headers are sent from `HTTP1ToRawGRPCServerCode`, but it was getting to complex.

Sources/SwiftGRPCNIO/GRPCHTTPProtocolSwitcher.swift
sergiocampama: done

Sources/SwiftGRPCNIO/GRPCHTTPProtocolSwitcher.swift
sergiocampama: Done.

I'd like for this to be provided by `NIOHTTP2`, but I didn't want to wait for it for implementing this. @Lukasa, wdyt?

Sources/SwiftGRPCNIO/GRPCHTTPProtocolSwitcher.swift
sergiocampama: gRPC web exists because browsers don't support HTTP2 :) If/when that happens, we can add the CORS to the HTTP2 pipeline too. I can't think of any case where an HTTP2 client would implement CORS, since it's a security feature for browsers.

Sources/SwiftGRPCNIO/GRPCHTTPProtocolSwitcher.swift
sergiocampama: yup, hate it, but didn't want to hardcode if checks

Sources/SwiftGRPCNIO/GRPCWebCorsHandler.swift
sergiocampama: done

sergiocampama: done

Sources/SwiftGRPCNIO/GRPCWebCorsHandler.swift
sergiocampama: done

Sources/SwiftGRPCNIO/GRPCWebCorsHandler.swift
sergiocampama: done

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
sergiocampama: the headers object doesn't care for the case of the field, lowercased to be consistent.

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
sergiocampama: I'll have to test more, haven't thought of a chunked mechanism, will keep open for now

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
sergiocampama: Couldn't find it in NIO, do you remember where you saw it?

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
sergiocampama: Reused the original buffer to avoid another malloc

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
sergiocampama: No issues on the string's I've tried

Tests/SwiftGRPCNIOTests/NIOServerTests.swift
sergiocampama: Done

Tests/SwiftGRPCNIOTests/NIOServerTests.swift
sergiocampama: Done

Tests/SwiftGRPCNIOTests/NIOServerTests.swift
sergiocampama: I think so? The Javascript library does `-text`. Still support for web could be improved/tested more. Want me to add a FIXME?

sergiocampama: Same as other

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
MrMage: Could you add a `TODO` to add tests and/or investigate this more?

Sources/SwiftGRPCNIO/GRPCChannelHandler.swift
MrMage: Hmm, is it possible to send multiple header frames in HTTP1 without problems?

Sources/SwiftGRPCNIO/GRPCHTTPProtocolSwitcher.swift
MrMage: SGTM; leaving open for @Lukasa to comment.

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
MrMage: Can't find it anymore; I was probably mistaken, sorry.

Tests/SwiftGRPCNIOTests/NIOServerTests.swift
MrMage: Yes, please. Or just a `TODO`, whatever you think appropriate :-)

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
MrMage: Might be necessary to have two buffers; one for Base64 input and another one for the already-decoded (binary) parts.

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -113,19 +132,40 @@ extension HTTP1ToRawGRPCServerCodec: ChannelInboundHandler {
...
+          responseBuffer.write(string: encodedData)
+        }
+      }
```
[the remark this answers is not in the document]
sergiocampama: You're correct. Had to process the messages in a different way, since to have the base64 encoding properly aligned, I had to receive all of the streaming messages plus the trailers in order to aggregate them together and calculate the base64 of the complete response.

I realize that the complexity of HTTP1ToRawGRPC is getting out of hand, so after this is merged, the first thing I'll do is extract the grpcweb base64 wrapping into a different handler and only add it for the HTTP1.1 pipeline. I've added a todo to do that.

Tests/SwiftGRPCNIOTests/NIOServerTests.swift
sergiocampama: Added

sergiocampama:
https://github.com/apple/swift-nio-http2/pull/36 was merged, we'll need to wait for a release now

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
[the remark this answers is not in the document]
MrMage: I like the idea of doing the base64 decoding/encoding in a separate handler; that's a great way to abstract this away.

Tests/SwiftGRPCNIOTests/NIOServerWebTests.swift
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  private func gRPCEncodedMessage(_ string: String) -> Data {
+    // Construct an EchoRequest message manually, by setting 0x0a as the field byte,
+    // corresponding to 0b(00001)(010) where the first part is the field number (1) and
+    // the second part corresponds to the field type (length delimited), and finally wrapping
+    // the proto message in the gRPC compression/length wrapper (5 bytes).
+    var data = string.data(using: .utf8)!
+    data.insert(contentsOf: [UInt8]([0x0a, UInt8(data.count)]), at: 0)
+    data.insert(contentsOf: [UInt8]([0x00, 0x00, 0x00, 0x00, UInt8(data.count)]), at: 0)
```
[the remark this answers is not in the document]
MrMage: Duh, I didn't thank of that. Thanks for pointing that out.

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -43,29 +50,64 @@ public final class HTTP1ToRawGRPCServerCodec {
...
-      buffer.write(buffer: &body)
+
+      // If the contentType is text, then decode the incoming bytes as base64 encoded, and append
+      // it to the binary buffer. If the request is chunked, this section will process the text
+      // in the biggest chunk that is multiple of 4, leaving the unread bytes in the textBuffer
+      // where it will expect a new incoming chunk.
+      if contentType == .text {
+        precondition(requestTextBuffer != nil)
+        requestTextBuffer.write(buffer: &body)
+        // Read in chunks of 4 bytes , as base64 encoded strings will always be multiples of 4.
```
MrMage: super-nit: extraneous comma.

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -43,29 +50,64 @@ public final class HTTP1ToRawGRPCServerCodec {
...
+        let readyBytes = requestTextBuffer.readableBytes - (requestTextBuffer.readableBytes % 4)
+        if let base64Encoded = requestTextBuffer.readString(length: readyBytes),
+          let decodedData = Data(base64Encoded: base64Encoded) {
```
MrMage: Should we cancel the call if the data can not be decoded as base64?

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -113,25 +155,82 @@ extension HTTP1ToRawGRPCServerCodec: ChannelInboundHandler {
...
+      if let contentType = contentType {
+        headers.add(name: "content-type", value: contentType.rawValue)
+        switch contentType {
+        case .text, .web:
+          version = .init(major: 1, minor: 1)
+        case .binary:
+          version = .init(major: 2, minor: 0)
+        }
+      } else {
+        version = .init(major: 2, minor: 0)
```
MrMage: How about setting the content type to `.binary` if no header is present, to avoid logic like this?

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -113,25 +155,82 @@ extension HTTP1ToRawGRPCServerCodec: ChannelInboundHandler {
...
       responseBuffer.write(integer: UInt32(messageBytes.readableBytes))
       responseBuffer.write(buffer: &messageBytes)
-      ctx.write(self.wrapOutboundOut(.body(.byteBuffer(responseBuffer))), promise: promise)
+
+      if contentType == .text {
+        precondition(responseTextBuffer != nil)
+        // Store the response into an independent buffer. We can't return the message directly as
+        // it needs to be aggregated with all the responses plus the trailers, in order to have
+        // the base64 response properly encoded in a single byte stream.
+        responseTextBuffer!.write(buffer: &responseBuffer)
```
MrMage: I understand the necessity to cache all responses to ensure a well-formed base64 stream, but doesn't buffering all of them defeat the purpose of streaming? How about instead trying to send all but the last `length - (length % 3)` bytes, as those should always result in four well-formed base64 bytes? Please at least consider that when refactoring the base64 handling out of this class in a separate PR.

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -113,25 +155,82 @@ extension HTTP1ToRawGRPCServerCodec: ChannelInboundHandler {
...
+        precondition(responseTextBuffer != nil)
+
+        // Encode the trailers into the response byte stream as a length delimited message, as per
+        // https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md
+        let textTrailers = trailers.map { name, value in "\(name): \(value)" }.joined(separator: "\r\n")
+        responseTextBuffer.write(integer: UInt8(0x80))
+        responseTextBuffer.write(integer: UInt32(textTrailers.utf8.count))
+        responseTextBuffer.write(string: textTrailers)
+
+        // TODO: Binary responses that are non multiples of 3 will end = or == when encoded in
```
MrMage: I think if the `=`s only appear at the end of the stream (but not in the middle), we should be fine. Not an expert on gRPC-Web though, so thanks for the TODO.

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -16,6 +17,12 @@ public enum RawGRPCServerResponsePart {
   case status(GRPCStatus)
 }
 
+enum ContentType: String {
+  case binary = "application/grpc"
```
MrMage: Please add headerdocs for the enum overall and the individual cases, maybe with a short explanation of the different stream semantics.

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
sergiocampama: closed the connection, but not sure how to handle it properly from the server side, or what to send to the client, since at this point the headers have already been sent. should we wait until the request has been fully decoded before starting to send headers? (out of scope for this change though)

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
sergiocampama: DOne

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
sergiocampama: Done.

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
sergiocampama: Isn't that the issue with HTTP1.1, that there is no proper server streaming? I may be wrong though. if we did do that, we'd still have an issue with some streams, as if there are 2 bytes missing for a full base64 frame, we'd have to wait until the trailers or the next messages to arrive before sending it, otherwise we'd be sending incomplete messages. Perhaps that's ok? not sure

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
sergiocampama: Ack :)

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
sergiocampama: Done.

sergiocampama:
Set the commit with the fix in Package.swift as well

wenbozhu:
I have been working on grpc/grpc-web and the grpc-web spec. Tim asked me to have a look at this - and thanks for doing it!

I don't really know swift ... but I'd like to propose a few general requirements here from the point view of grpc-web.

1. Use the "standard" HTTP or web framework in Swift to handle the grpc-web traffic. I.e.  handle the HTTP traffic using the high-level server API provided by the language platform, which may support only HTTP/1.1 or both HTTP/1.1 and HTTP/2.
2. Proxy the grpc-web traffic using something like in-process channels (if supported) or domain sockets. I.e. make this an in-process proxy, v.s. some kind of alternative "transport" to grpc runtime. 
3. Implement only the core spec and a minimum set of [browser-specific features](https://github.com/grpc/grpc-web/blob/master/BROWSER-FEATURES.md). The latter will evolve, e.g. security features,  JSON-compatible text encoding for protobuf, keep-alive support etc.

Could you comment on how you would support the above requirements in grpc-swift, or if you have any concerns or questions on any of the requirements?

Thanks.

MrMage:
Thanks for the insights, @wenbozhu! The point of the way we implemented the gRPC protocol is actually that it is based on Swift‘s recommended way of handling HTTP. This also lets us handle HTTP1 and HTTP2 almost identically, avoiding wasted effort due to duplicated code. So I guess that would have us conform to point 1.

Could you please elaborate more on points 2 and 3, however? I don’t quite understand yet what they mean, and what they are opposed to, i.e. how things should _not_ be done.

MrMage:
Also, could you elaborate how points 1 and 2 are related? Point 1 sounds to me like „use the language-specific HTTP serving functionality“, while point 2 sounds like „use the gRPC-Core runtime“ which would be the opposite. I’m probably misunderstanding one of the two points.

sergiocampama:
Re: point 3, this PR already supports CORS. Missing is the compression section. Could you elaborate on what [HTTP status code mapping](https://github.com/grpc/grpc-web/blob/master/BROWSER-FEATURES.md#http-status-code-mapping) means?

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
MrMage: Not sure, either. Thanks for adding the `FIXME`!

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
MrMage: HTTP1.1 already has [chunking](https://en.wikipedia.org/wiki/Chunked_transfer_encoding), which is essentially server streaming. Also, I think most gRPC clients should be capable of waiting for the next chunk mid-message, as the gRPC spec says that clients should be chunking-agnostic. I'd assume the same applies for HTTP1.1 — the client just receives a byte stream and waits for a full message to arrive before trying to decode it.

Would you mind adding a `FIXME` for this?

sergiocampama:
Thanks!

Lukasa:
Incidentally, the fix has landed in swift-nio-http2 0.2.1.

sergiocampama:
even better, thanks @Lukasa!

wenbozhu:
So point 1 shouldn't be a concern for grpc-swift. 

Point 2 is about the request path using a grpcs-swift client to call into the grpc server end-point. I.e. some HTTP handler terminates the browser request, parses the data, and then uses a generic grpc client to call into the server. This might introduce some overhead but will make the whole request path more compatible (to the server). I am open to hear about objections, esp. for a gRPC implementation which uses the language-platform provided HTTP stack (for HTTP/2). 

Point 3 is about not implementing advanced features and leaving those features to the "official" grpc-web proxy. The thinking is that most web-tier deployment involves a reverse proxy, and we are expecting Envoy will dominate, e.g. with k8s etc. I.e. in-process grpc-web support (per language) only implements a minimum set of mandatory features to enable grpc-web but leave to Envoy for the production support of grpc-web clients. If this make sense, we can start to identify those features on the spec doc.

MrMage:
> Point 2 is about the request path using a grpcs-swift client to call into the grpc server end-point. I.e. some HTTP handler terminates the browser request, parses the data, and then uses a generic grpc client to call into the server. This might introduce some overhead but will make the whole request path more compatible (to the server). I am open to hear about objections, esp. for a gRPC implementation which uses the language-platform provided HTTP stack (for HTTP/2).

Sorry, I still don't understand. I don't think we have plans to support gRPC-Web on the client-side in SwiftGRPC, only on the server-side. How would this affect us, then?

FYI, on the server-side, we use a (SwiftNIO-)framework-provided HTTP2-to-HTTP1 mapper such that our gRPC implementation only needs to deal with HTTP1 primitives, which are easier to reason about and give us gRPC-Web support for (almost) free.

sergiocampama:
1: OK

2: Isn't that an implementation detail on how to process the request? How does that relate to the protocol? If grpc-swift adheres to the spec, why does it matter how it is implemented? Swift NIO allows us to package each piece of the logic in very clear blocks that can be added/removed on the fly, so IMO the extra complexity is not required as long as the spec is respected.

3: While that may be true for most deployments, I don't think it should be translated to "The only supported way to use gRPC-Web is through a reverse proxy", which means that each language should be free to decide to implement any features above the minimum required set of features to support gRPC-Web, including not needing (but supporting) a reverse proxy frontend. 

In general, I think I'm confused when you join both points 2 and 3, since one is saying _how_ to implement the server to support gRPC-Web, and the other is saying that Envoy should be used as the frontend, which means that there is no need to support gRPC-Web in server libraries at all.

MrMage:
> 2: Isn't that an implementation detail on how to process the request? How does that relate to the protocol? If grpc-swift adheres to the spec, why does it matter how it is implemented? Swift NIO allows us to package each piece of the logic in very clear blocks that can be added/removed on the fly, so IMO the extra complexity is not required as long as the spec is respected.

💯

> 3: While that may be true for most deployments, I don't think it should be translated to "The only supported way to use gRPC-Web is through a reverse proxy", which means that each language should be free to decide to implement any features above the minimum required set of features to support gRPC-Web, including not needing (but supporting) a reverse proxy frontend.

FWIW, as a user I would be fine with using a reverse-proxy in front of the actual service. However, I agree with @sergiocampama that if the gRPC-Web implementation built into SwiftGRPCNIO is 'good enough' for a particular user, they should be free to just use that.

wenbozhu:
@sergiocampama 

2. All fair points, as I mentioned, esp. if the grpc implementation shares the same HTTP stack as grpc-web. I agree this is mostly a language-specific implementation choice, and we are yet to identify cases when using grpc clients to proxy requests is more beneficial, e.g. "advanced flow-control" ... (just making up one).

I'd be OK to suggest the swift approach is desired only when the grpc and grpc-web (http/1) share the same stack provided by the language platform directly. 

3. This is mostly a maintenance concern. A lot of grpc-web use cases are for local development, debugging, for which the in-process server will work just fine, with a minimum set of features.  But for more advanced features e.g. a different protobuf encoding, we want to limit the implementation to ideally just Envoy. Some features could be under-specified too so there will be compatibility concerns too. E.g. all the gRFCs.

Again, I am just throwing out some thoughts here, and your opinions are very valuable.

sergiocampama:
Thanks for the explanations! The last one is a valid point. It'd be great if grpc-web specs were expanded to describe 3 different sets of functionalities:
* required spec to interop with grpc-web clients at a minimum (e.g. grpc-web/grpc-web-text decoding/encoding, CORS)
* optional grpc-web spec that grpc servers can implement if they want (e.g. grpc-web-text+json)
* list of features _recommended_ to _not_ implement in grpc servers that proxies like Envoy should provide instead (e.g. compression, other features that are not as well defined)

What I'm thinking is that there could be a minimum set of features that can be used when developing/debugging, and for production apps, client code can be built in an optimized mode that uses the advanced features that Envoy should provide.

I'm also spitballing here, so please let me know if these make sense. In the mean time I don't believe this change conflicts with your original concerns.

wenbozhu:
@sergiocampama  All agreed. I will go ahead and put together a "guideline" on language-specific implementations of grpc-web proxy, along with minimum/optional features identified on the spec. Will put you as reviewers of the doc changes.  Thanks.  @timburks

wenbozhu:
https://github.com/grpc/grpc-web/blob/master/IN-PROCESS-PROXY.md

timburks:
I'm not sure what the right response here should be (or if this is a specific gRPC-web problem), but currently I can crash EchoNIO by connecting to the server with `curl http://localhost:8080`:
```
$ swift run EchoNIO serve
starting insecure server
Precondition failed: : file /Users/timburks/Desktop/grpcweb/grpc-swift/.build/checkouts/swift-nio.git-3446246794268775558/Sources/NIOHTTP1/HTTPServerProtocolErrorHandler.swift, line 67
Illegal instruction: 
```

Lukasa:
That's a NIO precondition, and it gets hit if the HTTPServerRequestParts are being sent out-of-order. Specifically, in this case, we received a sequence of parts where we received a `.end` before a `.head`.

Examples/EchoWeb/README.md
```diff
@@ -0,0 +1,17 @@
+# Echo gRPC-Web Sample App
+
+The Echo gRPC-Web is a node project that creates a website that
+connects to a Swift gRPC NIO server to display messages. To build
+it, just run `make` inside this directory, and then open the
+`index.html` file in a web browser. Remember to start the Echo
+service by executing `swift run Echo -- serve_nio` before opening
```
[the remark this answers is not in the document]
sergiocampama: Yes, that's correct. Updated

sergiocampama:
That error you're seeing @timburks is because we're not currently handling unsupported requests, in this case,  `GET` requests.

If you try:
```
curl -H "Content-Type: application/grpc-web-text" -d AAAAAAsKCWNhY2EgY2FjYQ== http://localhost:8080/echo.Echo/Get
```

You should get data back, which, when base64 decoded, represents the binary proto response.

timburks:
Merging this today. @sergiocampama thanks for this great work, and @MrMage and @wenbozhu for your thorough reviews!

MrMage:
Glad to see this merged 🎉🚀

@lukasa crashing when we see unexpected requests sounds dangerous; is this planned to change in future iterations of NIOHTTP2, or maybe has already changed with your pure-Swift implementation? I haven't gotten around to update our dependency to that implementation yet and verify that our tests still pass when using it.

Lukasa:
@MrMage That crash is coming from NIO, but I believe it’s not our fault. As I said, it’s when user code writes out HTTP parts in an invalid order, so it’s worth investigating how that error path issues writes.

PR #352, opened 2019-01-03
Upgrade swift-nio to 1.12

glbrntt:
@MrMage not sure if this is the desired ordering? @lukasa wasn't sure about the specifics.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -21,7 +21,7 @@ var packageDependencies: [Package.Dependency] = [
   .package(url: "https://github.com/apple/swift-protobuf.git", .upToNextMinor(from: "1.1.1")),
   .package(url: "https://github.com/kylef/Commander.git", .upToNextMinor(from: "0.8.0")),
   .package(url: "https://github.com/apple/swift-nio-zlib-support.git", .upToNextMinor(from: "1.0.0")),
-  .package(url: "https://github.com/apple/swift-nio.git", .upToNextMinor(from: "1.11.0")),
+  .package(url: "https://github.com/apple/swift-nio.git", .upToNextMinor(from: "1.12.0")),
   .package(url: "https://github.com/apple/swift-nio-nghttp2-support.git", .upToNextMinor(from: "1.0.0")),
   .package(url: "https://github.com/apple/swift-nio-http2.git", .revision("dd9339e6310ad8537a271f3ff60a4f3976ca8e4d"))
 ]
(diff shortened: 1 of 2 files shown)


MrMage:
Hmm; tough question. To be honest, I don't quite remember.

I think we'd want to _instantly_ add the call handler, to avoid accidentally dropping messages sent in between removing `self` and adding the call handler. Given that this change only adds the call handler _after_ removing `self`, would that still be guaranteed after merging this change? /ccing @Lukasa

Also, I'm not sure how this would interact with https://github.com/grpc/grpc-swift/pull/350. Would you mind if I waited with merging this until that PR has been merged?

Lukasa:
> I think we'd want to instantly add the call handler, to avoid accidentally dropping messages sent in between removing self and adding the call handler.

No such timing window should exist.

The callback (when attached to a promise) from removing `self` will fire when `self` is no longer in the pipeline, but while `self`'s `ChannelHandlerContext` is still valid. Given that removing `self` always runs synchronously when run from on the event loop, we should be just fine.

*With that said*, currently there is no good interface that *actually promises this behaviour*. We should probably consider adding some explicitly thread-unsafe interfaces to `ChannelHandlerContext` that promise that you can actually remove `self` synchronously (and add other handlers to `ctx.pipeline` synchronously) such that we don't encounter this kind of speculation.

glbrntt:
> Also, I'm not sure how this would interact with #350. Would you mind if I waited with merging this until that PR has been merged?

No problem!

MrMage:
> > I think we'd want to instantly add the call handler, to avoid accidentally dropping messages sent in between removing self and adding the call handler.
> 
> No such timing window should exist.
> 
> The callback (when attached to a promise) from removing `self` will fire when `self` is no longer in the pipeline, but while `self`'s `ChannelHandlerContext` is still valid. Given that removing `self` always runs synchronously when run from on the event loop, we should be just fine.
> 
> _With that said_, currently there is no good interface that _actually promises this behaviour_. We should probably consider adding some explicitly thread-unsafe interfaces to `ChannelHandlerContext` that promise that you can actually remove `self` synchronously (and add other handlers to `ctx.pipeline` synchronously) such that we don't encounter this kind of speculation.

Sounds good enough to me, thanks for the clarification!

@glbrntt I'm happy with the PR; will merge one #350 has been merged (which might require changes in this PR). Just ping me when that's done :-)

sergiocampama:
Actually, #350 still needs to wait until https://github.com/apple/swift-nio-http2/pull/36 is merged and released. You should go ahead and merge this one and I'll rebase it onto my PR.

Issue #190, opened 2018-03-19
Allow client to specify metadata per call


cvanderschuere:
Currently the client is only able to modify the metadata at the service level, but metadata is meant to be per-call.

https://github.com/grpc/grpc-swift/blob/dd4d1cb64d080a60e64f15b80e0090da6f5104f3/Sources/protoc-gen-swiftgrpc/Generator-Client.swift#L189-L194

Changes:
- Allow an optional metadata to be passed per call
- Add `merge` to Metadata
- Merge service metadata and call metadata on `call.start`

MrMage:
Note that we can't really make arguments optional, as that is currently not possible on protocols (and we switched to protocols for better test ability). Although, it might be possible to add protocol extensions with optional arguments; that could be a viable workaround.

Do you need per-call metadata at the moment? I would have supposed it's more of an edge case, with the most common case being per-channel authentication metadata.

cvanderschuere:
Yes we do. The most direct use case is for request tracing which must be via per-rpc metadata.

PR #355, opened 2019-01-11
Add RxSwift option to protoc-gen-swiftgrpc

xissy:
This PR adds RxSwift options to protoc-gen-swiftgrpc. It also provides a way to pass `metadata` argument per call.

Usage:
`protoc user.proto --swiftgrpc_out=RxSwift=true:.`

Sample:
``` swift
internal protocol User_UserService: ServiceClient {
  ...
  /// RxSwift. Unary.
  func get(_ request: User_GetRequest) -> Observable<User_GetResponse>
  /// RxSwift with metadata. Unary.
  func get(_ request: User_GetRequest, metadata: Metadata) -> Observable<User_GetResponse>
  ...
}

internal final class User_UserServiceClient: ServiceClientBase, User_UserService {
  ...
  /// RxSwift. Unary.
  internal func register(_ request: User_RegisterRequest) -> Observable<User_RegisterResponse> {
    return register(request, metadata: metadata)
  }
  /// RxSwift with metadata. Unary.
  internal func register(_ request: User_RegisterRequest, metadata: Metadata) -> Observable<User_RegisterResponse> {
    return Observable.just(request)
      .flatMap { request -> Observable<User_RegisterResponse> in
        return Observable.create { observer in
          do {
            let response = try User_UserRegisterCallBase(self.channel)
              .run(request: request, metadata: metadata)
            observer.onNext(response)
          } catch {
            observer.onError(error)
          }
          return Disposables.create()
        }
      }
  ...
}
```


xissy:
A travis-ci build job keeps failing with this error. 😭 
```
The command "if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install nghttp2; fi" failed and exited with 1 during .
```

Sources/protoc-gen-swiftgrpc/Generator-Client.swift
```diff
@@ -216,6 +227,44 @@ extension Generator {
...
+          indent()
+          println(".flatMap { request -> Observable<\(methodOutputName)> in")
+          indent()
+          println("return Observable.create { observer in")
+          indent()
+          println("do {")
+          indent()
+          println("let response = try \(callName)Base(self.channel)")
+          indent()
+          println(".run(request: request, metadata: metadata)")
```
MrMage: I think in this case it would be better to use the non-blocking variant of the unary call, and in the corresponding completion handler call `response.onNext/onError` as needed. Currently this might actually block the calling thread.

Sources/protoc-gen-swiftgrpc/Generator-Client.swift
```diff
@@ -216,6 +227,44 @@ extension Generator {
...
+          println("/// RxSwift. Unary.")
+          println("\(access) func \(methodFunctionName)(_ request: \(methodInputName)) -> Observable<\(methodOutputName)> {")
+          indent()
+          println("return \(methodFunctionName)(request, metadata: metadata)")
+          outdent()
+          println("}")
+          println("/// RxSwift with metadata. Unary.")
+          println("\(access) func \(methodFunctionName)(_ request: \(methodInputName), metadata: Metadata) -> Observable<\(methodOutputName)> {")
+          indent()
+          println("return Observable.just(request)")
```
MrMage: Why not return `Observable.create` directly? :-)

Sources/protoc-gen-swiftgrpc/Generator-Client.swift
```diff
@@ -164,6 +168,12 @@ extension Generator {
           println("/// Asynchronous. Unary.")
           println("func \(methodFunctionName)(_ request: \(methodInputName), completion: @escaping (\(methodOutputName)?, CallResult) -> Void) throws -> \(callName)")
         }
+        if rxSwiftCode {
```
MrMage: Consider calling through to the protocol's official asynchronous unary implementation (see line 169); that way all this code could be put into a protocol extension and would not need to be part of the main protocol. That would also have the advantage of letting you provide a default argument for `metadata` (provided the `metadata` argument gets added to the unary call's method signature first):

```
func \(methodFunctionName)(_ request: \(methodInputName), metadata customMetadata: Metadata?) -> Observable<\(methodOutputName)> {
  // use metadata: customMetadata ?? self.metadata when calling through
}
```

MrMage:
I think switching that visibility to internal would be fine.

_However,_ would it be possible to only call through to call the publicly-exposed asynchronous unary method declared in the protocol from the extension? I.e. you would just add a protocol extension that calls through to the “official” method. That would need adding the option to pass custom metadata to that method, but that would be a nice addition anyway (see https://github.com/grpc/grpc-swift/issues/190 <https://github.com/grpc/grpc-swift/issues/190>).

bohachevskyy:
Thank you @xissy it looks interesting. I'm going to use it. 
Please, let me know if you need help with creating plugin. I'd like to join you 😃

xissy:
@bohachevskyy I've already done some works on it here. https://github.com/xissy/protoc-gen-swiftgrpcrx
Free free to contribute anything you want.

PR #356, opened 2019-01-13
Allow client to specify metadata per call

xissy:
To address https://github.com/grpc/grpc-swift/pull/355#issuecomment-453744353. It also closes https://github.com/grpc/grpc-swift/issues/190.


xissy:
@MrMage Let me know if I missed or did something wrong. As a swift noob, I would follow all your suggestions to learn. Thank you. :)

MrMage:
Sorry, I thought I had seen changes where you made `metadata` non-optional, but now I can’t find them. Please make that change.

Also, why did the `.pb.swift` file change?

Sources/Examples/Echo/Generated/echo.pb.swift
```diff
@@ -33,20 +33,42 @@ fileprivate struct _GeneratedWithProtocGenSwiftVersion: SwiftProtobuf.ProtobufAP
   typealias Version = _2
 }
 
-struct Echo_EchoRequest: SwiftProtobuf.Message {
-  static let protoMessageName: String = _protobuf_package + ".EchoRequest"
+struct Echo_EchoRequest {
```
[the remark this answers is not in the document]
rebello95: Possibly using a different version of SwiftProtobuf? I think this file's changes should probably be dropped

MrMage:
I agree.

rebello95:
Any updates on this @xissy? Would be great to get this feature in!

xissy:
Sorry, I've been busy, will update soon.

Sources/Examples/Echo/Generated/echo.grpc.swift
```diff
@@ -117,93 +117,120 @@ class Echo_EchoUpdateCallTestStub: ClientCallBidirectionalStreamingTestStub<Echo
...
 
   /// Asynchronous. Bidirectional-streaming.
   /// Use methods on the returned object to stream messages,
   /// to wait for replies, and to close the connection.
-  func update(completion: ((CallResult) -> Void)?) throws -> Echo_EchoUpdateCall
+  func update(metadata customMetadata: Metadata, completion: ((CallResult) -> Void)?) throws -> Echo_EchoUpdateCall
+
+}
+
+internal extension Echo_EchoService {
```
rebello95: @MrMage I know you had suggested putting these functions in a separate extension, but it seems like this adds a bit of boilerplate. Would it make sense to update any generated callers of these functions instead of generating more functions as an extension?

MrMage: @rebello95 I think these methods are actually called by gRPC users, not by generated code. I'm not super happy about this boilerplate, either, but I don't know how to avoid it if we don't want to force our library's consumers to include this argument with each gRPC call they are going to call. I might be missing something, though; let me know.

rebello95: Yea that makes sense 👍

PR #363, opened 2019-01-30
Proposal to fix #362.

### LINKED ISSUE #362: Why I can't use default root certificates when I create a channel with client certificates and key?
### Question Checklist

### Question Subject
The question consists of two parts.
1. In ObjC GRPC I can set client certificates and private key and still use default "roots.pem" certificates as root certificates. In SwiftGRPC default root certificates are used only when I create channel using `address:secure:arguments:` initializer. Is this intentional and if it is, what reason for?
2. In other GRPC-libraries, I still have access to default certificates, usually through "roots.pem" file. 
In SwiftGRPC there is only a function that returns regular "roots.pem" content. This function is internal so I have no access to default root certificates and could not use them to pass them into an `address:certificates:clientCertificates:clientKey:args:` initializer. Is this intentional and if it is, what reason for?

#### Question Description
As for now the only decision that allows me to create a channel with default root certificates is to download "roots.pem"  and put it into my app, while I know that SwiftGRPC framework already contains it.
To be honest, I'm not very familiar with GRPC properly usage, and my expectations are mostly based on other GRPC frameworks knowledge. The question may be dumb and I apologize in advance.

---

Update: I've discovered that grpc-swift project also have "roots.pem" file which does not included as a resource into a pod.
It looks weird for me that "roos.pem" used by a grpc undercover, grpc-swift uses `roots_pem()` method in swift, and a pod user has no access to any of this file and no ability to the use default root certificates with his own client certificates and key.
In my opinion either "roots.pem" should be added to pod resources in a podspec, or roots_pem() must be a public, or Channel initializer should have a parameter to use default root certificates. The third way allows to keep private `roots_pem()` method and does not include "roots.pem" as a resourse of SwiftGRPC pod.

Vladlex:
Allows a user to set up a channel with a client certificate and key but still use default root certificates.


Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -75,17 +75,18 @@ public class Channel {
   /// Initializes a gRPC channel
   ///
   /// - Parameter address: the address of the server to be called
-  /// - Parameter certificates: a PEM representation of certificates to use
+  /// - Parameter certificates: a PEM representation of certificates to use. If nil, use defaults.
```
MrMage: nit: "If nil, the default root certificates provided by gRPC are used."

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -75,17 +75,18 @@ public class Channel {
...
   /// - Parameter clientCertificates: a PEM representation of the client certificates to use
   /// - Parameter clientKey: a PEM representation of the client key to use
   /// - Parameter arguments: list of channel configuration options
-  public init(address: String, certificates: String, clientCertificates: String? = nil, clientKey: String? = nil, arguments: [Argument] = []) {
+  public init(address: String, certificates: String?, clientCertificates: String? = nil, clientKey: String? = nil, arguments: [Argument] = []) {
```
MrMage: `certificates: String? = nil`, or (if that compiles, please try it and let me know if it doesn't) `certificates: String = roots_pem()`.

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -75,17 +75,18 @@ public class Channel {
...
     gRPC.initialize()
     host = address
     let argumentWrappers = arguments.map { $0.toCArg() }
     var argumentValues = argumentWrappers.map { $0.wrapped }
 
-    underlyingChannel = cgrpc_channel_create_secure(address, certificates, clientCertificates, clientKey, &argumentValues, Int32(arguments.count))
+    let certificatesToUs = certificates ?? roots_pem()
```
MrMage: Would you mind renaming `certificatesToUs` to `rootCertificates` (or dropping it entirely if `certificates: String = roots_pem()` works)?

Sources/SwiftGRPC/Core/Channel.swift
Vladlex: @MrMage 
`roots_pem()` returns an optional, so we have to either force unwrap `roots_pem()` either leave argument optional: `certificates: String? = roots_pem()`
I'm not sure which way better. Looking forward to your decision.

MrMage: Thanks for the explanation! I think we should change `roots_pem()` to forcibly unwrap the results of its two calls instead, at which point it wouldn't need to return an optional. Would you mind adding that?

Vladlex: @MrMage 

> Would you mind adding that?
Sure, thank you.

> `certificates: String? = nil`, or (if that compiles, please try it and let me know if it doesn't)
It doesn't.
We should change `roots_pem()` access modifier to `public` to use it as a default value.
To be more specific, compiler fails with following error:
`Global function 'roots_pem()' is internal and cannot be referenced from a default argument value`.

If this way is inappropriate we can also do something like that to hide roots_pem:
```
enum RootCertificates {
  case custom(String)
  case `default`
}
```
But making `roots_pem()` public looks better to me.

MrMage: I think that we should definitely make `roots_pem()` public.

Vladlex: Ok, so `Roots.swift` is a generated file which should not be edited according to comment in it.
I suggest to add a static method to `gRPC` class like that:

```
public final class gRPC {
    
  /// Default root certificates provided by gRPC
  public static let defaultRootCertificates = roots_pem()!

...
```

And after that, all should be ok finally.

MrMage: That file can be edited if you edit https://github.com/grpc/grpc-swift/blob/master/Sources/RootsEncoder/main.swift at the same time ;-)

Vladlex: Got it! New commit ready for review.

Vladlex:
Requested changes made are made.

MrMage:
As requested in https://github.com/grpc/grpc-swift/pull/363#discussion_r257622703, would you mind changing `roots_pem()` to return a non-optional `String` as well? Can be done by changing lines 41-46 in `Roots.swift` (and similarly in `RootsEncoder.swift`) to `return String(data: Data(base64Encoded: roots, options:[])!, encoding: .utf8)!`.

Vladlex:
> As requested in [#363 (comment)](https://github.com/grpc/grpc-swift/pull/363#discussion_r257622703), would you mind changing `roots_pem()` to return a non-optional `String` as well? Can be done by changing lines 41-46 in `Roots.swift` (and similarly in `RootsEncoder.swift`) to `return String(data: Data(base64Encoded: roots, options: [])!, encoding: .utf8)!`.

Got it. I've changed RootsEncoder's "main.swift", and generated a new one "Roots.swift" with RootsEncoder tool. Channel's initializer was changed accordingly. 
Thank you for the explanation!

PR #364, opened 2019-01-30
Improve error handling in NIO server.

glbrntt:
- Adds a user-configurable error handler to the server
- Updates NIO server codegen to provide an optional error handler
- Errors are handled by GRPCChannelHandler or BaseCallHandler,
  depending on the pipeline state
- Adds some error handling tests
- Tidies some logic in HTTP1ToRawGRPCServerCodec
- Extends message handling logic in HTTP1ToRawGRPCServerCodec
  to handle messages split across multiple ByteBuffers (i.e. when a
  message exceeds a the size of a frame)


Sources/SwiftGRPCNIO/CallHandlers/BaseCallHandler.swift
```diff
@@ -8,29 +8,79 @@ import NIOHTTP1
...
+  }
+}
+
+extension BaseCallHandler: ChannelOutboundHandler {
+  public typealias OutboundIn = GRPCServerResponsePart<ResponseMessage>
+  public typealias OutboundOut = GRPCServerResponsePart<ResponseMessage>
+
+  public func write(ctx: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
+    guard serverCanWrite else {
+      promise?.fail(error: GRPCStatus.processingError)
```
MrMage: Should we have a different kind of error for this (maybe not even of type `GRPCStatus`), e.g. "attempted to write more than one status"? (See below for a more detailed comment in a similar vein.)

Sources/SwiftGRPCNIO/CallHandlers/BaseCallHandler.swift
```diff
@@ -8,29 +8,79 @@ import NIOHTTP1
...
+      endOfStreamReceived()
+    }
+  }
+}
+
+extension BaseCallHandler: ChannelOutboundHandler {
+  public typealias OutboundIn = GRPCServerResponsePart<ResponseMessage>
+  public typealias OutboundOut = GRPCServerResponsePart<ResponseMessage>
+
+  public func write(ctx: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
```
MrMage: Nice approach!

Sources/SwiftGRPCNIO/CallHandlers/ServerStreamingCallHandler.swift
```diff
@@ -26,12 +26,10 @@ public class ServerStreamingCallHandler<RequestMessage: Message, ResponseMessage
...
   
   
-  public override func processMessage(_ message: RequestMessage) {
+  public override func processMessage(_ message: RequestMessage) throws {
     guard let eventObserver = self.eventObserver,
       let context = self.context else {
-      //! FIXME: Better handle this error?
-      print("multiple messages received on unary call")
-      return
+        throw GRPCStatus(code: .unimplemented, message: "multiple messages received on unary call")
```
MrMage: Is `.unimplemented` appropriate here?

Also, how about having a dedicated enum of all possible connection errors generated by gRPC (e.g. "attempted to write more than one status", "multiple messages received on unary call", `requestProtoParseError`, "unimplemented", `responseProtoSerializationError`, `unsupportedCompression` etc.) which conforms to `GRPCStatusTransformable`, instead of using `GRPCStatus` directly? The reasoning would be that this reduces the variety of errors the error delegate would need to expect/care about, but still lets us return the appropriate status to the client (thanks to the enum's conformance to `GRPCStatusTransformable`). Another advantage would be that we don't need to update this error string in two places, for example. (It is also used in `UnaryCallHandler.swift`.)

Sources/SwiftGRPCNIO/GRPCChannelHandler.swift
```diff
@@ -28,30 +28,32 @@ public protocol CallHandlerProvider: class {
...
+    self.errorDelegate = errorDelegate
   }
 }
 
 extension GRPCChannelHandler: ChannelInboundHandler {
   public typealias InboundIn = RawGRPCServerRequestPart
   public typealias OutboundOut = RawGRPCServerResponsePart
-  
+
+  public func errorCaught(ctx: ChannelHandlerContext, error: Error) {
```
MrMage: Mild code duplication from `BaseCallHandler` here, but not sure whether we can (or should) do anything about that.

Sources/SwiftGRPCNIO/GRPCServer.swift
```diff
@@ -39,20 +41,30 @@ public final class GRPCServer {
...
+  private var errorDelegate: ServerErrorDelegate?
 
-  private init(channel: Channel) {
+  private init(channel: Channel, errorDelegate: ServerErrorDelegate?) {
     self.channel = channel
+
+    // Maintain a strong reference to ensure it lives as long as the server.
+    self.errorDelegate = errorDelegate
+
+    // `BaseCallHandler` holds a weak reference to the delegate; nil out this reference to avoid retain cycles.
```
MrMage: How is the "`BaseCallHandler` holds a weak reference to the delegate;" part relevant to this line?

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -27,111 +27,200 @@ public enum RawGRPCServerResponsePart {
...
+  enum OutboundState {
+    case expectingHeaders
+    case expectingBodyOrStatus
+    case ignore
+  }
+}
 
-  private var buffer: NIO.ByteBuffer?
+extension HTTP1ToRawGRPCServerCodec {
+  struct StateMachineError: Error, GRPCStatusTransformable {
```
MrMage: Make this `struct` and `message` public (and/or `CustomStringConvertible`), so that library users can inspect the error?

Sources/SwiftGRPCNIO/GRPCStatus.swift
```diff
@@ -27,4 +27,19 @@ public struct GRPCStatus: Error {
...
     return GRPCStatus(code: .unimplemented, message: "unknown method " + method)
   }
+
+  // These status codes are informed by: https://github.com/grpc/grpc/blob/master/doc/statuscodes.md
+  static internal let requestProtoParseError = GRPCStatus(code: .internalError, message: "could not parse request proto")
+  static internal let responseProtoSerializationError = GRPCStatus(code: .internalError, message: "could not serialize response proto")
+  static internal let unsupportedCompression = GRPCStatus(code: .unimplemented, message: "compression is not supported on the server")
+}
+
+protocol GRPCStatusTransformable: Error {
```
MrMage: Using a protocol instead of having to use `GRPCStatus` directly is great! But should we make this `public`?

Sources/SwiftGRPCNIO/GRPCStatus.swift
```diff
@@ -27,4 +27,19 @@ public struct GRPCStatus: Error {
   public static func unimplemented(method: String) -> GRPCStatus {
     return GRPCStatus(code: .unimplemented, message: "unknown method " + method)
   }
+
+  // These status codes are informed by: https://github.com/grpc/grpc/blob/master/doc/statuscodes.md
+  static internal let requestProtoParseError = GRPCStatus(code: .internalError, message: "could not parse request proto")
```
MrMage: Why `internal`? In general, I prefer to have more open visibility. That could help low-level consumers of this library looking to re-implement/modify parts of it without re-inventing the wheel. Let me know if you think differently.

(If we decide to keep this internal, I'd still suggest dropping the `internal` keyword.)

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -27,111 +27,200 @@ public enum RawGRPCServerResponsePart {
...
-          //! FIXME: Use a slice of this buffer instead of copying to a new buffer.
-          var responseBuffer = ctx.channel.allocator.buffer(capacity: messageBytes.count)
-          responseBuffer.write(bytes: messageBytes)
-          ctx.fireChannelRead(self.wrapInboundOut(.message(responseBuffer)))
-          //! FIXME: Call buffer.discardReadBytes() here?
-          //! ALTERNATIVE: Check if the buffer has no further data right now, then clear it.
-
-          state = .expectingCompressedFlag
-        }
+    if case .ignore = inboundState { return }
```
MrMage: Would it be worth logging this (or sending it to an error delegate/calling `ctx.fireErrorCaught`) if hit? (Maybe not, given that something else would probably have gone wrong before this could even happen.)

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -27,111 +27,200 @@ public enum RawGRPCServerResponsePart {
...
+
+    do {
+      switch self.unwrapInboundIn(data) {
```
MrMage: Nice refactoring!

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -27,111 +27,200 @@ public enum RawGRPCServerResponsePart {
...
+
+    // Iterate over all available incoming data, trying to read length-delimited messages.
+    // Each message has the following format:
+    // - 1 byte "compressed" flag (currently always zero, as we do not support for compression)
+    // - 4 byte signed-integer payload length (N)
+    // - N bytes payload (normally a valid wire-format protocol buffer)
+    while true {
+      switch bodyState {
+      case .expectingCompressedFlag:
+        guard let compressionFlag: Int8 = messageBuffer.readInteger() else { return .expectingCompressedFlag }
```
MrMage: Nit: name discrepancy between `compressIONFlag` and `[expecting]CompressEDFlag` — rename either for consistency?

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -27,111 +27,200 @@ public enum RawGRPCServerResponsePart {
...
+            self.buffer = buffer
+          } else {
+            messageBuffer.reserveCapacity(numericCast(messageLength))
+            self.buffer = messageBuffer
+          }
+
+          return .receivedMessageLength(remainingBytes)
+        }
+
+        // We know buffer.readableBytes >= messageLength, so it's okay to force unwrap here.
```
MrMage: Good comment!

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -27,111 +27,200 @@ public enum RawGRPCServerResponsePart {
...
+
+      case .expectingMessageLength:
+        guard let messageLength: UInt32 = messageBuffer.readInteger() else { return .expectingMessageLength }
+        bodyState = .receivedMessageLength(messageLength)
+
+      case .receivedMessageLength(let messageLength):
+        // We need to account for messages being spread across multiple `ByteBuffer`s so buffer them
+        // into `buffer`. Note: when messages are contained within a single `ByteBuffer` we're just
+        // taking a slice so don't incur any extra writes.
+        guard messageBuffer.readableBytes >= messageLength else {
```
MrMage: Shouldn't this take into account the amount of data already read into `buffer`, if available?

Example:

Frame A contains message length (100) plus 50 bytes of the message. After receiving frame A, we will be in state `.receivedMessageLength(100)`, with 50 bytes in `buffer`.
If we now receive frame B with the remaining 50 bytes, this `guard` will bail, even though we can assemble the remaining data by concatenating `buffer` and `messageBuffer`. Let me know if I'm missing something.

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -27,111 +27,200 @@ public enum RawGRPCServerResponsePart {
...
+        self.buffer = nil
+        bodyState = .expectingCompressedFlag
       }
-      ctx.fireChannelRead(self.wrapInboundOut(.end))
     }
   }
+
+  private func processEnd(ctx: ChannelHandlerContext, trailers: HTTPHeaders?) throws -> InboundState {
+    guard trailers == nil else {
+      throw StateMachineError("unexpected trailers received \(String(describing: trailers))")
```
MrMage: change suggestion:
```
if let trailers = trailers {
	      throw StateMachineError( "client sent unexpected trailers \(trailers)")
}
```
Avoids printing "Optional(trailers...)" into the string.

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -27,111 +27,200 @@ public enum RawGRPCServerResponsePart {
...
     case .status(let status):
       var trailers = status.trailingMetadata
       trailers.add(name: "grpc-status", value: String(describing: status.code.rawValue))
       trailers.add(name: "grpc-message", value: status.message)
-      ctx.write(self.wrapOutboundOut(.end(trailers)), promise: promise)
+
+      // "Trailers-Only" response
+      if case .expectingHeaders = outboundState {
+        trailers.add(name: "content-type", value: "application/grpc")
+        let responseHead = HTTPResponseHead(version: .init(major: 2, minor: 0), status: .ok)
```
MrMage: Would you mind repeating `//! FIXME: Should return a different version if we want to support pPRC.` here?

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -27,111 +27,200 @@ public enum RawGRPCServerResponsePart {
...
+        ctx.write(self.wrapOutboundOut(.head(responseHead)), promise: nil)
```
MrMage: This will send two header frames — one header with content type and "OK", and a trailer one for the trailers. Should we merge the two instead? Is that even possible, or does HTTP2 always expect a header and a trailer frame?

Sources/SwiftGRPCNIO/ServerErrorDelegate.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+
+public protocol ServerErrorDelegate: class {
+  /// Called when an error thrown in the channel pipeline.
```
MrMage: nit: "is"

Sources/SwiftGRPCNIO/ServerErrorDelegate.swift
```diff
@@ -0,0 +1,23 @@
...
+  func observe(_ error: Error)
+
+  /// Transforms the given error into a new error.
+  ///
+  /// This allows framework to transform errors which may be out of their control
```
MrMage: nit: "framework" -> "framework users"?

Sources/SwiftGRPCNIO/GRPCChannelHandler.swift
```diff
@@ -28,30 +28,32 @@ public protocol CallHandlerProvider: class {
...
+    errorDelegate?.observe(error)
+
+    let transformedError = (errorDelegate?.transform(error) ?? error)
```
MrMage: super-nit: remove wrapping parentheses?

Tests/SwiftGRPCNIOTests/NIOServerTests.swift
```diff
@@ -122,6 +122,12 @@ extension NIOServerTests {
     XCTAssertEqual("Swift echo get: foo", try! client.get(Echo_EchoRequest(text: "foo")).text)
   }
 
+  func testUnaryWithLargeData() throws {
+    // Default max frame size is: 16,384. We'll exceed this as we also have to send the size and compression flag.
+    let request = Echo_EchoRequest.with { $0.text = String(repeating: "e", count: 16_384) }
+    XCTAssertNoThrow(try client.get(request))
```
MrMage: Could you please also assert that the response has the expected contents?

Tests/SwiftGRPCNIOTests/GRPCChannelHandlerResponseCapturingTestCase.swift
```diff
@@ -0,0 +1,67 @@
+import Foundation
+import NIO
+import NIOHTTP1
+@testable import SwiftGRPCNIO
+import XCTest
+
+internal struct CaseExtractError: Error {
```
MrMage: super-nit: drop `internal`, especially given that this is test-only anyway?

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -27,111 +27,200 @@ public enum RawGRPCServerResponsePart {
 ///
 /// The translation from HTTP2 to HTTP1 is done by `HTTP2ToHTTP1ServerCodec`.
 public final class HTTP1ToRawGRPCServerCodec {
-  private enum State {
+  internal var inboundState = InboundState.expectingHeaders
```
MrMage: super-nit: drop `internal`? (Also below)

Tests/SwiftGRPCNIOTests/GRPCChannelHandlerResponseCapturingTestCase.swift
```diff
@@ -0,0 +1,67 @@
...
+
+  func configureChannel(withHandlers handlers: [ChannelHandler]) -> EventLoopFuture<EmbeddedChannel> {
+    let channel = EmbeddedChannel()
+    return channel.pipeline.addHandlers(handlers, first: true)
+      .map { _ in channel }
+  }
+
+  /// Waits for `count` responses to be collected and then returns them. The test fails if the number
+  /// of collected responses does not match the expected.
+  func waitForGRPCChannelHandlerResponses(
```
MrMage: Would you mind adding a remark that `callback` is called _after_ the channel has been set up, but _before_ this method returns, and that it is meant to contain the calls that actually lead to the channel being "filled" with messages?

Tests/SwiftGRPCNIOTests/GRPCChannelHandlerResponseCapturingTestCase.swift
```diff
@@ -0,0 +1,67 @@
...
+class CollectingChannelHandler<OutboundIn>: ChannelOutboundHandler {
+  var responses: [OutboundIn] = []
+
+  func write(ctx: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
+    responses.append(unwrapOutboundIn(data))
+  }
+}
+
+class GRPCChannelHandlerResponseCapturingTestCase: XCTestCase {
+  static let echoProvider: [String: CallHandlerProvider] = ["echo.Echo": EchoProvider_NIO()]
```
MrMage: Make this `class`, so it can be overridden by subclasses?

Tests/SwiftGRPCNIOTests/GRPCChannelHandlerTests.swift
```diff
@@ -0,0 +1,196 @@
...
+
+      var buffer = channel.allocator.buffer(capacity: 3)
+      buffer.write(bytes: [1, 2, 3])
+      try channel.writeInbound(RawGRPCServerRequestPart.message(buffer))
+    }
+
+    XCTAssertNoThrow(try extractHeaders(responses[0]))
+    XCTAssertNoThrow(try extractStatus(responses[1])) { status in
+      let expectedStatus = GRPCStatus.requestProtoParseError
+      XCTAssertEqual(status.code, expectedStatus.code)
```
MrMage: Should we make `GRPCStatus` `Equatable` or `Hashable`, to simplify this?

Tests/SwiftGRPCNIOTests/GRPCChannelHandlerTests.swift
```diff
@@ -0,0 +1,196 @@
...
+
+    XCTAssertNoThrow(try extractHeaders(responses[0]))
+    XCTAssertNoThrow(try extractStatus(responses[1])) { status in
+      let expected = GRPCStatus.unsupportedCompression
+      XCTAssertEqual(status.code, expected.code)
+      XCTAssertEqual(status.message, expected.message)
+    }
+  }
+
+  func testMessageCanBeSentAcrossMultipleByteBuffers() throws {
```
MrMage: I'm curious — would this test have failed before this PR? I thought `HTTP1ToRawGRPCServerCodec` had already supported merging messages across frame boundaries, but I might be mistaken.

Tests/SwiftGRPCNIOTests/GRPCChannelHandlerTests.swift
```diff
@@ -0,0 +1,196 @@
...
+  ) throws -> [RawGRPCServerResponsePart] {
+    return try super.waitForGRPCChannelHandlerResponses(count: count, servicesByName: servicesByName) { channel in
+      _ = channel.pipeline.addHandlers(HTTP1ToRawGRPCServerCodec(), first: true)
+        .thenThrowing { _ in try callback(channel) }
+    }
+  }
+}
+
+// Assert the given expression does not throw, and validate the return value from that expression.
+public func XCTAssertNoThrow<T>(
```
MrMage: Terrific idea! I've always been bothered that `XCTAssertNoThrow` could not give you the result of its call. How about moving this to a `TestHelpers` file (or to `GRPCChannelHandlerResponseCapturingTestCase`), though?

Sources/SwiftGRPCNIO/CallHandlers/BaseCallHandler.swift
```diff
@@ -8,29 +8,79 @@ import NIOHTTP1
...
   /// Overridden by subclasses if the "end-of-stream" event is relevant.
   public func endOfStreamReceived() { }
+
+  /// Whether this handler can still write messages to the client.
+  private var serverCanWrite = true
+
+  /// Called for each error recieved in `errorCaught(ctx:error:)`.
+  private weak var errorDelegate: ServerErrorDelegate?
+
+  public init(errorDelegate: ServerErrorDelegate? = nil) {
```
MrMage: I think this initializer is only called by us, so drop the ` = nil` default argument?

Sources/SwiftGRPCNIO/GRPCChannelHandler.swift
```diff
@@ -28,30 +28,32 @@ public protocol CallHandlerProvider: class {
 /// for an `GRPCCallHandler` object. That object is then forwarded the individual gRPC messages.
 public final class GRPCChannelHandler {
   private let servicesByName: [String: CallHandlerProvider]
+  private weak var errorDelegate: ServerErrorDelegate?
 
-  public init(servicesByName: [String: CallHandlerProvider]) {
+  public init(servicesByName: [String: CallHandlerProvider], errorDelegate: ServerErrorDelegate? = nil) {
```
MrMage: I think this initializer is only called by us, so drop the ` = nil` default argument?

Sources/protoc-gen-swiftgrpc/Generator-Server.swift
```diff
@@ -85,7 +85,7 @@ extension Generator {
     if options.generateNIOImplementation {
       println("/// Determines, calls and returns the appropriate request handler, depending on the request's method.")
       println("/// Returns nil for methods not handled by this service.")
-      println("\(access) func handleMethod(_ methodName: String, request: HTTPRequestHead, serverHandler: GRPCChannelHandler, channel: Channel) -> GRPCCallHandler? {")
+      println("\(access) func handleMethod(_ methodName: String, request: HTTPRequestHead, serverHandler: GRPCChannelHandler, channel: Channel, errorDelegate: ServerErrorDelegate? = nil) -> GRPCCallHandler? {")
```
MrMage: Drop the ` = nil` default argument?

Sources/SwiftGRPCNIO/GRPCServer.swift
```diff
@@ -12,7 +12,9 @@ public final class GRPCServer {
     hostname: String,
     port: Int,
     eventLoopGroup: EventLoopGroup,
-    serviceProviders: [CallHandlerProvider]) -> EventLoopFuture<GRPCServer> {
+    serviceProviders: [CallHandlerProvider],
+    errorDelegate: ServerErrorDelegate? = nil
```
MrMage: How about a simple "LoggingErrorDelegate" (that just logs errors) instead of `nil` as the default?

Sources/SwiftGRPCNIO/CallHandlers/ServerStreamingCallHandler.swift
glbrntt: The documentation (https://github.com/grpc/grpc/blob/master/doc/statuscodes.md) suggests unimplemented. See:

> "Request cardinality violation (method requires exactly one request but client sent some other number of requests)"

--- 

Re: the dedicated enum of errors; I like that idea a lot. Another advantage of having them defined in one place is that it's a clear list of states to test for.




>

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
glbrntt: We update the state with the number of bytes we still need (see L153). 

I can see how the naming leads to confusion though, perhaps renaming `receivedMessageLength` to `expectingBytes` and `messageLength` to `requiredBytes` would help?

```swift
case .expectingBytes(let requiredBytes):
  // ...
```

Sources/SwiftGRPCNIO/CallHandlers/ServerStreamingCallHandler.swift
MrMage: Thanks for the elaboration, and sorry for missing that!

And glad you like the idea of the error enum. Note that my enumeration might not be exhaustive; we might want to search the repo for all `GRPCStatus(` calls to figure out which ones we should add to that enum (probably only the ones that can possibly be sent to the error delegate, though).

Tests/SwiftGRPCNIOTests/GRPCChannelHandlerTests.swift
glbrntt: It would, short of a couple of small issues; https://github.com/grpc/grpc-swift/blob/0195dfd5b3c7953dcff433a3eb86c27f0f5ff43a/Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift#L89 should have been `break requestProcessing` (otherwise you don't exit the while loop)

The other (which threw me for a while) was not reassigning to `self.buffer`, I _think_ what was happening is that when `buffer` was being written to here:

https://github.com/grpc/grpc-swift/blob/0195dfd5b3c7953dcff433a3eb86c27f0f5ff43a/Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift#L68

The underlying storage has references from `self.buffer` and `buffer`, so the storage was copied, we wrote into `buffer`s storage, so `self.buffer`s storage would still be empty.

MrMage:
Two more thoughts:

1. Should we have a `CollectingServerErrorDelegate` (similar to `CollectingChannelHandler`) in our tests and assert that the error delegate is sent the expected of errors?
2. It might be nice to provide some more error context to `ErrorDelegate`, e.g. which `Server/CallHandler/etc.` (and the corresponding method path) has caught the current error, so we know more precisely where the error occurred.

2. might require too many modifications to be incorporated into this PR, though. WDYT?

Tests/SwiftGRPCNIOTests/GRPCChannelHandlerTests.swift
MrMage: That makes sense, thank you for the fixes!

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
MrMage: Ah, thanks for the clarification! I indeed missed that. We should definitely do the renames you suggested, but maybe use something to the tune of `expectingMoreMessageBytes` and `bytesOutstanding`?

glbrntt: No problem -- that naming works for me!

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
glbrntt: That's correct -- it's not possible in NIO HTTP2, unfortunately. 

The "Trailers-Only" comment is misleading though: the spec says that "Trailers-Only" should be sent in a single frame:

> Response-Headers & Trailers-Only are each delivered in a single HTTP2 HEADERS frame block. Most responses are expected to have both headers and trailers but Trailers-Only is permitted for calls that produce an immediate error.

This is really response headers _and_ trailers. I'll fixup the comment.

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
glbrntt: Good call!

Tests/SwiftGRPCNIOTests/GRPCChannelHandlerTests.swift
glbrntt: We can do that, however, we'd also be checking the trailing metadata. These tests collect the responses _before_ the `grpc-{status, message}` get added to the trailers so they should be fine.

MrMage: Makes sense, resolving this.

Sources/SwiftGRPCNIO/CallHandlers/BaseCallHandler.swift
glbrntt: Absolutely, this will never actually reach the caller so doesn't even need to be a `GRPCStatus`.

Sources/SwiftGRPCNIO/GRPCChannelHandler.swift
glbrntt: I don't think there's a lot we can (reasonably) do!

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
glbrntt: I keep going back and forth on this. I don't feel strongly either way so will leave it for now. 

You're right in that this will happen if something's gone wrong or we've sent the status. In either case sending the status should close the stream, so we should stop receiving messages at that point anyway.

Sources/SwiftGRPCNIO/GRPCStatus.swift
glbrntt: > 1. Should we have a `CollectingServerErrorDelegate` (similar to `CollectingChannelHandler`) in our tests and assert that the error delegate is sent the expected of errors?

This should be pretty straightforward to do. I'll include this in the next revision.

> 2. It might be nice to provide some more error context to `ErrorDelegate`, e.g. which `Server/CallHandler/etc.` (and the corresponding method path) has caught the current error, so we know more precisely where the error occurred.

This is _potentially_ useful for us. However, with the new `enum` of error cases, each one is only thrown in one or two places so I think we can leave this for another day.

glbrntt:
> 1. Should we have a `CollectingServerErrorDelegate` (similar to `CollectingChannelHandler`) in our tests and assert that the error delegate is sent the expected of errors?

This should be pretty straightforward to do. I'll include this in the next revision.

> 2. It might be nice to provide some more error context to `ErrorDelegate`, e.g. which `Server/CallHandler/etc.` (and the corresponding method path) has caught the current error, so we know more precisely where the error occurred.

This is _potentially_ useful for us. However, with the new `enum` of error cases, each one is only thrown in one or two places so I think we can leave this for another day.

Sources/SwiftGRPCNIO/GRPCError.swift
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import Foundation
+
+public enum GRPCError: Error, Equatable {
```
MrMage: Thank you, and well done!

Idea: Should we have the server-only errors in a separate enum, so that client-side users don't need to care for server-only errors, and vice versa?

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -27,111 +27,187 @@ public enum RawGRPCServerResponsePart {
...
+      // Write out a length-delimited message payload. See `processBodyState` for the corresponding format.
+      var responseBuffer = ctx.channel.allocator.buffer(capacity: messageBytes.readableBytes + protobufMetadataSize)
       responseBuffer.write(integer: Int8(0))  // Compression flag: no compression
       responseBuffer.write(integer: UInt32(messageBytes.readableBytes))
       responseBuffer.write(buffer: &messageBytes)
       ctx.write(self.wrapOutboundOut(.body(.byteBuffer(responseBuffer))), promise: promise)
+      outboundState = .expectingBodyOrStatus
+
     case .status(let status):
+      // If we error before sending the initial headers (e.g. unimplemtned method) then we won't have sent the request head.
```
MrMage: typo: unimplemtned

Sources/SwiftGRPCNIO/LoggingServerErrorDelegate.swift
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import Foundation
+
+public class LoggingServerErrorDelegate: ServerErrorDelegate {
+  public init() {}
+
+  public func observe(_ error: Error) {
+    print("[grpc-server] \(error)")
```
MrMage: You could add a timestamp here, but not mandatory.

Tests/SwiftGRPCNIOTests/GRPCChannelHandlerResponseCapturingTestCase.swift
```diff
@@ -0,0 +1,63 @@
...
+  func configureChannel(withHandlers handlers: [ChannelHandler]) -> EventLoopFuture<EmbeddedChannel> {
+    let channel = EmbeddedChannel()
+    return channel.pipeline.addHandlers(handlers, first: true)
+      .map { _ in channel }
+  }
+
+  var errorCollector: CollectingServerErrorDelegate = CollectingServerErrorDelegate()
+
+  override func setUp() {
+    errorCollector.errors.removeAll()
```
MrMage: Is this necessary? I thought that a separate test class instance gets instantiated for each test case.

If you do need to keep this, please make sure to call `super.setUp`, though.

Tests/SwiftGRPCNIOTests/GRPCChannelHandlerTests.swift
```diff
@@ -0,0 +1,195 @@
...
+
+class GRPCChannelHandlerTests: GRPCChannelHandlerResponseCapturingTestCase {
+  func testUnimplementedMethodReturnsUnimplementedStatus() throws {
+    let responses = try waitForGRPCChannelHandlerResponses(count: 1) { channel in
+      let requestHead = HTTPRequestHead(version: .init(major: 2, minor: 0), method: .POST, uri: "unimplementedMethodName")
+      try channel.writeInbound(RawGRPCServerRequestPart.head(requestHead))
+    }
+
+    let expectedError = GRPCError.unimplementedMethod("unimplementedMethodName")
+    XCTAssertEqual(expectedError, errorCollector.errors.first as? GRPCError)
```
MrMage: `XCTAssertEqual([expectedError], errorCollector.errors as? [GRPCError])`, to ensure that we aren't logging extra errors that we don't expect?

Sources/SwiftGRPCNIO/GRPCStatus.swift
MrMage: I'm fine with keeping the `internal`, will try not to mention this in the future. I understand the value of not exposing these methods to avoid confusing users; let's keep this stuff internal until users ask to open this up, I guess.

In general, my reasoning behind making things public/open is to let users drop in at a lower level while still re-using as much of our work as possible. `internal/private` makes this much harder, but of course makes the exposed interface "cleaner". Tough trade-off. What's your stance on this? (I'm unresolving comment this to keep the discussion flowing, but no need for you to make further changes in the code related to this.)

Sources/SwiftGRPCNIO/GRPCError.swift
glbrntt: I was thinking about splitting them too; is having all three actually an issue?

Tests/SwiftGRPCNIOTests/GRPCChannelHandlerTests.swift
glbrntt: Good idea!

Sources/SwiftGRPCNIO/GRPCError.swift
MrMage: Might make it slightly harder to `switch` over them, but it would probably still make sense to split them, to avoid duplication for common errors.

MrMage:
Looking good! @glbrntt, I think all that's left would be comments from you on the two remaining unresolved discussions and fixing the merge conflict.

Sources/SwiftGRPCNIO/GRPCError.swift
glbrntt: For the time being I just renamed it to `GRPCServerError`. The common cases can be extracted when I resolve conflicts with the client PR.

At the moment I think the only place we switch over them is in the user-provided error delegate. We _could_ have different methods for the different enum types:

```
func observe(error: GRPCServerError)
func observe(error: GRPCCommonError)
// fallback to this:
func observe(error: Error)
```

We can default to just calling the `Error` observer for the server and common cases.

MrMage: I like the idea of having different methods; agreed that this should be the topic of a later PR, though.

Tests/LinuxMain.swift
```diff
@@ -38,4 +38,7 @@ XCTMain([
 
   // SwiftGRPCNIO
   testCase(NIOServerTests.allTests)
+  testCase(NIOServerWebTests.allTests)
```
MrMage: Missing commas here, which breaks Linux CI.

glbrntt: 🤦‍♂️

glbrntt:
Looks like we have a fatal error in the NIO Web tests:

```
Test Case 'NIOServerWebTests.testUnaryLotsOfRequests' started at 2019-02-26 13:42:02.497
Fatal error: Trying to remove task, but it's not in the registry.: file Foundation/URLSession/TaskRegistry.swift, line 76
```

@MrMage given these tests weren't included in LinuxMain before do you have any objections to just disabling the failing test (and opening an issue) for now?

MrMage:
> Looks like we have a fatal error in the NIO Web tests:
> 
> ```
> Test Case 'NIOServerWebTests.testUnaryLotsOfRequests' started at 2019-02-26 13:42:02.497
> Fatal error: Trying to remove task, but it's not in the registry.: file Foundation/URLSession/TaskRegistry.swift, line 76
> ```
> @MrMage given these tests weren't included in LinuxMain before do you have any objections to just disabling the failing test (and opening an issue) for now?

Thanks for spotting that! I agree with disabling them for now and opening an issue. This is probably not even related to our code directly, but rather due to a flaw in the Linux implementation of `URLSession`.

Issue #340, opened 2018-12-06
Memory Leak in ServiceClientBase.init(address:secure:arguments:)


crousselle2:
### New Issue Checklist

- [x] Updated grpc-swift to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/grpc/grpc-swift/blob/master/CONTRIBUTING.md)
- [x] I searched for [existing GitHub issues](https://github.com/grpc/grpc-swift/issues)(especially, there are many [resolved issues related to SSL](https://github.com/grpc/grpc-swift/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+ssl))

### Issue Description
`gpr_malloc` blocks are leaking heavily, when trying to instantiate a class subclassing `ServiceClientBase`.
The object is created as such 
`Services_ProtoActionsServiceClient(address: "server_url_here", secure: true)` where `Services_ProtoActionsServiceClient` is a simple `ServiceClientBase` subclass
Everytime we hit that creation and the server api is hit, a `gpr_malloc`is leaked, of pretty large size (250 to 530 KB)
This eventually adds up and leaks a huge amount memory.

Screenshot of profiling: 
<img width="2544" alt="screen shot 2018-12-05 at 4 56 02 pm" src="https://user-images.githubusercontent.com/45021389/49553892-af122500-f8ae-11e8-87c5-cc4cde501598.png">
Stack trace:
<img width="405" alt="screen shot 2018-12-05 at 4 56 34 pm" src="https://user-images.githubusercontent.com/45021389/49553908-be916e00-f8ae-11e8-8d85-2d73c0e350a4.png">
```
   grpc gpr_malloc Pods/gRPC-Core/src/core/lib/gpr/alloc.cc:57
   grpc gpr_strdup Pods/gRPC-Core/src/core/lib/gpr/string.cc:44
   grpc grpc_ssl_credentials_create Pods/gRPC-Core/src/core/lib/security/credentials/ssl/ssl_credentials.cc:92
   SwiftGRPC cgrpc_channel_create_secure /Pods/SwiftGRPC/Sources/CgRPC/shim/channel.c:59
   SwiftGRPC specialized Channel.init(address:secure:arguments:) /Pods/SwiftGRPC/Sources/SwiftGRPC/Core/Channel.swift:51
   SwiftGRPC Channel.init(address:secure:arguments:) /Pods/SwiftGRPC/Sources/SwiftGRPC/Core/Channel.swift:0
[...]

Issue #134, opened 2018-02-23
Should we use `gpr_malloc` in the C shim and/or double-check it for memory leaks?


MrMage:
Given that I found a memory leak in `metadata.c` (and fixed it with #133), would it be advisable to do a full review of the entire C shim library for memory leaks?

Also, the GRPC de-initializers that sometimes get called use `gpr_free`. To be consistent with that, should we allocate all memory using `gpr_malloc` instead of `malloc`? I assume the two are currently interchangeable, but it might make sense to be consistent there to avoid trouble down the road.

WDYT?

jingtianyilong:
Found memory leak with caller `gpr_zalloc` in `alloc.c`, when I call server with high frequency. Memory never get freed.

MrMage:
Thanks for the info. Is that an issue specific to SwiftGRPC or in the underlying GRPC library, though?

jingtianyilong:
I'm not sure actually. But the `alloc.c` located in `SwifGRPC.xcodeproj/Sources/CgRPC/src/core/lib/support/`
So I assume it's with SwiftGRPC.
When calling with high frequency, it generates tons of `Malloc 32.0Kib` according to the Profile Tool on XCode.
I've noticed that there is a `free` function in the `alloc.h`. But memory never actually get freed.

MrMage:
I think I fixed most leaks I could find recently; closing this.

PR #368, opened 2019-02-14
Improve memory management in Channel

kevints:
* Use the gpr family of allocation functions instead of malloc/free (as
  these cannot return `NULL`).
* Explicitly extend lifetime of `argumentWrappers` - Swift doesn't
  guarantee that objects will live to the end of their scope, so
  the optimizer is free to call `Channel.Argument.Wrapper.deinit` before
  the call to `cgrpc_channel_create_secure`, which would result in a
  use-after-free.

### DIFF
--- a/Sources/CgRPC/shim/channel.c
+++ b/Sources/CgRPC/shim/channel.c
@@ -18,15 +18,12 @@
 #include <grpc/support/string_util.h>
 #include <grpc/support/alloc.h>
 
-#include <assert.h>
-#include <stdio.h>
 #include <stdlib.h>
-#include <string.h>
 
 cgrpc_channel *cgrpc_channel_create(const char *address,
                                     grpc_arg *args,
                                     int num_args) {
-  cgrpc_channel *c = (cgrpc_channel *) malloc(sizeof (cgrpc_channel));
+  cgrpc_channel *c = (cgrpc_channel *) gpr_zalloc(sizeof (cgrpc_channel));
 
   grpc_channel_args channel_args;
   channel_args.args = args;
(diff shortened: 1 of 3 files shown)


Sources/CgRPC/shim/channel.c
```diff
@@ -105,8 +102,7 @@ cgrpc_call *cgrpc_channel_create_call(cgrpc_channel *channel,
                                                      NULL);
   grpc_slice_unref(host_slice);
   grpc_slice_unref(method_slice);
-  cgrpc_call *call = (cgrpc_call *) malloc(sizeof(cgrpc_call));
-  memset(call, 0, sizeof(cgrpc_call));
+  cgrpc_call *call = (cgrpc_call *) gpr_zalloc(sizeof(cgrpc_call));
```
MrMage: I assume `gpr_zalloc` is guaranteed to return zeroed memory?

kevints: Yep, here are the docs: https://grpc.io/grpc/cpp/alloc_8h.html#ad9a7a782f4f00cad27d64d27ebbd1a72

kevints:
> Nice! I meant to change the `malloc/free` calls (#134), but apparently never got around to actually do that.
> 
> While at it, would you mind adding the one-line change mentioned in #340 as well, given that it is in a similar spirit?

It's not clear to me from the docs that grpc_secure_channel_create copies the creds object, but tracing through the code on master it does appear to. Since a memory leak is much better than a use-after-free I'll post that change as a follow-up so that someone more knowledgable can chime in on the review.

MrMage:
> > Nice! I meant to change the `malloc/free` calls (#134), but apparently never got around to actually do that.
> > While at it, would you mind adding the one-line change mentioned in #340 as well, given that it is in a similar spirit?
> 
> It's not clear to me from the docs that grpc_secure_channel_create copies the creds object, but tracing through the code on master it does appear to. Since a memory leak is much better than a use-after-free I'll post that change as a follow-up so that someone more knowledgable can chime in on the review.

Good point! It does feel like ownership semantics like this are poorly documented in the gRPC-Core API.

PR #369, opened 2019-02-14
Fix channel credentials memory leak in shims

kevints:


kevints:
Tracing through the code it does seem that `grpc_secure_channel_create` copies the credentials object, but I can't find that documented anywhere so I need someone more knowledgable about the grpc-core implementation to review this PR. Cc @timburks @MrMage 

In the case that it doesn't we could maintain it as a property on cgrpc_channel and release it in cgrpc_channel_destroy.

MrMage:
Given that @kevints and I have both traced through the code and think that releasing channel credentials after use is "correct" (because they are retained/copied by the channel), shall we merge this? @rebello95, what do you think?

rebello95:
Sure, I’m fine with merging if you’re good with the change 👍 (not super familiar with this piece of code myself)

MrMage:
Merging this now.

PR #372, opened 2019-02-22
Store the root certificates as a multi-line string literal.

MrMage:
This makes reviews easier and avoids expensive memory allocations and Base64 conversions at runtime.

The signature of `roots_pem` changes from `String`-returning function to `String` constant; however, that method has only been made public a few days ago, so no users should be affected.


MrMage:
My hunch is that CI is starting to get flaky again (no idea about the reasons, that should probably be investigated soon); closing and re-opening to re-trigger CI.

Sources/RootsEncoder/main.swift
```diff
@@ -98,13 +98,9 @@ if CommandLine.arguments.contains("test") {
   s += "import Foundation\n"
   s += "\n"
 
-  s += "public func roots_pem() -> String {\n"
+  s += "public let roots_pem: String = \"\"\"\n"
```
rebello95: Should we make this a constant using the `k` prefix (i.e., `kRootsPem`)?

MrMage: Great idea! I've renamed the constant to `kRootCertificates` now; what do you think of that name?

rebello95: Yea sounds great!

PR #375, opened 2019-02-23
Make ConnectivityObserver class final

rebello95:
This is `private` to the file and will never be subclassed. I believe updating it with the `final` annotation will make this clearer to readers.

### DIFF
--- a/Sources/SwiftGRPC/Core/Channel.swift
+++ b/Sources/SwiftGRPC/Core/Channel.swift
@@ -135,7 +135,7 @@ private extension Channel {
   final class ConnectivityObserver {
     private let completionQueue: CompletionQueue
     private let underlyingChannel: UnsafeMutableRawPointer
-    private let underlyingCompletionQueue: UnsafeMutableRawPointer = cgrpc_completion_queue_create_for_next()
+    private let underlyingCompletionQueue: UnsafeMutableRawPointer
     private let callback: (ConnectivityState) -> Void
     private var lastState: ConnectivityState
     private var hasBeenShutdown = false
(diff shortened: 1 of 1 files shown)


Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -132,18 +132,17 @@ public class Channel {
 }
 
 private extension Channel {
-  class ConnectivityObserver {
+  final class ConnectivityObserver {
```
MrMage: Given that the containing extension is `private` already, making the class `final` won't make a difference to the compiler and it's unlikely that anyone is going to subclass `ConnectivityObserver` in the same file. However, it won't hurt, either.

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -132,18 +132,17 @@ public class Channel {
...
     private let completionQueue: CompletionQueue
     private let underlyingChannel: UnsafeMutableRawPointer
-    private let underlyingCompletionQueue: UnsafeMutableRawPointer
+    private let underlyingCompletionQueue: UnsafeMutableRawPointer = cgrpc_completion_queue_create_for_next()
```
MrMage: Not sure I like this move, given that it is closely related to the initialization of `self.completionQueue`.

rebello95: Fair, reverted

Sources/SwiftGRPC/Core/Channel.swift
rebello95: Agreed that this won't make a difference to the compiler, but I do think it's nice to annotate classes with `final` by default as opposed to making them subclass-able by default, since it is clearer from the reader's perspective whether something is _currently_ being subclassed (since it could have another subclass elsewhere in the file in this particular example)

MrMage: That makes sense, thank you for explaining.

PR #379, opened 2019-02-24
Run ChannelCrashTests on Linux/CI

rebello95:
When `ChannelCrashTests` was added, it was never configured to run on Linux/CI. This PR:

- Renames the test case to `ChannelConnectivityTests`
- Adds an entry to `LinuxMain` so this is run on CI
- Makes the class `final`
- Updates indentation to be consistent with the rest of the project (2 spaces)


rebello95:
My primary reasoning behind updating it to `final` is to make it clearer to contributors as to which test cases are designed to be subclassed in this project. As-is, not all of the test cases are designed to be subclassed, and I think this helps make the distinction a bit more obvious

MrMage:
> My primary reasoning behind updating it to `final` is to make it clearer to contributors as to which test cases are designed to be subclassed in this project. As-is, not all of the test cases are designed to be subclassed, and I think this helps make the distinction a bit more obvious

👍

PR #380, opened 2019-02-25
Refactor channel connectivity to avoid multiple spin loops

rebello95:
The current Swift implementation of the gRPC channel's connectivity observer spins up a new ConnectivityObserver instance which then starts a new "spin loop" thread and continually runs, observing changes to the underlying gRPC Core channel's connectivity and piping those back through a callback closure. This means that there's a new spin loop spun up for each observer for each channel.

We can avoid having to spin up multiple spin loops for each observer (keeping only 0 or 1 per channel) by allowing a single ConnectivityObserver instance to pipe changes back to multiple callbacks.

### DIFF
--- a/Sources/SwiftGRPC/Core/Channel.swift
+++ b/Sources/SwiftGRPC/Core/Channel.swift
@@ -95,7 +95,9 @@ public class Channel {
   }
 
   deinit {
-    self.connectivityObserver?.shutdown()
+    self.mutex.synchronize {
+      self.connectivityObserver?.shutdown()
+    }
     cgrpc_channel_destroy(self.underlyingChannel)
     self.completionQueue.shutdown()
   }
--- a/Sources/SwiftGRPC/Core/Channel.swift
+++ b/Sources/SwiftGRPC/Core/Channel.swift
@@ -15,7 +15,6 @@
  */
 #if SWIFT_PACKAGE
 import CgRPC
-import Dispatch
 #endif
 import Foundation
 
@@ -126,81 +125,5 @@ public class Channel {
   /// - Parameter callback: block executed every time a new connectivity state is detected
   public func addConnectivityObserver(callback: @escaping (ConnectivityState) -> Void) {
     connectivityObserver.addConnectivityObserver(callback: callback)
-  }
-}
-
-private extension Channel {
-  final class ConnectivityObserver {
-    private let completionQueue: CompletionQueue
-    private let underlyingChannel: UnsafeMutableRawPointer
-    private let underlyingCompletionQueue: UnsafeMutableRawPointer
-    private var callbacks = [(ConnectivityState) -> Void]()
-    private var hasBeenShutdown = false
-    private let stateMutex = Mutex()
-
-    init(underlyingChannel: UnsafeMutableRawPointer) {
-      self.underlyingChannel = underlyingChannel
-      self.underlyingCompletionQueue = cgrpc_completion_queue_create_for_next()
-      self.completionQueue = CompletionQueue(underlyingCompletionQueue: self.underlyingCompletionQueue,
-                                             name: "Connectivity State")
-      self.run()
-    }
-
-    deinit {
-      shutdown()
-    }
-
-    private func run() {
-      let spinloopThreadQueue = DispatchQueue(label: "SwiftGRPC.ConnectivityObserver.run.spinloopThread")
-      var lastState = ConnectivityState(cgrpc_channel_check_connectivity_state(self.underlyingChannel, 0))
-      spinloopThreadQueue.async {
-        while true  {
-          guard (self.stateMutex.synchronize { !self.hasBeenShutdown }) else {
-            return
-          }
-
-          guard let underlyingState = lastState.underlyingState else { return }
-
-          let deadline: TimeInterval = 0.2
-          cgrpc_channel_watch_connectivity_state(self.underlyingChannel, self.underlyingCompletionQueue,
-                                                 underlyingState, deadline, nil)
-
-          let event = self.completionQueue.wait(timeout: deadline)
-          guard (self.stateMutex.synchronize{ !self.hasBeenShutdown }) else {
-            return
-          }
-
-          switch event.type {
-          case .complete:
-            let newState = ConnectivityState(cgrpc_channel_check_connectivity_state(self.underlyingChannel, 0))
-            guard newState != lastState else { continue }
-
-            lastState = newState
-            self.stateMutex.synchronize {
-              self.callbacks.forEach { callback in callback(newState) }
-            }
-
-          case .queueShutdown:
-            return
-
-          default:
-            continue
-          }
-        }
-      }
-    }
-
-    func addConnectivityObserver(callback: @escaping (ConnectivityState) -> Void) {
-      self.stateMutex.synchronize {
-        self.callbacks.append(callback)
-      }
-    }
-
-    func shutdown() {
-      stateMutex.synchronize {
-        hasBeenShutdown = true
-      }
-      completionQueue.shutdown()
-    }
   }
 }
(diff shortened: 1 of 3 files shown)


Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -136,18 +134,16 @@ private extension Channel {
     private let completionQueue: CompletionQueue
```
MrMage: Could we move `class ConnectivityObserver` into a dedicated file?

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -136,18 +134,16 @@ private extension Channel {
...
     private let underlyingChannel: UnsafeMutableRawPointer
     private let underlyingCompletionQueue: UnsafeMutableRawPointer
-    private let callback: (ConnectivityState) -> Void
-    private var lastState: ConnectivityState
+    private var callbacks = [(ConnectivityState) -> Void]()
     private var hasBeenShutdown = false
-    private let stateMutex: Mutex = Mutex()
+    private let stateMutex = Mutex()
```
MrMage: Not sure if the name `stateMutex` still makes sense, given that it guards `callbacks` and `hasBeenShutdown`.

Tests/SwiftGRPCTests/ChannelConnectivityTests.swift
```diff
@@ -46,4 +47,21 @@ extension ChannelConnectivityTests {
 
     waitForExpectations(timeout: 0.5)
   }
+
+  func testMultipleConnectivityObserversAreCalled() {
+    let completionHandlerExpectation = expectation(description: "completion handler called")
+    var firstObserverCalled = false
+    var secondObserverCalled = false
+
+    client.channel.addConnectivityObserver { _ in firstObserverCalled = true }
```
MrMage: How about using expectations for these two as well?

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -47,12 +45,12 @@ public class Channel {
     let argumentWrappers = arguments.map { $0.toCArg() }
 
     underlyingChannel = withExtendedLifetime(argumentWrappers) {
-        var argumentValues = argumentWrappers.map { $0.wrapped }
-        if secure {
-          return cgrpc_channel_create_secure(address, kRootCertificates, nil, nil, &argumentValues, Int32(arguments.count))
-        } else {
-          return cgrpc_channel_create(address, &argumentValues, Int32(arguments.count))
-        }
+      var argumentValues = argumentWrappers.map { $0.wrapped }
```
MrMage: Thanks for the housekeeping!

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -14,28 +14,26 @@
...
 
 /// A gRPC Channel
 public class Channel {
   /// Pointer to underlying C representation
   private let underlyingChannel: UnsafeMutableRawPointer
-
   /// Completion queue for channel call operations
   private let completionQueue: CompletionQueue
+  /// Observer for connectivity state changes.
+  private lazy var connectivityObserver = ConnectivityObserver(underlyingChannel: self.underlyingChannel)
```
MrMage: I'm concerned about making this lazy; this means we would always spin up an observer in `deinit`. Should we make this an optional and populate the variable in `addConnectivityObserver` if it is still nil? While at it, should we synchronize access to this variable? (Depends on whether we want to support adding connectivity observers from different threads.)

Sources/SwiftGRPC/Core/Channel.swift
rebello95: Yep!

Tests/SwiftGRPCTests/ChannelConnectivityTests.swift
rebello95: I had done that initially, but these actually get called multiple times since the state changes more than once. Last I checked, Linux didn't support `assertForOverFulfill` or `expectedFulfillmentCount`, so I think we're stuck with this since CI doesn't use macOS. More info here: https://bugs.swift.org/browse/SR-6249

Sources/SwiftGRPC/Core/Channel.swift
rebello95: Can rename to simply `mutex`

Sources/SwiftGRPC/Core/Channel.swift
rebello95: Ahh yes great catch. I'll definitely update

rebello95:
@MrMage thank you for the review! Updated with feedback addressed if you'd like to take another look

rebello95:
Just noticed what looks like an existing issue where `completionQueue.shutdown()` would be called on `deinit` even if the observer was already shut down. Fixed that while I'm here

Tests/SwiftGRPCTests/ChannelConnectivityTests.swift
MrMage: Ah, that makes sense. Would you mind adding a comment explaining this?

Sources/SwiftGRPC/Core/ChannelConnectivityObserver.swift
```diff
@@ -0,0 +1,97 @@
...
+          }
+
+          switch event.type {
+          case .complete:
+            let newState = ConnectivityState(cgrpc_channel_check_connectivity_state(self.underlyingChannel, 0))
+            guard newState != lastState else { continue }
+
+            lastState = newState
+            self.mutex.synchronize {
+              self.callbacks.forEach { callback in callback(newState) }
```
MrMage: How about `let callbacks = self.mutex.synchronize { self.callbacks }; callbacks.forEach { ... }`? That way, we are holding the lock for a shorter period of time.

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -89,17 +87,17 @@ public class Channel {
...
     }
     completionQueue = CompletionQueue(underlyingCompletionQueue: cgrpc_channel_completion_queue(underlyingChannel), name: "Client")
     completionQueue.run() // start a loop that watches the channel's completion queue
   }
 
   deinit {
-    connectivityObservers.forEach { $0.shutdown() }
-    cgrpc_channel_destroy(underlyingChannel)
-    completionQueue.shutdown()
+    self.connectivityObserver?.shutdown()
```
MrMage: Wrap this in a `mutex.synchronize`? I guess it's unlikely that some other code is accessing `connectivityObserver` while we are already in `deinit`, but better safe than sorry?

rebello95: Yep, I have that change locally in another branch - meant to push it here too

Sources/SwiftGRPC/Core/ChannelConnectivityObserver.swift
rebello95: That makes sense. Don’t we also need to make a copy in that case, though? I.e., `let callbacks = Array(self.mutex.synchronize { self.callbacks })`?

Tests/SwiftGRPCTests/ChannelConnectivityTests.swift
rebello95: Sure

Sources/SwiftGRPC/Core/ChannelConnectivityObserver.swift
MrMage: I would have supposed that copy-on-write takes care of that for us, but to be safe, you could do `self.mutex.synchronize { Array(self.callbacks) }` (doing the copy inside the mutex).

Issue #337, opened 2018-12-03
Stream stops receiving messages after changing connection from Wifi to Cellular


slavabulgakov:
### Question Checklist

- [x] Updated grpc-swift to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/grpc/grpc-swift/blob/master/CONTRIBUTING.md)
- [x] I read [README](https://github.com/grpc/grpc-swift/blob/master/README.md)
- [x] I searched for [existing GitHub issues](https://github.com/grpc/grpc-swift/issues) (especially, there are many [resolved issues related to SSL](https://github.com/grpc/grpc-swift/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+ssl))

### Question Subject
How to restart a stream after changing connection from Wifi to Cellular?

#### Question Description
I use SwiftGRPC in iOS app and I have the client like:
```
service UpdateService {
    rpc GetUpdates (UpdatesConnection) returns (stream Update);
}
```
I subscribe on connectivity state of channel:
```
updateClient.channel.subscribe { connectivityState in
...
}
```
Then I make request to create stream from server:
```
let call = try updateClient.getUpdates(UpdatesConnection()) { result in
	switch result.statusCode {
		…
	} 
}
```
When I change iPhone connection from Wifi to Cellular, the stream stops receiving messages from server but the ```connectivityState``` (subscription to connectivity state) and ```result.statusCode``` keeping silent. After few minutes ```connectivityState``` gets ```.idle``` and ```result.statusCode``` gets ```. unavailable```. When ```result.statusCode``` gets ```. unavailable``` I trying to call ```let call = try updateClient.getUpdates(UpdatesConnection())``` again, but stream doesn't receive messages anyway.
So I've added subscription on iOS network state (Reachability) and I destroy the ```updateClient``` and recreate it when connection get changed from WIfi to Cellular. Only after destroying and recreating the ```updateClient``` the stream works properly - it receive messages.
Am I doing right? Maybe my way is wrong and there is another way to restart stream after changing connection?

rebello95:
The current issue is extensively problematic because end consumers/calls are never notified if a connection goes offline. Instead, gRPC streams/channels stay active even though they're silently offline (though not destroyed). This can result in streams appearing to hang.

[...]

Issue #198, opened 2018-03-30
How do I close the connection?


bigodines:
I've a go server and I'm playing around with using this library for my client code. I haven't found anywhere a way to close a  channel connection I don't need anymore. What is the right way to close the channel between my client and server?

MrMage:
Good point! The channel and its completion queue get destroyed and shut down once the channel is deallocated. So once the last remaining reference to your channel or service client is released, the connection should be closed automatically.

rebello95:
Closing this out, it should be addressed by https://github.com/grpc/grpc-swift/pull/384.

PR #384, opened 2019-02-26
Add ability to manually shut down channels

rebello95:
Addresses https://github.com/grpc/grpc-swift/issues/198.

It's currently not possible to manually shut down a gRPC channel, which means that the only way to shut down a channel is to deallocate it. This requirement is a bit fragile for cases where a consumer wants to manually shut down a channel, since it requires confidence that nothing else is holding a strong reference to the channel.

This PR:
- Adds a `shutdown` function to `Channel`, allowing consumers to arbitrarily shut down connections
- Cancels all existing calls that are using the channel upon shutdown
- Validates that existing calls will throw errors when attempting to read/write from a previously shut down channel
- Ensures creating new calls using previously shut down channels will result in the initializer throwing (already handled by code generators)

This change is increasingly relevant because consumers will need to shut down channels and restart them to help mitigate issues when switching between wifi/cellular as mentioned here: https://github.com/grpc/grpc-swift/issues/337.

### DIFF
--- a/Sources/Examples/Simple/main.swift
+++ b/Sources/Examples/Simple/main.swift
@@ -30,7 +30,7 @@ func client() throws {
 
     let method = (i < steps - 1) ? "/hello" : "/quit"
     print("calling " + method)
-    let call = try! c.makeCall(method)
+    let call = try c.makeCall(method)
 
     let metadata = try Metadata([
       "x": "xylophone",
@@ -38,7 +38,7 @@ func client() throws {
       "z": "zither"
     ])
 
-    try! call.start(.unary, metadata: metadata, message: message) {
+    try call.start(.unary, metadata: metadata, message: message) {
       response in
       print("status:", response.statusCode)
       print("statusMessage:", response.statusMessage!)
--- a/Sources/Examples/Simple/main.swift
+++ b/Sources/Examples/Simple/main.swift
@@ -30,7 +30,7 @@ func client() throws {
 
     let method = (i < steps - 1) ? "/hello" : "/quit"
     print("calling " + method)
-    let call = c.makeCall(method)
+    let call = try! c.makeCall(method)
 
     let metadata = try Metadata([
       "x": "xylophone",
--- a/Sources/SwiftGRPC/Core/Channel.swift
+++ b/Sources/SwiftGRPC/Core/Channel.swift
@@ -21,12 +21,12 @@ import Foundation
 /// A gRPC Channel
 public class Channel {
   private let mutex = Mutex()
-  /// Weak references to API calls using this channel that are in-flight
-  private let activeCalls = NSHashTable<Call>.weakObjects()
   /// Pointer to underlying C representation
   private let underlyingChannel: UnsafeMutableRawPointer
   /// Completion queue for channel call operations
   private let completionQueue: CompletionQueue
+  /// Weak references to API calls using this channel that are in-flight
+  private var activeCalls = [WeakReference<Call>]()
   /// Observer for connectivity state changes. Created lazily if needed
   private var connectivityObserver: ConnectivityObserver?
   /// Whether the gRPC channel has been shut down
@@ -51,39 +51,32 @@ public class Channel {
   /// - Parameter address: the address of the server to be called
   /// - Parameter secure: if true, use TLS
   /// - Parameter arguments: list of channel configuration options
-  public init(address: String, secure: Bool = true, arguments: [Argument] = []) {
+  public convenience init(address: String, secure: Bool = true, arguments: [Argument] = []) {
     gRPC.initialize()
-    host = address
-    let argumentWrappers = arguments.map { $0.toCArg() }
 
-    underlyingChannel = withExtendedLifetime(argumentWrappers) {
+    let argumentWrappers = arguments.map { $0.toCArg() }
+    self.init(host: address, underlyingChannel: withExtendedLifetime(argumentWrappers) {
       var argumentValues = argumentWrappers.map { $0.wrapped }
       if secure {
         return cgrpc_channel_create_secure(address, kRootCertificates, nil, nil, &argumentValues, Int32(arguments.count))
       } else {
         return cgrpc_channel_create(address, &argumentValues, Int32(arguments.count))
       }
-    }
-    completionQueue = CompletionQueue(underlyingCompletionQueue: cgrpc_channel_completion_queue(underlyingChannel), name: "Client")
-    completionQueue.run() // start a loop that watches the channel's completion queue
+    })
   }
 
   /// Initializes a gRPC channel
   ///
   /// - Parameter address: the address of the server to be called
   /// - Parameter arguments: list of channel configuration options
-  public init(googleAddress: String, arguments: [Argument] = []) {
+  public convenience init(googleAddress: String, arguments: [Argument] = []) {
     gRPC.initialize()
-    host = googleAddress
-    let argumentWrappers = arguments.map { $0.toCArg() }
 
-    underlyingChannel = withExtendedLifetime(argumentWrappers) {
+    let argumentWrappers = arguments.map { $0.toCArg() }
+    self.init(host: googleAddress, underlyingChannel: withExtendedLifetime(argumentWrappers) {
       var argumentValues = argumentWrappers.map { $0.wrapped }
       return cgrpc_channel_create_google(googleAddress, &argumentValues, Int32(arguments.count))
-    }
-
-    completionQueue = CompletionQueue(underlyingCompletionQueue: cgrpc_channel_completion_queue(underlyingChannel), name: "Client")
-    completionQueue.run() // start a loop that watches the channel's completion queue
+    })
   }
 
   /// Initializes a gRPC channel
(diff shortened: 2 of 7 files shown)


Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -21,19 +21,31 @@ import Foundation
 /// A gRPC Channel
 public class Channel {
   private let mutex = Mutex()
+  /// Weak references to API calls using this channel that are in-flight
+  private let activeCalls = NSHashTable<Call>.weakObjects()
```
rebello95: Definitely open to suggestions here. Some other options I had considered:
1. Instead storing a `private var shutdownCallbacks: [() -> Void]()` and adding a `register` function to `Channel` which would then be called where the `Call` is instantiated outside of the `Channel`. This felt a bit wasteful since that means each call would have to be manually registered with a channel, and even after a call completes, a long-lived channel would end up with a large list of closures
2. To solve the large list of closures in `#1`, could also add an `unregister` function for a given call which is called when the call is deallocated, but that would get a bit messy with all the required state tracking

The solution I stuck with (`NSHashTable` with `weak` references) seemed like the cleanest

Sources/Examples/Simple/main.swift
```diff
@@ -30,7 +30,7 @@ func client() throws {
 
     let method = (i < steps - 1) ? "/hello" : "/quit"
     print("calling " + method)
-    let call = c.makeCall(method)
+    let call = try! c.makeCall(method)
```
MrMage: How about just `try`?

Sources/SwiftGRPC/Core/Channel.swift
MrMage: Unfortunately, `NSHashTable` is not available on Linux. Another alternative could be to have something like this:

```
// Might need to use `class` instead of `struct`.
struct WeakReference<T: class> {
  weak var value: T?

  init(value: T) {
    self.value = value
  }
}

...

private var activeCalls: [WeakReference<Call>] = []
```

and prune out `nil` references (e.g. via `self.activeCalls = self.activeCalls.filter { $0.value != nil }`) on a regular basis, e.g. every 60 seconds.

Or, do we actually even need this at all? Maybe gRPC-Core automatically shuts down all calls when the corresponding channel has been destroyed, so we wouldn't need to do anything?

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -108,11 +129,20 @@ public class Channel {
...
-    let underlyingCall = cgrpc_channel_create_call(underlyingChannel, method, host, timeout)!
-    return Call(underlyingCall: underlyingCall, owned: true, completionQueue: completionQueue)
+  public func makeCall(_ method: String, host: String? = nil, timeout: TimeInterval? = nil) throws -> Call {
+    guard (self.mutex.synchronize { !self.hasBeenShutdown }) else {
+      throw Error.alreadyShutdown
+    }
+
+    guard let underlyingCall = cgrpc_channel_create_call(
+      self.underlyingChannel, method, host ?? self.host, timeout ?? self.timeout) else
+    {
```
MrMage: nit: move `{` to the preceding line, or have `else { throw ... }` all on the same line?

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -108,11 +129,20 @@ public class Channel {
...
-    let host = host.isEmpty ? self.host : host
-    let timeout = timeout ?? self.timeout
-    let underlyingCall = cgrpc_channel_create_call(underlyingChannel, method, host, timeout)!
-    return Call(underlyingCall: underlyingCall, owned: true, completionQueue: completionQueue)
+  public func makeCall(_ method: String, host: String? = nil, timeout: TimeInterval? = nil) throws -> Call {
+    guard (self.mutex.synchronize { !self.hasBeenShutdown }) else {
+      throw Error.alreadyShutdown
+    }
+
+    guard let underlyingCall = cgrpc_channel_create_call(
```
MrMage: Potential race condition: the channel might have been shut down in between the check above and this call here. You could avoid that by moving the `cgrpc_channel_create_call` into the synchronized closure. (At that point, I would probably just put essentially the entire method's contents into one big sync block, should still be fast enough.)

rebello95: Yep, good call

Sources/SwiftGRPC/Core/Channel.swift
rebello95: > Another alternative could be to have something like this:

Yes, that could work. I can update to use that functionality. 

> prune out `nil` references (e.g. via `self.activeCalls = self.activeCalls.filter { $0.value != nil }`) on a regular basis, e.g. every 60 seconds

Could do that. Not a huge fan of having to run a "cleanup" task, but the only other viable alternatives that come to mind are doing it every time a new call is made (which is probably too often) or every X times a new call is made 😕 

> do we actually even need this at all? Maybe gRPC-Core automatically shuts down all calls when the corresponding channel has been destroyed, so we wouldn't need to do anything?

Unfortunately, although gRPC-Core might be shutting down the underlying calls, it doesn't see, to notify us about them as far as I can tell. For example, if I remove the `self.activeCalls.allObjects.forEach { $0.cancel() }` line, the `testCancelsActiveCallWhenShutdownIsCalled` test fails. I think we're stuck with doing it manually unless I'm missing something

Sources/SwiftGRPC/Core/Channel.swift
rebello95: Moved to using a `lock()` and `unlock()` here since it makes throwing/returning simpler

Sources/Examples/Simple/main.swift
rebello95: Updated - was going off of the other `try!` below. Fixed that one too

Sources/SwiftGRPC/Core/Channel.swift
rebello95: Updated if you can take another look 🙏

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -139,4 +160,39 @@ public class Channel {
...
+    self.mutex.synchronize {
+      self.activeCalls = self.activeCalls.filter { $0.value != nil }
+    }
+    self.scheduleActiveCallCleanUp()
+  }
+}
+
+/// Used to hold weak references to objects since `NSHashTable<T>.weakObjects()` isn't available on Linux.
+/// If/when this type becomes available on Linux, this should be replaced.
+private final class WeakReference<T: AnyObject> {
```
MrMage: nit: move to the top of the file?

Sources/SwiftGRPC/Core/Channel.swift
```diff
@@ -139,4 +160,39 @@ public class Channel {
...
+  weak var value: T?
```
MrMage: `private(set) weak`, to make it more clear that this can only be changed by a deallocation?

Tests/SwiftGRPCTests/ChannelShutdownTests.swift
```diff
@@ -0,0 +1,97 @@
...
+    XCTAssertThrowsError(try self.client.channel.makeCall("foobar")) { error in
+      XCTAssertEqual(.alreadyShutdown, error as? Channel.Error)
+    }
+  }
+
+  func testCallReceiveThrowsWhenChannelIsShutDown() {
+    let call = try! self.client.channel.makeCall("foo")
+    self.client.channel.shutdown()
+
+    XCTAssertThrowsError(try call.receiveMessage(completion: { _ in })) { error in
```
MrMage: would `completion: nil` work as well? Don't know the interface off the top of my head. (also below)

Tests/SwiftGRPCTests/ChannelShutdownTests.swift
```diff
@@ -0,0 +1,97 @@
...
+      XCTAssertEqual(.completionQueueShutdown, error as? CallError)
+      XCTAssertNotNil(self.client.channel)
```
MrMage: What's this for?

Tests/SwiftGRPCTests/ChannelShutdownTests.swift
```diff
@@ -0,0 +1,97 @@
...
+    let errorExpectation = self.expectation(description: "error is returned to call when channel is shut down")
+    let call = try! self.client.channel.makeCall("foo")
+
+    try! call.receiveMessage { result in
+      XCTAssertFalse(result.success)
+      errorExpectation.fulfill()
+    }
+
+    self.client.channel.shutdown()
+    self.waitForExpectations(timeout: 0.1)
```
MrMage: This is less than the regular call timeout, ensuring that the channel shutdown shuts down the call _before it would have been finished due to a timeout, anyway_, right? Would this still pass if you remove `self.activeCalls.forEach { $0.value?.cancel() }` from `channel.shutdown()`? In that case, maybe we could get rid of `Channel.activeCalls` altogether?

Sources/SwiftGRPC/Core/Channel.swift
rebello95: good call

Tests/SwiftGRPCTests/ChannelShutdownTests.swift
rebello95: These closures are non-optional but I can make them trailing closures instead

Tests/SwiftGRPCTests/ChannelShutdownTests.swift
rebello95: Original thought was to make sure that the channel was shut down manually instead of because it was deallocated, but now that you mention it I don't think that really makes sense to test. Removing

Tests/SwiftGRPCTests/ChannelShutdownTests.swift
rebello95: >  This is less than the regular call timeout, ensuring that the channel shutdown shuts down the call _before it would have been finished due to a timeout, anyway_, right?

Yes, correct

>  Would this still pass if you remove `self.activeCalls.forEach { $0.value?.cancel() }` from `channel.shutdown()`?

Nope, commenting out that line causes this test to fail

Tests/SwiftGRPCTests/ChannelShutdownTests.swift
MrMage: Yes, also, the channel is retained at the test case level, so it can't be deallocated before the test case is.

Tests/SwiftGRPCTests/ChannelShutdownTests.swift
MrMage: Thanks for checking!

PR #387, opened 2019-03-01
Add ClientNetworkMonitor for tracking network changes

rebello95:
The SwiftGRPC implementation that is backed by [gRPC-Core(https://github.com/grpc/grpc) (and not SwiftNIO) is known to have some connectivity issues on iOS clients - namely, silently disconnecting (making it seem like active calls/connections are hanging) when switching between wifi <> cellular. The root cause of these problems is that the backing gRPC-Core doesn't get the optimizations made by iOS' networking stack when these types of changes occur, and isn't able to handle them itself.

There is also documentation of this behavior in [this gRPC-Core readme](https://github.com/grpc/grpc/blob/v1.19.0/src/objective-c/NetworkTransitionBehavior.md).

To aid in this problem, we're adding a [`ClientNetworkMonitor`](./Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift) that monitors the device for events that can cause gRPC to disconnect silently. We recommend utilizing this component to call `shutdown()` (or destroy) any active `Channel` instances, and start new ones when the network is reachable.

Details:
- **Switching between wifi <> cellular:** Channels silently disconnect
- **Switching between 3G <> LTE (etc.):** Channels silently disconnect
- **Network becoming unreachable:** Most times channels will time out after a few seconds, but `ClientNetworkMonitor` will notify of these changes much faster
- **Switching between background <> foreground:** No known issues

Original issue: https://github.com/grpc/grpc-swift/issues/337.

### DIFF
--- a/Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
+++ b/Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
@@ -28,16 +28,15 @@ import SystemConfiguration
 open class ClientNetworkMonitor {
     private let queue: DispatchQueue
     private let callback: (State) -> Void
-    private var reachability: SCNetworkReachability?
-    private var appWasInForeground: Bool?
+    private let reachability: SCNetworkReachability
 
     /// Instance of network info being used for obtaining cellular technology names.
-    public private(set) lazy var cellularInfo = CTTelephonyNetworkInfo()
+    public let cellularInfo = CTTelephonyNetworkInfo()
     /// Whether the network is currently reachable. Backed by `SCNetworkReachability`.
     public private(set) var isReachable: Bool?
     /// Whether the device is currently using wifi (versus cellular).
     public private(set) var isUsingWifi: Bool?
-    /// Name of the cellular technology being used (i.e., `CTRadioAccessTechnologyLTE`).
+    /// Name of the cellular technology being used (e.g., `CTRadioAccessTechnologyLTE`).
     public private(set) var cellularName: String?
 
     /// Represents a state of connectivity.
@@ -56,29 +55,32 @@ open class ClientNetworkMonitor {
         case cellularToWifi
         /// The device switched from wifi to cellular.
         case wifiToCellular
-        /// The cellular technology changed (i.e., 3G <> LTE).
+        /// The cellular technology changed (e.g., 3G <> LTE).
         case cellularTechnology(technology: String)
     }
 
-    /// Designated initializer for the network monitor.
+    /// Designated initializer for the network monitor. Initializer fails if reachability is unavailable.
     ///
+    /// - Parameter host:     Host to use for monitoring reachability.
     /// - Parameter queue:    Queue on which to process and update network changes. Will create one if `nil`.
     ///                       Should always be used when accessing properties of this class.
-    /// - Parameter host:     Host to use for monitoring reachability.
     /// - Parameter callback: Closure to call whenever state changes.
-    public init(host: String = "google.com", queue: DispatchQueue? = nil, callback: @escaping (State) -> Void) {
+    public init?(host: String = "google.com", queue: DispatchQueue? = nil, callback: @escaping (State) -> Void) {
+        guard let reachability = SCNetworkReachabilityCreateWithName(nil, host) else {
+            return nil
+        }
+
         self.queue = queue ?? DispatchQueue(label: "SwiftGRPC.ClientNetworkMonitor.queue")
         self.callback = callback
-        self.startMonitoringReachability(host: host)
+        self.reachability = reachability
+        self.startMonitoringReachability(reachability)
         self.startMonitoringCellular()
     }
 
     deinit {
-        if let reachability = self.reachability {
-            SCNetworkReachabilitySetCallback(reachability, nil, nil)
-            SCNetworkReachabilityUnscheduleFromRunLoop(reachability, CFRunLoopGetMain(),
-                                                       CFRunLoopMode.commonModes.rawValue)
-        }
+        SCNetworkReachabilitySetCallback(self.reachability, nil, nil)
+        SCNetworkReachabilityUnscheduleFromRunLoop(self.reachability, CFRunLoopGetMain(),
+                                                   CFRunLoopMode.commonModes.rawValue)
         NotificationCenter.default.removeObserver(self)
     }
 
@@ -86,7 +88,7 @@ open class ClientNetworkMonitor {
 
     private func startMonitoringCellular() {
         let notificationName: Notification.Name
-        if #available(iOS 13, *) {
+        if #available(iOS 12.0, *) {
             notificationName = .CTServiceRadioAccessTechnologyDidChange
         } else {
             notificationName = .CTRadioAccessTechnologyDidChange
@@ -100,7 +102,7 @@ open class ClientNetworkMonitor {
     private func cellularDidChange(_ notification: NSNotification) {
         self.queue.async {
             let newCellularName: String?
-            if #available(iOS 13, *) {
+            if #available(iOS 12.0, *) {
                 let cellularKey = notification.object as? String
                 newCellularName = cellularKey.flatMap { self.cellularInfo.serviceCurrentRadioAccessTechnology?[$0] }
             } else {
@@ -117,11 +119,7 @@ open class ClientNetworkMonitor {
 
     // MARK: - Reachability
 
-    private func startMonitoringReachability(host: String) {
-        guard let reachability = SCNetworkReachabilityCreateWithName(nil, host) else {
-            return
-        }
-
+    private func startMonitoringReachability(_ reachability: SCNetworkReachability) {
         let info = Unmanaged.passUnretained(self).toOpaque()
         var context = SCNetworkReachabilityContext(version: 0, info: info, retain: nil,
                                                    release: nil, copyDescription: nil)
@@ -134,8 +132,6 @@ open class ClientNetworkMonitor {
         SCNetworkReachabilityScheduleWithRunLoop(reachability, CFRunLoopGetMain(),
                                                  CFRunLoopMode.commonModes.rawValue)
         self.queue.async { [weak self] in
-            self?.reachability = reachability
-
             var flags = SCNetworkReachabilityFlags()
             SCNetworkReachabilityGetFlags(reachability, &flags)
             self?.reachabilityDidChange(with: flags)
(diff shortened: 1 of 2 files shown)


Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
```diff
@@ -0,0 +1,162 @@
...
+                                                       CFRunLoopMode.commonModes.rawValue)
+        }
+        NotificationCenter.default.removeObserver(self)
+    }
+
+    // MARK: - Cellular
+
+    private func startMonitoringCellular() {
+        let notificationName: Notification.Name
+        if #available(iOS 13, *) {
```
MrMage: iOS 13?

Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
```diff
@@ -0,0 +1,162 @@
...
+
+    /// A change in network condition.
+    public enum Change: Equatable {
+        /// Reachability changed (online <> offline).
+        case reachability(isReachable: Bool)
+        /// The device switched from cellular to wifi.
+        case cellularToWifi
+        /// The device switched from wifi to cellular.
+        case wifiToCellular
+        /// The cellular technology changed (i.e., 3G <> LTE).
```
MrMage: s/i.e./e.g./

Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
```diff
@@ -0,0 +1,162 @@
...
+        case cellularTechnology(technology: String)
+    }
+
+    /// Designated initializer for the network monitor.
+    ///
+    /// - Parameter queue:    Queue on which to process and update network changes. Will create one if `nil`.
```
MrMage: nit: change parameter order?

Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
```diff
@@ -0,0 +1,162 @@
...
+
+    private func reachabilityDidChange(with flags: SCNetworkReachabilityFlags) {
+        self.queue.async {
+            let isReachable = flags.contains(.reachable)
+            if let wasReachable = self.isReachable, wasReachable != isReachable {
+                self.callback(State(lastChange: .reachability(isReachable: isReachable), isReachable: isReachable))
+            }
+            self.isReachable = isReachable
+
+            let isUsingWifi = !flags.contains(.isWWAN)
```
MrMage: This might send two callbacks in short succession. Not sure if it matters, just an observation.

Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
```diff
@@ -0,0 +1,162 @@
...
+open class ClientNetworkMonitor {
+    private let queue: DispatchQueue
+    private let callback: (State) -> Void
+    private var reachability: SCNetworkReachability?
+    private var appWasInForeground: Bool?
+
+    /// Instance of network info being used for obtaining cellular technology names.
+    public private(set) lazy var cellularInfo = CTTelephonyNetworkInfo()
+    /// Whether the network is currently reachable. Backed by `SCNetworkReachability`.
+    public private(set) var isReachable: Bool?
```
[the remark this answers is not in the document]
MrMage: Okay, now I see that this is also used to call the callback right after initialization, when we get initialized with the first state. I'm not a fan of optional bools, but I guess it works here.

Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
```diff
@@ -0,0 +1,162 @@
...
+import CoreTelephony
+import Dispatch
+import SystemConfiguration
+
+/// This class may be used to monitor changes on the device that can cause gRPC to silently disconnect (making
+/// it seem like active calls/connections are hanging), then manually shut down / restart gRPC channels as
+/// needed. The root cause of these problems is that the backing gRPC-Core doesn't get the optimizations
+/// made by iOS' networking stack when changes occur on the device such as switching from wifi to cellular,
+/// enabling/disabling airplane mode, etc.
+/// Read more: https://github.com/grpc/grpc-swift/tree/master/README.md#known-issues
```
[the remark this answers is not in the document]
rebello95: [`- SeeMore`](https://nshipster.com/swift-documentation/) might make sense here, but I didn't notice it changing anything when rendering docs for me in Xcode. Can change if you feel strongly

Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
```diff
@@ -0,0 +1,162 @@
...
+/// Original issue: https://github.com/grpc/grpc-swift/issues/337
+open class ClientNetworkMonitor {
+    private let queue: DispatchQueue
+    private let callback: (State) -> Void
+    private var reachability: SCNetworkReachability?
+    private var appWasInForeground: Bool?
```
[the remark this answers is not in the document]
rebello95: yes, meant to remove this

Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
```diff
@@ -0,0 +1,162 @@
...
+
+    /// Instance of network info being used for obtaining cellular technology names.
+    public private(set) lazy var cellularInfo = CTTelephonyNetworkInfo()
```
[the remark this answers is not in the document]
rebello95: Can change it to happen at launch

Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
```diff
@@ -0,0 +1,162 @@
...
+
+        NotificationCenter.default.addObserver(self, selector: #selector(self.cellularDidChange(_:)),
+                                               name: notificationName, object: nil)
+    }
+
+    @objc
+    private func cellularDidChange(_ notification: NSNotification) {
+        self.queue.async {
+            let newCellularName: String?
+            if #available(iOS 13, *) {
```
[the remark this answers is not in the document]
rebello95: Should be 12 😬 was testing locally

Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
rebello95: Should be 12

Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
```diff
@@ -0,0 +1,162 @@
...
+            let isReachable = flags.contains(.reachable)
+            if let wasReachable = self.isReachable, wasReachable != isReachable {
+                self.callback(State(lastChange: .reachability(isReachable: isReachable), isReachable: isReachable))
+            }
+            self.isReachable = isReachable
+
+            let isUsingWifi = !flags.contains(.isWWAN)
+            if let wasUsingWifi = self.isUsingWifi, wasUsingWifi != isUsingWifi {
+                self.callback(State(lastChange: isUsingWifi ? .cellularToWifi : .wifiToCellular,
+                                    isReachable: self.isReachable ?? false))
```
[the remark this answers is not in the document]
rebello95: good call

Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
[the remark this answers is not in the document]
rebello95: Right, these are only optional to avoid notifying when the initial state is set asynchronously. Not a huge fan of the optionals, but I thought this would yield the best behavior 🤷‍♂️

Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
```diff
@@ -0,0 +1,162 @@
...
+/// it seem like active calls/connections are hanging), then manually shut down / restart gRPC channels as
+/// needed. The root cause of these problems is that the backing gRPC-Core doesn't get the optimizations
+/// made by iOS' networking stack when changes occur on the device such as switching from wifi to cellular,
+/// enabling/disabling airplane mode, etc.
+/// Read more: https://github.com/grpc/grpc-swift/tree/master/README.md#known-issues
+/// Original issue: https://github.com/grpc/grpc-swift/issues/337
+open class ClientNetworkMonitor {
+    private let queue: DispatchQueue
+    private let callback: (State) -> Void
+    private var reachability: SCNetworkReachability?
```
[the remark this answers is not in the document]
rebello95: Sounds good to me.

Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
rebello95: Yea, I was considering omitting the second one if the first one occurred. Going to update this

rebello95:
Thanks for all the great feedback @MrMage! Updated.

Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
```diff
@@ -0,0 +1,160 @@
...
+        case reachability(isReachable: Bool)
+        /// The device switched from cellular to wifi.
+        case cellularToWifi
+        /// The device switched from wifi to cellular.
+        case wifiToCellular
+        /// The cellular technology changed (e.g., 3G <> LTE).
+        case cellularTechnology(technology: String)
+    }
+
+    /// Designated initializer for the network monitor. Initializer fails if reachability is unavailable.
```
MrMage: Good job explaining the possible failures!

Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
```diff
@@ -0,0 +1,160 @@
...
+import SystemConfiguration
+
+/// This class may be used to monitor changes on the device that can cause gRPC to silently disconnect (making
+/// it seem like active calls/connections are hanging), then manually shut down / restart gRPC channels as
+/// needed. The root cause of these problems is that the backing gRPC-Core doesn't get the optimizations
+/// made by iOS' networking stack when changes occur on the device such as switching from wifi to cellular,
+/// enabling/disabling airplane mode, etc.
+/// Read more: https://github.com/grpc/grpc-swift/tree/master/README.md#known-issues
+/// Original issue: https://github.com/grpc/grpc-swift/issues/337
+open class ClientNetworkMonitor {
```
MrMage: I think you are using 4-spaces indentation here; would you mind changing it to 2-spaces as in the rest of the project? (`make project` is your friend)

Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
```diff
@@ -0,0 +1,160 @@
...
+            let isUsingWifi = !flags.contains(.isWWAN)
+            let isReachable = flags.contains(.reachable)
+
+            let notifyForWifi = self.isUsingWifi != nil && self.isUsingWifi != isUsingWifi
+            let notifyForReachable = self.isReachable != nil && self.isReachable != isReachable
+
+            self.isReachable = isReachable
+            self.isUsingWifi = isUsingWifi
+
+            if notifyForWifi {
```
MrMage: I think now this won't fire at all at launch. That's fine with me, but I guess then we could indeed fire twice if both `isReachable` and `isUsingWifi` have changed at the same time.

Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
rebello95: Ah, yup. Was using SwiftPM directly and forgot to update the preferences 🤦‍♂️

Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
rebello95: Yea that makes sense. I'm leaning towards the latter since notifying on launch is probably not what the consumer wants if they're restarting channels based on this (since they'll likely end up restarting a brand new channel if we notify immediately)

rebello95: Updated to notify on either change

rebello95:
@MrMage thanks. I’m happy to merge manually after that other PR goes in - want to do a couple final checks with this branch on my end.

MrMage:
@rebello95 alright, merge at your convenience!

PR #389, opened 2019-03-03
Refactor some properties shared by all server-side call contexts into a common `ServerCallContext` protocol …

MrMage:
… , while renaming the existing `ServerCallContext` class to `ServerCallContextBase`.

Motivation: As opposed to all three other call types, unary calls are not passed an instance of the previous `ServerCallContext` class, but rather of the `StatusOnlyCallContext` protocol. If one wanted to create a method that uses only a call's event loop and request headers but wants to process all *four* types of calls, that was not possible before. With this change, such a method could simply accept an instance of `ServerCallContext`, which the call arguments provided to all four methods now conform to.

### DIFF
--- a/Sources/SwiftGRPCNIO/ServerCallContexts/UnaryResponseCallContext.swift
+++ b/Sources/SwiftGRPCNIO/ServerCallContexts/UnaryResponseCallContext.swift
@@ -31,7 +31,7 @@ open class UnaryResponseCallContext<ResponseMessage: Message>: ServerCallContext
 /// be fulfilled by the user.
 ///
 /// We can use a protocol (instead of an abstract base class) here because removing the generic `responsePromise` field
-/// lets us avoid associated-type requirements on the protol.
+/// lets us avoid associated-type requirements on the protocol.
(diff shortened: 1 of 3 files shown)


MrMage:
@glbrntt, looks like I can't request a review from you, but please take a look regardless.

Sources/SwiftGRPCNIO/ServerCallContexts/UnaryResponseCallContext.swift
```diff
@@ -32,10 +32,7 @@ open class UnaryResponseCallContext<ResponseMessage: Message>: ServerCallContext
 ///
 /// We can use a protocol (instead of an abstract base class) here because removing the generic `responsePromise` field
 /// lets us avoid associated-type requirements on the protol.
```
rebello95: Not related to your change, but just noticed a typo of "protol" -> "protocol" if you want to fix it while you're here

MrMage: Done, thanks!

PR #392, opened 2019-03-05
Enforce request cardinality for unary-request calls also for the case of zero request messages being sent.

MrMage:
Otherwise, the server will never respond to a call that gets closed without the client sending a response.

In addition, we introduce a method `sendErrorStatus` (happy to discuss naming) on `BaseCallHandler` that sends an error status to the client while ensuring that all call context promises are fulfilled. This method is required (and needs to be overridden) because only the concrete call subclass knows which promises need to be fulfilled.

### DIFF
--- a/Sources/SwiftGRPCNIO/CallHandlers/BaseCallHandler.swift
+++ b/Sources/SwiftGRPCNIO/CallHandlers/BaseCallHandler.swift
@@ -19,7 +19,7 @@ public class BaseCallHandler<RequestMessage: Message, ResponseMessage: Message>:
   /// Called when the client has half-closed the stream, indicating that they won't send any further data.
   ///
   /// Overridden by subclasses if the "end-of-stream" event is relevant.
-  public func endOfStreamReceived() { }
+  public func endOfStreamReceived() throws { }
 
   /// Whether this handler can still write messages to the client.
   private var serverCanWrite = true
@@ -60,7 +60,11 @@ extension BaseCallHandler: ChannelInboundHandler {
       }
 
     case .end:
-      endOfStreamReceived()
+      do {
+        try endOfStreamReceived()
+      } catch {
+        self.errorCaught(ctx: ctx, error: error)
+      }
     }
   }
 }
--- a/Sources/SwiftGRPCNIO/CallHandlers/BaseCallHandler.swift
+++ b/Sources/SwiftGRPCNIO/CallHandlers/BaseCallHandler.swift
@@ -30,6 +30,12 @@ public class BaseCallHandler<RequestMessage: Message, ResponseMessage: Message>:
   public init(errorDelegate: ServerErrorDelegate?) {
     self.errorDelegate = errorDelegate
   }
+  
+  /// Sends an error status to the client while ensuring that all call context promises are fulfilled.
+  /// Because only the concrete call subclass knows which promises need to be fulfilled, this method needs to be overridden.
+  func sendErrorStatus(_ status: GRPCStatus) {
+    fatalError("needs to be overridden")
+  }
 }
 
 extension BaseCallHandler: ChannelInboundHandler {
@@ -40,10 +46,10 @@ extension BaseCallHandler: ChannelInboundHandler {
   /// return a status with code `.internalError`.
   public func errorCaught(ctx: ChannelHandlerContext, error: Error) {
     errorDelegate?.observe(error)
-
+    
     let transformed = errorDelegate?.transform(error) ?? error
     let status = (transformed as? GRPCStatusTransformable)?.asGRPCStatus() ?? GRPCStatus.processingError
-    self.write(ctx: ctx, data: NIOAny(GRPCServerResponsePart<ResponseMessage>.status(status)), promise: nil)
+    sendErrorStatus(status)
   }
 
   public func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
--- a/Sources/SwiftGRPCNIO/CallHandlers/BaseCallHandler.swift
+++ b/Sources/SwiftGRPCNIO/CallHandlers/BaseCallHandler.swift
@@ -46,7 +46,7 @@ extension BaseCallHandler: ChannelInboundHandler {
   /// return a status with code `.internalError`.
   public func errorCaught(ctx: ChannelHandlerContext, error: Error) {
     errorDelegate?.observe(error)
-    
+
     let transformed = errorDelegate?.transform(error) ?? error
     let status = (transformed as? GRPCStatusTransformable)?.asGRPCStatus() ?? GRPCStatus.processingError
     sendErrorStatus(status)
--- a/Sources/SwiftGRPCNIO/CallHandlers/BidirectionalStreamingCallHandler.swift
+++ b/Sources/SwiftGRPCNIO/CallHandlers/BidirectionalStreamingCallHandler.swift
@@ -36,7 +36,7 @@ public class BidirectionalStreamingCallHandler<RequestMessage: Message, Response
     }
   }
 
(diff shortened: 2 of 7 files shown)


Sources/SwiftGRPCNIO/GRPCError.swift
```diff
@@ -143,7 +146,10 @@ extension GRPCServerError: GRPCStatusTransformable {
     case .responseProtoSerializationFailure:
       return GRPCStatus(code: .internalError, message: "could not serialize response proto")
 
-    case .requestCardinalityViolation:
+    case .requestCardinalityViolationTooFew:
```
glbrntt: nit: these names don't sound quite right now, `too{Few,Many}Requests`?

MrMage: Done, thanks!

MrMage:
@glbrntt I've slightly extended the scope of this PR; please take another look.

Sources/SwiftGRPCNIO/CallHandlers/BidirectionalStreamingCallHandler.swift
```diff
@@ -36,9 +36,13 @@ public class BidirectionalStreamingCallHandler<RequestMessage: Message, Response
...
 
-  public override func endOfStreamReceived() {
+  public override func endOfStreamReceived() throws {
     eventObserver?.whenSuccess { observer in
       observer(.end)
     }
   }
+  
+  override func sendErrorStatus(_ status: GRPCStatus) {
+    context?.statusPromise.fail(error: status)
```
glbrntt: Where are we `write`-ing the status now?

Sources/SwiftGRPCNIO/GRPCError.swift
glbrntt: Ah, I meant just replacing the names!
- `requestCardinalityViolationTooFew` -> `tooFewRequests`
- `requestCardinalityViolationTooMany` -> `tooManyRequests`

MrMage: Renamed again :-)

Sources/SwiftGRPCNIO/CallHandlers/BidirectionalStreamingCallHandler.swift
MrMage: The call contexts schedule closures writing the status off the corresponding promises; see https://github.com/grpc/grpc-swift/blob/master/Sources/SwiftGRPCNIO/ServerCallContexts/UnaryResponseCallContext.swift and https://github.com/grpc/grpc-swift/blob/master/Sources/SwiftGRPCNIO/ServerCallContexts/StreamingResponseCallContext.swift.

glbrntt: Ah of course, thanks!

Sources/SwiftGRPCNIO/CallHandlers/BaseCallHandler.swift
```diff
@@ -40,10 +46,10 @@ extension BaseCallHandler: ChannelInboundHandler {
   /// return a status with code `.internalError`.
   public func errorCaught(ctx: ChannelHandlerContext, error: Error) {
     errorDelegate?.observe(error)
-
+
```
rebello95: nit: Revert?

Sources/SwiftGRPCNIO/GRPCError.swift
```diff
@@ -87,8 +87,11 @@ public enum GRPCServerError: Error, Equatable {
   /// It was not possible to serialize the response protobuf.
   case responseProtoSerializationFailure
 
+  /// Zero requests were sent for a unary-request call.
```
rebello95: Will this always be zero? If so, maybe the error should be `noRequestsSent`

Tests/SwiftGRPCNIOTests/NIOServerWebTests.swift
```diff
@@ -42,24 +42,30 @@ class NIOServerWebTests: NIOBasicEchoTestCase {
     data.insert(UInt8(0), at: 0)
     return data
   }
-
-  private func gRPCWebOKTrailers() -> Data {
-    var data = "grpc-status: 0\r\ngrpc-message: OK".data(using: .utf8)!
+
```
rebello95: Revert

Tests/SwiftGRPCNIOTests/NIOServerWebTests.swift
```diff
@@ -42,24 +42,30 @@ class NIOServerWebTests: NIOBasicEchoTestCase {
...
     // Add the gRPC prefix with the compression byte and the 4 length bytes.
     for i in 0..<4 {
       data.insert(UInt8((data.count >> (i * 8)) & 0xFF), at: 0)
     }
     data.insert(UInt8(0x80), at: 0)
     return data
   }
+  
+  private func gRPCWebOKTrailers() -> Data {
+    return gRPCWebTrailers(status: 0, message: "OK")
```
rebello95: Curious, why make this 1-line function instead of inlining the other `gRPCWebOKTrailers` function?

Sources/SwiftGRPCNIO/GRPCError.swift
MrMage: Renamed to `noRequestsButOneExpected` (to make it clear what the expectation is); let me know if that works for you.

rebello95: sgtm

Tests/SwiftGRPCNIOTests/NIOServerWebTests.swift
MrMage: Not sure what you mean with that, but I changed this to `fun gRPCWebTrailers(status: Int = 0, message: String = "OK")` now.

MrMage:
@rebello95 please take another look, especially at the unresolved conversations. Feel free to merge if you are satisfied with the current revision.

Tests/SwiftGRPCNIOTests/NIOServerWebTests.swift
rebello95: Looks like there's still trailing whitespace here

Tests/SwiftGRPCNIOTests/NIOServerWebTests.swift
rebello95: Sorry, typo. I was basically wondering why `gRPCWebTrailers(status: Int, message: String)` is a separate function since this is the only thing that calls it. This is fine though since `gRPCWebOKTrailers` provides clarity as to the intent

Tests/SwiftGRPCNIOTests/NIOServerWebTests.swift
```diff
@@ -85,7 +91,25 @@ extension NIOServerWebTests {
         completionHandlerExpectation.fulfill()
       }
     }
-
+    
+    waitForExpectations(timeout: defaultTestTimeout)
+  }
+  
+  func testUnaryWithoutRequestMessage() {
```
rebello95: Just realized that this also needs to be added to `allTests` to run on Linux

MrMage:
Looks like I forgot to actually commit and push my fixes, sorry about that! @rebello95, please take one final look.

Tests/SwiftGRPCNIOTests/NIOServerWebTests.swift
MrMage: FYI, it's actually used with a different argument in `testUnaryWithoutRequestMessage`.

PR #394, opened 2019-03-06
Add public factory methods for CallResult.

mpetrov:
This enables instantiating CallResult structs for test stubs.


fandyfyf:
👍

mpetrov:
Thanks Daniel!

This could be used for stubbing out the async calls as follows:

```
  override func get(_ request: Echo_EchoRequest, metadata customMetadata: Metadata, completion: @escaping (Echo_EchoResponse?, CallResult) -> Void) throws -> Echo_EchoGetCall {
    completion(nil, .error(statusCode: .notFound))
```

The bazel rules don't currently enable stub generation, but once that's in place I'll likely also change the default stub code as well.

MrMage:
Makes sense, thank you for the elaboration!

Issue #390, opened 2019-03-04
Split package for CgRPC and NIO variants


glbrntt:
The CgRPC and NIO variants are disjoint (well, almost: `StatusCode.swift` and `echo.proto` are shared) but live in the same Swift package and repository.

This has a couple of drawbacks:
- Both are tied to the same Swift version.
- They share a build system and CI.

In the NIO implementation we would like to use a newer version of swift-nio-http2 to drop the nghttp2 dependency. Doing this requires using NIO 2.0 which is [Swift 5 only](https://github.com/apple/swift-nio#prerequisites).

Two obvious approaches are:
1. separate repositories (i.e. "grpc-swift" and "grpc-swift-nio"), and
1. separate branches (with different major versions).

Part of this rests on how long the CgRPC implementation will be around. If support is dropped in favour of the NIO implementation then separate branches with different major release versions should be fine. If both are supported long term then separate repositories would probably make more sense since they are separate projects.

MrMage:
I agree that this is probably going to become necessary, and having separate repositories would probably make it easier to use both libraries at the same time for a transition period. My only concern about opening a separate repository would be possible "bureaucratic overhead" and a loss of mind share, because it would get harder for people to discover the NIO variant of the client.

On the other hand, I do hope that we'll be able to at least deprecate the non-NIO variant this year, at which point the NIO variant would become the "recommended" variant.

glbrntt:
> On the other hand, I do hope that we'll be able to at least deprecate the non-NIO variant this year, at which point the NIO variant would become the "recommended" variant.

If that's the case then separate branches should work well. We could always move the CgRPC variant out of master to increase discoverability for the NIO variant ;)

glbrntt:
What are your thoughts on this @timburks @rebello95?

MrMage:
@glbrntt to unblock you, I could create a `nio` branch that we merge all upcoming work on SwiftGRPCNIO into. It's unlikely that we'll need to cut a release based on NIO changes anytime soon, so this would let us continue working on the NIO variant while leaving us the option to converge both branches again in the future (if possible) — even if you need to make a few small changes to disable the gRPC-Core version on the `nio` branch for the time being. (That could be done by simply removing the corresponding target from `Package.swift`.) Alternatively, we could still split out the NIO work into a separate repository later on.

I think this would be an unbureaucratic way forward which still leaves all options open for the future. WDYT?

glbrntt:
[...]

PR #395, opened 2019-03-06
Fix crashes due to mismatching responses sent to the channel when event observer factories fail.

MrMage:


MrMage:
A review from @glbrntt would be appreciated as well :-)

rebello95:
Happy to take a look after @glbrntt since he probably has a bit more context on this.

Tests/LinuxMain.swift
```diff
@@ -34,11 +34,15 @@ XCTMain([
   testCase(MetadataTests.allTests),
   testCase(ServerCancellingTests.allTests),
   testCase(ServerTestExample.allTests),
-  testCase(ServerThrowingTests.allTests),
+  testCase(SwiftGRPCTests.ServerThrowingTests.allTests),
   testCase(ServerTimeoutTests.allTests),
 
   // SwiftGRPCNIO
+  //! TODO: Is there a better way to split tests for our two library targets?
```
glbrntt: How about #390?

Tests/SwiftGRPCNIOTests/ServerThrowingTests.swift
```diff
@@ -0,0 +1,152 @@
...
+import Foundation
+import NIO
+import NIOHTTP1
+import NIOHTTP2
+@testable import SwiftGRPCNIO
+import XCTest
+
+private let expectedError = GRPCStatus(code: .internalError, message: "expected error")
+
+// Motivation for two different provides: Throwing immediately causes the event observer future (in the
```
glbrntt: typo: provides -> providers

Tests/SwiftGRPCNIOTests/ServerThrowingTests.swift
```diff
@@ -0,0 +1,152 @@
...
+// client-streaming and bidi-streaming cases) to throw immediately, _before_ the corresponding handler has even added
+// to the channel. We want to test that case as well as the one were we throw only _after_ the handler has been added
```
glbrntt: typo: "one were we" -> "one where we"

Sources/SwiftGRPCNIO/CallHandlers/BidirectionalStreamingCallHandler.swift
```diff
@@ -21,14 +21,23 @@ public class BidirectionalStreamingCallHandler<RequestMessage: Message, Response
...
       // When done, reset references to avoid retain cycles.
       self.eventObserver = nil
       self.context = nil
     }
   }
+  
+  public override func handlerAdded(ctx: ChannelHandlerContext) {
+    guard let eventObserver = eventObserver,
+      let context = context else { return }
+    // Terminate the call if providing an observer fails.
```
glbrntt: I'm not sure I follow this line, do you mean "Terminate the call if the provided observer fails"?

The below explanation is helpful though! 👍

Tests/LinuxMain.swift
MrMage: Of course 😅

Sources/SwiftGRPCNIO/CallHandlers/BidirectionalStreamingCallHandler.swift
MrMage: Added "the future [providing an observer]", hope that makes it more clear: Users return a future containing the observer block. This lets them simply fail that future (with a `GRPCStatus`) if they do not want to receive messages on this call at all, e.g. in case of authentication failure (e.g. if the request metadata doesn't contain user credentials).

MrMage:
Updated the PR :-)

Tests/SwiftGRPCNIOTests/ServerThrowingTests.swift
```diff
@@ -0,0 +1,152 @@
...
+ */
+import Dispatch
+import Foundation
+import NIO
+import NIOHTTP1
+import NIOHTTP2
+@testable import SwiftGRPCNIO
+import XCTest
+
+private let expectedError = GRPCStatus(code: .internalError, message: "expected error")
```
rebello95: nit: Maybe prefix as `kExpectedError` since this is a global constant?

Tests/SwiftGRPCNIOTests/ServerThrowingTests.swift
```diff
@@ -0,0 +1,152 @@
...
+  }
+  
+  func update(context: StreamingResponseCallContext<Echo_EchoResponse>) -> EventLoopFuture<(StreamEvent<Echo_EchoRequest>) -> Void> {
+    return context.eventLoop.newFailedFuture(error: expectedError)
+  }
+}
+
+private extension EventLoop {
+  func newFailedFuture<T>(error: Error, delay: TimeInterval) -> EventLoopFuture<T> {
+    return self.scheduleTask(in: .nanoseconds(TimeAmount.Value(delay * 1000 * 1000 * 1000))) { () }.futureResult
```
rebello95: Why not use `seconds` here instead of multiplying?

Tests/SwiftGRPCNIOTests/ServerThrowingTests.swift
```diff
@@ -0,0 +1,152 @@
...
+    XCTAssertNoThrow(try call.sendEnd().wait())
+    XCTAssertEqual(expectedError, try call.status.wait())
+    
+    if type(of: makeEchoProvider()) != ErrorReturningEchoProviderNIO.self {
+      // With `ErrorReturningEchoProviderNIO` we actually _return_ a response, which means that the `response` future
+      // will _not_ fail.
+      XCTAssertThrowsError(try call.response.wait()) {
+        XCTAssertEqual(expectedError, $0 as? GRPCStatus)
+      }
+    }
```
rebello95: Should we fail on the `else` case?

rebello95: If so, might make more sense to `guard type(of: makeEchoProvider()) != ErrorReturningEchoProviderNIO.self  {} else { XCTFail(...) }`

Tests/SwiftGRPCNIOTests/ServerThrowingTests.swift
```diff
@@ -0,0 +1,152 @@
...
+  }
+  
+  func testServerStreaming() {
+    let call = client.expand(Echo_EchoRequest(text: "foo")) { _ in XCTFail("no message expected") }
```
rebello95: Could be useful to output the message in the `XCTFail("no message expected, got \($0)")`

Tests/SwiftGRPCNIOTests/ServerThrowingTests.swift
```diff
@@ -0,0 +1,152 @@
...
+    // Nothing to throw here, but the `status` should be the expected error.
+    XCTAssertEqual(expectedError, try call.status.wait())
+  }
+  
+  func testBidirectionalStreaming() {
+    let call = client.update() { _ in XCTFail("no message expected") }
```
rebello95: Same here, up to you

Tests/SwiftGRPCNIOTests/ServerThrowingTests.swift
MrMage: Given that this is a private fixture used in just one file, I prefer to not use the `k` prefix in this case.

Tests/SwiftGRPCNIOTests/ServerThrowingTests.swift
MrMage: `TimeInterval` is floating-point, while `TimeAmount.Value` is integer-valued, so we'd lose a ton of precision.

Tests/SwiftGRPCNIOTests/ServerThrowingTests.swift
MrMage: No, we do not want to do this check at all in the case of `ErrorReturningEchoProviderNIO`, so failing would be counter-productive. I hopefully clarified the comment.

PR #397, opened 2019-03-08
Rewrite LengthPrefixedMessageReader and tests

glbrntt:
- Hit a bug when testing against NIO 2 and the pure Swift nio-http2
- Rewrote the message reader to allow the buffer to be split at any point
- Added tests

### DIFF
--- a/Sources/SwiftGRPCNIO/HTTP1ToRawGRPCClientCodec.swift
+++ b/Sources/SwiftGRPCNIO/HTTP1ToRawGRPCClientCodec.swift
@@ -52,7 +52,7 @@ public final class HTTP1ToRawGRPCClientCodec {
   }
 
   private var state: State = .expectingHeaders
-  private let messageReader = LengthPrefixedMessageReader(mode: .client)
+  private let messageReader = LengthPrefixedMessageReader(mode: .client, compressionMechanism: .none)
   private let messageWriter = LengthPrefixedMessageWriter()
   private var inboundCompression: CompressionMechanism = .none
 }
@@ -93,13 +93,15 @@ extension HTTP1ToRawGRPCClientCodec: ChannelInboundHandler {
       throw GRPCError.client(.HTTPStatusNotOk(head.status))
     }
 
-    if let encodingType = head.headers["grpc-encoding"].first {
-      self.inboundCompression = CompressionMechanism(rawValue: encodingType) ?? .unknown
-    }
+    let inboundCompression: CompressionMechanism = head.headers["grpc-encoding"]
+      .first
+      .map { CompressionMechanism(rawValue: $0) ?? .unknown } ?? .none
 
     guard inboundCompression.supported else {
       throw GRPCError.client(.unsupportedCompressionMechanism(inboundCompression.rawValue))
     }
+
+    self.messageReader.compressionMechanism = inboundCompression
 
     ctx.fireChannelRead(self.wrapInboundOut(.headers(head.headers)))
     return .expectingBodyOrTrailers
@@ -114,7 +116,8 @@ extension HTTP1ToRawGRPCClientCodec: ChannelInboundHandler {
       throw GRPCError.client(.invalidState("received body while in state \(state)"))
     }
 
-    for message in try self.messageReader.consume(messageBuffer: &messageBuffer, compression: inboundCompression) {
+    self.messageReader.append(buffer: &messageBuffer)
+    while let message = try self.messageReader.nextMessage() {
       ctx.fireChannelRead(self.wrapInboundOut(.message(message)))
     }
 
--- a/Tests/SwiftGRPCNIOTests/LengthPrefixedMessageReaderTests.swift
+++ b/Tests/SwiftGRPCNIOTests/LengthPrefixedMessageReaderTests.swift
@@ -1,0 +1,247 @@
+import Foundation
+import XCTest
+import SwiftGRPCNIO
+import NIO
+
+class LengthPrefixedMessageReaderTests: XCTestCase {
+  static var allTests: [(String, (LengthPrefixedMessageReaderTests) -> () throws -> Void)] {
+    return [
+      ("testNextMessageReturnsNilWhenNoBytesAppended", testNextMessageReturnsNilWhenNoBytesAppended),
+      ("testNextMessageReturnsMessageIsAppendedInOneBuffer", testNextMessageReturnsMessageIsAppendedInOneBuffer),
+      ("testNextMessageReturnsMessageForZeroLengthMessage", testNextMessageReturnsMessageForZeroLengthMessage),
+      ("testNextMessageDeliveredAcrossMultipleByteBuffers", testNextMessageDeliveredAcrossMultipleByteBuffers),
+      ("testNextMessageWhenMultipleMessagesAreBuffered", testNextMessageWhenMultipleMessagesAreBuffered),
+      ("testNextMessageReturnsNilWhenNoMessageLengthIsAvailable", testNextMessageReturnsNilWhenNoMessageLengthIsAvailable),
+      ("testNextMessageReturnsNilWhenNotAllMessageLengthIsAvailable", testNextMessageReturnsNilWhenNotAllMessageLengthIsAvailable),
+      ("testNextMessageReturnsNilWhenNoMessageBytesAreAvailable", testNextMessageReturnsNilWhenNoMessageBytesAreAvailable),
+      ("testNextMessageReturnsNilWhenNotAllMessageBytesAreAvailable", testNextMessageReturnsNilWhenNotAllMessageBytesAreAvailable),
+      ("testNextMessageThrowsWhenCompressionMechanismIsNotSupported", testNextMessageThrowsWhenCompressionMechanismIsNotSupported),
+      ("testNextMessageThrowsWhenCompressionFlagIsSetButNotExpected", testNextMessageThrowsWhenCompressionFlagIsSetButNotExpected),
+      ("testNextMessageDoesNotThrowWhenCompressionFlagIsExpectedButNotSet", testNextMessageDoesNotThrowWhenCompressionFlagIsExpectedButNotSet),
+      ("testAppendReadsAllBytes", testAppendReadsAllBytes),
+    ]
+  }
+
+  var reader = LengthPrefixedMessageReader(mode: .client, compressionMechanism: .none)
+
+  var allocator = ByteBufferAllocator()
+
+  func byteBuffer(withBytes bytes: [UInt8]) -> ByteBuffer {
+    var buffer = allocator.buffer(capacity: bytes.count)
+    buffer.write(bytes: bytes)
+    return buffer
+  }
+
+  final let twoByteMessage: [UInt8] = [0x01, 0x02]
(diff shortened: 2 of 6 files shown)


Tests/SwiftGRPCNIOTests/LengthPrefixedMessageReaderTests.swift
```diff
@@ -0,0 +1,217 @@
...
+    ]
+
+    var buffer = byteBuffer(withBytes: bytes)
+    reader.append(buffer: &buffer)
+
+    XCTAssertThrowsError(try reader.nextMessage())
+  }
+
+  func testNextMessageThrowsWhenCompressionFlagIsNotSetButIsExpected() throws {
+    // `.identity` should always be supported and requires a flag.
```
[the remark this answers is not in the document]
glbrntt: Just re-checked the spec, you're totally right there!

weissi:
@glbrntt wouldn't this make sense as a `ChannelHandler`? Especially a `ByteToMessageDecoder` should make the implementation of this very straightforward. swift-nio-extras also has [`LengthFieldBasedFrameDecoder`](https://github.com/apple/swift-nio-extras/blob/master/Sources/NIOExtras/LengthFieldBasedFrameDecoder.swift) which is pretty similar already I guess?

glbrntt:
It _could_ be a channel handler, but what would the advantage be?

It's used within a two channel handlers already, and it seems like splitting up the logic in those handlers (i.e. dealing with the request/response head and end in one handler and the message in another) would make things harder to reason about?

weissi:
> It _could_ be a channel handler, but what would the advantage be?

- reusability
- correctness (mostly regarding re-entrancy but I'm not sure if you would have this problem here)
- visibility/debugability through the ChannelPipeline
- separation of concerns (only if they are actually separate concerns)
 
> It's used within a two channel handlers already, and it seems like splitting up the logic in those handlers (i.e. dealing with the request/response head and end in one handler and the message in another) would make things harder to reason about?

if it would make things harder, then fair enough. It just felt like some a task that's quite isolated and unrelated to the rest and therefore I thought it might be good to separate the concerns and have one channel handler do the length prefixing and the other handlers wouldn't need to bother with that. But I haven't actually read the code :)

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCClientCodec.swift
```diff
@@ -114,7 +116,8 @@ extension HTTP1ToRawGRPCClientCodec: ChannelInboundHandler {
       throw GRPCError.client(.invalidState("received body while in state \(state)"))
     }
 
-    for message in try self.messageReader.consume(messageBuffer: &messageBuffer, compression: inboundCompression) {
+    self.messageReader.append(buffer: &messageBuffer)
+    while let message = try self.messageReader.nextMessage() {
```
weissi: this really feels like this should be done as a `ByteToMessageDecoder`. I think my re-entrancy concerns are real here and I guess you might want to add a test case: Imagine this happens:

- you enter `channelRead` -> `processBody`
- let's assume `self.messageReader` contains 3 messages
- the first of the 3 messages is sent through the pipeline using the `ctx.fireChannelRead(self.wrapInboundOut(.message(message)))` below
- another `ChannelHandler` further down the pipeline decides to `channel.close`
- `channel.close` will force NIO to drain the socket buffer, ie. there might be more input coming from the socket
- the remaining bytes in the socket buffer are now fed back into `channelRead` but we're still on the same stack frame

what you'll now get is reordering. So let's assume there were 4 messages sent in total. As in the example above, 3 were delivered in one `channelRead`. When delivering the 1st, the next `ChannelHandler` will see this but as above, that next `ChannelHandler` will now cause a `close` to happen which might trigger another `channelRead` where we'd receive message 4 but that's now processed directly.

So a subsequent `ChannelHandler` would see `1, 4, 2, 3`

weissi: here's one of our test cases that fakes this scenario: https://github.com/apple/swift-nio/blob/87ee03c4ce4eb161519aff9b9dba65ed3f2ae23a/Tests/NIOTests/CodecTest.swift#L529-L564

in this test case instead of triggering the complicated case where `channelRead -> close -> channelRead` it just uses the `EmbeddedChannel` and fabricates and inbound messages whilst in `decode` which leads to exactly the same situation.

It's also not just `close` within `channelRead` that can cause re-entrancy. With the SSL handler I believe `write` can trigger a `channelRead`

weissi: @glbrntt / @MrMage what I just realise here is that unfortunately you would need a `MessageToMessageDecoder` here and not a `ByteToMessageDecoder` (because your `InboundIn == HTTPClientResponsePart` and not `ByteBuffer`). Unfortunately, NIO hasn't implemented a `MessageToMessageDecoder` yet... We should do that soon.

glbrntt: That makes sense, thanks for the detailed explanation @weissi!

Is `MessageToMessageDecoder` planned work? Do you have any suggestions for mitigating this in the meantime?

weissi: @glbrntt we haven't started work on it but this could be a good argument for doing it :). The mitigation is annoying but not too hard: You need to remember if you're currently calling out and if you are, you need to buffer any further arriving data... Something like

```swift
enum State {
    case normal
    case alreadyCallingOut(CircularBuffer<InboundIn>?)
}
```

then if `normal`, you'd just do everything as before but before calling out you would just enqueue the thing into the CircularBuffer for consumption when the `fireChannelRead` returns.

It's quite an annoying bit of code to write though. You could also decide to file a bug and fix this later. Please feel free to file a NIO feature request if you do so.

MrMage: @weissi wow, that re-ordering does sound _bad_. Sorry for the silly question, but how would this be mitigated if we implemented this as a `ChannelHandler` instead?

glbrntt: I don't think Implementing it as a `ChannelHandler` _without the above state tracking_ would mitigate this by itself.

`ByteToMessageDecoder` (well, `MessageToMessageDecoder` when it exists) would handle the state tracking for us.

weissi: indeed, what @glbrntt says. Today, you'd need to implement the buffering on re-entrancy yourself which is a bit annoying.

MrMage: Ouch. Does this mean that any `ChannelHandler` potentially needs to be re-entrant? Wouldn't it be easier to have that handled by the framework, or have the framework simply avoid such re-entrancy? (You have thought more about this than I of course; there are probably good reasons why the mentioned `Decoder`s are the better solution to this.)

weissi: > Ouch. Does this mean that any `ChannelHandler` potentially needs to be re-entrant? 

Correct, unless you have a `Channel` or `ChannelHandler` that guarantees that won't cause re-entrancy. You can also implement a handler up front that breaks re-entrancy originating from the `Channel` itself. [`HTTPServerPipelineHandler`](https://github.com/apple/swift-nio/blob/master/Sources/NIOHTTP1/HTTPServerPipelineHandler.swift) does things along those lines.

> Wouldn't it be easier to have that handled by the framework, or have the framework simply avoid such re-entrancy? (You have thought more about this than I of course; there are probably good reasons why the mentioned `Decoder`s are the better solution to this.)

It would make things easier but unfortunately also slow. That adds basically two buffers to each channel handler. We are thinking about making a thing which you can opt in to which solves re-entrancy for you. It could for example be a wrapper around any `ChannelHandler` breaking re-entrancy. There's a lot that we can improve here, for now we're not super helpful, just like Netty is. (unless you use `ByteToMessageDecoder`s)

For gRPC it's actually quite interesting because you're using `swift-nio-http2` which might already give you the guarantees you need. To implement HTTP/2 @lukasa already had to deal with reentrancy being tough a lot so it might be that what he implemented gives you no re-entrancy for handlers further back anyway and we can ignore all this here :).

---

side note: When I say a ChannelHandler could guarantee no re-entrancy to the back of it I only mean re-entrancy caused by the Channel or handlers before. No ChannelHandler will be able to protect about crazy code like

```swift
context.channel.pipeline.context(type: TypeOfHandlerInFrontOfMe.self).flatMap { context in
    context.fireChannelRead(...)
}
```

where a handler might inject data at any point to a handler before it.

MrMage:
Thanks for the explanations, @weissi! @weissi, @glbrntt, what's your stance on this PR? Should we merge this now and revisit refactoring this into a `ChannelHandler/Decoder` later, when needed/possible? Same about possible reentrancy issues, which might already be mitigated by the HTTP2 implementation in front of us?

glbrntt:
I'd say merge because it is more correct than before, even if it doesn't address the reentrancy issue. We can solve that problem separately whether it's now or later.

Given the comments re: HTTP/2 _possibly_ resolving this issue we should work out whether we actually need to address this as well. I have a Swift 5 update in the works so should be able to test this.

Lukasa:
> Same about possible reentrancy issues, which might already be mitigated by the HTTP2 implementation in front of us?

Sadly, they are not mitigated by the HTTP/2 implementation in front of you. In general it is not possible for a ChannelHandler to prevent the rest of the pipeline from behaving re-entrantly, and this implementation does not try.

What it does do is guarantee that the HTTP/2 protocol implementation *itself* is safe against arbitrary-reentrancy. What this means is that, regardless of whatever madness occurs in the rest of the pipeline, the HTTP/2 protocol invariants will be respected by the core HTTP/2 implementation. This is a bit gnarly internally, so it's hard to reproduce here, but as @weissi says this may be a good impetus to investigate providing better general solutions to this.

Lukasa:
So @weissi suggested I clarify by expressing the actual behaviour.

The new `HTTP2Handler` can trigger re-entrancy in the following ways:

1. When your `ChannelHandler` calls `write`, the `HTTP2Handler` may call `fireUserInboundEventTriggered`. If you don't catch user events, you'll never notice this.
2. During `channelRead`, the `HTTP2Handler` may invoke `write` to write an automatic frame.

If you never pay attention to user events, and you never place any handlers of your own between the `HTTP2Handler` and the network, you never need to worry about re-entrancy triggered by that handler.

weissi:
> The new `HTTP2Handler` can trigger re-entrancy in the following ways:

The new HTTP/1 decoders also won't call `channelRead` on you re-entrantly. So I think you're actually good here.

MrMage:
Okay, sounds like we are good here.  Last chance for any objections before this gets merged ;-)

glbrntt:
Can you merge this @MrMage? I think we'd have heard any objections by now ;)

MrMage:
@glbrntt yes, sorry!

PR #398, opened 2019-03-09
Generate ClientCallUnary test stubs.

mpetrov:
This simplifies testing for async calls, e.g. a failing request could
be simulated with something like:
```
  override func get(_ request: Echo_EchoRequest, metadata customMetadata: Metadata, completion: @escaping (Echo_EchoResponse?, CallResult) -> Void) throws -> Echo_EchoGetCall {
    completion(nil, .error(statusCode: .notFound))
    return Echo_EchoGetCallTestStub()
  }
```

### DIFF
--- a/Sources/SwiftGRPC/Runtime/ClientCallUnary.swift
+++ b/Sources/SwiftGRPC/Runtime/ClientCallUnary.swift
@@ -58,7 +58,7 @@ open class ClientCallUnaryBase<InputType: Message, OutputType: Message>: ClientC
 }
 
 /// Simple fake implementation of `ClientCallUnary`.
-open class ClientCallUnaryTestStub<InputType: Message, OutputType: Message>: ClientCallUnary {
+open class ClientCallUnaryTestStub: ClientCallUnary {
   open class var method: String { fatalError("needs to be overridden") }
 
   open func cancel() {}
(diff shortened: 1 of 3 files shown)


Sources/SwiftGRPC/Runtime/ClientCallUnary.swift
```diff
@@ -56,3 +56,10 @@ open class ClientCallUnaryBase<InputType: Message, OutputType: Message>: ClientC
     return self
   }
 }
+
+/// Simple fake implementation of `ClientCallUnary`.
+open class ClientCallUnaryTestStub<InputType: Message, OutputType: Message>: ClientCallUnary {
```
MrMage: Does this need to be generic at all?

mpetrov: Not strictly necessary at the moment, was primarily going for consistency with `ClientCallClientStreamingTestStub`.  The generics would be necessary if we want similar request queueing or response recording, thought TBH I find that approach a bit to constrained for my use-cases. In unit tests I'd generally want to test a lot of error cases and in integration tests I'd generally want the stub to have business logic (as in, I really want an in-memory fake).

I don't have strong preferences regarding the generics, let me know if you'd prefer I remove them. Happy to update the PR either way.

MrMage: Thanks for the elaboration! For now, let's remove them; I'm open to re-adding them later on if necessary.

mpetrov: Done.

PR #399, opened 2019-03-11
Mark unary async calls as @discardableResult.

mpetrov:
The result is only necessary if you need cancellation.

### DIFF
--- a/Sources/protoc-gen-swiftgrpc/Generator-Client.swift
+++ b/Sources/protoc-gen-swiftgrpc/Generator-Client.swift
@@ -175,6 +175,7 @@ extension Generator {
         }
         if asynchronousCode {
           println("/// Asynchronous. Unary.")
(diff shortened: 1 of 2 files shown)


Sources/protoc-gen-swiftgrpc/Generator-Client.swift
```diff
@@ -175,6 +175,7 @@ extension Generator {
         }
         if asynchronousCode {
           println("/// Asynchronous. Unary.")
+	  println("@discardableResult")
```
MrMage: nit: whitespace?

mpetrov: Good catch. I think I lost `set noindent` when swapping machines and the tap was only visible in the diff but not the implementation.

PTAL

PR #402, opened 2019-03-12
Allow generating test stubs separately from the client implementation

mpetrov:
This allows generating test stubs in a separate file. Generating an implementation can be disabled with 
`--swiftgrpc_opt=Imlpementations=false`.

A new `ExtraModuleImports` option allows adding import statement for the implementation. This enables generating test stubs in a separate module from implementation files. The name of the implementation module might very depending on the build system, so it can't be reliably inferred.

One use case for this would be generating different bazel build targets: a `client` target that's usable in production, and a `stub` target that's `testonly = True` (and only usable from test targets).

For example, a client could be generated in a `echo_swift_grpc_client` module follows:
```
protoc Sources/Examples/Echo/echo.proto \
--proto_path=Sources/Examples/Echo \
--plugin=.build/debug/protoc-gen-swift \
--plugin=.build/debug/protoc-gen-swiftgrpc \
--swiftgrpc_out=/tmp/echo_swift_grpc_client \
--swiftgrpc_opt=Client=true \
--swiftgrpc_opt=Server=false \
--swiftgrpc_opt=Visibility=Public
```

With a separate module for test stubs:
```
protoc Sources/Examples/Echo/echo.proto \
--proto_path=Sources/Examples/Echo \
--plugin=.build/debug/protoc-gen-swift \
--plugin=.build/debug/protoc-gen-swiftgrpc\
--swiftgrpc_out=/tmp/echo_swift_grpc_stub \
--swiftgrpc_opt=TestStubs=true \
--swiftgrpc_opt=Imlpementations=false \
--swiftgrpc_opt=Client=true \
--swiftgrpc_opt=Server=false \
--swiftgrpc_opt=ExtraModuleImports=echo_swift_grpc_client
```


MrMage:
Sorry, it's still unclear to me how the logic has changed, especially with regards to how it used to behave depending on the Client/Server/TestStubs flags and how it behaves now. Would you mind elaborating further? It also looks like your changes only affect the generated code for the client, not the server? Should this be done for both?

mpetrov:
There are effectively 3 changes here:

1. Rather than interleaving generated clients and stubs, this separates them into two distinct sections of the generated source file. The primary goal of this is to allow turning either client generation or stub generation off with a single options check rather than sprinkling around if/else checks throughout the client generator source code. This is similar to how `printNIOGRPCClient` only gets called if `options.generateNIOImplementation` is true, rather than adding `if options.generateNIOImplementation { ... }` checks all throughout the generator.
2. Allow injecting import statements into the generated source via the `ExtraModuleImports` option. This is to allow splitting generated source into various modules that can import each other. It would also help handle any weirdnesses specific to a particular build system that might come up.
3. Generate client stubs when `TestStubs=true` is on, even if client code isn't included `Client=false`. I don't think I was fully factoring server-test stubs here, so this is probably the most half-baked part of this PR. I could alternatively (a) Introduce a `Implementation=false` flag to turn off actual implementation code OR (b) Introduce `TestClientStubs` and `TestServerStubs` options to allow finer-grained control. Explicitly setting `TestStubs` could modify both for backwards compatibility.

Feedback is welcome. I'm happy to farm off smaller PRs if you feel any parts of this are less controversial, or would make it easier to review.

MrMage:
@mpetrov thank you for the elaboration! 1. and 2. sound reasonable to me. Same for 3., but I agree that we should have more granular options for this.

Sounds like we have four different binary flags here, but not all combinations would make sense, for example this one:

||Client|Server|
|-|-|-|
|Impls|✓|✗|
|TestStubs|✗|✓|

So far, I tend towards having the following options:

- `Impls` (naming to be discussed): Defaults to true. If set, the regular interfaces implementations will be generated.
- `TestStubs`: If set, the test stubs implementations will be generated. _Can not stand on its own without the interfaces generated by `Impls.`_
- `Client/Server`: Whether implementations/stubs should be generated for client, server, both, or none (with the last option not being very useful, same for `Impls = false` combined with `TestStubs = false`).

So these would be the conditions to generate the corresponding code:

||Client|Server|
|-|-|-|
|Impls|`Client` && `Impls`|`Server` && `Impls`|
|TestStubs|`Client` && `TestStubs`|`Server` && `TestStubs`|

Thoughts?

mpetrov:
Thanks for the feedback @MrMage 

I split out the client generator updates (1.) in #403 since that's the bulk of the deltas in this PR.

An `Impls` or `Implementations` flag sounds good to me. I don't have strong preferences on naming, so happy to take suggestions.

MrMage:
@rebello95 any preference on naming? Otherwise, I would suggest `Implementations`. (@mpetrov, don't forget to document the flag in README.md.)

mpetrov:
Rebased to include #403 and add new option.

README.md
```diff
@@ -103,7 +103,9 @@ separated from the output directory by a colon.
 | `Async` |  `true`/`false` | `true` | Whether to generate asynchronous code |
 | `Sync` |  `true`/`false` | `true` | Whether to generate synchronous code |
 | `TestStubs` |  `true`/`false` | `false` | Whether to generate test stub code |
+| `Implementations` |  `true`/`false` | `true` | Whether to generate imlpementation code |
```
MrMage: typo: imlpem...

Sources/protoc-gen-swiftgrpc/Generator.swift
```diff
@@ -91,6 +91,9 @@ class Generator {
     for moduleName in moduleNames {
       println("import \(moduleName)")
     }
+    for moduleName in options.extraModuleImports {
```
MrMage: use `moduleNames + options.extraModuleImports` three lines above instead?

Sources/protoc-gen-swiftgrpc/Generator.swift
```diff
@@ -106,18 +109,32 @@ class Generator {
     println()
     
     if options.generateClient {      
-      for service in file.services {
+      for service in file.services where options.generateImplementations {
```
MrMage: I'm not a fan of `where` clauses independent of the iterated variable; would you mind wrapping the loop in a plain old `if`? (also elsewhere, if applicable)

Sources/protoc-gen-swiftgrpc/options.swift
```diff
@@ -131,6 +140,13 @@ final class GeneratorOptions {
           throw GenerationError.invalidParameterValue(name: pair.key, value: pair.value)
         }
 
+      case "ExtraModuleImports":
+        if !pair.value.isEmpty {
+          extraModuleImports += [pair.value]
```
MrMage: `.append`?

README.md
MrMage: nit: move this above `TestStubs`, and elaborate that this is about generating the regular non-test interfaces for using the corresponding service? (Could also mention that this options is mostly useful for generating files that _only_ contain the test stubs, without the corresponding "regular" implementation.) Hope that makes sense, "Implementation" alone is not specific enough IMHO.

mpetrov: Updated the text, please take a look.

README.md
```diff
@@ -102,8 +102,10 @@ separated from the output directory by a colon.
 | `Client` |  `true`/`false` | `true` | Whether to generate client code |
 | `Async` |  `true`/`false` | `true` | Whether to generate asynchronous code |
 | `Sync` |  `true`/`false` | `true` | Whether to generate synchronous code |
+| `Implementations` |  `true`/`false` | `true` | Whether to generate protocols and non-test service code. Toggling this to `false` is mostly useful when combined with `TestStubs=true` to generate files containing only test stub code.  |
 | `TestStubs` |  `true`/`false` | `false` | Whether to generate test stub code |
 | `FileNaming` | `FullPath`/`PathToUnderscores`/`DropPath` | `FullPath` | How to handle the naming of generated sources |
+| `ExtraModuleImports` |  `String` | `` | Extra module to import in generated code |
```
rebello95: The description is `Extra module to import in generated code`, but I assume multiple imports can be passed? If so, can we provide details on how that works?

Sources/protoc-gen-swiftgrpc/Generator-Client.swift
```diff
@@ -58,8 +54,8 @@ extension Generator {
                                      synchronousCode: synchronousCode)
   }
 
-  private func printCGRPCClientTestStubs(asynchronousCode: Bool,
-                                         synchronousCode: Bool) {
+  internal func printCGRPCClientTestStubs(asynchronousCode: Bool,
```
rebello95: No need for the `internal` keyword

Sources/protoc-gen-swiftgrpc/Generator.swift
```diff
@@ -88,7 +88,7 @@ class Generator {
         "SwiftGRPC",
         "SwiftProtobuf"]
     }
-    for moduleName in moduleNames {
+    for moduleName in moduleNames +  options.extraModuleImports {
```
rebello95: Might be nice to sort these

Sources/protoc-gen-swiftgrpc/Generator.swift
```diff
@@ -118,6 +131,9 @@ class Generator {
       if options.generateTestStubs && options.generateNIOImplementation {
         fatalError("Generating test stubs is not yet supported for SwiftGRPC-NIO.")
       }
+      if options.generateTestStubs && !options.generateImplementations {
+        fatalError("Generating server test stubs without an implementation is not yet supported.")
```
rebello95: Do we plan to support this? If not, maybe remove "yet"

Sources/protoc-gen-swiftgrpc/options.swift
```diff
@@ -131,6 +140,13 @@ final class GeneratorOptions {
           throw GenerationError.invalidParameterValue(name: pair.key, value: pair.value)
         }
 
+      case "ExtraModuleImports":
+        if !pair.value.isEmpty {
+          extraModuleImports.append(pair.value)
```
rebello95: Do we need to split the value here to handle multiple import statements? I.e., `pair.value.components(separatedBy: ",")`?

README.md
```diff
@@ -102,8 +102,10 @@ separated from the output directory by a colon.
 | `Client` |  `true`/`false` | `true` | Whether to generate client code |
 | `Async` |  `true`/`false` | `true` | Whether to generate asynchronous code |
 | `Sync` |  `true`/`false` | `true` | Whether to generate synchronous code |
+| `Implementations` |  `true`/`false` | `true` | Whether to generate protocols and non-test service code. Toggling this to `false` is mostly useful when combined with `TestStubs=true` to generate files containing only test stub code.  |
```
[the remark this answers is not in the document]
mpetrov: Removed trailing periods.

README.md
mpetrov: Updated the text a bit. Imports could look as follows:
`--swiftgrpc_opt=ExtraModuleImports=foo,ExtraModuleImports=qux --swiftgrpc_opt=ExtraModuleImports=baz`

Sources/protoc-gen-swiftgrpc/Generator-Client.swift
mpetrov: Done.

Sources/protoc-gen-swiftgrpc/Generator.swift
mpetrov: Removed for now.

Sources/protoc-gen-swiftgrpc/options.swift
mpetrov: No, per other comment multiple imports would be done by repeating the parameter.

`--swiftgrpc_opt=ExtraModuleImports=foo,ExtraModuleImports=qux --swiftgrpc_opt=ExtraModuleImports=baz`

The options are already comma separated in `parseParameter(string:)`, and changing the semantics there could break existing callers.

mpetrov:
Took another pass, thanks for the feedback!

PR #406, opened 2019-03-19
Add TLS to the NIO implementation

glbrntt:
- Tests to follow in a separate PR
- Certificate and key from Echo have been updated since the certificate had expired
- EchoNIO uses the same certificate and key
- Also adds diagram to explain the server pipeline


Sources/SwiftGRPCNIO/GRPCServer.swift
```diff
@@ -2,8 +2,79 @@ import Foundation
 import NIO
 import NIOHTTP1
 import NIOHTTP2
+import NIOSSL
 
 /// Wrapper object to manage the lifecycle of a gRPC server.
+///
+/// The pipeline is configured in three stages detailed below. Note: handlers marked with
+/// a '*' are responsible for handling errors.
```
weissi: 💜 the ascii art

glbrntt:
I also submitted a PR against master with the updated Echo certificate and key (#407)

Sources/Examples/EchoNIO/main.swift
```diff
@@ -29,11 +31,33 @@ let messageOption = Option("message",
                            default: "Testing 1 2 3",
                            description: "message to send")
 
+func makeClientTLSConfiguration() throws -> TLSConfiguration {
+  let certificate = try NIOSSLCertificate(file: "ssl.crt", format: .pem)
+  // The certificate common name is "example.com", so skip verification.
```
MrMage: nit: mind inserting "hostname" after "skip"?

glbrntt:
> Great work, especially the flow chart! I do wonder, though, if we should provide a more convenient way of setting up "vanilla" SSL, especially on the client (see the gRPC-Core implementation, where you'd essentially just need to add `, secure: true`). Happy to have that in a later PR, though.

That's a good point. It's worth adding now since it should be fairly straightforward.

> When you add tests in a follow-up PR, consider doing it similar to the gRPC-Core implementation, where we essentially run all tests a second time with SSL enabled.

Exactly what I had in mind ;)

rebello95:
@MrMage I trust your review - I'm not very familiar with the NIO implementation 😄

PR #407, opened 2019-03-19
Update Echo certificate and key

glbrntt:
- Generated new certificate and key as certificate was expired.


MrMage:
CI failure looks like a flake; merging this.

PR #409, opened 2019-03-21
Temporarily disable `NIOClientTimeoutTests.testBidirectionalStreamingTimeoutAfterSending` on Linux for now

MrMage:
This test tends to crash randomly on Linux (making CI flaky), so it is disabled there for now.

Most of the NIO work is done on the `nio` branch now, anyway, where this test is still being run.


glbrntt:
@MrMage can you create an issue to fix this?

MrMage:
> @MrMage can you create an issue to fix this?

@glbrntt I'm not even sure if it's still occurring with the non-`nghttp2` variant of SwiftNIO at all; let's keep monitoring commits to that branch for CI failures.

glbrntt:
Okay, that's fine with me!

MrMage:
Looks like this CI failure is to temporary unavailability of Homebrew, as evidenced by the fact that one of the two CI builds succeeded. I think this can be reviewed anyway.

PR #411, opened 2019-03-21
TLS Tests

glbrntt:
- Initial tests only include variants for the "normal" functional tests
- Adds an additional target with self signed certificates; by avoiding the filesystem we also avoid having to run the Echo demo from the right directory
- Tidy up test helpers


Sources/SwiftGRPCNIOTestData/GRPCSwiftCertificate.swift
```diff
@@ -0,0 +1,191 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import Foundation
+import NIOSSL
+
+/// Wraps `NIOSSLCertificate` to provide the certificate common name and expiry date.
+public struct GRPCSwiftCertificate {
```
MrMage: nit: is there a way to make it even clearer naming-wise that these are just intended as sample certificates? E.g. `SampleCertificate`? Just an idea.

Similarly, the `TestData` module name feels a bit weird to me. How about `SwiftGRPCNIO{Ex,S}ampleData`? Still not yet sure myself how much better these would be, let's discuss this.

Sources/SwiftGRPCNIOTestData/GRPCSwiftCertificate.swift
```diff
@@ -0,0 +1,191 @@
...
+  /// Returns whether the certificate has expired.
+  public var isExpired: Bool {
+    return notAfter < Date()
+  }
+}
+
+/// Provides convenience methods to make `NIOSSLPrivateKey`s for corresponding `GRPCSwiftCertificate`s.
+public struct GRPCSwiftPrivateKey {
+  private init() {
+  }
```
MrMage: super-nit: move to the above line?

Sources/SwiftGRPCNIOTestData/GRPCSwiftCertificate.swift
```diff
@@ -0,0 +1,191 @@
...
+  public static func server() throws -> NIOSSLPrivateKey {
+    return try NIOSSLPrivateKey(buffer: Array(serverKey.utf8CString), format: .pem)
+  }
+
+  public static func client() throws -> NIOSSLPrivateKey {
+    return try NIOSSLPrivateKey(buffer: Array(clientKey.utf8CString), format: .pem)
+  }
+}
+
+// MARK:- Certificates and private keys
```
MrMage: super nit: add a space after the colon?

Sources/SwiftGRPCNIOTestData/GRPCSwiftCertificate.swift
```diff
@@ -0,0 +1,191 @@
...
+import Foundation
+import NIOSSL
+
+/// Wraps `NIOSSLCertificate` to provide the certificate common name and expiry date.
+public struct GRPCSwiftCertificate {
+  public var certificate: NIOSSLCertificate
+  public var commonName: String
+  public var notAfter: Date
+
+  public static func ca() throws -> GRPCSwiftCertificate {
```
MrMage: make these static properties that only get initialized once instead?

Tests/SwiftGRPCNIOTests/NIOBasicEchoTestCase.swift
```diff
@@ -31,41 +33,132 @@ extension Echo_EchoResponse {
...
 
-class NIOBasicEchoTestCase: XCTestCase {
+enum TransportSecurity {
+  case none
+  case anonymousClient
+  case mutualAuthentication
+}
+
+extension TransportSecurity {
+  private func caCert() throws -> GRPCSwiftCertificate {
```
MrMage: the _could_ also be static properties I guess

Tests/SwiftGRPCNIOTests/NIOBasicEchoTestCase.swift
```diff
@@ -31,41 +33,132 @@ extension Echo_EchoResponse {
...
+  }
+}
+
+class NIOEchoTestCaseBase: XCTestCase {
   var defaultTestTimeout: TimeInterval = 1.0
 
-  var serverEventLoopGroup: EventLoopGroup!
-  var server: GRPCServer!
+  var serverEventLoopGroup: EventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+  var clientEventLoopGroup: EventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
```
MrMage: `let`? (Also below)

Tests/SwiftGRPCNIOTests/NIOFunctionalTests.swift
```diff
@@ -0,0 +1,266 @@
+/*
+ * Copyright 2018, gRPC Authors All rights reserved.
```
MrMage: super-nit: 2019? (also elsewhere)

Tests/SwiftGRPCNIOTests/NIOFunctionalTests.swift
```diff
@@ -0,0 +1,266 @@
...
+ */
+import Dispatch
+import Foundation
+import NIO
+import NIOHTTP1
+import NIOHTTP2
+@testable import SwiftGRPCNIO
+import XCTest
+
+protocol NIOFunctionalTests: class {
```
MrMage: What's the advantage of putting these into a protocol?

Tests/SwiftGRPCNIOTests/EventLoopFuture+Assertions.swift
```diff
@@ -0,0 +1,42 @@
...
+import XCTest
+
+extension EventLoopFuture where Value: Equatable {
+  /// Registers a callback which asserts the value promised by this future is equal to
+  /// the expected value. Causes a test failure if the future returns an error.
+  ///
+  /// - Parameters:
+  ///   - expected: The expected value.
+  ///   - expectation: A test expectation to fulfill once the future has completed.
+  func assertEqual(_ expected: Value, fulfill expectation: XCTestExpectation? = nil, file: StaticString = #file, line: UInt = #line) {
```
MrMage: is there a reasonable case where `expectation` would be nil? In that case we could never be sure that our check ever runs, I think.

Sources/SwiftGRPCNIOTestData/GRPCSwiftCertificate.swift
glbrntt: `SampleCertificate` is better. `SwiftGRPCNIOSampleData` works for me too, I wasn't keen on the naming before!

Sources/SwiftGRPCNIOTestData/GRPCSwiftCertificate.swift
glbrntt: We'd have to `try!` if they were static properties; I _guess_ this isn't so much an issue as we know it works with the current certificates. What do you think?

Tests/SwiftGRPCNIOTests/NIOFunctionalTests.swift
glbrntt: We can use `Self` in the default implementation:

```swift
static var allTests: [(String, (Self) -> () throws -> Void)]
```

If we didn't then I think we'd have to override this in each subclass.

I was also thinking we should use `swift test --generate-linuxmain` or similar script instead but wanted to save that for a different PR.

Tests/SwiftGRPCNIOTests/EventLoopFuture+Assertions.swift
glbrntt: Good point, there probably isn't a case. If we find one we can easily add this back.

glbrntt:
> Initial impression is very good. I must admit I haven't yet through each test in `NIOFunctionalTests.swift`, though.

They're more or less the `NIOServerTests` but relying on `XCTestExpectation`s instead of call timeouts instead.

Sources/SwiftGRPCNIOTestData/GRPCSwiftCertificate.swift
MrMage: I think the `try!` should be fine here.

Tests/SwiftGRPCNIOTests/NIOFunctionalTests.swift
MrMage: Ah, thanks for the elaboration. Let's stick with the protocol in this PR and run `swift test --generate-linuxmain` in a follow-up PR (possibly adding it to the `Makefile`; we could even check that this doesn't update anything as part of CI).

PR #415, opened 2019-03-27
Disable iOS 12 cellular connectivity monitor by default

rebello95:
We've started seeing some crashes when using the `CTServiceRadioAccessTechnologyDidChange` notification within `ClientNetworkMonitor ` in production on iOS 12. This is an issue that's been reported in this radar: https://openradar.appspot.com/46873673

To provide a workaround for this problem, I'm adding the option to opt out of the new iOS 12 API and to instead use the stable pre-iOS 12 API.

**I'm planning to release a new version of SwiftGRPC with this fix after it's merged, which is what the second commit of this PR reflects.**


MrMage:
@rebello95 feel free to merge and cut a new release at your convenience.

rebello95:
Thanks @MrMage. Going to merge-commit (instead of squashing) since these 2 changes are pretty distinct.

PR #419, opened 2019-04-01
SR-10128 workaround

glbrntt:
- Workaround to avoid Swift 5 compiler crash: https://bugs.swift.org/browse/SR-10128
- Handles Swift 5 `String` API deprecation: https://github.com/apple/swift-evolution/blob/master/proposals/0241-string-index-explicit-encoding-offset.md

### DIFF
--- a/Sources/SwiftGRPCNIO/ClientCalls/BaseClientCall.swift
+++ b/Sources/SwiftGRPCNIO/ClientCalls/BaseClientCall.swift
@@ -83,6 +83,12 @@ extension BaseClientCall: ClientCall {
     return self.clientChannelHandler.statusPromise.futureResult
   }
 
+  // Workaround for: https://bugs.swift.org/browse/SR-10128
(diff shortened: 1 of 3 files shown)


glbrntt:
@timburks

Sources/SwiftGRPCNIO/ClientCalls/BaseClientCall.swift
```diff
@@ -83,6 +83,12 @@ extension BaseClientCall: ClientCall {
     return self.clientChannelHandler.statusPromise.futureResult
   }
 
+  // Workaround for: https://bugs.swift.org/browse/SR-10128
+  // Once resolved this can become a default implementation on `ClientCall`.
```
MrMage: Would you mind adding a TODO to put this back in place?

glbrntt: Done!

PR #421, opened 2019-04-02
Allow clients to make calls via factory methods

glbrntt:
- Allows clients to use factory methods to make calls
- This was possible before but was a little unwieldy
- Makes it easier for clients to call services which haven't had clients generated for them

```swift
// This could be a generated client (conforming to `GRPCServiceClient`) or an instance of 
// something conforming to GRPCClient.
let client: GRPCClient = ... 

// will be of type UnaryClientCall<Echo_EchoRequest, Echo_EchoResponse>
let unary = client.makeUnaryCall(path: "/echo.Echo/Get",
	                         request: Echo_EchoRequest(),
	                         callOptions: CallOptions(),
                                 responseType: Echo_EchoResponse.self)
```


Sources/Examples/EchoNIO/Generated/echo.grpc.swift
```diff
@@ -35,19 +35,19 @@ internal protocol Echo_EchoService_NIO {
...
   internal var defaultCallOptions: CallOptions
 
   /// Creates a client for the echo.Echo service.
   ///
   /// - Parameters:
   ///   - connection: `GRPCClientConnection` to the service host.
-  ///   - defaultCallOptions: Options to use for each service call if the user doesn't provide them. Defaults to `client.defaultCallOptions`.
-  internal init(connection: GRPCClientConnection, defaultCallOptions: CallOptions? = nil) {
+  ///   - defaultCallOptions: Options to use for each service call if the user doesn't provide them.
+  internal init(connection: GRPCClientConnection, defaultCallOptions: CallOptions = CallOptions()) {
```
MrMage: (Scratch what I said below, just noticed that `GRPCClientConnection.defaultCallOptions` got removed. Would you mind explaining the intentions behind that change?)


Why this change? I assume this would make us ignore `GRPCClientConnection.defaultCallOptions` altogether; should we remove that one, then? Not 100% sure where the "right" place for the user to set default options should be, on the connection or on the client level.

Or should we even make `defaultCallOptions` optional altogether, using `connection.defaultCallOptions` if nil? (Similar to before, but always uses the connection's latest `defaultCallOptions` if none are set on the client.)

glbrntt: Separation of concerns; it seemed strange to me having the call options on the connection as the connection just provides the transport, it doesn't know that it's being used to make gRPC calls.

MrMage: Makes sense, thanks!

PR #422, opened 2019-04-04
Allow users to fulfill status promises with `GRPCStatusTransformable` errors.

MrMage:
We then take care of converting these to `GRPCStatus`.

### DIFF
--- a/Sources/SwiftGRPCNIO/GRPCStatus.swift
+++ b/Sources/SwiftGRPCNIO/GRPCStatus.swift
@@ -25,11 +25,11 @@ public struct GRPCStatus: Error, Equatable {
 }
 
 public protocol GRPCStatusTransformable: Error {
-  func asGRPCStatus() -> GRPCStatus
+  func asGRPCStatus() -> GRPCStatus?
 }
 
 extension GRPCStatus: GRPCStatusTransformable {
-  public func asGRPCStatus() -> GRPCStatus {
+  public func asGRPCStatus() -> GRPCStatus? {
     return self
   }
 }
--- a/Sources/SwiftGRPCNIO/GRPCStatus.swift
+++ b/Sources/SwiftGRPCNIO/GRPCStatus.swift
@@ -25,11 +25,11 @@ public struct GRPCStatus: Error, Equatable {
 }
 
 public protocol GRPCStatusTransformable: Error {
-  func asGRPCStatus() -> GRPCStatus?
+  func asGRPCStatus() -> GRPCStatus
 }
 
 extension GRPCStatus: GRPCStatusTransformable {
-  public func asGRPCStatus() -> GRPCStatus? {
+  public func asGRPCStatus() -> GRPCStatus {
     return self
   }
 }
--- a/Sources/SwiftGRPCNIO/ServerCallContexts/StreamingResponseCallContext.swift
+++ b/Sources/SwiftGRPCNIO/ServerCallContexts/StreamingResponseCallContext.swift
@@ -6,8 +6,9 @@ import NIOHTTP1
 /// Abstract base class exposing a method to send multiple messages over the wire and a promise for the final RPC status.
 ///
 /// - When `statusPromise` is fulfilled, the call is closed and the provided status transmitted.
-/// - If `statusPromise` is failed and the error is of type `GRPCStatus`, that error will be returned to the client.
-/// - For other errors, `GRPCStatus.processingError` is returned to the client.
+/// - If `statusPromise` is failed and the error is of type `GRPCStatusTransformable`,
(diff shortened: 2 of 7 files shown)


MrMage:
@glbrntt, please review at your convenience.

Sources/SwiftGRPCNIO/GRPCStatus.swift
```diff
@@ -25,11 +25,11 @@ public struct GRPCStatus: Error, Equatable {
 }
 
 public protocol GRPCStatusTransformable: Error {
-  func asGRPCStatus() -> GRPCStatus
+  func asGRPCStatus() -> GRPCStatus?
```
glbrntt: Why does this need to return `GRPCStatus?` now?

Sources/SwiftGRPCNIO/ServerCallContexts/StreamingResponseCallContext.swift
```diff
@@ -35,7 +35,7 @@ open class StreamingResponseCallContextImpl<ResponseMessage: Message>: Streaming
     statusPromise.futureResult
       // Ensure that any error provided is of type `GRPCStatus`, using "internal server error" as a fallback.
```
glbrntt: Can you update this and the class level documentation as well?

Sources/SwiftGRPCNIO/ServerCallContexts/UnaryResponseCallContext.swift
```diff
@@ -55,7 +55,7 @@ open class UnaryResponseCallContextImpl<ResponseMessage: Message>: UnaryResponse
       }
       // Ensure that any error provided is of type `GRPCStatus`, using "internal server error" as a fallback.
```
glbrntt: Can you update this and the class level documentation as well?

Sources/SwiftGRPCNIO/GRPCStatus.swift
MrMage: Internally, I'm using this to wrap any `Error` in a `WrappedError` that includes the location where it was thrown. However, `WrappedError` might or might not be wrapping another `GRPCStatusTransformable` — in the latter case it would have to return `nil` in its `asGRPCStatus()` implementation.

MrMage:
@glbrntt, please take another look.

MrMage:
Also, I noticed this in ServerErrorDelegate:

```
  /// Transforms the given error into a new error.
  ///
  /// This allows framework users to transform errors which may be out of their control
  /// due to third-party libraries, for example, into more meaningful errors or
  /// `GRPCStatus` errors. Errors returned from this protocol are not passed to
  /// `observe`.
  ///
  /// - note:
  /// This defaults to returning the provided error.
  func transform(_ error: Error) -> Error
```

I'm wondering whether we should try calling `serverErrorDelegate.transform` on the errors in `*ServerCallContext` before trying to convert them to a status. However, that would change the current semantics of that method a bit (and might need a bit of refactoring), given that it would now receive errors thrown in the pipeline _and_ errors thrown by the user themselves in their request handlers. But then, the documentation already says `This allows framework users to transform errors which may be out of their control due to third-party libraries` — and that would be a strong point for calling this in `*ServerCallContext`.

What do you think?

Sources/SwiftGRPCNIO/GRPCStatus.swift
glbrntt: Thanks for the explanation!

The current naming of `GRPCStatusTransformable` is at odds with with `asGRPCStatus()` returning an optional;  i.e. something conforming to `GRPCStatusTransformable` doesn't necessarily have to be transformable to `GRPCStatus`.

Could you have `WrappedError` conform to `GRPCStatusTransformable` and handle the underlying error not conforming to `GRPCStatusTransformable` there?

glbrntt:
@MrMage that a sensible idea, I didn't realise that we didn't do this already! I think we should be observing the errors with the delegate as well.

Sources/SwiftGRPCNIO/GRPCStatus.swift
MrMage: Done; I guess I can simply return the "internal server error" myself.

MrMage:
> @MrMage that a sensible idea, I didn't realise that we didn't do this already! I think we should be observing the errors with the delegate as well.

Sounds good. Would you object if we did that in a different PR, however?

glbrntt:
> > @MrMage that a sensible idea, I didn't realise that we didn't do this already! I think we should be observing the errors with the delegate as well.
> 
> Sounds good. Would you object if we did that in a different PR, however?

Absolutely fine with me!

PR #423, opened 2019-04-08
Ask `ServerErrorDelegate` to transform status/response promise failures before returning them to the client.

MrMage:
See https://github.com/grpc/grpc-swift/pull/422#issuecomment-480174616 for discussion.


Sources/SwiftGRPCNIO/ServerCallContexts/StreamingResponseCallContext.swift
```diff
@@ -28,15 +28,22 @@ open class StreamingResponseCallContext<ResponseMessage: Message>: ServerCallCon
...
     self.channel = channel
 
     super.init(eventLoop: channel.eventLoop, request: request)
 
     statusPromise.futureResult
       // Ensure that any error provided can be transformed to `GRPCStatus`, using "internal server error" as a fallback.
-      .recover { error in
-        (error as? GRPCStatusTransformable)?.asGRPCStatus() ?? .processingError
+      .recover { [weak errorDelegate] error in
+        ((errorDelegate?.transform(error) ?? error) as? GRPCStatusTransformable)?.asGRPCStatus() ?? .processingError
```
glbrntt: I think we should be observing the error with the delegate here as well as transforming it.

Sources/SwiftGRPCNIO/ServerCallContexts/UnaryResponseCallContext.swift
```diff
@@ -55,8 +60,8 @@ open class UnaryResponseCallContextImpl<ResponseMessage: Message>: UnaryResponse
         self.responseStatus
       }
       // Ensure that any error provided can be transformed to `GRPCStatus`, using "internal server error" as a fallback.
-      .recover { error in
-        (error as? GRPCStatusTransformable)?.asGRPCStatus() ?? .processingError
+      .recover { [weak errorDelegate] error in
+        ((errorDelegate?.transform(error) ?? error) as? GRPCStatusTransformable)?.asGRPCStatus() ?? .processingError
```
glbrntt: And here as well

Sources/SwiftGRPCNIO/ServerCallContexts/StreamingResponseCallContext.swift
MrMage: So far all errors passed to `ServerErrorDelegate.observe` are those that occur somewhere inside the pipeline, while the errors thrown into this (and the other) future have usually been thrown in "user-space" instead of "gRPC-library-space". I don't object to observing these errors as well (we're getting into "middleware" territory here, but that's not necessarily a bad thing), but maybe we should have a separate `observe` method to distinguish them? (That method could also take more context, given that e.g. the request's headers are known at this point.)

What do you think?

glbrntt: Separate `observe` and `transform` methods sound good to me.

Not sure on naming; my current best suggestion is `{observe,transform}{Library,User}Error(_ error: Error)` (with any additional context for the user errors)

MrMage: I've tried incorporating this now; please let me know what you think.

MrMage:
The test failure is in the non-NIO variant and thus unrelated; merging this.

PR #425, opened 2019-04-11
gRPC interoperability tests

glbrntt:
This change introduces the [gRPC interoperability tests](https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md).

This includes:
- A CLI for running the test server and test client
- `XCTest`s to run the NIO-based server against the NIO-based client
- Some fixes to our code based on running the tests:
  - Client handling of trailers-only responses
  - Handle non-ASCII characters in status messages correctly (i.e. with percent encoding)
  - Removing an incorrect space from the "grpc-timeout" header value
  - Trailing metadata was never actually passed downstream with the status to the client
  - Client pipeline is configured correctly
- This also surfaced a bug in NIO HTTP/2 which has now been [fixed](https://github.com/apple/swift-nio-http2/pull/88), we'll need pick that up when 1.0.2 is released.

So far I've only run the tests in a few configurations:

### Swift gRPC Client

Server    | TLS | Plain text
----------|-----|-----------
Java      | ✅   | ✅
Go        | ✅   | ✅
Dart      | ⚠️ (see below) | ✅
C++       | ⚠️ (see below) | ✅

⚠️ The Dart and C++ server aren't sending close notify on TLS shutdown so the client is hanging on shutdown after each test, the actual tests seem fine. I've filed an issue [here](https://github.com/apple/swift-nio-ssl/issues/96) to add a timeout in NIO SSL, and also one in gRPC core. 

### Swift gRPC Server

Client    | TLS | Plain text
----------|-----|-----------
Java      | ✅   | ✅
Go        | ✅   | ✅
Dart      | ✅   | ✅
C++       | ✅   | ✅


glbrntt:
@MrMage please review when you get a chance 👍

glbrntt:
Closing in favour of lots of smaller PRs

PR #426, opened 2019-04-11
Run CI on both Swift 4.2 and 5, and fix Swift 5 warning

rebello95:
Fixes a warning when compiling with Swift 5:

> 'withUnsafeBytes' is deprecated: use `withUnsafeBytes<R>(…)

Also updates CI to run tests on both Swift 4.2 and Swift 5.0.

### DIFF
--- a/.travis-install.sh
+++ b/.travis-install.sh
@@ -18,10 +18,10 @@
 #
 # Install dependencies that aren't available as Ubuntu packages (or already present on macOS).
 #
-# Everything goes into $HOME/local. 
+# Everything goes into $HOME/local.
 #
-# Scripts should add 
-# - $HOME/local/bin to PATH 
+# Scripts should add
+# - $HOME/local/bin to PATH
 # - $HOME/local/lib to LD_LIBRARY_PATH
 #
 
@@ -32,9 +32,9 @@ if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
   PROTOC_URL=https://github.com/google/protobuf/releases/download/v3.5.1/protoc-3.5.1-osx-x86_64.zip
 else
   # Install swift
-  SWIFT_URL=https://swift.org/builds/swift-4.1.1-release/ubuntu1404/swift-4.1.1-RELEASE/swift-4.1.1-RELEASE-ubuntu14.04.tar.gz
+  SWIFT_URL=https://swift.org/builds/swift-5.0-release/ubuntu1404/swift-5.0-RELEASE/swift-5.0-RELEASE-ubuntu14.04.tar.gz
   echo $SWIFT_URL
-  curl -fSsL $SWIFT_URL -o swift.tar.gz 
+  curl -fSsL $SWIFT_URL -o swift.tar.gz
   tar -xzf swift.tar.gz --strip-components=2 --directory=local
 
   PROTOC_URL=https://github.com/google/protobuf/releases/download/v3.5.1/protoc-3.5.1-linux-x86_64.zip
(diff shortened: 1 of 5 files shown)


rebello95:
Updated @timburks - we'll see if CI takes the change. @MrMage would also appreciate it if you could take a look whenever you get a chance 😃

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:4.0
+// swift-tools-version:5.0
```
MrMage: I think this change means that SPM users on Swift 4 will no longer be able to use SwiftGRPC. I don't think we should cross that bridge yet.

.travis-install.sh
```diff
@@ -32,9 +32,9 @@ if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
   PROTOC_URL=https://github.com/google/protobuf/releases/download/v3.5.1/protoc-3.5.1-osx-x86_64.zip
 else
   # Install swift
-  SWIFT_URL=https://swift.org/builds/swift-4.1.1-release/ubuntu1404/swift-4.1.1-RELEASE/swift-4.1.1-RELEASE-ubuntu14.04.tar.gz
```
MrMage: I would prefer if we could have CI run with both Swift 4.2 and Swift 5, but I don't see a good way of achieving that besides a lot of duplication. What do you think?

keith: It's possible to setup a matrix of travis builds that have separate install commands, https://github.com/kylef/PathKit/blob/73f8e9dca9b7a3078cb79128217dc8f2e585a511/.travis.yml then you could pass the version to this script

Package.swift
keith: @rebello95 instead you can leave this at 4 (or maybe it has to be 4.2 for this feature) and add explicit language versions https://github.com/apple/swift-protobuf/blob/3dd275f36c2089aca0446d75c55b7c325fb36cf2/Package.swift#L34

Then with the Swift 5 tools I _believe_ it will pass `-swift-version 5`

rebello95:
Looks like a CI flake, re-running

.travis-install.sh
rebello95: Nice, looks like your suggestion is working @keith 👍

Package.swift
rebello95: Updated!

rebello95:
Thanks for the feedback @MrMage @timburks @keith! 🙏 

CI is green now, and ran on 4.2 and 5.0.

.travis.yml
```diff
@@ -15,9 +15,28 @@
 #
 # Travis CI build file for Swift gRPC.
 
-os:
-  - linux
-  - osx
+matrix:
+  include:
+    - os: linux
+      osx_image: xcode10
```
MrMage: Can `osx_image: xcode10` be omitted for the Linux runs?

.travis.yml
```diff
@@ -15,9 +15,28 @@
...
+      env:
+        - SWIFT_VERSION=4.2
+        - MAJOR_VERSION=4.2
+    - os: osx
+      osx_image: xcode10
```
MrMage: Should we use Xcode 10.2 for the Swift 4.2 runs as well?

rebello95: Yep makes sense

.travis.yml
rebello95: Done

rebello95:
Sounds good. Going to go ahead and merge.

PR #428, opened 2019-04-12
Add interoperability test provider.

glbrntt:
- Service provider for the gRPC interop testing service
- Convenience methods for generated responses
- Note: method documentation is lifted from the test spec: https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md

### DIFF
--- a/Sources/SwiftGRPCNIOInteroperabilityTests/TestServiceProvider_NIO.swift
+++ b/Sources/SwiftGRPCNIOInteroperabilityTests/TestServiceProvider_NIO.swift
@@ -23,7 +23,17 @@ import NIO
 public class TestServiceProvider_NIO: Grpc_Testing_TestServiceProvider_NIO {
   public init() { }
 
+  private static let echoMetadataNotImplemented = GRPCStatus(
+    code: .unimplemented,
+    message: "Echoing metadata is not yet supported")
+
   /// Features that this server implements.
+  ///
+  /// Some 'features' are methods, whilst others optionally modify the outcome of those methods. The
+  /// specification is not explicit about where these modifying features should be implemented (i.e.
+  /// which methods should support them) and they are not listed in the individual metdod
+  /// descriptions. As such implementation of these modifying features within each method is
+  /// determined by the features required by each test.
   public static var implementedFeatures: Set<ServerFeature> {
     return [
       .emptyCall,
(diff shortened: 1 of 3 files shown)


Sources/SwiftGRPCNIOInteroperabilityTests/TestServiceProvider_NIO.swift
```diff
@@ -0,0 +1,203 @@
...
+  ///
+  /// If the server does not support the `responseType`, then it should fail the RPC with
+  /// `INVALID_ARGUMENT`.
+  public func unaryCall(
+    request: Grpc_Testing_SimpleRequest,
+    context: StatusOnlyCallContext
+  ) -> EventLoopFuture<Grpc_Testing_SimpleResponse> {
+    let promise = context.eventLoop.makePromise(of: Grpc_Testing_SimpleResponse.self)
+    if request.shouldEchoStatus {
+      let code = StatusCode(rawValue: numericCast(request.responseStatus.code)) ?? .unknown
```
MrMage: I can't find this behavior in the test case description?

Sources/SwiftGRPCNIOInteroperabilityTests/TestServiceProvider_NIO.swift
```diff
@@ -0,0 +1,203 @@
...
+      promise.fail(GRPCStatus(code: code, message: request.responseStatus.message))
```
MrMage: How about `return context.eventLoop.makeFailedFuture` instead? (Also for all the other cases.)

Sources/SwiftGRPCNIOInteroperabilityTests/TestServiceProvider_NIO.swift
```diff
@@ -0,0 +1,203 @@
...
+      switch event {
+      case .message(let request):
+        aggregatePayloadSize += request.payload.body.count
+
+      case .end:
+        context.responsePromise.succeed(Grpc_Testing_StreamingInputCallResponse.with { response in
+          response.aggregatedPayloadSize = numericCast(aggregatePayloadSize)
+        })
+      }
+
```
MrMage: nit: extra newline

Sources/SwiftGRPCNIOInteroperabilityTests/TestServiceProvider_NIO.swift
```diff
@@ -0,0 +1,203 @@
...
+  /// of size `ResponseParameter.size` bytes, as specified by its respective `ResponseParameter`s.
+  /// After receiving half close and sending all responses, it closes with OK.
+  public func fullDuplexCall(
+    context: StreamingResponseCallContext<Grpc_Testing_StreamingOutputCallResponse>
+  ) -> EventLoopFuture<(StreamEvent<Grpc_Testing_StreamingOutputCallRequest>) -> Void> {
+    // We don't have support for this yet so just fail the call.
+    if context.request.headers.shouldEchoMetadata {
+      let status = GRPCStatus(
+        code: .unimplemented,
+        message: "'cacheableUnaryCall' requires control of the initial/trailing metadata which isn't supported"
```
MrMage: nit: message copy/pasted

Sources/SwiftGRPCNIOInteroperabilityTests/TestServiceProvider_NIO.swift
```diff
@@ -0,0 +1,203 @@
...
+      return context.eventLoop.makeFailedFuture(status)
+    }
+
+    var sendQueue = context.eventLoop.makeSucceededFuture(())
+
+    func streamHandler(_ event: StreamEvent<Grpc_Testing_StreamingOutputCallRequest>) {
+      switch event {
+      case .message(let message):
+        if message.shouldEchoStatus {
+          let code = StatusCode(rawValue: numericCast(message.responseStatus.code))
```
MrMage: (Similar to above) Where is this behavior documented?

Sources/SwiftGRPCNIOInteroperabilityTests/TestServiceProvider_NIO.swift
glbrntt: The interoperability spec isn't very good at explaining this, but as far as I can tell there are two types of "features" that servers can implement; the actual methods (unaryCall, emptyCall, etc.) and some modifying features (echo status, echo message). 

These modifying features can exist across different methods, this isn't explicitly called out anywhere. I've used their requirement in client tests to include which methods they should be implemented in on the server.

I guess it could also be implemented at a different level, i.e. as an interceptor before reaching the server handler, but I think that's a larger change.

MrMage: Fine with me as long as this is documented in a comment :-)

glbrntt: Sure -- it wasn't obvious where to document this to avoid duplicating the explanation so I added it to `implementedFeatures`, let me know if you have a better suggestion!

PR #429, opened 2019-04-12
Add gRPC interoperability tests

glbrntt:
- Adds implementations for most of the gRPC interoperability test cases. Some are missing due to features we don't (yet) support such as compression and authentication with GCE.
- Similar to the server provider, `InteroperabilityTestCases.swift` has documentation lifted from the test descriptions.
- This change also includes a change allow calls to be marked as cacheable, in which case GET is used instead of POST. This is described in the test descriptions but not in the specification.

### DIFF
--- a/Sources/SwiftGRPCNIO/ClientCalls/ClientCall.swift
+++ b/Sources/SwiftGRPCNIO/ClientCalls/ClientCall.swift
@@ -20,12 +20,7 @@ import NIOHTTP2
 import SwiftProtobuf
 
 /// Base protocol for a client call to a gRPC service.
-public protocol ClientCall {
-  /// The type of the request message for the call.
-  associatedtype RequestMessage: Message
-  /// The type of the response message for the call.
-  associatedtype ResponseMessage: Message
-
+public protocol UntypedClientCall {
   /// HTTP/2 stream that requests and responses are sent and received on.
   var subchannel: EventLoopFuture<Channel> { get }
 
@@ -52,6 +47,14 @@ public protocol ClientCall {
   /// Any unfulfilled promises will be failed with a cancelled status (excepting `status` which will be
   /// succeeded, if not already succeeded).
   func cancel()
+}
+
+/// Base protocol for a client call to a gRPC service.
+public protocol ClientCall: UntypedClientCall {
+  /// The type of the request message for the call.
+  associatedtype RequestMessage: Message
+  /// The type of the response message for the call.
+  associatedtype ResponseMessage: Message
 }
 
 /// A `ClientCall` with request streaming; i.e. client-streaming and bidirectional-streaming.
--- a/Sources/SwiftGRPCNIOInteroperabilityTests/Assertions.swift
+++ b/Sources/SwiftGRPCNIOInteroperabilityTests/Assertions.swift
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2019, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import Foundation
+import NIO
+
+/// Assertion error for interoperability testing.
+public struct AssertionError: Error {
+  let message: String
+  let file: StaticString
+  let line: UInt
+}
+
+/// Asserts that the two given values are equal.
+public func assertEqual<T: Equatable>(
+  _ value1: T,
+  _ value2: T,
+  file: StaticString = #file,
+  line: UInt = #line
+) throws {
+  guard value1 == value2 else {
+    throw AssertionError(message: "'\(value1)' is not equal to '\(value2)'", file: file, line: line)
+  }
+}
+
+/// Asserts that the value returned by the future is equal to the given value.
+///
+/// - Important: This should not be run on an event loop since this function calls `wait()` on the
+///   given future.
+public func assertEqual<T: Equatable>(
+  _ future: EventLoopFuture<T>,
+  _ value: T,
+  file: StaticString = #file,
+  line: UInt = #line
+) throws {
+  try assertEqual(try future.wait(), value, file: file, line: line)
+}
+
+/// Asserts that values retuned by each future are equal.
+///
+/// - Important: This should not be run on an event loop since this function calls `wait()` on the
+///   given future.
+public func assertEqual<T: Equatable>(
+  _ future1: EventLoopFuture<T>,
+  _ future2: EventLoopFuture<T>,
+  file: StaticString = #file,
+  line: UInt = #line
+) throws {
+  try assertEqual(try future1.wait(), try future2.wait(), file: file, line: line)
+}
--- a/Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
+++ b/Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
@@ -1,0 +1,679 @@
+/*
+ * Copyright 2019, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import Foundation
+import SwiftGRPCNIO
+import NIOHTTP1
+
+/// This test verifies that implementations support zero-size messages. Ideally, client
+/// implementations would verify that the request and response were zero bytes serialized, but
+/// this is generally prohibitive to perform, so is not required.
+///
+/// Server features:
+/// - EmptyCall
+///
+/// Procedure:
+/// 1. Client calls EmptyCall with the default Empty message
+///
+/// Client asserts:
+/// - call was successful
+/// - response is non-null
+///
+/// It may be possible to use UnaryCall instead of EmptyCall, but it is harder to ensure that the
+/// proto serialized to zero bytes.
+class EmptyUnary: InteroperabilityTest {
+  func run(using connection: GRPCClientConnection) throws {
+    let client = Grpc_Testing_TestServiceService_NIOClient(connection: connection)
+    let call = client.emptyCall(Grpc_Testing_Empty())
+
+    try assertEqual(call.response, Grpc_Testing_Empty())
+    try assertEqual(call.status.map { $0.code }, .ok)
+  }
+}
+
+/// This test verifies that gRPC requests marked as cacheable use GET verb instead of POST, and
+/// that server sets appropriate cache control headers for the response to be cached by a proxy.
+/// This test requires that the server is behind a caching proxy. Use of current timestamp in the
+/// request prevents accidental cache matches left over from previous tests.
+///
+/// Server features:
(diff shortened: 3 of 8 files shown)


MrMage:
Re-running CI...

Sources/SwiftGRPCNIO/ClientCalls/ClientCall.swift
```diff
@@ -54,6 +49,14 @@ public protocol ClientCall {
   func cancel()
 }
 
+/// Base protocol for a client call to a gRPC service.
```
MrMage: Duplication of documentation; should this be explained slightly different from `UntypedClientCall`?

Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,679 @@
...
+      request.payload = .zeros(count: 217_828)
+    }
+
+    let initialName = "x-grpc-test-echo-initial"
+    let initialValue = "test_initial_metadata_value"
+
+    let trailingName = "x-grpc-test-echo-trailing-bin"
+    let trailingValue = Data([0xab, 0xab, 0xab])
+
+    func checkMetadata(call: UntypedClientCall) throws {
```
MrMage: An alternative to declaring `UntypedClientCall` could be to declare this method as follows (also for the other method):

```
func checkMetadata<SpecificClientCall>(call: SpecificClientCall) throws where SpecificClientCall: ClientCall {
```

Also, how about moving this from `run` scope to class scope? (Also below)

Sources/SwiftGRPCNIOInteroperabilityTests/Assertions.swift
```diff
@@ -0,0 +1,62 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import Foundation
+import NIO
+
+/// Assertion error for interoperability testing.
```
MrMage: These are needed because these tests are not actual test cases and thus can't depend on `XCTest`, right?

Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,679 @@
...
+/// - EmptyCall
+///
+/// Procedure:
+/// 1. Client calls EmptyCall with the default Empty message
+///
+/// Client asserts:
+/// - call was successful
+/// - response is non-null
+///
+/// It may be possible to use UnaryCall instead of EmptyCall, but it is harder to ensure that the
```
MrMage: Sorry, could you elaborate?

Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,679 @@
...
+/// proto serialized to zero bytes.
+class EmptyUnary: InteroperabilityTest {
+  func run(using connection: GRPCClientConnection) throws {
+    let client = Grpc_Testing_TestServiceService_NIOClient(connection: connection)
+    let call = client.emptyCall(Grpc_Testing_Empty())
+
+    try assertEqual(call.response, Grpc_Testing_Empty())
```
MrMage: Seeing this without a `.wait()` after `call.response` confused me at first. Personally, I'd prefer to make the `.wait()` explicit here (and elsewhere, of course), but not feeling strongly given that this is "only" test code.

Thinking about this more, I agree that cluttering the tests with `try ....wait()` statements would probably be annoying.

Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,679 @@
...
+    let request = Grpc_Testing_SimpleRequest.withPayload(of: .bytes(of: &timestamp))
+
+    var headers = HTTPHeaders()
+    headers.add(name: "x-user-ip", value: "1.2.3.4")
+    let callOptions = CallOptions(customMetadata: headers, cacheable: true)
+
+    let call1 = client.cacheableUnaryCall(request, callOptions: callOptions)
+    let call2 = client.cacheableUnaryCall(request, callOptions: callOptions)
+
+    try assertEqual(call1.response.map { $0.payload }, call2.response.map { $0.payload })
```
MrMage: Should we also compare this to the payload in `request`? Also, given that we are not behind a caching proxy, I guess these two are only identical because the corresponding requests had identical payloads as well?

Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,679 @@
...
+///        payload:{
+///            body: 271828 bytes of zeros
+///        }
+///    }
+///    ```
+///
+/// Client asserts:
+/// - call was successful
+/// - response payload body is 314159 bytes in size
+/// - clients are free to assert that the response payload body contents are zero and comparing
```
MrMage: Should we assert that the response payload is zero, for good measure?

Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,679 @@
...
+      call.sendMessage(.withPayload(of: .zeros(count: 27_182)))
+    }.flatMap {
+      call.sendMessage(.withPayload(of: .zeros(count: 8)))
+    }.flatMap {
+      call.sendMessage(.withPayload(of: .zeros(count: 1_828)))
+    }.flatMap {
+      call.sendMessage(.withPayload(of: .zeros(count: 45_904)))
+    }.whenSuccess {
+      call.sendEnd(promise: nil)
+    }
```
MrMage: Should we wait on this _before_ waiting on the response and status below?

Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,679 @@
...
+    let request = Grpc_Testing_StreamingOutputCallRequest.with { request in
+      request.responseParameters = responseSizes.map { .size($0) }
+    }
+
+    var actualResponseSizes: [Int] = []
+    let call = client.streamingOutputCall(request) { response in
+      actualResponseSizes.append(response.payload.body.count)
+    }
+
+    // Wait for the status first to ensure we've finished collecting responses.
```
MrMage: good comment!

Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,679 @@
...
+    let responseReceived = DispatchSemaphore(value: 0)
+
+    var actualResponseSizes: [Int] = []
+    let call = client.fullDuplexCall { response in
+      actualResponseSizes.append(response.payload.body.count)
+      responseReceived.signal()
+    }
+
+    try zip(requestSizes, responseSizes).map { requestSize, responseSize in
+      return Grpc_Testing_StreamingOutputCallRequest.with { request in
```
MrMage: super-nit: remove `return`?

Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,679 @@
...
+        request.payload = .zeros(count: requestSize)
+        request.responseParameters = [.size(responseSize)]
+      }
+    }.forEach { request in
+      call.sendMessage(request, promise: nil)
+      try assertEqual(responseReceived.wait(timeout: .now() + .seconds(1)), .success)
```
MrMage: nice, never though of using `signal.wait()` like this :-)

Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,679 @@
...
+/// Client asserts:
+/// - Call completed with status DEADLINE_EXCEEDED.
+class TimeoutOnSleepingServer: InteroperabilityTest {
+  func run(using connection: GRPCClientConnection) throws {
+    let client = Grpc_Testing_TestServiceService_NIOClient(connection: connection)
+
+    let callOptions = CallOptions(timeout: try .milliseconds(1))
+    let call = client.fullDuplexCall(callOptions: callOptions) { _ in }
+
+    try assertEqual(call.status.map { $0.code }, .deadlineExceeded)
```
MrMage: Nice, I had completely forgotten that we actually enforce deadlines. Do you remember if we enforce them in the client, on the server, or both?

Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
glbrntt: That's the same thought process I went through! I think some renaming could help:

```swift
try assertEqual(future: call.response, Grpc_Testing_Empty())
// or...
try assertEqual(future: call.response, value: Grpc_Testing_Empty())
```

For two futures:
```swift
try assertEqual(future: call1.response, future: call2.response)
// or maybe
try assertFuturesEqual(call1.response, call2.response)
```

It might be worth explicitly calling these out as synchronous as well (`assertEqualSync` or similar).

What do you think?

MrMage:
How about `assertFutureEqual` or `waitAndAssertEqual`? `assertEqualSync` would also be fine with me, but I would avoid the `future:` in all three cases.

Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
glbrntt: I didn't write this; it's lifted from the official test descriptions. I assume this is just an implementation option. I can remove it though so we're using empty call anyway.

Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
glbrntt: > Should we also compare this to the payload in `request`?

The payload is ignored by the server: 
https://github.com/grpc/grpc-swift/blob/283754e80bb30a630e5e06508af654bc1d68fd7d/Sources/SwiftGRPCNIOInteroperabilityTests/TestServiceProvider_NIO.swift#L89

I'll add a comment to elaborate.

> Also, given that we are not behind a caching proxy, I guess these two are only identical because the corresponding requests had identical payloads as well?

This is just the client end of the test, we won't run this against a NIO server (for now). Some server implementations support this test by faking a caching proxy; this is probably something we can do if we allow additional handlers to be added to the server.

Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
glbrntt: Might as well; I think this is recommended in a few places to so I'll update them all.

Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
glbrntt: Good idea!

Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
glbrntt: Only on the client at the moment. We do send the timeout to the server though, so other implementations which support it can use it.

Sources/SwiftGRPCNIO/ClientCalls/ClientCall.swift
glbrntt: Reverted this change in favour of using generics in the test.

glbrntt:
> Looks good, mostly a few clarification questions, some of which might warrant slightly more documentation. Also, would you mind documenting which tests we do _not_ yet support?

Sure!

> Lastly, I remember mention of the C# client somewhere — is that actually going to be used somewhere?

Hmm, I think it was the C++ client. Once everything is merged we should set up the tests to run as part of the CI against a different implementation, it doesn't really matter which, it could even be more than one. The Java version might be a good choice since it's easy to build but we can cross that bridge when we get to it.

PR #432, opened 2019-04-16
Document missing tests and server features

glbrntt:


glbrntt:
@MrMage this was missing from PR #429

Sources/SwiftGRPCNIOInteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -17,6 +17,18 @@ import Foundation
...
+/// - server_compressed_unary
+/// - client_compressed_streaming
+/// - server_compressed_streaming
+/// - compute_engine_creds
+/// - jwt_token_creds
+/// - oauth2_auth_token
+/// - per_rpc_creds
+/// - google_default_credentials
+/// - compute_engine_channel_credentials
+
```
MrMage: Super-nit: Should there be a `///` on this line, too?

glbrntt: No, it's is distinct from `EmptyUnary`. When we have a CLI, we'll have something to map test names (as listed in the spec, e.g. `empty_unary`) to test instances; that will be a better place for the list of missing tests to live.

MrMage: All right, thanks!

PR #435, opened 2019-04-17
Decrease a few timeouts in the hopes of fixing a test flake.

MrMage:
@glbrntt for review.

If this doesn't fix the flake, we could consider temporarily disabling the non-NIO tests altogether on the `nio` branch.


MrMage:
Closing and re-opening a few times to trigger several CI runs.

MrMage:
Looks like this fixed the flake; merging.

PR #442, opened 2019-04-18
Various fixes found during interop testing

glbrntt:
Interoperability testing found a number of small bugs:
- We didn't handle "Trailers-Only" responses
- We didn't handle non-ASCII status messages correctly
- The timeout passed to the server incorrectly had an space in it
- Trailing metadata was never passed down the pipeline to the caller


Sources/SwiftGRPCNIO/GRPCHeaderName.swift
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import Foundation
+
+public struct GRPCHeaderName {
```
kevints: Since this is never instantiated consider an empty enum instead

MrMage: For what it's worth I'm always torn between an empty enum and a struct with a private initializer.

glbrntt: Other than not being able to instantiate the `enum` is there any difference?

glbrntt:
> LGTM. Should we have extra tests for e.g. passing the trailing metadata and/or non-ASCII status messages, or are these covered sufficiently by the interop tests?

I think we're covered sufficiently by the interop tests.

Sources/SwiftGRPCNIO/GRPCHeaderName.swift
MrMage: I don't think so. Maybe @kevints knows more?

glbrntt:
@MrMage can this be merged?

MrMage:
Yes, sorry. I had just been waiting for a comment from @kevints, but I think this is good to go as-is.

PR #443, opened 2019-04-18
Verify TLS/ALPN were successful before returning a connection

glbrntt:
- Verify that a handshake was successful and that ALPN returned a valid protocol before returning a client connection


Sources/SwiftGRPCNIO/GRPCClientConnection.swift
```diff
@@ -49,8 +50,23 @@ open class GRPCClientConnection {
...
+      // Check the handshake succeeded and a valid protocol was negotiated via ALPN.
+      let tlsVerified: EventLoopFuture<Void>
+
+      if case .none = tlsMode {
+        tlsVerified = channel.eventLoop.makeSucceededFuture(())
+      } else {
+        tlsVerified = channel.pipeline.context(handlerType: GRPCTLSVerificationHandler.self).map {
+          $0.handler as! GRPCTLSVerificationHandler
+        }.flatMap {
+          $0.verification
```
MrMage: What is this for? Seems like the result gets dropped?

Sources/SwiftGRPCNIO/GRPCClientConnection.swift
```diff
@@ -49,8 +50,23 @@ open class GRPCClientConnection {
...
-      channel.pipeline.context(handlerType: HTTP2StreamMultiplexer.self).map { context in
-        context.handler as! HTTP2StreamMultiplexer
+      // Check the handshake succeeded and a valid protocol was negotiated via ALPN.
+      let tlsVerified: EventLoopFuture<Void>
+
+      if case .none = tlsMode {
+        tlsVerified = channel.eventLoop.makeSucceededFuture(())
+      } else {
+        tlsVerified = channel.pipeline.context(handlerType: GRPCTLSVerificationHandler.self).map {
+          $0.handler as! GRPCTLSVerificationHandler
```
MrMage: I've seen this pattern in the previous "hang fix" PR as well. Should this functionality maybe be provided by NIO itself? /cc @weissi

Sources/SwiftGRPCNIO/GRPCTLSVerificationHandler.swift
```diff
@@ -0,0 +1,71 @@
...
+/// and that the negotiated application protocol is valid.
+///
+/// The handler holds a promise which is succeeded on successful verification of the negotiated
+/// application protocol and failed if any error is received by this handler or an invalid
+/// application protocol was negotiated.
+///
+/// Users of this handler should rely on the `verification` future held by this instance.
+///
+/// On fulfillment of the promise this handler is removed from the channel pipeline.
+public class GRPCTLSVerificationHandler: ChannelInboundHandler, RemovableChannelHandler {
```
MrMage: I assume this class is required _in addition to_ the `applicationProtocols: GRPCApplicationProtocolIdentifier.allCases.map { $0.rawValue }` above?

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCClientCodec.swift
```diff
@@ -15,6 +15,7 @@
  */
 import Foundation
 import NIO
+import NIOTLS
```
MrMage: Can this be removed?

Tests/SwiftGRPCNIOTests/EventLoopFuture+Assertions.swift
```diff
@@ -40,3 +40,29 @@ extension EventLoopFuture where Value: Equatable {
...
+  ///   - handler: A block to run additional verification on the error. Defaults to no-op.
+  func assertError(fulfill expectation: XCTestExpectation, file: StaticString = #file, line: UInt = #line, handler: @escaping (Error) -> Void = { _ in }) {
+    self.whenComplete { result in
+      defer {
+        expectation.fulfill()
+      }
+
+      switch result {
+      case .success:
+        XCTFail("Unexpectedly received \(Value.self))", file: file, line: line)
```
MrMage: nit: add `", expected an error"`?

Sources/SwiftGRPCNIO/GRPCClientConnection.swift
glbrntt: There's already a PR open for this but it will require NIO 2.1: https://github.com/apple/swift-nio/pull/974

Sources/SwiftGRPCNIO/GRPCClientConnection.swift
glbrntt: We drop the value, but it _could_ contain an `error` which will be propagated back to the caller instead of the connection.

Sources/SwiftGRPCNIO/GRPCTLSVerificationHandler.swift
glbrntt: Yes, that's correct. We also check that the handshake was successful here as well as the negotiated protocol.

Sources/SwiftGRPCNIO/GRPCClientConnection.swift
MrMage: But `tlsVerified` is declared as an `ELF<Void>`, not `ELF<Connection>` or something like that? LMK if I'm missing something.

Sources/SwiftGRPCNIO/GRPCClientConnection.swift
MrMage: Awesome; could you make a note to yourself to update our code once that becomes available?

Sources/SwiftGRPCNIO/GRPCClientConnection.swift
glbrntt: We `flatMap` off it a few lines below. If `tlsVerified` contains an error then we won't execute that chain of `map`s, but we will return an `ELF<GRPCClientConnection>` containing the error from `tlsVerified`.

Sources/SwiftGRPCNIO/HTTP1ToRawGRPCClientCodec.swift
glbrntt: Not sure what this is doing here in the first place!

Sources/SwiftGRPCNIO/GRPCClientConnection.swift
glbrntt: Added TODOs.

Sources/SwiftGRPCNIO/GRPCClientConnection.swift
MrMage: Stupid me; I mistook `flatMap` and `map`. Would you mind adding a short comment to this line nevertheless?

glbrntt: Easily done! Comment added.

PR #445, opened 2019-04-23
Diagrams for client connection and client call pipelines.

glbrntt:
@MrMage I thought this might be a useful addition given the previous questions about the TLS verification.


Sources/SwiftGRPCNIO/ClientCalls/BaseClientCall.swift
```diff
@@ -22,6 +22,30 @@ import SwiftProtobuf
 /// This class provides much of the boilerplate for the four types of gRPC call objects returned to framework
 /// users.
 ///
+/// Each call will be configured on a multiplexed channel on the given connection. The mutliplexed
```
MrMage: typo: mutliplexed

glbrntt: Thanks, fixed.

PR #358, opened 2019-01-16
Feature/support tvos

bimawa:
Related with grpc/grpc#14173

Pls, check these changes.

PR #446, opened 2019-04-26
Support tvOS

enricmacias:
Adds the changes from https://github.com/grpc/grpc-swift/pull/358


MrMage:
Thanks! Can this be applied to the Carthage project as well? And why did the BoringSSL dependency need to be deleted?

enricmacias:
The official BoringSSL doesn't support tvOS. But the one included in gRPC-Core library (BoringSSL-GRPC) it does. It is working fine with that library.

enricmacias:
About Carthage. I tried adding `config.build_settings["TVOS_DEPLOYMENT_TARGET"] = "10.0"` to `fix-project-settings.rb`, but for some reason it is not working as expected. 

We may need to check which SDK target is using and set the appropriate deployment target for each one. I need some time/help for that, I don't really know how this ruby files are working at all.

MrMage:
> About Carthage. I tried adding `config.build_settings["TVOS_DEPLOYMENT_TARGET"] = "10.0"` to `fix-project-settings.rb`, but for some reason it is not working as expected.
> 
> We may need to check which SDK target is using and set the appropriate deployment target for each one. I need some time/help for that, I don't really know how this ruby files are working at all.

What is the error you are getting with that when running `make project-carthage `?

MrMage:
@enricmacias thank you! Two more things:

1. We now have a merge conflict because the Carthage project on the master branch has changed in the meantime. I would suggest merging `master` into your branch, then running `make project-carthage` a second time.
2. Could you please verify that the Carthage project builds on your machine? You should be able to open it in Xcode and then click "Build".

enricmacias:
@MrMage I merged the master into my branch and run `make project-carthage` .

About the Carthage project building on my machine. I've already created a new tvOS project and installed SwiftGRPC through Carthage. The project compiled and run without a problem. Is that what you wanted for me to check?

MrMage:
> About the Carthage project building on my machine. I've already created a new tvOS project and installed SwiftGRPC through Carthage. The project compiled and run without a problem. Is that what you wanted for me to check?

Yes, that's perfect. Thank you!

I will close and re-open the PR to re-run CI.

PR #447, opened 2019-04-26
Fix libtool warning when compiling with bazel

keith:
If you build CgRPC with bazel, as done [here](https://github.com/bazelbuild/rules_swift/blob/master/third_party/com_github_grpc_grpc_swift/BUILD.overlay), and you end up creating a library with libtool that contains it, you get this warning for multiple files:

```
libtool: warning same member name (byte_buffer.o) in output file used for input files: .../libCgRPC.a(byte_buffer.o) and: .../libCgRPC.a(byte_buffer.o) due to use of basename, truncation and blank padding
```

This is because bazel translates both `server.c` and `server.cc` to `server.o`. This fixes these warnings by appending `_shim` to the shim files.


keith:
FYI using `ar` instead of `libtool` doesn't produce this warning, by does produce the exact same binary, here's the relevant code from the open source dump of libtool:

```c++
	/* if this is libtool warn about duplicate member names */
	if(cmd_flags.ranlib == FALSE)
	    warn_duplicate_member_names();
```

PR #448, opened 2019-04-27
Add sample Google API call using SwiftGRPCNIO.

timburks:
This adds a small sample app that calls a production Google API using SwiftGRPCNIO.

The most interesting/important thing that it demonstrates is how to set up the TLS connection.

More elaborate examples will follow here and elsewhere.


timburks:
@glbrntt I can't add you as a reviewer but please take a look, especially at the TLS setup.

Examples/Google/NaturalLanguageNIO/Sources/main.swift
```diff
@@ -0,0 +1,122 @@
...
+import NIOHTTP1
+import NIOSSL
+
+func makeClientTLS() throws -> GRPCClientConnection.TLSMode {
+  let configuration = TLSConfiguration.forClient(applicationProtocols: ["h2"])
+  let context = try NIOSSLContext(configuration: configuration)
+  return .custom(context)
+}
+
+let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 2)
```
glbrntt: Why 2?

Examples/Google/NaturalLanguageNIO/Sources/main.swift
```diff
@@ -0,0 +1,122 @@
...
+
+/// Create a client and wait for it to initialize. Returns nil if initialisation fails.
+func makeServiceClient(address: String, port: Int) -> Google_Cloud_Language_V1_LanguageServiceService_NIOClient? {
+  do {
+    return try GRPCClientConnection.start(host: address,
+                                port: port,
+                                eventLoopGroup: eventLoopGroup,
+                                tls:makeClientTLS())
```
MrMage: nit: missing a space after the colon

Examples/Google/NaturalLanguageNIO/Sources/main.swift
```diff
@@ -0,0 +1,122 @@
...
+  let context = try NIOSSLContext(configuration: configuration)
+  return .custom(context)
+}
+
+let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 2)
+
+/// Create a client and wait for it to initialize. Returns nil if initialisation fails.
+func makeServiceClient(address: String, port: Int) -> Google_Cloud_Language_V1_LanguageServiceService_NIOClient? {
+  do {
+    return try GRPCClientConnection.start(host: address,
```
MrMage: @glbrntt what do you think of having another `GRPCClientConnection.start` method that performs the `makeClientTLS` setup for the user? Or make `makeClientTLS` part of the main `SwiftGRPCNIO` library?

Or maybe we should even have a class that takes care of spinning up/shutting down the event loop group for the user, if they don't want to deal with it?

Examples/Google/NaturalLanguageNIO/Sources/main.swift
```diff
@@ -0,0 +1,122 @@
...
+      .wait()
+  } catch {
+    print("Unable to create a client: \(error)")
+    return nil
+  }
+}
+
+let scopes = ["https://www.googleapis.com/auth/cloud-language"]
+
+var authToken : String?
```
MrMage: nit: please remove the space in front of the colon

Examples/Google/NaturalLanguageNIO/Sources/main.swift
```diff
@@ -0,0 +1,122 @@
...
+func makeClientTLS() throws -> GRPCClientConnection.TLSMode {
+  let configuration = TLSConfiguration.forClient(applicationProtocols: ["h2"])
+  let context = try NIOSSLContext(configuration: configuration)
+  return .custom(context)
+}
+
+let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 2)
+
+/// Create a client and wait for it to initialize. Returns nil if initialisation fails.
+func makeServiceClient(address: String, port: Int) -> Google_Cloud_Language_V1_LanguageServiceService_NIOClient? {
```
MrMage: How about returning an `EventLoopFuture<Google_Cloud_Language_V1_LanguageServiceService_NIOClient>` here and `wait()`ing and handling errors in the "main" code?

Examples/Google/NaturalLanguageNIO/Sources/main.swift
```diff
@@ -0,0 +1,122 @@
...
+  address: "language.googleapis.com",
+  port: 443)
+  else {
+    print("ERROR: Unable to create service client.")
+    exit(-1)
+}
+
+
+let headers = HTTPHeaders([("authorization", "Bearer " + authToken)])
+let callOptions = CallOptions(customMetadata:headers,
```
MrMage: nit: please add spaces after the colons

Examples/Google/NaturalLanguageNIO/Sources/main.swift
glbrntt: We definitely need _something_ because the TLS setup is unwieldy. However, I'm not sure what that looks like at the moment. 

At a minimum I think we should replace the `Mode` enums with `NIOSSLContext?` and have the `start` methods be the lowest level available. We can have more "friendly" methods or classes as you suggest to make this easier for those that need it.

Examples/Google/NaturalLanguageNIO/Sources/main.swift
timburks: Arbitrary - I've reduced it to 1 and everything seems fine. Thanks!

Examples/Google/NaturalLanguageNIO/Package.swift
```diff
@@ -0,0 +1,37 @@
+// swift-tools-version:4.0
```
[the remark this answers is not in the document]
timburks: fixed, thanks

Examples/Google/NaturalLanguageNIO/Sources/main.swift
```diff
@@ -0,0 +1,122 @@
...
+  return .custom(context)
+}
+
+let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 2)
+
+/// Create a client and wait for it to initialize. Returns nil if initialisation fails.
+func makeServiceClient(address: String, port: Int) -> Google_Cloud_Language_V1_LanguageServiceService_NIOClient? {
+  do {
+    return try GRPCClientConnection.start(host: address,
+                                port: port,
```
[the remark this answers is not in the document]
timburks: fixed, thanks

Examples/Google/NaturalLanguageNIO/Sources/main.swift
timburks: fixed, thanks

Examples/Google/NaturalLanguageNIO/Sources/main.swift
timburks: fixed, thanks

Examples/Google/NaturalLanguageNIO/Sources/main.swift
timburks: fixed, thanks

Examples/Google/NaturalLanguageNIO/Sources/main.swift
```diff
@@ -0,0 +1,138 @@
...
+    return nil
+  }
+}
+
+enum AuthError: Error {
+  case providerFailed
+  case noProvider
+}
+
+func getAuthToken(_ eventLoopGroup: MultiThreadedEventLoopGroup,
```
MrMage: How about passing in an `eventLoop` instead?

Examples/Google/NaturalLanguageNIO/Sources/main.swift
```diff
@@ -0,0 +1,138 @@
...
+        } else {
+          if let error = error {
+            promise.fail(error)
+          } else {
+            promise.fail(AuthError.providerFailed)
+          }
+        }
+      }
+    }
+    catch {
```
MrMage: super-nit: move the `catch` to the preceding line?
Also, consider failing the promise with the caught error instead?

Examples/Google/NaturalLanguageNIO/Sources/main.swift
```diff
@@ -0,0 +1,138 @@
...
+      promise.fail(AuthError.providerFailed)
+    }
+  } else {
+    promise.fail(AuthError.noProvider)
+  }
+  return promise.futureResult
+}
+
+let scopes = ["https://www.googleapis.com/auth/cloud-language"]
+guard let authToken = try? getAuthToken(eventLoopGroup, scopes: scopes).wait()
```
MrMage: How about using `try` and wrapping the entire `main` code in a try/catch block that prints the error and calls `exit(1)` in the catch block?

Examples/Google/NaturalLanguageNIO/Sources/main.swift
```diff
@@ -0,0 +1,138 @@
...
+}
+
+annotateText.response.whenFailure { error in
+  print("annotateText failed with error: \(error)")
+}
+
+print("running")
+// wait() on the status to stop the program from exiting.
+do {
+  print("waiting")
```
MrMage: This will be printed right after `running` in every case — remove it?

Examples/Google/NaturalLanguageNIO/Sources/main.swift
```diff
@@ -0,0 +1,138 @@
...
+
+enum AuthError: Error {
+  case providerFailed
+  case noProvider
+}
+
+func getAuthToken(_ eventLoopGroup: MultiThreadedEventLoopGroup,
+                  scopes: [String]) -> EventLoopFuture<String> {
+  let promise = eventLoopGroup.next().makePromise(of: String.self)
+  if let provider = DefaultTokenProvider(scopes: scopes) {
```
MrMage: you _could_ use `guard let provider = ... else { promise.fail; return promise.futureResult }` instead (optional)

Examples/Google/NaturalLanguageNIO/Sources/main.swift
timburks: fixed (both)

Examples/Google/NaturalLanguageNIO/Sources/main.swift
timburks: done

Examples/Google/NaturalLanguageNIO/Sources/main.swift
timburks: done

Examples/Google/NaturalLanguageNIO/Sources/main.swift
```diff
@@ -0,0 +1,122 @@
...
+    return nil
+  }
+}
+
+let scopes = ["https://www.googleapis.com/auth/cloud-language"]
+
+var authToken : String?
+
+if let provider = DefaultTokenProvider(scopes: scopes) {
+  let sem = DispatchSemaphore(value: 0)
```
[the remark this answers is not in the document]
timburks: done -- at some point we'll also rewrite the token providers to return futures.

Examples/Google/NaturalLanguageNIO/Sources/main.swift
timburks: done

Examples/Google/NaturalLanguageNIO/Sources/main.swift
timburks: done

Examples/Google/NaturalLanguageNIO/Sources/main.swift
timburks: done

Examples/Google/NaturalLanguageNIO/Sources/main.swift
timburks: Unwieldy and mysterious: I needed [this stackoverflow post](https://stackoverflow.com/questions/53466997/using-swiftnio-and-swiftniohttp2-as-an-http2-client) to figure it out.

For Google API clients, we might eventually wrap all this in service constructors that hide all the networking details: TLS, address, port, etc.

glbrntt: Agh, that's pretty bad that you had to resort to stack overflow! Created #452 to address the TLS setup.

Examples/Google/NaturalLanguageNIO/Sources/main.swift
```diff
@@ -0,0 +1,134 @@
...
+                       address: String,
+                       port: Int)
+  -> EventLoopFuture<Google_Cloud_Language_V1_LanguageServiceService_NIOClient> {
+    let promise = eventLoopGroup.next().makePromise(of: Google_Cloud_Language_V1_LanguageServiceService_NIOClient.self)
+    do {
+      try GRPCClientConnection.start(host: address,
+                                     port: port,
+                                     eventLoopGroup: eventLoopGroup,
+                                     tls: makeClientTLS())
+        .map { client in
```
glbrntt: Nit: `cascade` the result of the future into the promise:

```swift
.map { client in
    Google_Cloud_Language_V1_LanguageServiceService_NIOClient(connection: client)
}.cascade(to: promise)
```

timburks: Updated. Thanks for the guidance!

timburks:
@MrMage Any remaining concerns? If not, feel free to merge this (I think it's ready).

Issue #449, opened 2019-04-30
Carthage not working for version 0.8.2 and beyond


WilliamIzzo83:
### New Issue Checklist

- [ ] Updated grpc-swift to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/grpc/grpc-swift/blob/master/CONTRIBUTING.md)
- [x] I searched for [existing GitHub issues](https://github.com/grpc/grpc-swift/issues)(especially, there are many [resolved issues related to SSL](https://github.com/grpc/grpc-swift/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+ssl))

### Issue Description

Hi!

I get the following linker error when trying to get grpc-swift through carthage

```
ld: warning: directory not found for option '-L/usr/local/Cellar/nghttp2/1.37.0/lib'
ld: library not found for -lnghttp2
```

It seems that nghttp2 library can't be found. I tried to get it with mac ports first and the error is still there. Then by looking at the flag ```-L/usr/local/Cellar/nghttp2/1.37.0/lib``` I thought that maybe with brew I could do it, but I had no luck because brew will install 1.38.0

Any ideas?

MrMage:
I think Homebrew has an option to install older versions; did you try that?

WilliamIzzo83:
Yes I was looking into it and it seems quite [complex](https://zoltanaltfatter.com/2017/09/07/Install-a-specific-version-of-formula-with-homebrew/), not to mention that a simple brew update will break it again. Also in my case I had to install specifically brew: usually I can do everything with macports.

I was thinking if there could be other options, because this setup seems really brittle. Of course I'm willing to contribute.

MrMage:
That makes sense. Thinking about this more, I think the dependency on `nghttp2` is only an artifact of the Carthage project being generated by SwiftPM — our CocoaPods variant does not require `nghttp2`, after all. I think it should be possible to modify https://github.com/grpc/grpc-swift/blob/master/patch-carthage-project.rb to remove this dependency from the resulting project.

WilliamIzzo83:
All right, I can give this a shot

MrMage:
Fixed by #450.

PR #450, opened 2019-04-30
Avoids linking nghttp2 in Carthage project

WilliamIzzo83:
This should address #449

### DIFF
--- a/SwiftGRPC-Carthage.xcodeproj/GeneratedModuleMap/CNIONghttp2/module.modulemap
+++ b/SwiftGRPC-Carthage.xcodeproj/GeneratedModuleMap/CNIONghttp2/module.modulemap
@@ -1,4 +1,4 @@
 module CNIONghttp2 {
-    umbrella "/Users/William/Documents/work/clients/perceptolab/safehome/grpc-swift/.build/checkouts/swift-nio-http2/Sources/CNIONghttp2/include"
+    umbrella "/Users/William/Documents/work/clients/perceptolab/safehome/grpc-swift/.build/checkouts/swift-nio-http2.git--6139390414245659487/Sources/CNIONghttp2/include"
     export *
 }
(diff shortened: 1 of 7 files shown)


patch-carthage-project.rb
```diff
@@ -8,9 +8,23 @@
 targets_to_remove.each do |target|
   target.remove_from_project
 end
+
+# 2) Prevent linking of nghttp2 library
+carthage_nghttp2_unlink_targets = [ "BoringSSL", "CgRPC", "SwiftGRPC" ]
+targets_to_unlink = project.targets.select { |target| carthage_nghttp2_unlink_targets.include?(target.name) }
+ldflags_to_remove = [ "-L/usr/local/Cellar/nghttp2/1.38.0/lib", "-lnghttp2" ]
```
MrMage: Would it be possible to make this more generic, removing any `LDFLAG` containing `nghttp2`?

SwiftGRPC-Carthage.xcodeproj/GeneratedModuleMap/CNIONghttp2/module.modulemap
```diff
@@ -1,4 +1,4 @@
 module CNIONghttp2 {
-    umbrella "/Users/mrebello/Development/grpc-swift/.build/checkouts/swift-nio-http2.git-1684232237084789971/Sources/CNIONghttp2/include"
```
MrMage: I'm a bit surprised by all the `.git-1684232237084789971` suffixes disappearing. I'd expect the values to have changed, but that there would still be a suffix present. Which Swift version did you use to generate the new project? (I think we should stick with Swift 4.2 for now, to ensure that we don't break the build for Swift 4.2 users by generating the project with Swift 5.) See https://medium.com/xcblog/switching-swift-versions-inside-xcode-using-toolchains-755b28831c43 on selecting the Swift 4.2 toolchain.

patch-carthage-project.rb
WilliamIzzo83: I can scan every target in the scheme if that is all right with you

SwiftGRPC-Carthage.xcodeproj/GeneratedModuleMap/CNIONghttp2/module.modulemap
WilliamIzzo83: Of course: I'm using swift 5. I'll fix that

patch-carthage-project.rb
MrMage: Sounds good, but I was more thinking of removing any flag that contains `nghttp2`, even if there is not an exact match.

WilliamIzzo83: Done

SwiftGRPC-Carthage.xcodeproj/GeneratedModuleMap/CNIONghttp2/module.modulemap
WilliamIzzo83: Done

PR #451, opened 2019-04-30
Fail calls on closed connections.

glbrntt:
Motivation:

Calls on a connection which was already closed or closed during the call
would never fulfill the response/status/metadata futures.

Modifications:

- Observe failures on newly creating stream channels and errors received
  from NIOHTTP2
- Add tests


glbrntt:
@MrMage -- since we now _always_ create a promise for `send{Head,Message,End}` should we remove the variants which accept a `promise` and mark the returned futures as `@discardableResult`?

MrMage:
> Calls on a connection which was already closed or closed during the call
> would never fulfill the response/status/metadata futures.

Yikes, thanks for spotting!

Sources/SwiftGRPCNIO/GRPCStatus.swift
```diff
@@ -33,3 +35,27 @@ extension GRPCStatus: GRPCStatusTransformable {
...
+}
+
+extension ChannelError: GRPCStatusTransformable {
+  public func asGRPCStatus() -> GRPCStatus {
+    switch self {
+    case .inputClosed, .outputClosed, .ioOnClosedChannel:
+      return .init(code: .unavailable, message: "The connection is closed")
+
+    default:
+      return .processingError
```
MrMage: What are the other cases here?

Tests/SwiftGRPCNIOTests/NIOClientClosedChannelTests.swift
```diff
@@ -0,0 +1,167 @@
...
+
+    self.client.connection.close().map {
+      self.client.get(Echo_EchoRequest(text: "foo"))
+    }.whenSuccess { get in
+      get.initialMetadata.assertError(fulfill: initialMetadataExpectation)
+      get.response.assertError(fulfill: responseExpectation)
+      get.status.map { $0.code }.assertEqual(.unavailable, fulfill: statusExpectation)
+    }
+
+    self.wait(for: [initialMetadataExpectation, responseExpectation, statusExpectation],
```
MrMage: Does this work on Linux now (probably since Swift 5)?

Sources/SwiftGRPCNIO/GRPCStatus.swift
glbrntt: Source: https://github.com/apple/swift-nio/blob/768c3ab328a79cb5a4a4f91094050206aaf11328/Sources/NIO/Channel.swift#L294

```swift
/// An error that can occur on `Channel` operations.
public enum ChannelError: Error {
    /// Tried to connect on a `Channel` that is already connecting.
    case connectPending

    /// Connect operation timed out
    case connectTimeout(TimeAmount)

    /// Unsupported operation triggered on a `Channel`. For example `connect` on a `ServerSocketChannel`.
    case operationUnsupported

    /// An I/O operation (e.g. read/write/flush) called on a channel that is already closed.
    case ioOnClosedChannel

    /// Close was called on a channel that is already closed.
    case alreadyClosed

    /// Output-side of the channel is closed.
    case outputClosed

    /// Input-side of the channel is closed.
    case inputClosed

    /// A read operation reached end-of-file. This usually means the remote peer closed the socket but it's still
    /// open locally.
    case eof

    /// A `DatagramChannel` `write` was made with a buffer that is larger than the MTU for the connection, and so the
    /// datagram was not written. Either shorten the datagram or manually fragment, and then try again.
    case writeMessageTooLarge

    /// A `DatagramChannel` `write` was made with an address that was not reachable and so could not be delivered.
    case writeHostUnreachable

    /// The local address of the `Channel` could not be determined.
    case unknownLocalAddress

    /// The address family of the multicast group was not valid for this `Channel`.
    case badMulticastGroupAddressFamily

    /// The address family of the provided multicast group join is not valid for this `Channel`.
    case badInterfaceAddressFamily

    /// An attempt was made to join a multicast group that does not correspond to a multicast
    /// address.
    case illegalMulticastAddress(SocketAddress)

    /// Multicast is not supported on Interface
    case multicastNotSupported(NIONetworkInterface)

    /// An operation that was inappropriate given the current `Channel` state was attempted.
    case inappropriateOperationForState

    /// An attempt was made to remove a ChannelHandler that is not removable.
    case unremovableHandler
}
```

Tests/SwiftGRPCNIOTests/NIOClientClosedChannelTests.swift
glbrntt: Yes, since Swift 5.

MrMage:
Thanks for the replies!

PR #453, opened 2019-05-01
Batch requests in testUnaryLotsOfRequests

glbrntt:
@MrMage this seems like a cleaner workaround and something we may want to do in the library at some point.

I did a few more runs (release mode):

requests|  time
--------|------
  1,000 | 0.548
  2,000 | 1.062
  4,000 | 2.148
  8,000 | 4.308
 16,000 | 8.485

### DIFF
--- a/Tests/SwiftGRPCNIOTests/NIOFunctionalTests.swift
+++ b/Tests/SwiftGRPCNIOTests/NIOFunctionalTests.swift
@@ -77,33 +77,44 @@ extension NIOFunctionalTestsInsecureTransport {
   }
 
   func testUnaryLotsOfRequests() throws {
-    self.defaultTestTimeout = 60.0
-
     // Sending that many requests at once can sometimes trip things up, it seems.
     let clockStart = clock()
     let numberOfRequests = 2_000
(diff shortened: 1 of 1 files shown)


Tests/SwiftGRPCNIOTests/NIOFunctionalTests.swift
```diff
@@ -77,33 +77,44 @@ extension NIOFunctionalTestsInsecureTransport {
...
+
+      for i in lowerBound..<upperBound {
+        let request = Echo_EchoRequest(text: "foo \(i)")
+        let response = Echo_EchoResponse(text: "Swift echo get: foo \(i)")
+
+        let get = client.get(request)
+        get.response.assertEqual(response, fulfill: responseExpectation)
+        get.status.map { $0.code }.assertEqual(.ok, fulfill: statusExpectation)
+
+        if (i+1) % 1_000 == 0 {
```
MrMage: How about checking lowerBound in the outer loop instead?

glbrntt: Done!

PR #456, opened 2019-05-03
Add a client error delegate

glbrntt:


Sources/Examples/EchoNIO/Generated/echo.grpc.swift
```diff
@@ -39,15 +39,18 @@ internal final class Echo_EchoService_NIOClient: GRPCServiceClient, Echo_EchoSer
...
+  internal var errorDelegate: ClientErrorDelegate?
 
   /// Creates a client for the echo.Echo service.
   ///
   /// - Parameters:
   ///   - connection: `GRPCClientConnection` to the service host.
   ///   - defaultCallOptions: Options to use for each service call if the user doesn't provide them.
-  internal init(connection: GRPCClientConnection, defaultCallOptions: CallOptions = CallOptions()) {
+  ///   - errorDelegate: Delegate to use when receiving errors.
+  internal init(connection: GRPCClientConnection, defaultCallOptions: CallOptions = CallOptions(), errorDelegate: ClientErrorDelegate? = nil) {
```
MrMage: Should we use a "logging client error delegate" by default, similar to the server case?

Sources/SwiftGRPCNIO/ClientErrorDelegate.swift
```diff
@@ -0,0 +1,34 @@
...
+  /// Called when the client catches an error.
+  ///
+  /// - Parameters:
+  ///   - error: The error which was caught.
+  ///   - file: The file where the error was raised.
+  ///   - line: The line within the file where the error was raised.
+  func didCatchError(_ error: Error, file: StaticString, line: Int)
+}
+
+public class LoggingClientErrorDelegate: ClientErrorDelegate {
```
MrMage: Should we offer a shared instance of this? (Related to the weak reference comment above.)

Sources/Examples/EchoNIO/Generated/echo.grpc.swift
```diff
@@ -39,15 +39,18 @@ internal final class Echo_EchoService_NIOClient: GRPCServiceClient, Echo_EchoSer
   internal let connection: GRPCClientConnection
   internal var serviceName: String { return "echo.Echo" }
   internal var defaultCallOptions: CallOptions
+  internal var errorDelegate: ClientErrorDelegate?
```
MrMage: Should this be a weak reference?

Sources/SwiftGRPCNIO/GRPCClientChannelHandler.swift
```diff
@@ -104,6 +107,7 @@ internal class GRPCClientChannelHandler<RequestMessage: Message, ResponseMessage
   ///
   /// - Parameter error: the error to observe.
   internal func observeError(_ error: GRPCError) {
+    self.errorDelegate?.didCatchError(error.error, file: error.file, line: error.line)
```
MrMage: When exactly is this method called, could you refresh my memory?
(Might also need to update the documentation for this method now.)

Sources/SwiftGRPCNIOInteroperabilityTests/protos/Generated/test.grpc.swift
```diff
@@ -43,15 +43,18 @@ public final class Grpc_Testing_TestServiceService_NIOClient: GRPCServiceClient,
   public let connection: GRPCClientConnection
   public var serviceName: String { return "grpc.testing.TestService" }
   public var defaultCallOptions: CallOptions
+  public var errorDelegate: ClientErrorDelegate?
```
MrMage: weak? (in the codegen)

Sources/SwiftGRPCNIOInteroperabilityTests/protos/Generated/test.grpc.swift
```diff
@@ -43,15 +43,18 @@ public final class Grpc_Testing_TestServiceService_NIOClient: GRPCServiceClient,
...
 
   /// Creates a client for the grpc.testing.TestService service.
   ///
   /// - Parameters:
   ///   - connection: `GRPCClientConnection` to the service host.
   ///   - defaultCallOptions: Options to use for each service call if the user doesn't provide them.
-  public init(connection: GRPCClientConnection, defaultCallOptions: CallOptions = CallOptions()) {
+  ///   - errorDelegate: Delegate to use when receiving errors.
+  public init(connection: GRPCClientConnection, defaultCallOptions: CallOptions = CallOptions(), errorDelegate: ClientErrorDelegate? = nil) {
```
MrMage: default to a logging delegate? (in the codegen)

Sources/Examples/EchoNIO/Generated/echo.grpc.swift
glbrntt: I don't _think_ so. We need someone to hold a strong reference to the delegate and the client seems like the obvious owner here since it's the "useful" object the client interacts with. If the user needs to reference the client in the delegate then they should use a weak reference in their delegate.

Sources/SwiftGRPCNIO/GRPCClientChannelHandler.swift
glbrntt: At the moment it's called by `errorCaught(context:error:)` in the final handler in the pipeline. It's also called when the call times out and when the call is cancelled.

I have a follow up PR to change those last two usages into user events (i.e. handled with `userInboundEventTriggered(context:event:)`). This would stop `observeError` from being called outside of this class.

Sources/SwiftGRPCNIO/ClientErrorDelegate.swift
glbrntt: Sure, we can do that.

Sources/Examples/EchoNIO/Generated/echo.grpc.swift
glbrntt: I wasn't sure it made as much sense on the client as it does on the server but I don't feel strongly about it either way.

glbrntt:
> Looks good; I have no idea whether the error delegate gets called in enough places, though ;-)

That's actually a good callout, thinking about it a little more we don't have a handler _after_ the multiplexer on the "main" channel to catch errors. I think that's worth adding.

Sources/Examples/EchoNIO/Generated/echo.grpc.swift
MrMage: Makes sense. How about an error delegate that logs in debug builds (to ease development and help us help users encountering issues), but not release builds? I think the NIO team has a function we could steal that does something like that.

Sources/SwiftGRPCNIO/ClientErrorDelegate.swift
MrMage: Let's do that and make `init` private :-)

Sources/Examples/EchoNIO/Generated/echo.grpc.swift
MrMage: "If the user needs to reference the client in the delegate then they should use a weak reference in their delegate." feels a bit like an inversion of responsibilities. But AFAICT the error delegate should exist before the client, anyway, so it can't really know about the client. So I guess this would be a non-issue in practice. It would also be a non-issue when using shared instances of the logger (or retaining it somewhere in "user land"), too.

Still not 100% sure about the right approach here, though.

Sources/SwiftGRPCNIO/GRPCClientChannelHandler.swift
MrMage: Makes sense. Are there any other _additional_ places where we could/should call this? (Not sure myself; if you say that this is the only place, I trust you.)

MrMage:
> That's actually a good callout, thinking about it a little more we don't have a handler _after_ the multiplexer on the "main" channel to catch errors. I think that's worth adding.

👍

Sources/Examples/EchoNIO/Generated/echo.grpc.swift
glbrntt: It would be possible to reference it in a delegate afterwards if the delegate stored a `Client?`, but I think that's a bit of an edge case and I'm not sure what we can do to prevent it (apart from documenting it). I'm also not 100% sure here but this the best candidate at the moment.

Sources/Examples/EchoNIO/Generated/echo.grpc.swift
glbrntt: Sounds good!

Sources/SwiftGRPCNIO/GRPCClientChannelHandler.swift
glbrntt: It should be called in handlers at the end of the NIO pipeline. In this case we have the HTTP/2 multiplexed channels and pipelines (i.e. in the snippet above) and behind the multiplexer for errors that aren't at the stream level.

glbrntt:
> LGTM. It looks like the changes to the codegen have disappeared; are these no longer necessary?

Correct; I bumped the error delegate up a level to the `GRPCClientConnection`.

PR #458, opened 2019-05-07
Support tvos for cocoapods

bimawa:
@MrMage I updated pod spec.


MrMage:
@bimawa we already have #446 for this, which also includes Carthage support.

bimawa:
ups thanks!

PR #459, opened 2019-05-08
Use user events for timing out and cancelling client calls

glbrntt:
- Use user events for timing out and cancelling calls
- Cancelling a call is now no longer treated as an error
- `GRPCError.error` was renamed to `GRPCError.wrappedError`

### DIFF
--- a/Sources/SwiftGRPCNIO/ClientCalls/BaseClientCall.swift
+++ b/Sources/SwiftGRPCNIO/ClientCalls/BaseClientCall.swift
@@ -122,7 +122,7 @@ extension BaseClientCall: ClientCall {
   public func cancel() {
     self.connection.channel.eventLoop.execute {
       self.subchannel.whenSuccess { channel in
-        channel.close(mode: .all, promise: nil)
+        channel.pipeline.fireUserInboundEventTriggered(GRPCClientUserEvent.cancelled)
       }
     }
   }
@@ -233,7 +233,9 @@ extension BaseClientCall {
     if timeout == .infinite { return }
 
     self.connection.channel.eventLoop.scheduleTask(in: timeout.asNIOTimeAmount) { [weak self] in
-      self?.clientChannelHandler.observeError(GRPCError.client(.deadlineExceeded(timeout)))
+      self?.subchannel.whenSuccess { stream in
+        stream.pipeline.fireUserInboundEventTriggered(GRPCClientUserEvent.timedOut(timeout))
+      }
     }
   }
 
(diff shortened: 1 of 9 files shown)


Sources/SwiftGRPCNIO/ClientCalls/BaseClientCall.swift
```diff
@@ -233,7 +233,9 @@ extension BaseClientCall {
     if timeout == .infinite { return }
 
     self.connection.channel.eventLoop.scheduleTask(in: timeout.asNIOTimeAmount) { [weak self] in
```
MrMage: Would you mind adding timeout support to the server, too? Not in this PR, of course. When adding that, we should also make sure that _all_ promises (status and response) are being fulfilled (probably with an error) when the timeout occurs.

It might also be nice to set a maximum timeout on the server object, to prevent malicious clients from opening many requests with very long timeouts.

Sources/SwiftGRPCNIO/GRPCClientChannelHandler.swift
```diff
@@ -192,8 +208,6 @@ extension GRPCClientChannelHandler: ChannelOutboundHandler {
 extension GRPCClientChannelHandler {
   /// Closes the HTTP/2 stream. Inbound and outbound state are set to ignore.
   public func close(context: ChannelHandlerContext, mode: CloseMode, promise: EventLoopPromise<Void>?) {
-    self.observeError(GRPCError.client(.cancelledByClient))
```
MrMage: Should we send a user event here instead?

Sources/SwiftGRPCNIO/GRPCClientChannelHandler.swift
```diff
@@ -206,3 +220,12 @@ extension GRPCClientChannelHandler {
     context.close(mode: .all, promise: nil)
   }
 }
+
+/// Client user evenets.
+///
+/// - cancelled: The call has been cancelled.
```
MrMage: nit: document each enum case in-line? (Not sure if it makes a difference)

Sources/SwiftGRPCNIO/GRPCClientChannelHandler.swift
glbrntt: No; the previous mechanism for cancelling was sending a `close` which we'd pick up here which was a bit of a hack.

MrMage: Duh, right. I missed that above. Sorry!

Sources/SwiftGRPCNIO/GRPCClientChannelHandler.swift
glbrntt: Sure, Xcode defaults to putting them above like this (with "⌥⌘/") but inline is a little easier to read.

Sources/SwiftGRPCNIO/ClientCalls/BaseClientCall.swift
glbrntt: We should definitely add this. I think we have higher priorities at the moment but I created #462 so we don't lose it!

PR #463, opened 2019-05-14
Getting initialMetadata at the beginning of streaming

Dimat112:
I needed to get data from initialMetada at the beginning of streaming(bidirectional and server) in order to see progress, but I couldn't do it.


MrMage:
This causes the server-streaming/bidi-streaming call completion handler to get called _twice_ with completely different semantics. I don't think it would be a good idea to mix the semantics of the two completion handles ("initial metadata received/status received") like that. If anything, we might want to pass a separate "initial metadata handler" to these calls, but that might require far more changes in several parts of the library.

MrMage:
Closing this for now, as it can not be merged in its current stage and we are planning to deprecate the "legacy" gRPC-Core-based implementation soon, anyway.

PR #465, opened 2019-05-20
Remove CgRPC implementation

glbrntt:
- Removes the CgRPC implementation from the `nio` branch
- Also removes related assets, scripts etc.
- CI updated for just NIO


weissi:
Screenshot 2019-05-20 at 1 56 27 pm
 🙌

MrMage:
🎉  (I'd like to wait for at least one more review before merging.)

Did you test making the Carthage project? Did that work?

glbrntt:
It generates without an issue and Xcode opens it fine. 

I didn't test integrating it with a project, but given we've never had Carthage support for the NIO implementation I thought we should address it when we sort out Cocoapods.

MrMage:
Sounds good to me. It would be nice if you could check that the Carthage project builds (just open it in Xcode and build, nothing more), but no obligation.

rebello95:
Since we're removing CocoaPods in this PR, I assume we plan to add it back to the `nio` branch in a separate change prior to making NIO the default on `master`?

MrMage:
Wow, I had completely missed that change. It makes sense, though, given that we don't yet have a Podspec for the NIO implementation. We'll definitely need to add one for NIO before making this the new master, though.

glbrntt:
> Sounds good to me. It would be nice if you could check that the Carthage project builds (just open it in Xcode and build, nothing more), but no obligation.

It builds, _however_ it is only building SwiftProtobuf; I removed Carthage related things in favour of addressing it properly with CocoaPods.

glbrntt:
> Since we're removing CocoaPods in this PR, I assume we plan to add it back to the `nio` branch in a separate change prior to making NIO the default on `master`?

Absolutely!

PR #467, opened 2019-05-23
Add a performance test runner and benchmarks

glbrntt:
Initial results (perf gains to follow!):

bench | time (s)
-|-
10k unary requests of size 8 | 2.394
10k unary requests of size 65536 | 8.463
10k bidirectional messages of size 8 | 0.579
10k bidirectional messages of size 65536 | 5.377
100 connections created | 0.024

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -28,7 +28,7 @@ var packageDependencies: [Package.Dependency] = [
   // Main SwiftNIO package
   .package(url: "https://github.com/apple/swift-nio.git", from: "2.0.0"),
   // HTTP2 via SwiftNIO
-  .package(url: "https://github.com/apple/swift-nio-http2.git", from: "1.0.0"),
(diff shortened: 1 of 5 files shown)


MrMage:
Looks good! Are the results with or without SSL?

Also, I would suggest slightly increasing request count for "10k bidirectional messages of size 8" to make the numbers larger (and thus more meaningful). Same for "100 connections created".

glbrntt:
These are without SSL. 

I wouldn't read _too_ much into these results since they're not run in a quiet environment (I'll try to do that at some point). I should've also added that these times are the mean of 10 runs.

How about 20k for the bidi size 8? 

Creating significantly more than 100 connections at a time leads to some of them failing, 100 seems to be reliable.

MrMage:
20k sounds good 👍 Will merge once that's changed.

PR #468, opened 2019-05-23
Update codegen to include only NIO implementation.

glbrntt:
- Also regenerated clients/servers to reflect this change (i.e. the removal of `_NIO` from type names).
- Also fixed some Swift 5 warnings re: deprecated APIs in Sources/protoc-gen-swiftgrpc/io.swift


rebello95:
I'll let @MrMage review in detail - but can we also make sure that the options supported in the README are updated accordingly?

glbrntt:
Sure - the whole README needs a rewrite but I’ll update the codegen part of it as part of this PR.

Examples/Google/NaturalLanguageNIO/RUNME
```diff
@@ -23,7 +23,6 @@ protoc \
 	-I../common/include \
 	--swift_out=googleapis \
 	--swiftgrpc_out=googleapis \
```
MrMage: Need to remove the `\` as well?

Sources/protoc-gen-swiftgrpc/Generator.swift
```diff
@@ -106,35 +99,14 @@ class Generator {
...
-        for service in file.services {
-          self.service = service
-          printClient(asynchronousCode: options.generateAsynchronous,
-                      synchronousCode: options.generateSynchronous)
-        }
-      }
-
-      if options.generateTestStubs {
-        if options.generateNIOImplementation {
-          fatalError("Generating test stubs is not yet supported for SwiftGRPC-NIO.")
```
MrMage: I think this is still true. (At least we will probably need to generate client test stubs at some point.)

Sources/protoc-gen-swiftgrpc/Generator.swift
```diff
@@ -106,35 +99,14 @@ class Generator {
...
-        }
+      for service in file.services {
+        self.service = service
+        printClient()
       }
     }
     println()
     
     if options.generateServer {
-      if options.generateTestStubs && options.generateNIOImplementation {
```
MrMage: Any idea why this was in here? I did think we have server test stubs already. Or do we have them, but don't need to generate any code for them?

Sources/protoc-gen-swiftgrpc/Generator.swift
glbrntt: It is true, but I didn't find any NIO test stubs so I thought it was better to remove the option than have an option which didn't do anything!

Sources/protoc-gen-swiftgrpc/Generator.swift
glbrntt: No idea, git tells me you wrote this line though, so you may be better positioned to answer that question! 🙃

As far as I can tell there are no test stubs for NIO.

Examples/Google/NaturalLanguageNIO/RUNME
glbrntt: good catch, fixed!

Sources/protoc-gen-swiftgrpc/Generator.swift
MrMage: Right, I think the test stub so far simply has you use `UnaryResponseCallContextTestStub` and the like directly; no need to generate code for these.

PR #470, opened 2019-05-23
Don't remove status and message from trailing metadata

glbrntt:
Removing the status and message triggers an unnecessary copy of the trailers. This is obviously worse the more trailers there are!


glbrntt:
This is in the client code so it shouldn't affect gRPC-Web (unless I'm mistaken)?

I initially removed it because we pulled it out to make it more accessible via `GRPCStatus`.

MrMage:
Thanks for the elaboration!

PR #471, opened 2019-05-24
Refactor client channel handlers into request and response handlers

glbrntt:
Refactors `GRPCClientChannelHandler` into separate handlers for requests and responses, with unary and streaming variants of each. Each call type (unary, server streaming, client streaming, bidi streaming) is then just a case of configuring the pipeline with the correct handlers.

This allows us to separate out some of the logic into something which is a little easier to reason about.


Sources/SwiftGRPCNIO/GRPCClientRequestChannelHandler.swift
```diff
@@ -0,0 +1,67 @@
...
+
+  /// The request head to send.
+  internal let requestHead: HTTPRequestHead
+
+  init(requestHead: HTTPRequestHead) {
+    self.requestHead = requestHead
+  }
+
+  public func channelActive(context: ChannelHandlerContext) {
+    // If we don't provide a method here the default implementation on protocol (i.e. no-op) will be
```
MrMage: Good documentation! Stuff like this has bitten me in the past.

Sources/SwiftGRPCNIO/GRPCClientResponseChannelHandler.swift
```diff
@@ -14,34 +14,27 @@
...
-/// response (if applicable) are failed with first error received. The status promise is __succeeded__
-/// with the error as the result of `GRPCStatusTransformable.asGRPCStatus()`, if available.
-/// The stream is also closed and any inbound or outbound messages are ignored.
-internal class GRPCClientChannelHandler<RequestMessage: Message, ResponseMessage: Message> {
+/// This includes hold promises for the initial metadata and status of the gRPC call. This handler
+/// is also responsible for error handling, via an error delegate and by appropriately failing the
+/// aforementioned promises.
+internal class GRPCClientResponseChannelHandler<ResponseMessage: Message>: ChannelInboundHandler {
+  public typealias InboundIn = GRPCClientResponsePart<ResponseMessage>
+  public typealias OutboundIn = Any
```
MrMage: Could you explain this?

glbrntt: It wasn't meant to be there, thanks for catching!

PR #473, opened 2019-05-25
Update README.md for examples to point to correct folders

asif13:
This commit updates Readme.md to point correct directories.


MrMage:
Thank you!

PR #475, opened 2019-05-29
Enable TCP_NODELAY for client connections.

glbrntt:


glbrntt:
Latency. This disables Nagle's algorithm which buffers data before writing to the socket to avoid sending small TCP packets. Since we're careful about when we choose to flush it can actually get in the way.

I was running the performance tests on Linux and the large unary requests took _significantly_ longer (~40-50x) without `TCP_NODELAY`.

PR #476, opened 2019-05-31
Reduce allocations with NIOAny

glbrntt:
This change reduces a bunch of allocations:
- We should be careful when wrapping types in `NIOAny`, anything larger than 24 bytes will lead to an additional heap allocation (see: https://github.com/apple/swift-nio-http2/commit/4097c3a807a83661f0add383edef29b426e666cb)
- This provides a `Box` class which just holds a value and makes the `GRPCClientRequestPart` and `GRPCClientResponsePart` `enum`s small enough to avoid extra heap allocations.
- This also turns `GRPCStatus` into a class (the reason is documented in code).


Sources/GRPC/Box.swift
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import Foundation
+
+/// Provides a "box" to put a value in.
+///
+/// Allows large values to be passed around without being copied.
+public final class Box<T> {
```
weissi: do you really want this to be public?

glbrntt: Not really, but it's used by a few public enums which are used in public handlers. I'll prefix it with an underscore.

Sources/GRPC/GRPCClientCodec.swift
```diff
@@ -21,14 +21,14 @@ import SwiftProtobuf
 /// Outgoing gRPC package with a fixed message type.
 public enum GRPCClientRequestPart<RequestMessage: Message> {
   case head(HTTPRequestHead)
-  case message(RequestMessage)
+  case message(Box<RequestMessage>)
```
MrMage: I'd add a comment here (and below) explaining the purpose of this.

Tests/GRPCTests/GRPCTypeSizeTests.swift
```diff
@@ -0,0 +1,56 @@
...
+import XCTest
+
+/// These test check the size of types which get wrapped in `NIOAny`. If the size of the type is
+/// greater than 24 bytes (the size of the value buffer in an existential container) then it will
+/// incur an additional heap allocation.
+///
+/// This commit message explains the problem and one way to mitigate the issue:
+/// https://github.com/apple/swift-nio-http2/commit/4097c3a807a83661f0add383edef29b426e666cb
+///
+/// Session 416 of WWD6 2016 also provides a good explanation of existential containers.
```
MrMage: typo: WWDC?

Tests/GRPCTests/GRPCTypeSizeTests.swift
```diff
@@ -0,0 +1,56 @@
...
+class GRPCTypeSizeTests: XCTestCase {
+  let existentialContainerBufferSize = 24
+
+  func checkSize<T>(of: T.Type) {
```
MrMage: nit: `add file:line:` parameters? (could be bike-shedding)

Tests/GRPCTests/GRPCTypeSizeTests.swift
```diff
@@ -0,0 +1,56 @@
...
+
+/// These test check the size of types which get wrapped in `NIOAny`. If the size of the type is
+/// greater than 24 bytes (the size of the value buffer in an existential container) then it will
+/// incur an additional heap allocation.
+///
+/// This commit message explains the problem and one way to mitigate the issue:
+/// https://github.com/apple/swift-nio-http2/commit/4097c3a807a83661f0add383edef29b426e666cb
+///
+/// Session 416 of WWD6 2016 also provides a good explanation of existential containers.
+class GRPCTypeSizeTests: XCTestCase {
```
MrMage: Add this file to LinuxMain.swift?

PR #477, opened 2019-06-05
Refactor the client connection for easier creation.

glbrntt:
Refactor the client connection for easier creation. The creation is now a little bit more flexible (since you can now use e.g. a Unix domain socket).

### DIFF
--- a/Sources/Examples/Echo/main.swift
+++ b/Sources/Examples/Echo/main.swift
@@ -32,11 +32,11 @@ let messageOption = Option("message",
                            default: "Testing 1 2 3",
                            description: "message to send")
 
-func makeClientTLS(enabled: Bool) throws -> GRPCClientConnection.TLSMode {
+func makeClientSSLContext(enabled: Bool) throws -> NIOSSLContext? {
   guard enabled else {
-    return .none
+    return nil
   }
-  return .custom(try NIOSSLContext(configuration: try makeClientTLSConfiguration()))
+  return try NIOSSLContext(configuration: try makeClientTLSConfiguration())
 }
 
 func makeServerTLS(enabled: Bool) throws -> GRPCServer.TLSMode {
@@ -74,8 +74,21 @@ func makeServerTLSConfiguration() throws -> TLSConfiguration {
 /// Create en `EchoClient` and wait for it to initialize. Returns nil if initialisation fails.
 func makeEchoClient(address: String, port: Int, ssl: Bool) -> Echo_EchoServiceClient? {
   let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+
   do {
-    return try GRPCClientConnection.start(host: address, port: port, eventLoopGroup: eventLoopGroup, tls: try makeClientTLS(enabled: ssl))
+    let tlsConfiguration: GRPCClientConnection.TLSConfiguration?
+    if let sslContext = try makeClientSSLContext(enabled: ssl) {
+      tlsConfiguration = .init(sslContext: sslContext)
+    } else {
+      tlsConfiguration = nil
+    }
+
+    let configuration = GRPCClientConnection.Configuration(
+      target: .hostAndPort(address, port),
+      eventLoopGroup: eventLoopGroup,
+      tlsConfiguration: tlsConfiguration)
+
+    return try GRPCClientConnection.start(configuration)
       .map { Echo_EchoServiceClient(connection: $0) }
       .wait()
   } catch {
--- a/Sources/Examples/Echo/main.swift
+++ b/Sources/Examples/Echo/main.swift
@@ -32,21 +32,18 @@ let messageOption = Option("message",
                            default: "Testing 1 2 3",
                            description: "message to send")
 
-func makeClientSSLContext(enabled: Bool) throws -> NIOSSLContext? {
-  guard enabled else {
-    return nil
-  }
-  return try NIOSSLContext(configuration: try makeClientTLSConfiguration())
+func makeClientSSLContext() throws -> NIOSSLContext {
+  return try NIOSSLContext(configuration: makeClientTLSConfiguration())
 }
 
 func makeServerTLS(enabled: Bool) throws -> GRPCServer.TLSMode {
   guard enabled else {
     return .none
   }
-  return .custom(try NIOSSLContext(configuration: try makeServerTLSConfiguration()))
+  return .custom(try NIOSSLContext(configuration: makeServerTLSConfiguration()))
 }
 
-func makeClientTLSConfiguration() throws -> TLSConfiguration {
+func makeClientTLSConfiguration() -> TLSConfiguration {
   let caCert = SampleCertificate.ca
   let clientCert = SampleCertificate.client
   precondition(!caCert.isExpired && !clientCert.isExpired,
@@ -59,7 +56,7 @@ func makeClientTLSConfiguration() throws -> TLSConfiguration {
                     applicationProtocols: GRPCApplicationProtocolIdentifier.allCases.map { $0.rawValue })
 }
 
-func makeServerTLSConfiguration() throws -> TLSConfiguration {
+func makeServerTLSConfiguration() -> TLSConfiguration {
   let caCert = SampleCertificate.ca
   let serverCert = SampleCertificate.server
   precondition(!caCert.isExpired && !serverCert.isExpired,
@@ -77,8 +74,8 @@ func makeEchoClient(address: String, port: Int, ssl: Bool) -> Echo_EchoServiceCl
 
   do {
     let tlsConfiguration: GRPCClientConnection.TLSConfiguration?
-    if let sslContext = try makeClientSSLContext(enabled: ssl) {
-      tlsConfiguration = .init(sslContext: sslContext)
+    if ssl {
+      tlsConfiguration = .init(sslContext: try makeClientSSLContext())
     } else {
       tlsConfiguration = nil
     }
--- a/Sources/GRPC/GRPCClientConnection.swift
+++ b/Sources/GRPC/GRPCClientConnection.swift
@@ -54,6 +54,228 @@ import NIOTLS
 ///
 /// See `BaseClientCall` for a description of the remainder of the client pipeline.
 open class GRPCClientConnection {
+  /// Makes and configures a `ClientBootstrap` using the provided configuration.
+  ///
+  /// Enables `SO_REUSEADDR` and `TCP_NODELAY` and configures the `channelInitializer` to use the
+  /// handlers detailed in the documentation for `GRPCClientConnection`.
+  ///
+  /// - Parameter configuration: The configuration to prepare the bootstrap with.
+  public class func makeBootstrap(configuration: Configuration) -> ClientBootstrap {
+    let bootstrap = ClientBootstrap(group: configuration.eventLoopGroup)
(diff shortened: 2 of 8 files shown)


glbrntt:
@MrMage I've only updated the Echo example to use the new API for now. I'll update the reset and remove what I've marked as "Legacy API" in a follow-up PR once these changes have been approved.

Sources/GRPC/GRPCClientConnection.swift
```diff
@@ -171,7 +153,163 @@ open class GRPCClientConnection {
   }
 }
 
+// MARK:- Configuration structures
+
+/// A target to connect to.
+public enum ConnectionTarget {
```
MrMage: Move this inside `GRPCClientConnection`, or `GRPCClientConnection.Configuration`? Just an idea, might make more sense to keep it outside.

Sources/GRPC/GRPCClientConnection.swift
```diff
@@ -171,7 +153,163 @@ open class GRPCClientConnection {
...
+
+  var host: String? {
+    guard case .hostAndPort(let host, _) = self else {
+      return nil
+    }
+    return host
+  }
+}
+
 extension GRPCClientConnection {
```
MrMage: Not really related to this PR, but do you think we should rename `GRPCClientConnection` to `ClientConnection`? Might be too generic, though.

Sources/GRPC/GRPCClientConnection.swift
```diff
@@ -171,7 +153,163 @@ open class GRPCClientConnection {
...
+  /// The configuration for a connection.
+  public struct Configuration {
+    /// The target to connect to.
+    public var target: ConnectionTarget
```
MrMage: Should all these be `var` or `let`? (The struct itself is probably stored as `let`, anyway, but still.)

Sources/GRPC/GRPCClientConnection.swift
```diff
@@ -171,7 +153,163 @@ open class GRPCClientConnection {
...
+
+    /// The event loop group to run the connection on.
+    public var eventLoopGroup: EventLoopGroup
+
+    /// An error delegate which is called when errors are caught. Provided delegates **must not
+    /// maintain a strong reference to this `GRPCClientConnection`**. Doing so will cause a retain
```
MrMage: Good documentation! 😍

Sources/GRPC/GRPCClientConnection.swift
```diff
@@ -54,111 +54,93 @@ import NIOTLS
...
   }
 
   public let channel: Channel
   public let multiplexer: HTTP2StreamMultiplexer
-  public let host: String
-  public let httpProtocol: HTTP2ToHTTP1ClientCodec.HTTPProtocol
-  public let errorDelegate: ClientErrorDelegate?
+  public let configuration: Configuration
+
+  internal var httpProtocol: HTTP2ToHTTP1ClientCodec.HTTPProtocol {
```
MrMage: Remove this variable altogether and use `configuration.httpProtocol` directly?

Sources/GRPC/GRPCClientConnection.swift
```diff
@@ -171,7 +153,163 @@ open class GRPCClientConnection {
...
+      self.eventLoopGroup = eventLoopGroup
+      self.errorDelegate = errorDelegate
+      self.tlsConfiguration = tlsConfiguration
+    }
+  }
+
+  /// The TLS configuration for a connection.
+  public struct TLSConfiguration {
+    /// The SSL context to use.
+    public var sslContext: NIOSSLContext
```
MrMage: Also `let`, as above?

Sources/GRPC/GRPCClientConnection.swift
```diff
@@ -171,7 +153,163 @@ open class GRPCClientConnection {
...
+      ) {
+      self.target = target
+      self.eventLoopGroup = eventLoopGroup
+      self.errorDelegate = errorDelegate
+      self.tlsConfiguration = tlsConfiguration
+    }
+  }
+
+  /// The TLS configuration for a connection.
+  public struct TLSConfiguration {
```
MrMage: Should we provide a "default" initializer similar to the one you created in the Echo example?

Sources/GRPC/GRPCClientConnection.swift
```diff
@@ -171,7 +153,163 @@ open class GRPCClientConnection {
...
+    public var hostnameOverride: String?
+
+    public init(sslContext: NIOSSLContext, hostnameOverride: String? = nil) {
+      self.sslContext = sslContext
+      self.hostnameOverride = hostnameOverride
+    }
+  }
+}
+
+// MARK:- Configuration helpers/extensions
```
MrMage: nit: space after colon? (also elsewhere)

Sources/GRPC/GRPCClientConnection.swift
```diff
@@ -171,7 +153,163 @@ open class GRPCClientConnection {
...
+
+fileprivate extension ClientBootstrap {
+  /// Connect to the given connection target.
+  ///
+  /// - Parameter target: The target to connect to.
+  func connct(to target: ConnectionTarget) -> EventLoopFuture<Channel> {
```
MrMage: typo: connct

Sources/GRPC/GRPCClientConnection.swift
```diff
@@ -171,7 +153,163 @@ open class GRPCClientConnection {
...
+  ///   - eventLoopGroup: Event loop group to run the connection on.
+  ///   - errorDelegate: An error delegate which is called when errors are caught. Provided
+  ///       delegates **must not maintain a strong reference to this `GRPCClientConnection`**. Doing
+  ///       so will cause a retain cycle. Defaults to a delegate which logs errors in debug builds
+  ///       only.
+  ///   - tlsMode: How TLS should be configured for this connection.
+  ///   - hostOverride: Value to use for TLS SNI extension; this must not be an IP address. Ignored
+  ///       if `tlsMode` is `.none`.
+  /// - Returns: A future which will be fulfilled with a connection to the remote peer.
+  public static func start(
```
MrMage: Should we mark this as deprecated?

Sources/Examples/Echo/main.swift
```diff
@@ -32,11 +32,11 @@ let messageOption = Option("message",
                            default: "Testing 1 2 3",
                            description: "message to send")
 
-func makeClientTLS(enabled: Bool) throws -> GRPCClientConnection.TLSMode {
+func makeClientSSLContext(enabled: Bool) throws -> NIOSSLContext? {
```
MrMage: Remove `enabled` and only call this if it needs to be called?

Would also replace the call site with

```
let tlsConfiguration = ssl ? GRPCClientConnection.TLSConfiguration(try makeClientSSLContext()) : nil
```

Sources/Examples/Echo/main.swift
```diff
@@ -32,11 +32,11 @@ let messageOption = Option("message",
...
   guard enabled else {
-    return .none
+    return nil
   }
-  return .custom(try NIOSSLContext(configuration: try makeClientTLSConfiguration()))
+  return try NIOSSLContext(configuration: try makeClientTLSConfiguration())
```
MrMage: The inner `try` can be omitted.

Sources/Examples/Echo/main.swift
glbrntt: Ah true, that makes more sense now.

Sources/GRPC/GRPCClientConnection.swift
glbrntt: Great question, I've been wondering about this too since _most_ of the objects are prefixed with `GRPC`. We need a naming policy of some kind, ideally before tagging 1.0. I'm not sure what this should be though!

Sources/GRPC/GRPCClientConnection.swift
glbrntt: `var`, since sometimes it's easier to conditionally set configuration after the fact (I think it's used somewhere in this PR to set TLS configuration).

More generally though I'd say if you want a struct to be immutable you should store it in a `let` as opposed to forcing it on the property level.

Sources/GRPC/GRPCClientConnection.swift
glbrntt: Copied from previous documentation but thanks!

Sources/GRPC/GRPCClientConnection.swift
glbrntt: I was going to remove it in the next PR so didn't bother :)

Sources/GRPC/GRPCClientConnection.swift
glbrntt: I'm not sure what you mean here?

MrMage: Essentially moving the `makeClientSSLContext` from the example into the framework :-)

Sources/GRPC/GRPCClientConnection.swift
MrMage: Agreed, let's discuss this later!

Sources/GRPC/GRPCClientConnection.swift
glbrntt: We should be able to reuse this for the server so I think separate makes more sense.

Sources/GRPC/GRPCClientConnection.swift
glbrntt: Oh I see, I'm going to hold off on this for now if you don't mind because it just occurred to me that setting up SSL for NIOTS is different, so no doubt this configuration will have to change anyway.

Sources/GRPC/GRPCClientConnection.swift
```diff
@@ -54,111 +54,90 @@ import NIOTLS
...
+  /// - Parameter configuration: The configuration to start the connection with.
+  public class func start(_ configuration: Configuration) -> EventLoopFuture<GRPCClientConnection> {
+    return makeBootstrap(configuration: configuration)
+      .connect(to: configuration.target)
+      .flatMap { channel in
+        let tlsVerified: EventLoopFuture<Void>?
+        if configuration.tlsConfiguration != nil {
+          tlsVerified = verifyTLS(channel: channel)
+        } else {
+          tlsVerified = nil
```
MrMage: How about simply using `= channel.eventLoop.makeSucceededFuture(())` in this case, and making `tlsVerified` non-optional?

glbrntt: I had the same thought but it felt weird naming it `tlsVerified` if TLS wasn't being used and I couldn't come up with a reasonable name! Happy to change it if you like, though.

glbrntt:
> Also, do you think we could have a short snippet of documentation in the README explaining how this should be used once it is done?

Absolutely, updating the entire README is on my TODO list. Some kind of tutorial would be good, maybe it should be done via the Echo example and pointed to from the README?

MrMage:
> Absolutely, updating the entire README is on my TODO list. Some kind of tutorial would be good, maybe it should be done via the Echo example and pointed to from the README?

Sounds good, but I'd still try to have a 10-line example in the README, I just find that nicer than having to read the example.

Sources/GRPC/GRPCClientConnection.swift
MrMage: Fine as is :-)

PR #480, opened 2019-06-10
Add connection backoff

glbrntt:
- If connections fail to start, they can now retry after some backoff. 
- Connection will fail eventually (once backoff becomes too large).


Sources/GRPC/ConnectionBackoff.swift
```diff
@@ -0,0 +1,169 @@
...
+    self.currentElement = self.makeElement(backoff: 0.0)
+  }
+
+  /// The configuration being used.
+  private let connectionBackoff: ConnectionBackoff
+
+  /// The backoff in seconds, without jitter.
+  private var unjitteredBackoff: TimeInterval
+
+  /// The state of production.
```
MrMage: Could you provide a more elaborate comment here?

Sources/GRPC/ConnectionBackoff.swift
```diff
@@ -0,0 +1,169 @@
...
+  private var productionState: ProductionState
+
+  /// The current timeout and backoff waiting to be returned. This will become `nil` when we run
+  /// out of values to return.
+  private var currentElement: Element?
+
+  /// The next timeout and backoff to be returned.
+  private var nextElement: Element?
```
MrMage: Why do we need to store this? Couldn't we generate it on the fly when calling `next()`?

Sources/GRPC/ConnectionBackoff.swift
```diff
@@ -0,0 +1,169 @@
...
+
+  // The current state of production.
```
MrMage: Same here about the comment.

Sources/GRPC/GRPCClientConnection.swift
```diff
@@ -114,9 +114,32 @@ open class GRPCClientConnection {
...
   public class func start(_ configuration: Configuration) -> EventLoopFuture<GRPCClientConnection> {
-    return makeBootstrap(configuration: configuration)
-      .connect(to: configuration.target)
-      .flatMap { channel in
+    return start(configuration, backoffIterator: configuration.connectionBackoff?.makeIterator())
+  }
+
+  /// Starts a client connection using the given configuration and backoff.
+  ///
+  /// In additiaon to the steps taken in `start(configuration:)`, we _may_ additionally set a
```
MrMage: typo: additiaon

Sources/GRPC/GRPCClientConnection.swift
```diff
@@ -114,9 +114,32 @@ open class GRPCClientConnection {
...
+  /// connection timeout and schedule a retry attempt (should the connection fail) if a
+  /// `ConnectionBackoff.Iterator` is provided.
+  ///
+  /// - Parameter configuration: The configuration to start the connection with.
+  /// - Parameter backoffIterator: A `ConnectionBackoff` iterator which generates connection
```
MrMage: Could we make this `backoffIterator` part of the connection somehow?

Tests/GRPCTests/ClientConnectionBackoffTests.swift
```diff
@@ -0,0 +1,92 @@
...
+
+    self.wait(for: [serverStarted, clientConnected], timeout: 2.0, enforceOrder: true)
+  }
+
+  func testClientEventuallyTimesOut() throws {
+    var configuration = self.makeClientConfiguration()
+    configuration.connectionBackoff = ConnectionBackoff(maximumBackoff: 0.1)
+
+    self.client = self.makeClientConnection(configuration)
+    XCTAssertThrowsError(try self.client.wait())
```
MrMage: Should we check the error thrown?

Tests/GRPCTests/ClientConnectionBackoffTests.swift
```diff
@@ -0,0 +1,92 @@
...
+
+  func testClientEventuallyConnects() throws {
+    let clientConnected = self.expectation(description: "client connected")
+    let serverStarted = self.expectation(description: "server started")
+
+    // Start the client first.
+    self.client = self.makeClientConnection(self.makeClientConfiguration())
+    self.client.assertSuccess(fulfill: clientConnected)
+
+    self.server = try self.makeServer()
```
MrMage: Should we pause slightly before creating the server, to ensure that the back-off actually reconnects after a short delay?

Tests/GRPCTests/ConnectionBackoffTests.swift
```diff
@@ -0,0 +1,93 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import Foundation
+import GRPC
+import XCTest
+
+class ConnectionBackoffTests: XCTestCase {
+  var backoff: ConnectionBackoff!
```
MrMage: Could we simply initialize this as `var backoff = ConnectionBackoff()` and remove `setUp()`?

Sources/GRPC/ConnectionBackoff.swift
glbrntt: Great question -- I was using this differently when starting the connection in a way which required the initial backoff to be zero (i.e. waiting and then trying instead of trying then waiting). Since the first _real_ backoff isn't jittered we needed to create it manually. Since that's not the case anymore we can simplify this.

Sources/GRPC/GRPCClientConnection.swift
glbrntt: How do you mean? We don't have a `Connection` instance yet since it's a `class func`.

MrMage: Part of the configuration, sorry.

glbrntt: It's should be possible but I don't think it belongs there: the iterator is state derived from configuration.

Tests/GRPCTests/ClientConnectionBackoffTests.swift
glbrntt: Good idea, done.

Tests/GRPCTests/ClientConnectionBackoffTests.swift
glbrntt: It shouldn't matter but there's no harm in doing, it certainly makes the test more "obvious".

Tests/GRPCTests/ConnectionBackoffTests.swift
glbrntt: 🤦‍♂️, done

PR #481, opened 2019-06-10
Add AnyServiceClient

glbrntt:
Motivation:
- Sometimes it's useful to call a service when you don't have a client
  generated for it.
- We had `GRPCClient` which contains all the boilerplate for creating
  calls, but no concrete implementation.

Modifications:
- Add `AnyServiceClient`, a minimal `GRPCClient` with no generated
  stubs.
- Added tests.


Sources/GRPC/GRPCClient.swift
```diff
@@ -29,39 +29,59 @@ extension GRPCClient {
   public func makeUnaryCall<Request: Message, Response: Message>(
     path: String,
     request: Request,
-    callOptions: CallOptions,
+    callOptions: CallOptions? = nil,
```
MrMage: What's the reason for making this optional?

glbrntt: The `GRPCClient` protocol also requires `defaultCallOptions`, so we should use those if the `callOptions` aren't specified explicitly. 

If this wasn't the case then if you use `AnyServiceClient` you'd always have to specify `anyClient.make<Type>Call(..., callOptions: anyClient.defaultCallOptions, ...)`

PR #483, opened 2019-06-13
Update README for NIO based implementation

glbrntt:
Motivation:

The README was focused mostly on the libgrpc implementation.

Modifications:

Rewrote large chunks of the README, including:
- How to integrate GRPC with an existing/new project
- Details on how to use GRPC with an example
- Differences between the call types
- How to make calls without a generated client

Result:

More up-to-date README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,13 +205,17 @@ logical cores on your system.
 We also `wait` for the server to start before setting up a client.
 
 ```swift
+let serverEventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
 let server = try Server.start(
   hostname: host,
   port: port,
-  eventLoopGroup: MultiThreadedEventLoopGroup(numberOfThreads: 1),
+  eventLoopGroup: serverEventLoopGroup,
   serviceProviders: [EchoProvider()]
 ).wait()
 ```
+
+Note that the at the end of the program, the `serverEventLoopGroup` whould be
+shutdown (`try serverEventLoopGroup.syncShutdownGracefully()`).
 
 Normally the client would be in a different binary to the server, however, for
 this example we will include them together.
@@ -222,9 +226,10 @@ means to make gRPC calls via a generated client. Note that
 `Echo_EchoServiceClient` is the client we generated from `echo.proto`.
 
 ```swift
+let clientEventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
 let configuration = ClientConnection.Configuration(
   target: .hostAndPort(host, port),
-  eventLoopGroup: MultiThreadedEventLoopGroup(numberOfThreads: 1)
+  eventLoopGroup: clientEventLoopGroup
 )
 
 let echo = try ClientConnection.start(configuration).map {
@@ -232,6 +237,9 @@ let echo = try ClientConnection.start(configuration).map {
   Echo_EchoServiceClient(connection: $0)
 }.wait()
 ```
+
+Note that the `clientEventLoopGroup` should also be shutdown when it is no
+longer required.
 
 We can also define some options on our call, such as custom metadata and a
 timeout. Note that options are not required at the call-site and may be omitted
@@ -368,6 +376,39 @@ These differences are summarised in the following table.
 | Server Streaming        | Handler  | At call time
 | Bidirectional Streaming | Handler  | On the call object
 
+### Using TLS
+
+gRPC calls can be made over a secure channel by configuring TLS. This requires
+specifying `tlsConfiguration` on `ClientConnection.Configuration` or
+`Server.Configuration`.
+
+Configuring TLS requires a [`NIOSSLContext`][nio-ref-sslcontext] which in turn
+requires a [`TLSConfiguration`][nio-ref-tlsconfig] (this is _not_ the same as
+`ClientConnection.TLSConfiguration` or `Server.TLSConfiguration`).
+
+For the client, [`TLSConfiguration`][nio-ref-tlsconfig] can be as simple as:
+
+```swift
+let configuration = TLSConfiguration.forClient(
+  minimumTLSVersion: TLSVersion = .tlsv12,
+  applicationProtocols: GRPCApplicationProtocolIdentifier.allCases.map { $0.rawValue }
+)
+```
+
+For the server, [`TLSConfiguration`][nio-ref-tlsconfig] is slightly more
+complicated as it requires a certificate chain and private key:
+
+```swift
+let configuration = TLSConfiguration.forServer(
+  certificateChain: [.file("cert.pem")],
+  privateKey: .file("key.pem"),
+  minimumTLSVersion: TLSVersion = .tlsv12,
+  applicationProtocols: GRPCApplicationProtocolIdentifier.allCases.map { $0.rawValue }
+)
+```
+
+Note that the gRPC specification requires TLS v1.2 or later.
+
 ### Plugin Options
 
 To pass extra parameters to the plugin, use a comma-separated parameter list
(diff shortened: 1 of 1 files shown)


glbrntt:
@MrMage @timburks @rebello95 Would appreciate all of your feedback on this one!

Also would it make sense to refer to this project as "gRPC Swift" (as opposed to "Swift gRPC") as it's a closer match to the repository name and the convention used by other gRPC implementations (as far as I can tell).

README.md
```diff
@@ -1,209 +1,404 @@
...
-and [Swift gRPC](https://github.com/grpc/grpc-swift) plugins to
-generate the necessary support code.
+APIs and generated code is provided for both gRPC clients and servers, and can
+be built either with Xcode or the Swift Package Manager. Support is provided for
+all four gRPC API styles (Unary, Server Streaming, Client Streaming, and
+Bidirectional Streaming) and connections can be made either over secure (TLS) or
+insecure channels.
 
-### Getting the plugins
+Swift gRPC is build on top of [Swift NIO][swift-nio] as opposed to the core
```
MrMage: nit: "built"

README.md
```diff
@@ -1,209 +1,404 @@
...
+]
+```
 
-#### Parameters
+##### Xcode
 
-To pass extra parameters to the plugin, use a comma-separated parameter list
-separated from the output directory by a colon.
+From Xcode 11 it is possible to [add Swift Package dependencies to Xcode
+projects][xcode-spm] and link targets to products of those packages, this is the
```
MrMage: super-nit: could we have a semicolon in front of "this"?

README.md
```diff
@@ -1,209 +1,404 @@
...
+above definition for the Echo service and models were saved as `echo.proto` and
+the `protoc-gen-swift` plugin is available in your `PATH` then the following
+will generate the models and write them to `echo.pb.swift` in the current
+directory:
+
+```sh
+protoc echo.proto --swift_out=.
+```
+
+Swift gRPC, provides a plugin (`protoc-gen-swiftgrpc`) to generate the client
```
MrMage: nit: remove the comma?

README.md
```diff
@@ -1,209 +1,404 @@
...
+server. Note that you can also use `System.coreCount` to get the number of
+logical cores on your system.
+
+We also `wait` for the server to start before setting up a client.
+
+```swift
+let server = try Server.start(
+  hostname: host,
+  port: port,
+  eventLoopGroup: MultiThreadedEventLoopGroup(numberOfThreads: 1),
```
MrMage: I assume this event loop group would never be shut down? (Fine for an example and most servers that run indefinitely, I guess)

README.md
```diff
@@ -1,209 +1,404 @@
...
+
+Once the server has started we can create a client by first defining its
+configuration, starting a connection, and then using that connection as a means
+to make gRPC calls via a generated client. Note that `Echo_EchoServiceClient` is
+the client we generated from `echo.proto`.
+
+```swift
+let configuration = ClientConnection.Configuration(
+  target: .hostAndPort(host, port),
+  eventLoopGroup: MultiThreadedEventLoopGroup(numberOfThreads: 1)
```
MrMage: same here regarding event loop group shutdown

README.md
```diff
@@ -1,209 +1,404 @@
...
+```swift
+let server = try Server.start(
+  hostname: host,
+  port: port,
+  eventLoopGroup: MultiThreadedEventLoopGroup(numberOfThreads: 1),
+  serviceProviders: [EchoProvider()]
+).wait()
+```
+
+Once the server has started we can create a client by first defining its
```
MrMage: I would make it explicit that the client is usually going to be a different binary.

README.md
```diff
@@ -1,209 +1,404 @@
...
+```swift
+var callOptions = CallOptions()
+// Add a request id header to the call.
+callOptions.customMetadata.add(name: "x-request-id", value: UUID().uuidString)
+// Set the timeout to 5 seconds. This may throw since the timeout is validated against
+// the gRPC specification, which limits timeouts to be at most 8 digits long.
+callOptions.timeout = try .seconds(5)
+```
+
+We can now may an asynchronous call to the service by using the functions on the
```
MrMage: typo: make

README.md
```diff
@@ -1,209 +1,404 @@
...
+  eventLoopGroup: MultiThreadedEventLoopGroup(numberOfThreads: 1)
+)
+
+let echo = try ClientConnection.start(configuration).map {
+  // This is our generated client, we only need to pass it a connection.
+  Echo_EchoServiceClient(connection: $0)
+}.wait()
+```
+
+We can also define some options on our call, like custom metadata and a timeout.
```
MrMage: Make it more explicit that the call options can be omitted altogether?

README.md
```diff
@@ -1,209 +1,404 @@
...
+
+The returned `get` object is of type `UnaryClientCall<Echo_EchoRequest, Echo_EchoResponse>`
+and has [futures][nio-ref-elf] for the initial metadata, response, trailing
+metadata and call status. The differences between call types is detailed in
+[API Types](#api-types).
+
+Note that the call can be made synchronous by waiting on the `response` on the
+`get` object:
+
+```
```
MrMage: Add `swift`

README.md
```diff
@@ -1,209 +1,404 @@
...
+
+let get = echo.get(request, callOptions: callOptions)
+```
+
+The returned `get` object is of type `UnaryClientCall<Echo_EchoRequest, Echo_EchoResponse>`
+and has [futures][nio-ref-elf] for the initial metadata, response, trailing
+metadata and call status. The differences between call types is detailed in
+[API Types](#api-types).
+
+Note that the call can be made synchronous by waiting on the `response` on the
```
MrMage: "response property of the get object"?

README.md
```diff
@@ -1,209 +1,404 @@
...
+`get` object:
+
+```
+let response = try echo.get(request).response.wait()
```
MrMage: just `try get.response.wait()`?

README.md
```diff
@@ -1,209 +1,404 @@
...
+  print("The Get call to the Echo service failed: \(error)")
+}
+```
+
+The call will always terminate with a status which includes a status code and
+optionally a message and trailing metadata.
+
+This is often the most useful way to determine the outcome of a call. However,
+it should be noted that __even if the call fails, the `status` future will be
+succeeded__.
```
MrMage: make "succeeded" italics?

README.md
```diff
@@ -1,209 +1,404 @@
...
+These methods are also available on generated clients, allowing you to call
+methods which have been added to the service since the client was generated.
+
+### API Types
+
+Swift gRPC provides all four API styles: Unary, Server Streaming, Client
+Streaming, and Bidirectional Streaming. Calls to the generated types will return
+an object of the approriate type:
+
+- `UnaryClientCall<Request, Response>`
```
MrMage: Should we link these to the corresponding files?

README.md
glbrntt: That's correct and this should definitely be fixed! 

In parallel, I've been doing the work to get NIOTS integrated. This includes giving the user an option to specify which model to use (NIO vs NIOTS), which is actually "always NIO" or "NIOTS when possible, otherwise NIO". Since that option fits quite nicely into `Configuration` (this needs to be done for the `Server` too) and the `EventLoopGroup` depends on that option, it might make more sense for the client/server to own the ELG, and shut it down when it closes.

What do you think?

glbrntt:
> The only thing I'd suggest adding would be a short paragraph on handling SSL support, especially on the client.

Great idea, totally slipped my mind, I'll add this.

README.md
glbrntt: We could, but I'm not sure how much value there is, since each of them just constructs a request and a response handler before calling the superclass initializer.

README.md
MrMage: Alternatively, we could add an argument like this to `Configuration`:

```swift
enum OperationMode {
  case nio(eventLoopGroup: EventLoopGroup)
  case nioTS
}
```

Not sure if that would work, though. I'm not feeling too strongly, as long as we can have compatibility with Vapor (although that would probably focused on Linux and thus regular NIO, anyway). I guess if NIOTS makes it even easier for client users because they don't have to worry about the event loop group, that would be great.

glbrntt: What are the requirements for compatibility with Vapor? 

To clarify: NIOTS needs an ELG but it's a `NIOTSEventLoopGroup` which is incompatible with `{Client,Server}Bootstrap`.

It will be easier to discuss this in the context of the NIOTS PR. We can then circle back to this example/README.

MrMage: Good question. I guess Vapor users would need to use the non-TS version for now, anyway. We can defer this conversation in any case.

PR #484, opened 2019-06-13
Add missing bracket in NaturalLanguage example

glbrntt:
This was broken in #479 but it wasn't picked up because it isn't a target in the main `Package.swift`.


glbrntt:
We should probably move all of the example (when rewritten) to `Sources/Examples` (where the Echo example lives) and add them as targets to the main `Package.swift`.

MrMage:
> We should probably move all of the example (when rewritten) to `Sources/Examples` (where the Echo example lives) and add them as targets to the main `Package.swift`.

Agreed.

PR #486, opened 2019-06-14
Tidy up and add comments to Package.swift

glbrntt:
Tidy up `Package.swift`:
- added comments to each target
- changed `Commander` from `.upToNextMinor(from: "0.8.0")` to `from: "0.8.0"`

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -35,7 +35,7 @@ let package = Package(
     .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.0.0"),
 
     // Official SwiftProtobuf library, for [de]serializing data to send on the wire.
-    .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.3.1"),
+    .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.5.0"),
 
     // Command line argument parser for our auxiliary command line tools.
     .package(url: "https://github.com/kylef/Commander.git", from: "0.8.0"),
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -41,46 +25,95 @@ let package = Package(
...
+    // GRPC dependencies:
+    // Main SwiftNIO package
+    .package(url: "https://github.com/apple/swift-nio.git", from: "2.2.0"),
+    // HTTP2 via SwiftNIO
+    .package(url: "https://github.com/apple/swift-nio-http2.git", from: "1.2.1"),
+    // TLS via SwiftNIO
+    .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.0.0"),
+
+    // Official SwiftProtobuf library, for [de]serializing data to send on the wire.
+    .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.3.1"),
```
MrMage: How about "from 1.5.0"?

Package.swift
```diff
@@ -41,46 +25,95 @@ let package = Package(
...
+    .target(
+      name: "protoc-gen-swiftgrpc",
+      dependencies: [
+        "SwiftProtobuf",
+        "SwiftProtobufPluginLibrary",
+        "protoc-gen-swift"
+      ]
+    ),
+
+    // Interoperability tests, this doesn't contain the CLI as this is reused
```
MrMage: What does the second "this" refer to? Could that "this" replaced for clarity?

Package.swift
glbrntt: Ah yes, good idea!

Package.swift
glbrntt: "this" refers to the target, will clarify!

MrMage: Sorry, mind further clarifying "this" to "the interoperability tests are..."=

MrMage:
Thank you for the changes!

PR #489, opened 2019-06-19
Add automatic reconnect to the client connection.

glbrntt:
Motivation:

Clients could lose their connections to the server for a number of
reasons, we offered no means to automatically reconnect.

Modification:

- `ClientConnection` now stores a future Channel and future multiplexer,
  as opposed to future `ClientConnection`s holding a channel and
  multiplexer.
- This allows us to replace the future channel and future multiplexer
  when the client is closed (but not explicitly via `close()`).
- The state can be monitored via a delegate or by registering callbacks
  on the next transition to a given state.
- Reconnection uses the same backoff logic used for the initial
  connection creation.

Results:

Clients can automatically reconnect when their connection goes away.


Sources/GRPC/ClientConnection.swift
```diff
@@ -53,139 +53,239 @@ import NIOTLS
...
+    channel.always { result in
+      // If we failed to get a channel then we've exhausted our backoff; we should `.shutdown`.
+      if case .failure = result {
+        self.connectivity.state = .shutdown
+      }
+    }.flatMap {
+      $0.closeFuture
+    }.whenComplete { _ in
+      // `.shutdown` is terminal so don't attempt a reconnection.
+      if self.connectivity.state != .shutdown {
```
MrMage: Could this be a `guard` instead?

Sources/GRPC/ClientConnection.swift
```diff
@@ -53,139 +53,239 @@ import NIOTLS
...
+  private func replaceChannelAndMultiplexerOnClose(channel: EventLoopFuture<Channel>) {
+    channel.always { result in
+      // If we failed to get a channel then we've exhausted our backoff; we should `.shutdown`.
+      if case .failure = result {
+        self.connectivity.state = .shutdown
+      }
+    }.flatMap {
+      $0.closeFuture
+    }.whenComplete { _ in
+      // `.shutdown` is terminal so don't attempt a reconnection.
```
MrMage: This looks very similar to `replaceChannelAndMultiplexer()`, any chance to merge the two? (If they are too different, just let me know.)

Also, when is `replaceChannelAndMultiplexer()` even called?

Sources/GRPC/ClientConnection.swift
```diff
@@ -53,139 +53,239 @@ import NIOTLS
...
-  /// handshake was successful (if TLS was configured) and creating the `ClientConnection`.
-  /// See the individual functions for more information:
-  ///  - `makeBootstrap(configuration:)`,
-  ///  - `verifyTLS(channel:)`, and
-  ///  - `makeClientConnection(channel:configuration:)`.
-  ///
-  /// - Parameter configuration: The configuration to start the connection with.
-  public class func start(_ configuration: Configuration) -> EventLoopFuture<ClientConnection> {
-    return start(configuration, backoffIterator: configuration.connectionBackoff?.makeIterator())
+  /// The `EventLoop` this connection is using. Note that this _may_ change over time.
```
MrMage: Not a fan of allowing this to change; could we re-use the previous channel's event loop instead?

Sources/GRPC/ConnectivityState.swift
```diff
@@ -0,0 +1,134 @@
...
+  /// error). Channels in this state will eventually switch to the `.connecting` state and try to
+  /// establish a connection again. Since retries are done with exponential backoff, channels that
+  /// fail to connect will start out spending very little time in this state but as the attempts
+  /// fail repeatedly, the channel will spend increasingly large amounts of time in this state.
+  case transientFailure
+
+  /// This channel has started shutting down. Any new RPCs should fail immediately. Pending RPCs
+  /// may continue running till the application cancels them. Channels may enter this state either
+  /// because the application explicitly requested a shutdown or if a non-recoverable error has
+  /// happened during attempts to connect communicate. Channels that enter this state never leave
```
MrMage: "connect communicate"?

MrMage: Rephrase as "Channels that **have** entered this state **will** never leave this state."?

Sources/GRPC/ConnectivityState.swift
```diff
@@ -0,0 +1,134 @@
...
+    }
+  }
+
+  /// Creates a new connectivity state monitor.
+  ///
+  /// - Parameter delegate: A delegate to call when the connectivity state changes.
+  /// - Parameter initialState: The initial connectivity state, defaults to `.idle`.
+  public init(delegate: ConnectivityStateDelegate?, initialState: ConnectivityState = .idle) {
+    self.delegate = delegate
+    self.state = .idle
```
MrMage: `= initialState`, or reme `initialState` argument

Sources/GRPC/ConnectivityState.swift
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  /// Registers a callback on the given state and calls it the next time that state is observed.
+  /// Subsequent transitions to that state will **not** trigger the callback.
+  ///
+  /// - Parameter state: The state on which to call the given callback.
+  /// - Parameter callback: The closure to call once the given state has been transitioned to. The
+  ///     `callback` can be removed by passing in `nil`.
+  public func onNext(state: ConnectivityState, callback: Callback?) {
```
MrMage: What is this for as opposed to just providing a delegate, and why is the callback reset after every change to that state?

Sources/GRPC/TLSVerificationHandler.swift
```diff
@@ -38,26 +37,21 @@ public class TLSVerificationHandler: ChannelInboundHandler, RemovableChannelHand
     return verificationPromise.futureResult
   }
 
-  public init(errorDelegate: ClientErrorDelegate?) {
```
MrMage: Why was the error delegate removed? Is that done through `context.fireErrorCaught(error)` now?

Tests/GRPCTests/ClientConnectionBackoffTests.swift
```diff
@@ -18,82 +18,169 @@ import GRPC
...
+    self.server = try self.makeServer()
+    let server = try self.server.wait()
+
+    let connectionReady = self.expectation(description: "connection ready")
+    self.client = self.makeClientConnection(self.makeClientConfiguration())
+    self.client.connectivity.onNext(state: .ready) {
+      connectionReady.fulfill()
     }
+
+    // Once the connection ready we can kill the server.
```
MrMage: nit: add "is"

Tests/GRPCTests/ClientConnectionBackoffTests.swift
```diff
@@ -18,82 +18,169 @@ import GRPC
...
+    }
+
+    // Start a new server.
+    self.serverGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+    self.server = try self.makeServer()
+
+    self.wait(for: [reconnectionReady], timeout: 1.0)
+    XCTAssertEqual(self.stateDelegate.clearStates(), [.connecting, .ready])
+
+    // Ensure we can actually make a call.
```
MrMage: Could we also test making a call after the first server has shut down, but before the second one has been started?

Sources/GRPC/ClientConnection.swift
glbrntt: Do you mean re-use it for the new channel? Or store it and always return it here? Because the former isn't possible.

Sources/GRPC/ClientConnection.swift
glbrntt: Agh, sorry, `replaceChannelAndMultiplexer` was meant to be removed.

Sources/GRPC/ConnectivityState.swift
glbrntt: Good question; my take on it is that they're useful for different periods of time. You can achieve exactly the same with a delegate but in some cases this is just _much_ more convenient (the tests are a good example of this). The delegate is useful for longer time periods (logging/monitoring). This is useful for single events (e.g. my connection has shutdown, I can now update my UI) which is why it resets after each change. Definitely open to not resetting after each change though as I realise that's a slightly strange decision!

Sources/GRPC/TLSVerificationHandler.swift
glbrntt: Exactly. We added an error handler with the error delegate after this a little while ago and never removed this.

Tests/GRPCTests/ClientConnectionBackoffTests.swift
glbrntt: We can, but the reconnect will already be taking place: if we `wait()` on that call to complete (before starting the server) it will eventually time out the reconnect (and cause a `.shutdown`). 

If we don't `wait()` on it, then it will eventually connect and that call will succeed.

I'll add the former as a separate test.

MrMage: > If we don't wait() on it, then it will eventually connect and that call will succeed.

I think that's what we should do — start the call while the server is offline, then `wait()` for it to succeed _after_ the server is back up.

Sources/GRPC/ClientConnection.swift
MrMage: I was hoping for the former; why would that be a problem?

(My concern is that Vapor does a lot of thread-local stuff and generally expects all "sub-requests" to run on the same event loop the original request is handled on. Open to keep this as-is for now, though, given that one could simply pass in the request's `EventLoop` as the client's `EventLoopGroup` argument.)

Sources/GRPC/ConnectivityState.swift
MrMage: Also, if we don't reset, we could have the user pass in a closure instead that could `switch` over the states they are interested in and discard the rest via `default:`.

At that point, one could also argue over removing the delegate and only letting the user provide callback closures. And at that point we could even keep an array of closures, in case the user wants more than one observer.

To be honest, having two different ways of passing in callback closures feels a bit weird right now, but this is not a blocker. We can revisit it later (before 1.0).

Sources/GRPC/ClientConnection.swift
glbrntt: I take that back -- I didn't realise `EventLoop` conformed to `EventLoopGroup`!

PR #493, opened 2019-06-26
fix carthage install

KyoheiG3:
It will be possible to build using bundler.
Then please add a new tag.


MrMage:
@KyoheiG3 a new tag is being released right now.

PR #498, opened 2019-07-02
Revert to Apache 2.0 license verbatim [skip ci]

mbrukman:
Replaced the license with the verbatim contents of the canonical license URL:
https://www.apache.org/licenses/LICENSE-2.0.txt

The license seems to be missing random lines, the entire paragraph at the end,
and should not be replacing the template with a specific copyright as this is the
license, not a copyright statement for the project. Each header has a copyright
statement at the top.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,13 +1,14 @@
 
                                  Apache License
                            Version 2.0, January 2004
-                        https://www.apache.org/licenses/
+                        http://www.apache.org/licenses/
 
    TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
 
    1. Definitions.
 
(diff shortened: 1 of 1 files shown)


glbrntt:
We'll also need to apply this patch to the `nio` branch

mbrukman:
Do you merge `master` into `nio` or vice versa during the course of development?

Or would you prefer I open a separate PR for the `nio` branch?

glbrntt:
No, `nio` will eventually replace `master`. A separate PR would be awesome if you don't mind (if not I can create one).

Thanks for spotting this by the way! 😬

mbrukman:
Sure, no problem! See https://github.com/grpc/grpc-swift/pull/504 for the PR to `nio` branch.

PR #506, opened 2019-07-04
Remove `onNext` from connectivity state delegate

glbrntt:


glbrntt:
Hmm not sure why that test is failing on CI. Looking into it...

PR #513, opened 2019-07-12
Provide restricted TLS configurations

glbrntt:
Motivation:

Providing TLS configuration is difficult without prior knowledge since
one must provide the application protocols identifiers for ALPN. In
addition the default minimum TLS version provided by NIOs
TLSConfiguration is lower than that specified by the gRPC protococl.

Modifications:

Add TLS configuration for both server and client which defaults which
better align to gRPC.

Add shims marking the old APIs as deprecated.

Result:

TLS is easier to configure.

### DIFF
--- a/Sources/GRPC/Shims.swift
+++ b/Sources/GRPC/Shims.swift
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2019, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import NIOSSL
+
+// This file contains shims to notify users of API changes between v1.0.0-alpha.1 and v1.0.0.
+
+// TODO: Remove these shims before v1.0.0 is tagged.
+
+extension ClientConnection.Configuration {
+  @available(*, deprecated, message: "use 'tls' and 'ClientConnection.Configuration.TLS'")
+  public var tlsConfiguration: TLSConfiguration? {
+    return nil
+  }
+}
+
+extension Server.Configuration {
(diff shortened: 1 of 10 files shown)


Sources/GRPC/Shims.swift
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+import NIOSSL
+
+// This file contains shims to notify users of API changes between v1.0.0-alpha.1 and v1.0.0.
+
+// TODO: Remove these shims before v1.0.0 is tagged.
+
+extension ClientConnection.Configuration {
+  @available(*, deprecated, message: "use 'tls' and 'ClientConnection.Configuration.TLS'")
```
kevints: You can use a `renamed: ` attribute here to get Xcode to suggest a migration fixit

glbrntt: It's a different type (albeit a very similar one) instead of a direct rename so this seemed more appropriate

Sources/GRPC/TLSConfiguration.swift
```diff
@@ -0,0 +1,195 @@
...
+      certificateVerification: CertificateVerification = .fullVerification,
+      hostnameOverride: String? = nil
+    ) {
+      self.configuration = .forClient(
+        minimumTLSVersion: .tlsv12,
+        certificateVerification: certificateVerification,
+        trustRoots: trustRoots,
+        certificateChain: certificateChain,
+        privateKey: privateKey,
+        applicationProtocols: GRPCApplicationProtocolIdentifier.allCases.map { $0.rawValue }
```
MrMage: Sorry, which identifiers does this include again?

Sources/GRPC/Shims.swift
glbrntt: We could, this makes it's easier to fix for anyone who is using the currently tagged version though.

Sources/GRPC/TLSConfiguration.swift
glbrntt: Off the top of my head: "h2" and "grpc-ext"

Sources/GRPC/TLSConfiguration.swift
```diff
@@ -0,0 +1,195 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import NIOSSL
+
+extension ClientConnection.Configuration {
+  /// TLS configuration for a `ClientConnection`.
+  ///
+  /// Note that this configuration is a subset of `NIOSSL.TLSConfiguration` where certain options
+  /// are removed from the users control to ensure the configuration complies with the gRPC
```
glbrntt:
```suggestion
  /// are removed from the user's control to ensure the configuration complies with the gRPC
```

PR #521, opened 2019-07-23
Make recording state in tests thread-safe

glbrntt:
Motivation:

When the tests are run with TSAN, it complained about
concurrent modification of states in the connectivity state
recording delegate.

Modifications:

Guard access to the recorded states via a lock.

Result:

TSAN doesn't complain when running the tests.


MrMage:
Is the test failure a flake?

glbrntt:
I think so, the `ClientTimeoutTests` have been pretty bad, especially on Linux recently. I'll look into this.

glbrntt:
@MrMage flake should be fixed in #524 (#522 and #523 failed for the same reason)

Issue #520, opened 2019-07-22
Client streaming RPC implementation: Cannot call context.responsePromise.fail() outside of the EventLoopFuture


vyshane:
### New Issue Checklist

- [x] Updated grpc-swift to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/grpc/grpc-swift/blob/master/CONTRIBUTING.md)
- [x] I searched for [existing GitHub issues](https://github.com/grpc/grpc-swift/issues)(especially, there are many [resolved issues related to SSL](https://github.com/grpc/grpc-swift/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+ssl))

### Issue Description

I'm using grpc-swift with NIO (gRPC Swift 1.0.0-alpha.1). When implementing a client streaming RPC, I get a fatal error if I don't call `context.responsePromise.fail(status)` inside of the `EventLoopFuture<(StreamEvent<EchoRequest>) -> Void>`. I have distilled the issue into the following.

#### The Protobuf

```protobuf
service ClientStreamingScenarios {
  rpc ClientStreamFailedPrecondition (stream EchoRequest) returns (Empty) {}
}

message EchoRequest {
  string message = 1;
}

message Empty {}
```

#### The RPC Implementation

```swift
class ClientStreamingTestsService: ClientStreamingScenariosProvider {

  func clientStreamFailedPrecondition(context: UnaryResponseCallContext<Empty>)
    -> EventLoopFuture<(StreamEvent<EchoRequest>) -> Void>
  {
    let status = GRPCStatus(code: .failedPrecondition, message: "Failed precondition message")
    context.responsePromise.fail(status)
    return context.eventLoop.makeSucceededFuture({ streamEvent in
      // Noop
    })
  }
}
```

#### The Error

When I exercise the RPC, I get:

```text
Fatal error: tried to decode as type RawGRPCServerResponsePart but found GRPCServerResponsePart<Empty> with contents other(GRPC.GRPCServerResponsePart<Empty>.status(GRPC.GRPCStatus)): file [...]/.build/checkouts/swift-nio/Sources/NIO/NIOAny.swift, line 200
```

### What Works

If I change the RPC implementation to the following it works:

```swift
  func clientStreamFailedPrecondition(context: UnaryResponseCallContext<Empty>)
    -> EventLoopFuture<(StreamEvent<EchoRequest>) -> Void>
  {
    return context.eventLoop.makeSucceededFuture({ streamEvent in
[...]

PR #523, opened 2019-07-23
Delay creating event observers for client streaming calls

glbrntt:
Motivation:

Providers for client and bidirectional streaming calls require
the user provide a future stream-event handler to handle requests
from the client. However, these methods get called as the pipeline
handling an incoming call is being configured, as these methods
also expose promises for response (for client streaming) and
call status (for bidirectional streaming) it is possible for these
to be fulfilled before the pipeline has been configured. Since
no handler is in place to deal with the promised types the server
will fatal error as the first handler in place will fail to unwrap
the promised type.

Modifications:

Delay the creation of event observers for client and
bidirectional streaming calls until their handlers have been added
to the pipeline.

Result:

Client streaming and bidirectional streaming calls can fulfill
their response and status promises outside of their stream handlers.


glbrntt:
Resolves #520

MrMage:
Do you think we could set the observer factories to nil once they are called (or if we determine that we won't need to call them at all), as a precaution against retain cycles?

glbrntt:
Sure, that's a good idea!

PR #528, opened 2019-07-24
Fallback to the hostname for TLS if a hostname override isn't provided

glbrntt:
Motivation:

TLS does not work if a hostname override was not provided and hostname
verification is enabled as only the override was provided to the
NIOSSLClientHandler.

Modifications:

Provide the SSL handler with the override, if one is provided, or the
hostname otherwise.

Result:

Fewer TLS failures.


MrMage:
Do we need tests for this? (Just asking.)

Sources/GRPC/ClientConnection.swift
```diff
@@ -325,21 +326,33 @@ extension ClientConnection {
...
+  /// - Parameter timeout: The connection timeout in seconds.
   private class func makeBootstrap(
     configuration: Configuration,
     group: EventLoopGroup,
     timeout: TimeInterval?,
     logger: Logger
   ) -> ClientBootstrapProtocol {
+    // Provide a server hostname if we're using TLS. Prefer the override.
+    let serverHostname: String? = configuration.tls.map {
+      if let override = $0.hostnameOverride {
```
kevints: nit: instead of shadowing the `override` keyword here does it make sense to call this hostnameOverride (it's legal in this context but GitHub's syntax coloring is confused)?

Sources/GRPC/ClientConnection.swift
```diff
@@ -325,21 +326,33 @@ extension ClientConnection {
...
+        logger.debug("using hostname override for TLS")
```
kevints: nit: include the value of the hostname override as logger metadata here

Sources/GRPC/ClientConnection.swift
```diff
@@ -325,21 +326,33 @@ extension ClientConnection {
...
+        return override
+      } else {
+        logger.debug("using host connection target for TLS")
```
kevints: nit: include the value of host here as logger metadata?

glbrntt:
@MrMage we probably should add tests here, yes.

Sources/GRPC/ClientConnection.swift
glbrntt: Sure, it's a better name anyway!

Issue #519, opened 2019-07-22
NIO client declares itself as `.ready` too soon


glbrntt:
### Issue Description

Issue #517 indicated that we can have an infinite reconnection loop if the connection closes immediately after the channel becomes active. We should transition the connection state to `.ready` after we have received e.g. the initial settings frame on the connection.

MrMage:
Should be fixed in https://github.com/grpc/grpc-swift/releases/tag/1.0.0-alpha.2.

PR #529, opened 2019-07-25
Delay marking connection as .ready until receiving initial settings f…

glbrntt:
…rame

Motivation:

It is possible for a connection to be immediately closed by the remote
after a successful TLS handshake. In this case, the connection would be
marked as ready and any backoff would be reset. That is, a reconnection
attempt would be made immediately and the cycle would repeat. The gRPC
core lib suggests resetting backoff after the initial settings frame has
been made.

Modifications:

Add a handler which on observing the initial settings frame:
1. marks the connectivity as ready and,
2. removes itself from the pipeline.

The client connection diagram was also updated to reflect this.

Result:

The connectivity state will only be marked as ready once the settings
frame has been received from the peer.

### DIFF
--- a/Sources/GRPC/ClientConnection.swift
+++ b/Sources/GRPC/ClientConnection.swift
@@ -27,9 +27,20 @@ import Logging
 /// The connection is initially setup with a handler to verify that TLS was established
 /// successfully (assuming TLS is being used).
 ///
-///                          ▲                       |
-///                HTTP2Frame│                       │HTTP2Frame
-///                        ┌─┴───────────────────────▼─┐
+///               ┌──────────────────────────┐
+///               │  DelegatingErrorHandler  │
+///               └──────────▲───────────────┘
+///                HTTP2Frame│
+///               ┌──────────┴───────────────┐
+///               │ SettingsObservingHandler │
+///               └──────────▲───────────────┘
+///                HTTP2Frame│
+///                          │                ⠇ ⠇   ⠇ ⠇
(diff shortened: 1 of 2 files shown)


glbrntt:
Fixes #519

Sources/GRPC/ClientConnection.swift
```diff
@@ -27,9 +27,20 @@ import Logging
 /// The connection is initially setup with a handler to verify that TLS was established
 /// successfully (assuming TLS is being used).
 ///
-///                          ▲                       |
-///                HTTP2Frame│                       │HTTP2Frame
-///                        ┌─┴───────────────────────▼─┐
+///               ┌──────────────────────────┐
```
MrMage: Should we also add the source files for these diagrams to version control at some point?

glbrntt:
> Is this something we should have tests for?

Ideally, yes. Realistically I think the only thing we can test for here is that the client _doesn't_ become ready before receiving the settings frame and I'm not sure how much value there is in that, WDYT?

MrMage:
Yeah, I wasn't sure about that myself, so let's not test this.

Once you update the conflicts I'll merge :-) And if you do that in the next hour and ping me I can even cut a release ;-)

Sources/GRPC/ClientConnection.swift
glbrntt: This is it I'm afraid! I created them with Monodraw originally and have just updated them manually since 😬

Pretty sure you can paste ASCII-art back into Monodraw though so it shouldn't be a problem.

Issue #534, opened 2019-07-29
Archive failed in Xcode 11 ("nio" package added) with arithmetic operations overflow.


Krivoblotsky:
### New Issue Checklist

- [x] I read the [Contribution Guidelines](https://github.com/grpc/grpc-swift/blob/master/CONTRIBUTING.md)
- [x] I searched for [existing GitHub issues](https://github.com/grpc/grpc-swift/issues)(especially, there are many [resolved issues related to SSL](https://github.com/grpc/grpc-swift/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+ssl))

### Issue Description
I have added Swift Package ("nio" branch) with Xcode 11 intergated SPM. Once I do "archive" with selected iOS-platform, the build fails with "arithmetic operations overflow" in `GRPCTimeoutUnit`.
I suppose this is 32-bit related issue because by default Xcode adds armv7 architecture to the build. 
I'm on 10.15, but I get same results on 10.14.
Archive is ok for macOS builds (i386/x86_64)

<img width="807" alt="Screen Shot 2019-07-29 at 9 59 04 PM" src="https://user-images.githubusercontent.com/1411778/62075371-8f6caa00-b24d-11e9-9381-b2fd888c92a9.png">

### Environment

<details>
  <pre>
  
| Key                           | Value                                                         |
| ----------------------------- | --------------------------------------------------------------|
| OS Version                    |   10.15                                                       |
| Swift Version                 |   5.1                                                       |
| Xcode Version                 |   11.0 beta (11M336w)                                                       |
  </pre>
</details>

weissi:
oh yes, the type for `asNanoseconds` should be `Int64` not `Int`, or else it won't work on 32bit platforms.

Krivoblotsky:
Thanks. I’ll open a PR.

Krivoblotsky:
Please, see the PR https://github.com/grpc/grpc-swift/pull/535

MrMage:
Should be fixed by #535.

PR #535, opened 2019-07-30
Fix building for 32-bit architectures. Fix Int overflow.

Krivoblotsky:
## What
Fix possible `Int` overflow when building for 32-bit architectures by increasing the size to `Int64`.
See related issue https://github.com/grpc/grpc-swift/issues/534

## Why
The build is failing and overflow issues are possible

## Affected Areas
Timestamp generation in `GRPCTimeout.swift`

### DIFF
--- a/Sources/GRPC/GRPCTimeout.swift
+++ b/Sources/GRPC/GRPCTimeout.swift
@@ -46,7 +46,7 @@ public struct GRPCTimeout: CustomStringConvertible, Equatable {
 
     // See "Timeout" in https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
     let description = "\(amount)\(unit.rawValue)"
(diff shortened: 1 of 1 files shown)


MrMage:
Thank you!

saudeoncovr:
@MrMage are we able to get a new alpha tagged release that includes this and #538?

Unfortunately our CI is failing to compile due to the Int overflow error.

MrMage:
> @MrMage are we able to get a new `1.0.0-alpha.3` tagged release that includes this and #538? Or even @glbrntt?
> 
> Unfortunately our CI is failing to compile due to the Int overflow error.

That might be possible. Until then, could you simply point your `Package.swift` to the `nio` branch instead of a specific version?

glbrntt:
Fine with me if you don't mind cutting it @MrMage :)

MrMage:
Done: https://github.com/grpc/grpc-swift/releases/tag/1.0.0-alpha.3

PR #538, opened 2019-08-02
Add sendMessages and flush to StreamingRequestClientCall

glbrntt:
Motivation:

The API for `StreamingRequestClientCall` does not make it easy for users
to send a batch of messages and then flush at the end.

Modifications:

Add methods to `StreamingRequestClientCall` to:
- send messages, and
- manually flush.

Result:

Users can send a batch of messages and manually flush for calls which
stream from the client.


glbrntt:
Resolves #538

Sources/GRPC/ClientCalls/ClientCall.swift
```diff
@@ -131,6 +153,33 @@ extension StreamingRequestClientCall {
...
+      }
+      return EventLoopFuture.andAllSucceed(writeFutures, on: channel.eventLoop)
+    }
+  }
+
+  public func sendMessages<S: Sequence>(_ messages: S, promise: EventLoopPromise<Void>?, flush: Bool = true) where S.Element == RequestMessage {
+    if let promise = promise {
+      self.sendMessages(messages, flush: flush).cascade(to: promise)
+    } else {
+      self.subchannel.whenSuccess { channel in
```
MrMage: Could we replace this with e.g. `self.sendMessages(messages, flush: flush).then { *do nothing* }`?

Tests/GRPCTests/StreamingRequestClientCallTests.swift
```diff
@@ -0,0 +1,77 @@
...
+    // Send more messages, but don't flush.
+    let secondBatchNotReceived = self.expectation(description: "second batch not received")
+    secondBatchNotReceived.isInverted = true
+    counter.expectation = secondBatchNotReceived
+
+    let secondBatch = (0..<3).map { Echo_EchoRequest(text: "\($0)") }
+    update.sendMessages(secondBatch, promise: nil, flush: false)
+
+    // Wait and check that the expectation hasn't been fulfilled (because we haven't flushed).
+    self.wait(for: [secondBatchNotReceived], timeout: 0.5)
```
MrMage: Can this timeout be shortened, to avoid delaying test runtime by 0.5 s?

Sources/GRPC/ClientCalls/ClientCall.swift
```diff
@@ -76,6 +76,25 @@ public protocol StreamingRequestClientCall: ClientCall {
...
+
+  /// Sends a sequence of messages to the service.
+  ///
+  /// - Important: Callers must terminate the stream of messages by calling `sendEnd()` or `sendEnd(promise:)`.
+  ///
+  /// - Parameters:
+  ///   - messages: The sequence of messages to send.
+  ///   - promise: A promise to be fulfilled when all messages have been sent successfully.
+  ///   - flush: Whether the buffer should be flushed after writing the message.
+  func sendMessages<S: Sequence>(_ messages: S, promise: EventLoopPromise<Void>?, flush: Bool) where S.Element == RequestMessage
```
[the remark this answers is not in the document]
glbrntt: I don't think so: in the default implementations we default `flush` to `true`. The usual convention is to have parameters with defaults after those without. Also use this ordering for `sendMessage`.

Sources/GRPC/ClientCalls/ClientCall.swift
```diff
@@ -131,6 +153,33 @@ extension StreamingRequestClientCall {
     }
   }
 
+  public func sendMessages<S: Sequence>(_ messages: S, flush: Bool = true) -> EventLoopFuture<Void> where S.Element == RequestMessage {
+    self.subchannel.flatMap { channel in
```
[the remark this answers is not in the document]
glbrntt: Good catch!

```suggestion
    return self.subchannel.flatMap { channel in
```

Tests/GRPCTests/StreamingRequestClientCallTests.swift
glbrntt: Sure!

```suggestion
    self.wait(for: [secondBatchNotReceived], timeout: 0.1)
```

Sources/GRPC/ClientCalls/ClientCall.swift
glbrntt: Why? Doing so would create an additional `messages.count` callbacks doing that.

glbrntt:
> Should we have the same methods on the server as well?

Potentially, but I think it would come with a broader set of changes: at the moment there's really no control for the user in how they send messages, only `sendResponse(message:) -> EventLoopFuture<Void>`.

MrMage:
Thinking about this more, what's the benefit of the extra `flush:` argument over calling `flush()`? Sorry if I missed that before.

I'm a bit worried that having both future-returning and promise-taking methods that both have a `flush:` argument might start to get a bit confusing and violate the single responsibility of just sending a message.

MrMage:
LGTM, save for my question about `flush:` before.

glbrntt:
Hmm I see your point.

We should keep the future/promise APIs. But I think we can do away with `flush()` and `flush:` and just always flush in `sendMessage`/`sendMessages`.

I think it's a reasonable assumption to make that most users will send a single message at a time in which case I think they'd (probably) want to flush as well. If they don't want to flush when writing a message then they can keep their own buffer of messages and then send them with `sendMessages`.

The API would then be something like:
- `sendMessage(_ message: RequestMessage) -> EventLoopFuture<Void>`
- `sendMessage(_ message: RequestMessage, promise: EventLoopPromise<Void>?)`
- `sendMessages<S: Sequence>(_ messages: S) -> EventLoopFuture<Void> where S.Element == RequestMessage`
- `sendMessages<S: Sequence>(_ messages: S, promise: EventLoopPromise<Void>?) where S.Element == RequestMessage`

What do you think?

MrMage:
> We should keep the future/promise APIs. But I think we can do away with `flush()` and `flush:` and just always flush in `sendMessage`/`sendMessages`.

Agreed.

> The API would then be something like:

Looks good to me :-)

Tests/GRPCTests/StreamingRequestClientCallTests.swift
```diff
@@ -0,0 +1,58 @@
...
+
+    let statusReceived = self.expectation(description: "status received")
+    update.status.map { $0.code }.assertEqual(.ok, fulfill: statusReceived)
+
+    // End the call.
+    update.sendEnd(promise: nil)
+
+    self.wait(for: [statusReceived], timeout: 0.5)
+  }
+
```
MrMage:
```suggestion
```

MrMage: Remove this newline?

Issue #530, opened 2019-07-25
Callback not getting called from ClientNetworkMonitor cellularDidChange method


tikekar:
### New Issue Checklist

- [ ] Updated grpc-swift to the latest version
- [ X] I read the [Contribution Guidelines](https://github.com/grpc/grpc-swift/blob/master/CONTRIBUTING.md)
- [X ] I searched for [existing GitHub issues](https://github.com/grpc/grpc-swift/issues)(especially, there are many [resolved issues related to SSL](https://github.com/grpc/grpc-swift/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+ssl))

### Issue Description
When I turn the airplane mode on and off from wifi, the cellularDidChange method of ClientNetworkMonitor class gets called. But the callback "self.callback(State..." doesn't get called. If I remove the if() statement surrounding it, it gets called.
```
// commenting the if statement works.
// if let newCellularName = newCellularName, self.cellularName != newCellularName {
        self.cellularName = newCellularName
        self.callback(State(lastChange: .cellularTechnology(technology: newCellularName ?? ""),
                            isReachable: self.isReachable ?? false))
      // }
```

##### Complete output when running grpc-swift, including the stack trace and command used

<details>
  <pre>[INSERT OUTPUT HERE]</pre>
</details>

### Environment

<details>
  <pre>
  
| Key                           | Value                                                         |
| ----------------------------- | --------------------------------------------------------------|
| OS Version                    |   x.x.x                                                       |
| Swift Version                 |   x.x.x                                                       |
| Xcode Version                 |   x.x.x                                                       |
| gRPC-Swift Version            |   x.x.x                                                       |
| gRPC-Swift Version            |   x.x.x                                                       |
| protoc Version                |   x.x.x                                                       |
| protoc-gen-swift Version      |   x.x.x                                                       |
| protoc-gen-swiftgrpc Version  |   x.x.x                                                       |
  </pre>
</details>

rebello95:
Thanks for filing, @tikekar. What problems is this causing for your application? This API is supposed to be used for restarting a channel when we know it needs to be switched to another reachable connection. In this case, there are no reachable connections.

From what I've seen in past experimentation, disabling airplane mode will:
- allow the connection to automatically resume if it hasn't timed out
- receive data that was missed while airplane mode was enabled (if the connection hasn't timed out)
- send a notification through this interface notifying you of a new connection

Are you not seeing this behavior?

[...]

PR #543, opened 2019-08-06
Notify state change for initial state

shoheiyokoyama:
Hi. :hand:

I fixed state notification logic in `ClientNetworkMonitor`.

In the initial state of the `ClientNetworkMonitor` instance, the internal `reachabilityDidChange(with:)` is called, but if `isUsingWifi` or `isReachable` is nil, `callBack` is not called.  
In addition, `SCNetworkReachabilityCallBack` will then notify `SCNetworkReachabilityFlags`, but `callBack` will not be called because of comparing the values.
Was this implementation intentionally?

```swift
let notifyForWifi = self.isUsingWifi != nil && self.isUsingWifi != isUsingWifi
let notifyForReachable = self.isReachable != nil && self.isReachable != isReachable
```

I think `callback` should be called anytime because `SCNetworkReachabilitySetCallback` notifies the first state. 
[Reachability.swift](https://github.com/ashleymills/Reachability.swift), a typical Reachability framework, also execute status notifications for the first state. Please [see](https://github.com/ashleymills/Reachability.swift/blob/master/Sources/Reachability.swift) here for the details.

### DIFF
--- a/Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
+++ b/Sources/SwiftGRPC/Core/ClientNetworkMonitor.swift
@@ -149,8 +149,8 @@ open class ClientNetworkMonitor {
       let isUsingWifi = !flags.contains(.isWWAN)
       let isReachable = flags.contains(.reachable)
 
-      let notifyForWifi = self.isUsingWifi != nil && self.isUsingWifi != isUsingWifi
-      let notifyForReachable = self.isReachable != nil && self.isReachable != isReachable
+      let notifyForWifi = self.isUsingWifi != isUsingWifi
+      let notifyForReachable = self.isReachable != isReachable
 
       self.isUsingWifi = isUsingWifi
       self.isReachable = isReachable


rebello95:
Thank you for the contribution! I posted a response on a similar issue this morning [here](https://github.com/grpc/grpc-swift/issues/530#issuecomment-520158681).

The goal of the `ClientNetworkMonitor` was to notify consumers when a gRPC channel needs to be restarted based on device conditions. Avoiding notifying on initial state was originally done by design under the assumption that consumers would only want to know about _future_ changes in order to restart the channel, but it could be that we missed a use case. Can you elaborate on the issue with gRPC that you're seeing by the network monitor not being called with the initial state?

shoheiyokoyama:
@rebello95 

> Can you elaborate on the issue with gRPC that you're seeing by the network monitor not being called with the initial state?

For example, observe to `ClientNetworkMonitor.State` when the device is airplane mode and wifi is off. In this initial state, `isUsingWifi` is `true` because `SCNetworkReachabilityFlags` includes `isWWAN`. At this time `callback` isn't invoked by optional evaluation.

```swift
let isUsingWifi = !flags.contains(.isWWAN) // true 🤔 
let isReachable = flags.contains(.reachable) // false

// notifyForWifi and notifyForReachable are false 
// because self.isUsingWifi and self.isReachable are Optional.none
let notifyForWifi = self.isUsingWifi != nil && self.isUsingWifi != isUsingWifi
let notifyForReachable = self.isReachable != nil && self.isReachable != isReachable
```

Next, turn on wifi while airplane mode. `SCNetworkReachabilityCallBack` is called immediately and `isUsingWifi` and `isReachable` are true. Therefore, `notifyForReachable` becomes true because `isReachable` is changed,  and `State (lastChange: .reachability (isReachable: true), isReachable: true)` is sent. However, since `isUsingWifi` hasn't changed, it isn't possible to know whether the iPhone is cellular or wifi.

```swift
let isUsingWifi = !flags.contains(.isWWAN) // true
let isReachable = flags.contains(.reachable) // true

// notifyForWifi is false 
// because self.isUsingWifi and isUsingWifi are true
let notifyForWifi = self.isUsingWifi != nil && self.isUsingWifi != isUsingWifi
let notifyForReachable = self.isReachable != nil && self.isReachable != isReachable
```

If there is no evaluation of optional, the first notification is invalid because `isReachable` is `false` and `isUsingWifi` is `true`😵. However, in the second notification, you can know that the iPhone turn on wifi because `isReachable` is `true` and `isUsingWifi` is true.

Device: iPhone X
OS: 12.3.1

> The goal of the ClientNetworkMonitor was to notify consumers when a gRPC channel needs to be restarted based on device conditions. 
Avoiding notifying on initial state was originally done by design under the assumption that consumers would only want to know about future changes in order to restart the channel, 

I think it should be left to the consumer to restart the gRPC channel so the device state should always be sent so that the consumer can respond flexibly to the various use case.
I hope that more general use cases are envisioned so that more users can use this wonderful implementation 👍 .

MrMage:
Closing this due to inactivity for now; please let us know if you would like to pick it up again.

PR #546, opened 2019-08-13
Fix typos in README

glbrntt:
Motivation:

Parts of the README were outdated and incorrect.

Modifications:

Update the README to correct known typographic errors. Update the
section on TLS since it was incorrect.

Removed old TLS code the README referred to but could not be used.

Result:

More helpful README


MrMage:
Thanks to Daniel Dunbar for reporting these!

PR #554, opened 2019-08-15
Make the backoff infinite

glbrntt:
Motivation:

The gRPC backoff continues ad infinitum (it is capped at a maximum
timeout instead of growing exponentially forever). Our backoff
terminates once it reaches the maximum backoff.

Modifications:

- Remove the upper limit on backoff generation so it is produced infinitely.
- Remove a test which is no longer relevant.
- Enable backoff by default

Result:

Client will try connecting until the user calls `close()`


MrMage:
Should we at least keep the option for users to stop retrying after the last backoff try? (Open to removing this now and re-adding this later only if it actually gets requested.)

glbrntt:
> Should we at least keep the option for users to stop retrying after the last backoff try? (Open to removing this now and re-adding this later only if it actually gets requested.)

I'd say remove and re-add if it gets requested.

PR #557, opened 2019-08-19
Add the connection backoff interop test

glbrntt:
Motivation:

We'd like our implementation to be compliant with the gRPC
specification. Passing the interop tests is an important part of this,
yet we don't have an implementation of the connectivity backoff test.

Modifications:

- Implement the connectivity backoff test.
- Fixes a bug where call timeouts would not be respected if the
  connection had not been established.
- Fixes a bug where the connection could be re-established after calling
  shutdown.

Result:

Our connection backoff code has better test coverage; fewer bugs.


glbrntt:
We should yes, I was going to set up the CI to run all interop tests against a different implementation (probably C++ since it's recommended for the backoff test). I'll look into this and post a separate PR.

PR #561, opened 2019-08-22
Update CI to run interoperability tests.

glbrntt:
Motivation:

We don't run the interoperability tests in CI at the moment, we should
run them against a different implementation to help ensure we don't
regress.

Modifications:

- Add two additional CI jobs to run the interoperability tests (which do
  not run the other tests).
- Update the Travis setup so the steps run are in .travis-script.sh
  instead of inline in .travis.yml
- Update .travis-install to conditionally install Bazel from source and
  build the gRPC C++ interop testing servers
- Update the Makefile to make it more configurable (i.e. 'swift' can be
  set to make it easier to run with different toolchains via xcrun)

Result:

- Interop tests will be run in CI in separate build jobs


.travis.yml
```diff
@@ -13,22 +13,49 @@
...
+      dist: bionic
+      script: ./.travis-script.sh
+      env: RUN_INTEROP_TESTS=true
+    - <<: *interop_tests
+      name: "macOS (Xcode 10.3)"
+      os: osx
+      osx_image: xcode10.3
+
+stages:
+  # Only run stage for PRs
```
MrMage: What other types are there? Should we maybe _always_ run tests?

.travis.yml
```diff
@@ -38,14 +65,11 @@ addons:
...
-
-script:
-  - echo -en 'travis_fold:start:script.environment\\r' && export PATH=$HOME/local/bin:$PATH && export LD_LIBRARY_PATH=$HOME/local/lib && echo -en 'travis_fold:end:script.environment\\r'
-  - echo -en 'travis_fold:start:swift.resolve-deps\\r' && swift package -v resolve && echo -en 'travis_fold:end:swift.resolve-deps\\r'
-  - echo -en 'travis_fold:start:make.all\\r' && make all && echo -en 'travis_fold:end:make.all\\r'
-  - echo -en 'travis_fold:start:make.test_generate_linuxmain\\r' && make test-generate-linuxmain && echo -en 'travis_fold:end:make.test_generate_linuxmain\\r'
-  - echo -en 'travis_fold:start:make.test\\r' && make test && echo -en 'travis_fold:end:make.test\\r'
-  - echo -en 'travis_fold:start:make.test_plugin\\r' && make test-plugin && echo -en 'travis_fold:end:make.test_plugin\\r'
-  - echo -en 'travis_fold:start:make.project\\r' && if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then make project; fi && echo -en 'travis_fold:end:make.project\\r'
+script: ./.travis-script.sh
```
MrMage: It's much nicer to have these commands outsourced :-)

Makefile
```diff
@@ -1,40 +1,87 @@
...
+SWIFT_PACKAGE:=${SWIFT} package ${SWIFT_FLAGS}
+
+# Name of generated xcodeproj
+XCODEPROJ:=GRPC.xcodeproj
+
+### Package and plugin build targets ###########################################
+
 all:
-	swift build
-	cp .build/debug/protoc-gen-swift .
```
MrMage: This means that we won't litter the main directory with binaries anymore, right? Nice!

Makefile
```diff
@@ -1,40 +1,87 @@
...
-endif
+### Misc. ######################################################################
 
 clean:
 	-rm -rf Packages
-	-rm -rf .build build
-	-rm -rf GRPC.xcodeproj
-	-rm -rf Package.pins Package.resolved
-	-rm -rf protoc-gen-swift protoc-gen-swiftgrpc
+	-rm -rf ${SWIFT_BUILD_PATH}
```
MrMage: Did you try running `make clean` to ensure that this is complete?

.travis.yml
glbrntt: It's only: `push`, `pull_request`, `api`, and `cron` (https://docs.travis-ci.com/user/conditions-v1#integration). 

Actually running them when we merge is a good idea: a clean merge doesn't necessarily mean the outcome was correct.

.travis.yml
glbrntt: _much_ easier to read!

Makefile
glbrntt: Yup!

Makefile
glbrntt: Yes, although I omitted `.swiftpm` from this (i.e. support for Swift PM in Xcode 11 when you're not using a generated `xcodeproj`). The rationale being that it's generated and managed by Xcode and not us.

.travis.yml
MrMage: Should we run this for both pushes and PRs then?

PR #567, opened 2019-08-27
Split the Echo example into model, implementation and runtime

glbrntt:
Motivation:

The generated code and implementation for the Echo example is used as
part of the example and as well as in the test suite. To make the
example more obvious we should split it into its constituent parts: the
model (i.e. the definition and generated code), the implementation (of
the server), and the runtime (CLI for client and server).

Modifications:

- Split the Echo example into three modules: `EchoModel` (generated code),
  `EchoImplementation` (server implementation), and `EchoRuntime` (CLI for
  client and server).
- Update the `Makefile` to target the updated paths
- Add missing imports in tests
- Make the generated code `public` (instead of `internal`) since we're
  no longer symlinking into the same module for tests

Result:

- No more symlinks for Echo
- Structure of example is more straightforward


MrMage:
Looks good, but do you think we could merge `EchoModel` and `EchoImplementation`? Shouldn't be much overhead, and saves us one extra target.

glbrntt:
I think it's better to keep them separate; it makes the separation of generation and implementation _really_ obvious (i.e. for someone who's new to gRPC). WDYT?

MrMage:
Fine with me :-)

PR #568, opened 2019-08-28
Update Swift packages

glbrntt:
Motivation:

One of interop test cases may hang prior to NIO 2.7.1 causing the CI to
fail.

Modifications:

Update Package.resolved so that NIO 2.7.1, which contains a fix for the
aforementioned bug, is used.

Result:

Interop tests are less likely to hang.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -33,8 +33,8 @@
         "repositoryURL": "https://github.com/apple/swift-nio.git",
         "state": {
           "branch": null,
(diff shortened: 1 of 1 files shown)


glbrntt:
See:
- https://travis-ci.org/grpc/grpc-swift/jobs/577764439
- https://github.com/apple/swift-nio/releases/tag/2.7.1
- https://github.com/apple/swift-nio/pull/1121

PR #569, opened 2019-08-28
Allow timeouts to be rounded so they may comply with the spec

glbrntt:
Motivation:

The gRPC spec dictates that timeouts may be no longer than 8 digits
long. As such methods for creating `GRPCTimeout`s all throw if the given
timeout is not valid. In the vast majority of cases this is not actually
an issue and having to handle the exception can be a pain.

Modifications:

Add non-throwing factory methods to `GRPCTimeout` which round the
timeout so that it may be encoded over the wire.

Result:

Timeouts are easier to create.


Package.resolved
```diff
@@ -24,8 +24,8 @@
         "repositoryURL": "https://github.com/apple/swift-log",
         "state": {
           "branch": null,
-          "revision": "e8aabbe95db22e064ad42f1a4a9f8982664c70ed",
-          "version": "1.1.1"
+          "revision": "f4240bf022a69815241a883c03645444b58ac553",
```
MrMage: Regression?

glbrntt: Oops, from a rebase I think; will revert.

PR #571, opened 2019-08-30
Update Travis script

glbrntt:
Motivation:

Recently the CLI parsing library was dropped in favour of our own CLI
parsing. However, the CI script for running interop tests was not
updated.

Modifications:

Update the CI script to run interop tests. Also update an off-by-one
error in the server CLI parsing that forced users to provide an optional
argument.

Result:

CI shouldn't error because of CLI failures.


glbrntt:
sorry about this one! 🤦‍♂️

PR #572, opened 2019-08-30
Separate trailers from GRPCStatus

glbrntt:
Motivation:

`GRPCStatus` contains the status code and message of the RPC along with
any trailing metadata returned from the call. However, trailers are not
semantically important to the status, they are just returned at the same
point in time (if the status originates from the server).

Modifications:

- Remove trailers from `GRPCStatus`
- Expand the `status` case of `RawGRPCClientResponsePart` and
  `RawGRPCServerResponsePart` from `.status(GRPCStatus)` to
  `.statusAndTrailers(GRPCStatus, HTTPHeaders?)`
- Add `Equatable` conformance to `GRPCStatus`
- Add `CustomStringConvertible` conformance to `GRPCStatus`
- Remove test helpers which checked for `GRPCStatus` status code and
  message equality (i.e. ignoring trailers)

Result:

- `GRPCStatus` makes more sense as a type
- `GRPCStatus` is equatable
- client streaming and bidirectional streaming calls on the server can
  no longer send trailing metadata via `GRPCStatus`


glbrntt:
> client streaming and bidirectional streaming calls on the server can
no longer send trailing metadata via GRPCStatus

@MrMage re: this point I think we need to reconsider the generated handlers for the server, we couldn't send trailers for unary and server streaming calls before this change and we can't send initial metadata for any of them.

Sources/GRPC/ClientResponseChannelHandler.swift
```diff
@@ -108,10 +111,11 @@ internal class ClientResponseChannelHandler<ResponseMessage: Message>: ChannelIn
   /// are failed with the given status.
   ///
   /// - Parameter status: the status to observe.
-  internal func observeStatus(_ status: GRPCStatus) {
+  internal func observeStatus(_ status: GRPCStatus, trailingMetadata: HTTPHeaders?) {
     if status.code != .ok {
       self.initialMetadataPromise.fail(status)
     }
+    self.trailingMetadataPromise.succeed(trailingMetadata ?? HTTPHeaders())
```
MrMage: Should we instead make the promise's result optional as well?

MrMage:
> @MrMage re: this point I think we need to reconsider the generated handlers for the server, we couldn't send trailers for unary and server streaming calls before this change and we can't send initial metadata for any of them.

Right. One option _could_ be a similar solution to what we already have for [returning trailing metadata on unary-response calls](https://github.com/grpc/grpc-swift/blob/f5178bbbace2db7a4046c3a2fdc22c59269ac99e/Sources/GRPC/ServerCallContexts/UnaryResponseCallContext.swift#L35) — a property on the context that can be filled with the desired leading metadata to return. _However,_ I'm not even sure if the request proto is guaranteed to be sent before the server responds with the initial metadata to "accept" the call. If not, we can't expect the user's call handlers to be called before the initial metadata is being sent. In that case, maybe it would be easier to offer a global interceptor (i.e. for all methods) that is provided the client's metadata headers (including the method name, of course) and returns any initial metadata that should be returned to the client?

MrMage:
Speaking of this, I think https://github.com/grpc/grpc-swift/blob/f5178bbbace2db7a4046c3a2fdc22c59269ac99e/Sources/GRPC/ServerCallContexts/UnaryResponseCallContext.swift#L35 should be extended with another field for the trailing metadata.

Sources/GRPC/ClientResponseChannelHandler.swift
glbrntt: I don't think so: the server will send trailers if it terminates the call (containing the status code at a minimum) so we only have `nil` here if the client terminates the call (by canceling it, timing out, or some other client side error).

glbrntt:
>  _However,_ I'm not even sure if the request proto is guaranteed to be sent before the server responds with the initial metadata to "accept" the call. 

It _isn't_ guaranteed.

> If not, we can't expect the user's call handlers to be called before the initial metadata is being sent. In that case, maybe it would be easier to offer a global interceptor (i.e. for all methods) that is provided the client's metadata headers (including the method name, of course) and returns any initial metadata that should be returned to the client?

I think that would work and shouldn't be too difficult to plug in. I think this will unlock a few more server-side interop tests for us as well.

MrMage:
> I think that would work and shouldn't be too difficult to plug in. I think this will unlock a few more server-side interop tests for us as well.

👍 I think this could also be extended into a full-blown middleware system later on.

Sources/GRPC/ClientResponseChannelHandler.swift
MrMage: But what's the advantage of returning an empty `HTTPHeaders` struct to the client user vs. just returning nil?

glbrntt: It's easier to deal with an empty collection than a `nil` one.

MrMage: Fair enough :-)

glbrntt:
> LGTM, but should we maybe document that `context.trailingMetadata` should be set before `statusPromise`/`responsePromise` is fulfilled, if the user wants to send trailing metadata?

Good point, done.

MrMage:
LGTM; merge at your convenience.

PR #573, opened 2019-09-03
Update minimum Swift NIO version

glbrntt:
Motivation:

The minimum NIO version is quite old and there have been a number of bug
fixes since which have fixed issues we've seen in gRPC. We should
enforce users to use a newer version so that they will not see these
issues.

Modifications:

- Update our minimum NIO version to 2.7.1.
- Also bump gRPC alpha version in README

Result:

NIO is more up-to-date.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -29,7 +29,7 @@ let package = Package(
   dependencies: [
     // GRPC dependencies:
     // Main SwiftNIO package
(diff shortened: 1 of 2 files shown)


glbrntt:
@MrMage mind if we cut another release following this PR?

MrMage:
Feel free to merge and cut a release :-)

Issue #574, opened 2019-09-03
Docker documentation is stale and inconsistent


Kirow:
### Question Checklist

- [x] Updated grpc-swift to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/grpc/grpc-swift/blob/master/CONTRIBUTING.md)
- [x] I read [README](https://github.com/grpc/grpc-swift/blob/master/README.md)
- [x] I searched for [existing GitHub issues](https://github.com/grpc/grpc-swift/issues) (especially, there are many [resolved issues related to SSL](https://github.com/grpc/grpc-swift/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+ssl))

### Question Subject
Running `protoc` from docker container.

#### Question Description
I've tried to follow [this guide](https://github.com/grpc/grpc-swift/blob/master/DOCKER.md), but it seems that image is outdated (swift version I guess).

Then I tried to build image from [Dockerfile](https://github.com/grpc/grpc-swift/blob/master/Docker/Dockerfile), but it also has missing dependencies (libz, curl, libssl). I've made some changes to let it work.

```bash
ENV PROTOC_VERSION=3.9.1 #updated

RUN apt-get -q update \
    && apt-get install curl \ #new
    && apt-get install libz-dev \ #new
    && apt-get install libssl-dev \ #new
    && apt-get -q install -y unzip libnghttp2-dev \
    && rm -r /var/lib/apt/lists/*
```

Please correct  this.

glbrntt:
@Kirow if the above changes to the Dockerfile fix this would you mind submitting a PR?

Kirow:
@glbrntt ok, I will do PR for Dockerfile. But I'm not sure what to do with [DOCKER.md](https://github.com/grpc/grpc-swift/blob/master/DOCKER.md).  How about to remove it and then update [README.md](https://github.com/grpc/grpc-swift/blob/master/README.md) to reference [Docker/README.md](https://github.com/grpc/grpc-swift/blob/master/Docker/README.md)?

glbrntt:
@Kirow apologies for the delay. 

If I remember correctly the Dockerfile just installs the `protoc` plugins, [DOCKER.md](DOCKER.md) leads you to a built version of `grpc-swift` to play around with. I think we should also update the instructions in [DOCKER.md](DOCKER.md) to include the missing packages.

Kirow:
@glbrntt 

There are few places in the docs about the same:
[...]

PR #579, opened 2019-09-06
Updated Dockerfile and docs

Kirow:
Following the context of #574 

- Updated `Dockerfile` to use multistage build process
  - `development` target - to run/build `grpc/swift`
  - `protoc-runner` target - to run `protoc` with swift plugins
- removed `DOCKER.md` 
- updated `Docker/README.md`
- updated reference in `README.md`

I've tried to build `protoc-runner` from `alpine`, but it seems that I've lost all my teeth while trying to run swift binaries there. Will be awesome if someone will make it work in smallest possible environment. But even in current implementation `protoc-runner` will need 206MB compared to 1.66GB `development` image.


Docker/README.md
```diff
@@ -1,13 +1,86 @@
...
+
+## Protoc-runner
+
+Build docker image with the following command:
+
+```bash
+docker build -t protoc-runner .
+```
+
+If you plan only to use `protoc` with plugins - run `docker image prune` to remove intermediate images.
```
glbrntt: I'd rather we kept this section as how to build the image and explaining what it provides instead of how to use Docker.

MrMage:
Closing this due to inactivity for now; please let us know if you would like to pick it up again.

PR #580, opened 2019-09-09
RPC state machine for client

glbrntt:
Motivation:

State for an RPC is tracked, and partially duplicated, across multiple
channel handlers. This can make it difficult to understand and map back
to the gRPC protocol.

Modifications:

- Create a state machine representing the RPC state from the perspective
  of the client.
- Add tests.

Result:

Nothing yet, since it isn't in use.


Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+  /// - `.clientStreamingServerIdle` → `.clientStreamingServerIdle`
+  /// - `.clientStreamingServerStreaming` → `.clientStreamingServerStreaming`
+  ///
+  /// It is invalid (but not fatal) for the client to attempt to send requests once the request
+  /// stream is closed. The states are:
+  /// - `.clientClosedServerIdle`
+  /// - `.clientClosedServerStreaming`
+  /// - `.clientClosedServerClosed`
+  ///
+  /// Sending messages is invald, and fatal, for the following states:
```
MrMage: typo: invald (also in some other cases)

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+  }
+
+  /// Formats a request to send to the server.
+  ///
+  /// The client must be streaming in order for this to return successfully. Therefore the valid
+  /// state transitions are:
+  /// - `.clientStreamingServerIdle` → `.clientStreamingServerIdle`
+  /// - `.clientStreamingServerStreaming` → `.clientStreamingServerStreaming`
+  ///
+  /// It is invalid (but not fatal) for the client to attempt to send requests once the request
```
MrMage: Could we define "invalid" and "fatal" somewhere?

Sources/GRPC/GRPCClientStateMachine.swift
MrMage: Why is this particular one fatal?

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+  /// - Parameter allocator: A `ByteBufferAllocator` to allocate the buffer into which the encoded
+  ///     request will be written.
+  mutating func sendRequest(
+    _ message: Request,
+    allocator: ByteBufferAllocator
+  ) -> Result<ByteBuffer, MessageWriteError> {
+    return self.state.sendRequest(message, allocator: allocator)
+  }
+
+  /// Terminates the request stream.
```
MrMage: "Closes" instead of "Terminates"? It feels a bit like we only _close_ half-streams (i.e. request or response) but _terminate_ whole RPCs.

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+  /// - `.clientStreamingServerStreaming` → `.clientStreamingServerStreaming`
+  ///
+  /// It is invalid and fatal to recieve a response in the following states:
+  /// - `.clientIdleServerIdle`
+  /// - `.clientClosedServerStreaming`
+  /// - `.clientStreamingServerStreaming`
+  /// - `.clientClosedServerClosed`
+  ///
+  /// - Parameter buffer: A buffer of bytes received from the server.
+  mutating func recieveResponse(
```
MrMage: Calls this "receiveResponse**s**" instead?

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+  mutating func sendRequestHeaders(
+    host: String,
+    path: String,
+    options: CallOptions,
+    requestID: String
+  ) -> Result<HTTPRequestHead, InvalidStateError> {
+    let result: Result<HTTPRequestHead, InvalidStateError>
+
+    switch self {
+    case let .clientIdleServerIdle(pendingWriteState, responseArity):
```
MrMage: Should we always call this `{Request,Response}Arity` instead of `{Request,Response}Count`, both the variable names and the type name?

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+    }
+
+    return result
+  }
+
+  /// See `GRPCClientStateMachine.recieveResponseHeaders(_:)`.
+  mutating func recieveResponseHeaders(
+    _ responseHead: HTTPResponseHead,
+    logger: Logger
+  ) -> Result<HTTPHeaders, RecieveResponseHeadError> {
```
MrMage: Make `Head/Headers` consistent in `receiveResponseHeaders` and `ReceiveResponseHeadError`?

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+
+     return result
+   }
+
+  /// Makes the request headers ("Request-Headers" in the specification) used to initiate an RPC
+  /// call.
+  ///
+  /// See: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
+  ///
+  /// - Important: pseudo-headers (":method", ":scheme", ":path") are not included here and should
```
MrMage: super-nit (optional): use backticks instead of quotes?

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+
+     case .clientIdleServerIdle,
+          .clientClosedServerClosed:
+      result = .failure(.invalidState())
+     }
+
+     return result
+   }
+
+  /// Makes the request headers ("Request-Headers" in the specification) used to initiate an RPC
```
MrMage: super-nit (optional): use backticks instead of quotes?

Sources/GRPC/GRPCClientStateMachine.swift
MrMage: add "except for `:host`"?

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+    path: String,
+    options: CallOptions,
+    requestID: String
+  ) -> HTTPRequestHead {
+    // Note: we don't currently set the 'grpc-encoding' header, if we do we will need to feed that
+    // encoded into the message writer.
+    var headers: HTTPHeaders = [
+      "content-type": "application/grpc",
+      "te": "trailers",  // Used to detect incompatible proxies.
+      "user-agent": "grpc-swift-nio",  //  TODO: Add a more specific user-agent.
```
MrMage: is the double-space after the `//` intentional?

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+    host: String,
+    path: String,
+    options: CallOptions,
+    requestID: String
+  ) -> HTTPRequestHead {
+    // Note: we don't currently set the 'grpc-encoding' header, if we do we will need to feed that
+    // encoded into the message writer.
+    var headers: HTTPHeaders = [
+      "content-type": "application/grpc",
+      "te": "trailers",  // Used to detect incompatible proxies.
```
MrMage: Mention that this is part of the gRPC spec?

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+    // Extract the "Status"
+    let code = trailers[GRPCHeaderName.statusCode].first
+      .flatMap(Int.init)
+      .flatMap(GRPCStatus.Code.init) ?? .unknown
+
+    // Extract and unmarshall the "Status-Message"
+    let message = trailers[GRPCHeaderName.statusMessage].first
+      .map(GRPCStatusMessageMarshaller.unmarshall)
+
+    return .init(code: code, message: message)
```
MrMage: Wouldn't we also need to extract other (custom) trailing metadata in case the server sent it, or is that done elsewhere?

Sources/GRPC/ReadWriteStates.swift
```diff
@@ -0,0 +1,191 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import NIO
+import SwiftProtobuf
+
+/// Number of messages expected on a stream.
+enum MessageCount {
+  case one
```
MrMage: Should we add `non` and transition into that once a unary call has received its first message? (This may be a dumb idea.)

Reading further, this might replace `canWrite/canRead`, but no strong feelings what is better.

Sources/GRPC/ReadWriteStates.swift
```diff
@@ -0,0 +1,191 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import NIO
+import SwiftProtobuf
+
+/// Number of messages expected on a stream.
+enum MessageCount {
```
MrMage: Rename to `MessageArity`?

Sources/GRPC/ReadWriteStates.swift
```diff
@@ -0,0 +1,191 @@
...
+/// The write state of a stream.
+struct WriteState {
+  /// Whether the stream may be written to.
+  internal private(set) var canWrite: Bool = true
+
+  /// The number of messages we expect to write to the stream.
+  var expectedCount: MessageCount
+
+  /// A writer to encode `Message`s into the gRPC wire-format.
+  var writer: LengthPrefixedMessageWriter
```
MrMage: Will this ever be initialized with something other than `LengthPrefixedMessageWriter()`?

Sources/GRPC/ReadWriteStates.swift
```diff
@@ -0,0 +1,191 @@
...
+    self.contentType = contentType
+  }
+}
+
+enum MessageWriteError: Error {
+  /// Too many messages were written.
+  case cardinalityViolation
+
+  /// Message serialization failed.
+  case serialzationFailed
```
MrMage: typo: serialzation

Sources/GRPC/ReadWriteStates.swift
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    guard let data = try? message.serializedData() else {
+      return .failure(.serialzationFailed)
+    }
+
+    // Zero is fine: the writer will allocate the correct amount of space.
+    var buffer = allocator.buffer(capacity: 0)
+    // TODO: add support for compression.
+    self.writer.write(data, into: &buffer, usingCompression: .none)
```
MrMage: Should the used compression be part of the writer instead?

Sources/GRPC/ReadWriteStates.swift
```diff
@@ -0,0 +1,191 @@
...
+  /// Message deserialization failed.
+  case deserializationFailed
+
+  /// An invalid state was encountered. This is a serious implementation error.
+  case invalidState
+}
+
+extension ReadState {
+  /// Consume the given `buffer` then attempt to read and subsequently decode length-prefixed
+  /// serialzed messages.
```
MrMage: typo: serialzed

Sources/GRPC/ReadWriteStates.swift
```diff
@@ -0,0 +1,191 @@
...
+    switch (self.expectedCount, messages.count) {
+    case (.one, 1):
+      self.canRead = false
+      // We shouldn't have any bytes leftover after reading a single message.
+      if self.reader.hasBytes {
+        return .failure(.leftOverBytes)
+      }
+
+    case (.one, 2...):
+      return .failure(.cardinalityViolation)
```
MrMage: Should we also set `self.canRead = false`, or is this implied by returning `.failure`?

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import Foundation
+import NIO
+import NIOHTTP1
+import Logging
+import SwiftProtobuf
+
+enum RecieveResponseHeadError: Error, Equatable {
```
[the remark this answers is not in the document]
glbrntt: "a few": 183 by my count 🤦‍♂️

glbrntt:
> Looks great already! I must admit I didn't have the time to look at the tests.

No problem, they're mostly boring and just prodding different states.

> Also, I noted that you are using `Result` instead of `try` on the state machine. I assume there is a good reason for this (better type-safety?), would love to know it.

Pretty much: having the exact type saves you from having to cast and have a catch-all should that cast fails.

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+  /// The encoding used by the server is not supported.
+  case unsupportedMessageEncoding
+
+  /// An invalid state was encountered. This is a serious implementation error.
+  case invalidState
+}
+
+enum SendEndOfRequestStreamError: Error {
+  /// The request stream has already been closed.
+  case endOfStreamCardinalityViolation
```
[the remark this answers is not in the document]
glbrntt: Agreed. I think renaming it is the better bet here since it should only be possible for the user to trigger it.

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+
+  /// An invalid state was encountered. This is a serious implementation error.
+  case invalidState
```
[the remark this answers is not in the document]
glbrntt: I don't think so: `invalidState` really only serves a purpose for testing (since it should be a `preconditionFailure` the rest of the time. The context we add in `InvalidStateError` ~probably~ isn't actually necessary and can be removed.

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+  let function: String
+  static func invalidState(_ function: String = #function) -> InvalidStateError {
+    return .init(function: function)
+  }
+}
+
+/// A state machine for a single gRPC call from the perspective of a client.
+///
+/// See: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md
+struct GRPCClientStateMachine<Request: Message, Response: Message> {
```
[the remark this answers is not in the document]
glbrntt: Correct!

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+
+    /// The client has initiated an RPC and has not had a response from the server. Holds the
+    /// writing state for requests and expected message count for responses.
+    ///
+    /// Valid transitions:
+    /// - `clientStreamingServerStreaming`: if the server acknowledges the RPC initiation,
+    /// - `clientClosedServerIdle`: if the client closes the request stream,
+    /// - `clientClosedServerClosed`: if the client terminates the RPC or the server terminates the
+    ///      RPC with a "trailers-only" response.
+    case clientStreamingServerIdle(WriteState, MessageCount)
```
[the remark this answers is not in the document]
glbrntt: Good point: perhaps a "open"/"active" or similar would be better still.

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+  /// - `.clientStreamingServerClosed`: The client may not stream if the server is closed.
+  enum State {
+    /// Initial state. Neither request stream nor response stream have been initiated. Holds the
+    /// pending write state for the request stream and expected message count for the response
+    /// stream, respectively.
+    ///
+    /// Valid transitions:
+    /// - `clientStreamingServerIdle`: if the client initiates the RPC,
+    /// - `clientClosedServerClosed`: if the client terminates the RPC.
+    case clientIdleServerIdle(PendingWriteState, MessageCount)
```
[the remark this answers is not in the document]
glbrntt: Usually I don't like them because the labels tend to just match the types so you end up with e.g.:

```swift
switch self {
  case .clientIdleServerIdle(pendingWriteState: let pendingWriteState, messageCount: let messageCount):
    // ...
}
```

Perhaps `client` and `server` labels would be useful though?
```swift
switch self {
  case .clientIdleServerIdle(client: let pendingWriteState, server: let messageCount):
    // ...
}
```

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+    /// - `clientClosedServerClosed`: if the client terminates the RPC or the server terminates the
+    ///      RPC with a "trailers-only" response.
+    case clientClosedServerIdle(MessageCount)
+
+    /// The client has initiated the RPC and the server has acknowledged it. Messages may have been
+    /// sent and/or recieved. Holds the request stream write state and response stream read state.
+    ///
+    /// Valid transitions:
+    /// - `clientClosedServerStreaming`: if the client closes the request stream,
+    /// - `clientClosedServerClosed`: if the client terminates the RPC.
```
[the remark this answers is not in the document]
glbrntt: Good catch :)

Sources/GRPC/GRPCClientStateMachine.swift
glbrntt: Ah, the previous iteration of this used `.nonFatal` and `.fatal` errors to determine whether the RPC should be terminated or not: I'll do a pass of the documentation to update this.

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,588 @@
...
+      case (.clientIdleServerIdle, .clientIdleServerIdle),
+           (.clientStreamingServerIdle, .clientStreamingServerIdle),
+           (.clientClosedServerIdle, .clientClosedServerIdle),
+           (.clientStreamingServerStreaming, .clientStreamingServerStreaming),
+           (.clientClosedServerStreaming, .clientClosedServerStreaming),
+           (.clientClosedServerClosed, .clientClosedServerClosed):
+        break
+
+      default:
+        preconditionFailure("invalid state transition from '\(self.state)' to '\(oldValue)'")
```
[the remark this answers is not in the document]
glbrntt: The transitions covered here _should not_ be possible. If we find ourselves _trying to make_ one of these transitions then we should be returning a `Result` indicating the invalid state in some form or another but _not_ updating `self.state`.

There are unit tests calling each of the functions on the state machine from each state to validate that we get an invalid state. If we regress -- i.e. actually update `self.state` then the tests will blow up hitting this precondition.

Sources/GRPC/GRPCClientStateMachine.swift
glbrntt: I think `recieveResponseBuffer` is probably more accurate still.

Sources/GRPC/GRPCClientStateMachine.swift
glbrntt: We should be consistent, yes. I originally had it as `MessageArity` but renamed to `MessageCount` since it's a "more obvious" (for want of a better term) name.

Sources/GRPC/GRPCClientStateMachine.swift
glbrntt: At the moment (here and in the current implementation) we don't remove the gRPC status code or message from the headers and I don't think there's much value in doing so (it would also require a copy of the headers). We can just forward on the same headers we feed into the state machine to extract the status.

Sources/GRPC/ReadWriteStates.swift
glbrntt: Yes: when we add compression support we'll have to initialize the writer with the compression algorithm being used.

Sources/GRPC/ReadWriteStates.swift
glbrntt: Yes: looks like we always pass in `.none` anyway!

Sources/GRPC/GRPCClientStateMachine.swift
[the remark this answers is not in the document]
MrMage: Should we still list in the comment all the "usual" reasons why the stream could already be closed?

Sources/GRPC/GRPCClientStateMachine.swift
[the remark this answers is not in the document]
MrMage: `clientOpen` sounds good although the channel is technically already open in the "idle" state. But I guess `clientActive` or `clientSending` would work. I think `clientActive` is my favourite.

Sources/GRPC/GRPCClientStateMachine.swift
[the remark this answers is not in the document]
MrMage: I think you can actually omit the label (e.g. `pendingWriteState: `) in `case` statements, could you check that?

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -0,0 +1,610 @@
...
+  }
+
+  /// Formats a request to send to the server.
+  ///
+  /// The client must be streaming in order for this to return successfully. Therefore the valid
+  /// state transitions are:
+  /// - `.clientStreamingServerIdle` → `.clientStreamingServerIdle`
+  /// - `.clientStreamingServerStreaming` → `.clientStreamingServerStreaming`
+  ///
+  /// The client should not to attempt to send requests once the request stream is closed, that is
```
MrMage: nit: double "to", did not check other occurrences of "should"

Sources/GRPC/GRPCClientStateMachine.swift
MrMage: Not done yet, I think.

Sources/GRPC/GRPCClientStateMachine.swift
MrMage: Has this been done? (Sorry, don't have the time for a full re-review right now.)

Sources/GRPC/GRPCClientStateMachine.swift
MrMage: Has this been answered? (Sorry, don't have the time for a full re-review right now.)

Sources/GRPC/GRPCClientStateMachine.swift
MrMage: Was this removed altogether?

Sources/GRPC/GRPCClientStateMachine.swift
MrMage: Not addressed yet.

Sources/GRPC/GRPCClientStateMachine.swift
MrMage: Sorry, what I meant is, where is the trailing metadata exposed to the client? Lower down the chain? I.e. we can just ignore it here, but it does get used elsewhere? (This is just for understanding, the current approach is most likely the right one.)

Sources/GRPC/ReadWriteStates.swift
```diff
@@ -0,0 +1,199 @@
...
+  case many
+}
+
+/// Encapsulates the state required to create a new write state.
+struct PendingWriteState {
+  /// The number of messages we expect to write to the stream.
+  var arity: MessageArity
+
+  /// The encoding we should use when writing the message.
+  var encoding: CompressionMechanism
```
MrMage: Feels like we are inconsistent with encoding vs compression here; not sure if we can improve on that, though.

Sources/GRPC/ReadWriteStates.swift
MrMage: A quick reply would be appreciated :-)

Sources/GRPC/GRPCClientStateMachine.swift
glbrntt: It's been reworded to make the expected errors clearer. In this case "fatal" because it would be an invalid state enabled by bad design (i.e. somehow we made it possible for the user to send a request before sending the request headers).

Sources/GRPC/GRPCClientStateMachine.swift
glbrntt: Yes: it's now `MessageArity` with an additional case for `none`.

Sources/GRPC/GRPCClientStateMachine.swift
glbrntt: Yes: they're added as part of the `HTTPRequestHead` and made into pseudo-headers by the H2 to H1 codec so I didn't think this was necessary.

Sources/GRPC/GRPCClientStateMachine.swift
glbrntt: The state machine will be used within a channel handler. When it reads the trailers they'll be passed into the state machine for validation, if they're successfully validated then we get a `GRPCStatus` in response which will be forwarded along with the trailers to the next handler. If they're not valid then an error will be fired and the next handler will have to synthesise a status so the handler can fail the trailing metadata promise and succeed the status promise.

Sources/GRPC/ReadWriteStates.swift
glbrntt: The header name for compression is "grpc-encoding", and both encoding and compression are used in the spec. I'm happy to change this to `compression` here though.

Sources/GRPC/ReadWriteStates.swift
glbrntt: It's implied by returning `.failure` but I think this should be more explicit. Same goes for `WriteState`.

Sources/GRPC/GRPCClientStateMachine.swift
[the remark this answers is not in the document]
glbrntt: Changed `Streaming` to `Active`

Sources/GRPC/GRPCClientStateMachine.swift
MrMage: Oops, sorry!

Sources/GRPC/GRPCClientStateMachine.swift
MrMage: Thank you for the explanation! I do wonder whether `receiveEndOfResponseStream` should return both `GRPCStatus` and the trailers to the channel handler calling it (feels like it would be the responsibility of the state machine to extract the trailers, not of the channel handler), but I'm happy to suspend this discussion once the channel handler is actually written.

glbrntt: No problem. 

> I'm happy to suspend this discussion once the channel handler is actually written.

Makes sense!

PR #590, opened 2019-09-25
Add HelloWorld example

glbrntt:
Motivation:

To help users get started quickly with gRPC Swift it would be useful to
have a "hello world" example. grpc.io has such examples for
a number of languages: https://grpc.io/docs/quickstart/ but not for Swift.

Another motivating factor for this change was feedback from the gRPC team
on the SSWG draft proposal suggesting that it would be useful to frame the
proposal from a canonical example, either "HelloWorld" or "RouteGuide".

Modifications:

- Add a "HelloWorld" example
- Create a "quick-start" guide modelled on those on grpc.io

Result:

- No functional changes
- Another example and tutorial


Sources/Examples/Echo/Model/echo.pb.swift
```diff
@@ -3,7 +3,7 @@
 // Generated by the Swift generator plugin for the protocol buffer compiler.
 // Source: echo.proto
 //
-// For information on using the generated types, please see the documentation:
+// For information on using the generated types, please see the documenation:
```
MrMage: typo, possibly in the original proto file?

docs/quick-start.md
```diff
@@ -0,0 +1,220 @@
+# gRPC Swift Quick Start
```
MrMage: Should we mention this example in the README?

docs/quick-start.md
```diff
@@ -0,0 +1,220 @@
...
+#### Install Protocol Buffers v3
+
+Install the protoc compiler that is used to generate gRPC service code. The
+simplest way to do this is to download pre-compiled binaries for your
+platform (`protoc-<version>-<platform>.zip`) from here:
+[https://github.com/google/protobuf/releases][protobuf-releases].
+
+* Unzip this file.
+* Update the environment variable `PATH` to include the path to the `protoc`
+  binary file.
```
MrMage: ", or move the `protoc` binary into a directory that is part of your `PATH`"?

docs/quick-start.md
```diff
@@ -0,0 +1,220 @@
...
+### Download the example
+
+You'll need a local copy of the example code to work through this quickstart.
+Download the example code from our GitHub repository (the following command
+clones the entire repository, but you just need the examples for this quickstart
+and other tutorials):
+
+```sh
+$ # Clone the repository at the latest release to get the example code:
+$ git clone -b 1.0.0-alpha.6 https://github.com/grpc/grpc-swift
```
MrMage: Should we really mention a specific tag here? How about just cloning the `nio` branch for now?

docs/quick-start.md
```diff
@@ -0,0 +1,220 @@
...
+   ```sh
+   $ swift run HelloWorldServer
+   server started on port 52200
+   $ # Note: the port may be different on your machine.
+   ```
+
+2. In another terminal, compile and run the client
+
+   ```sh
+   $ swift run HelloWorldClient 52200
```
MrMage: Add the output of that command as well?

docs/quick-start.md
```diff
@@ -0,0 +1,220 @@
...
+buffers; you can find out lots more about how to define a service in a `.proto`
+file in [What is gRPC?][grpc-guides]. For now all you need to know is that both
+the server and the client "stub" have a `SayHello` RPC method that takes a
+`HelloRequest` parameter from the client and returns a `HelloReply` from the
+server, and that this method is defined like this:
+
+```proto
+// The greeting service definition.
+service Greeter {
+  // Sends a greeting
```
MrMage: Should we be consistent with full stops at the end of comments? (Also below)

docs/quick-start.md
```diff
@@ -0,0 +1,220 @@
...
+### Update and run the application
+
+We need to regenerate
+`Sources/Examples/HelloWorld/Model/helloworld.grpc.swift`, which
+contains our generated gRPC client and server classes.
+
+From the `grpc-swift` directory run
+
+```sh
+make generate-helloworld
```
MrMage: I wonder if we should specify the `protoc` command here, so that users could better adapt that command to their own project?

Sources/Examples/Echo/Model/echo.pb.swift
glbrntt: Not the Proto file but the plugin! I fixed this a couple of days ago: https://github.com/apple/swift-protobuf/pull/902 so we should be able to fix this when they tag the next release.

docs/quick-start.md
glbrntt: We should, but I'm going to leave that until I've done the route guide tutorial (in progress) as well, I think it'll make the README a little clear (i.e. here's a quick start guide, here's a more in depth guide). WDYT?

docs/quick-start.md
glbrntt: I'd rather keep the wording as it is: this is the wording for all of the quick start guides on grpc.io

docs/quick-start.md
glbrntt: I don't mind: I just went for consistency with other quick start guides (and the README).

docs/quick-start.md
glbrntt: I think - as the most basic tutorial - we can leave this out. It will be included in the route guide tutorial which will be linked in the "What's next" section when it's created.

docs/quick-start.md
MrMage: SGTM

docs/quick-start.md
MrMage: Thanks for the info. Either is fine with me then.

docs/quick-start.md
MrMage: SGTM

MrMage:
LGTM

PR #593, opened 2019-09-26
Use "swift run" for interop tests in CI

glbrntt:
Motivation:

60205adb (#592) cleared up the product namespace by removing executables
which were not meant to be exposed to clients. However the post-merge CI
failed (i.e. interop tests) because the executable targets no longer
exist. This was missed because the CI for the PR was still able to build
the targets and because the products had different names to the targets.

Modifications:

Use 'swift run' in the CI script used after merging PRs.

Result:

Interop tests should be runnable when a PR is merged.

### DIFF
--- a/.travis-script.sh
+++ b/.travis-script.sh
@@ -86,7 +86,6 @@ make_project() {
 
 run_interop_tests() {
   echo -en 'travis_fold:start:test.interop_tests\\r'
-  make interop-test-runner
   INTEROP_TEST_SERVER_PORT=8080
 
   # interop_server should be on $PATH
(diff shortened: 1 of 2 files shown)


glbrntt:
@kevints I'm unable to request your review for this but please do so!

kevints:
Looks like this is a package manager bug, filed https://bugs.swift.org/browse/SR-11528

As a workaround, can the main build run `swift build` without any arguments? Then you shouldn't be required to use `swift run` here. If you still want to use `swift run` here, consider adding `--skip-build` so that any build failures are caught in the pre-commit step rather than the post-commit one.

glbrntt:
Closing in favour of: https://github.com/grpc/grpc-swift/pull/596

PR #596, opened 2019-09-27
Fix interoperability tests.

kevints:
Motivation:

In 60205ad I changed the Makefile to use `--target` instead of
`--product` to build the interoperability tests. However, it turns out
this only builds the objects and doesn't link the final executable.

Modifications:

Use `--product` to build the interoperability tests again.

Result:

The build is fixed.


kevints:
cc @glbrntt

glbrntt:
Thanks @kevints!

PR #598, opened 2019-09-27
Add a RouteGuide example and tutorial

glbrntt:
Motivation:

a30043a (#590) introduced a 'Hello World' example and tutorial to help
new gRPC Swift users. A more in-depth tutorial would be useful as a
next-step for those users.

Modifications:

Add a tutorial based on the "route guide" service as per
https://grpc.io/docs/tutorials/

Move parts of the `README.md` into separate documenation files in `docs/`

Result:

- Another gRPC tutorial
- Easier to find documenation


Sources/Examples/RouteGuide/Client/main.swift
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  for _ in 0..<featuresToVisit {
+    let index = Int.random(in: 0..<features.count)
+    let point = features[index].location
+    print("Visiting point \(point.latitude), \(point.longitude)")
+    call.sendMessage(point, promise: nil)
+
+    // Sleep for a bit before sending the next one.
+    usleep(UInt32.random(in: 0...1_000_000) + 500_000)
```
MrMage: Can we use e.g. Thread.sleep here, or some other more Swifty API?

Sources/Examples/RouteGuide/Model/route_guide.grpc.swift
```diff
@@ -0,0 +1,152 @@
+//
+// DO NOT EDIT.
+//
+// Generated by the protocol buffer compiler.
+// Source: route_guide.proto
+//
+
+//
+// Copyright 2018, gRPC Authors All rights reserved.
```
MrMage: (Unrelated to this PR) I guess we should update the copyright string in our protoc plugin soon

Sources/Examples/RouteGuide/Server/RouteGuideProvider.swift
```diff
@@ -0,0 +1,189 @@
...
+  func listFeatures(
+    request: Routeguide_Rectangle,
+    context: StreamingResponseCallContext<Routeguide_Feature>
+  ) -> EventLoopFuture<GRPCStatus> {
+    let left = min(request.lo.longitude, request.hi.longitude)
+    let right = max(request.lo.longitude, request.hi.longitude)
+    let top = max(request.lo.latitude, request.hi.latitude)
+    let bottom = max(request.lo.latitude, request.hi.latitude)
+
+    self.features.filter { feature in
```
MrMage: We could make this filter lazy, but not strictly necessary

Sources/GRPC/ClientCalls/ClientCall.swift
```diff
@@ -52,6 +52,15 @@ public protocol ClientCall {
   func cancel()
 }
 
+extension ClientCall {
+  public func waitForCompletion(errorCallback: (GRPCStatus) -> Void = { _ in }) {
```
MrMage: What's the advantage of this method over `try wait()`?

docs/basic-tutorial.md
```diff
@@ -0,0 +1,685 @@
+# Basic Tutorial
+
+This tutorial provides a basic Swift programmer's introduction to working with
```
MrMage: s/programmer/developer/?

docs/basic-tutorial.md
```diff
@@ -0,0 +1,685 @@
...
+    $0.location = location
+  }
+}
+```
+
+`getFeature()` takes two parameters:
+
+- `Routeguide_Point`: the request
+- `StatusOnlyCallContext`: a context which exposes status and trailing metadata
+  objects.
```
MrMage: s/objects/promises that you can fulfill/?

Sources/GRPC/ClientCalls/ClientCall.swift
glbrntt: Oops -- I didn't mean to commit this. I was playing around with ideas because waiting on the `GRPCStatus` can be a bit of a pain: even though we guarantee that it will never fail (i.e. `try! status.wait()` should not fail) it's something that users have to trust and is absolutely a code-smell. 

Also when the future is completed the user has to switch on the status code to check whether it's `.ok` or not. So there's actually quite a lot of boilerplate required to check the outcome of the call.

Not sold on any solution yet but I've been thinking that a never-failing future would be helpful (i.e. you would `recover` or similar into something with no `.failure` branch), as would futures with a particular error type (i.e. like `Result`).

Sources/Examples/RouteGuide/Model/route_guide.grpc.swift
glbrntt: Good point but should this even be our copyright header? SwiftProtobuf uses the copyright header from the source Proto file: https://github.com/apple/swift-protobuf/blob/da75a93ac017534e0028e83c0e4fc4610d2acf48/Sources/protoc-gen-swift/FileGenerator.swift#L71

I'll leave this unchanged for now and take a look into what other gRPC plugins do.

Sources/GRPC/ClientCalls/ClientCall.swift
MrMage: Sounds like we should revert this here and file an issue for a better future solution :-)

Sources/Examples/RouteGuide/Model/route_guide.grpc.swift
MrMage: SHTM

docs/basic-tutorial.md
glbrntt: The equivalent tutorials for different languages on grpc.io use programmer so I think we should stay as close to that as possible.

glbrntt:
> Great codelab! Are there plans to link to all the docs from one central location?

I spoke to @timburks a little while ago about getting gRPC Swift included on https://grpc.io -- so hopefully there!

docs/basic-tutorial.md
glbrntt: They're not `EventLoopPromise`s: "objects that you can update"?

MrMage:
> I spoke to @timburks a little while ago about getting gRPC Swift included on https://grpc.io -- so hopefully there!

Hmm, how about linking to them from README.md as well?

docs/basic-tutorial.md
MrMage: Sorry, my mistake. How about "a context which exposes status and trailing metadata
  fields that you can change if needed."?

glbrntt:
> > I spoke to @timburks a little while ago about getting gRPC Swift included on https://grpc.io -- so hopefully there!
> 
> Hmm, how about linking to them from README.md as well?

Already done in this PR!

PR #612, opened 2019-10-24
Update the Podspec for v0.10, with gRPC-Core 1.23 and Swift 5.

MrMage:


SwiftGRPC.podspec
```diff
@@ -32,7 +32,7 @@ Pod::Spec.new do |s|
   s.authors  = { 'The gRPC contributors' => 'grpc-packages@google.com' }
   s.source = { :git => 'https://github.com/grpc/grpc-swift.git', :tag => s.version }
 
-  s.swift_versions = [4.2]
+  s.swift_versions = ['4.2', '5']
```
glbrntt: Should 5.1 be included here too?

MrMage: Not sure if that is even an official Swift version as far as CocoaPods is concerned. I would prefer to keep it safe and only include 5. Shouldn't make a difference in practice, anyway.

PR #618, opened 2019-10-30
Use string comparison instead of regex in the HTTPProtocolSwitcher

glbrntt:
Motivation:

Regex is overkill to check which HTTP version is being used in the
HTTPProtocolSwitcher.

Modifications:

Determine the HTTP version by checking for a substring instead of using
a regex.

Result:

Simpler code.


glbrntt:
Credit to @Lukasa for pointing this out

PR #625, opened 2019-11-12
Update the client to use the new state machine backed channel handler

glbrntt:
Motivation:

We recently rewrote the client channel handling pipeline to be backed by
a single state machine which also does not require the conversion from
HTTP2 to HTTP1. This PR update the client to make use of this new
pipeline.

Modifications:

- Update the client request and response channel handlers as they now
  use different types and no longer require their own state tracking
  since it is provided by the GRPCClientChannelHandler
- Update the different call types and base client call to make use of
  these new types.
- Delete unused channel handlers.
- Updated tests to make use of new types.
- Fixed a bug in one of the interop test cases.

Result:

The client now uses our new channel handler and state machine and no
longer requires the conversion from HTTP2 to HTTP1.


Sources/GRPC/ClientCalls/BaseClientCall.swift
```diff
@@ -34,159 +34,121 @@ import Logging
...
+    self.subchannel = streamPromise.futureResult
+    self.initialMetadata = responseHandler.initialMetadataPromise.futureResult
+    self.trailingMetadata = responseHandler.trailingMetadataPromise.futureResult
+    self.status = responseHandler.statusPromise.futureResult
+
+    // If the stream (or multiplexer) fail we need to fail any responses.
+    self.multiplexer.cascadeFailure(to: streamPromise)
+    streamPromise.futureResult.whenFailure(responseHandler.onError)
+
+    // Create an http/2 stream and configure it with the gRPC handler.
```
MrMage: super-nit: uppercase HTTP?

Sources/GRPC/ClientCalls/ClientCall.swift
```diff
@@ -42,14 +43,15 @@ public protocol ClientCall {
...
+  var trailingMetadata: EventLoopFuture<HPACKHeaders> { get }
 
   /// Cancel the current call.
   ///
   /// Closes the HTTP/2 stream once it becomes available. Additional writes to the channel will be ignored.
   /// Any unfulfilled promises will be failed with a cancelled status (excepting `status` which will be
   /// succeeded, if not already succeeded).
-  func cancel()
+  func cancel() -> EventLoopFuture<Void>
+  func cancel(promise: EventLoopPromise<Void>?)
```
MrMage: Should we default the promise to nil? (Feel free to disregard if that would be inconsistent with the rest of NIO.)

glbrntt: We can't provide a default as it would make `cancel()` ambiguous (and would be inconsistent with NIO).

Sources/GRPC/ClientCalls/BaseClientCall.swift
glbrntt: Done!

glbrntt:
> Sweet!
> 
> I assume the CI failure is only for the PR, but the last commit succeeds? If not, please rerun before merging.

Forgot to update the tests :(

Rerunning CI now.

MrMage:
LGTM

Issue #342, opened 2018-12-13
Failed to create subchannel


pffan91:
#### Question Description
I've created request as:

```
        let client = Service_Frontend_FrontendServiceClient(address: "backend.server.com:443", secure: true)
        
        var request = Service_Frontend_LoginRequest()
        
        var facebook = Service_Frontend_FacebookAuthData()
        facebook.facebookAuthToken = user?.userToken ?? ""
        request.authData.facebookAuthData = facebook
    
        do {
            try client.login(request) { (response, result) in
                print("Response: \n \(response)")
                print("Result: - \(result)")
            }
        } catch {
            print(error)
        }

```

After that I receive response as `nil` and result is:

```
 - successful, status unknown: Failed to create subchannel
initialMetadata: [:]
trailingMetadata: [:]
```

So where I am wrong? Can you guide me? Thank you!

MrMage:
For me, this only happens when I don’t have an internet connection to that server. Would that explain it?

pffan91:
The problem with App Transport Security I think.. The same code runnig on MacOS without any problems.. Any thoughts how to fix it on iOS?

pffan91:
The issue was in calling wrong method. I used this one:

```
 try client.login(request) { (response, result) in
               print(“Response: \n \(response)“)
               print(“Result: - \(result)“)
           }
```

But looks like it does not worked.. When I changed it to 

  ```
let response = try client.login(request)
           print(response.authToken)
```

I successfully received response:)

**Comment by phlippieb** [comment]
@pffan91 I ran into the same issue, and figured out what is really happening. The problem wasn't that I was calling the wrong method; the sync and async versions of an rpc are equally valid, and in many cases we should prefer the async version. The problem was that I didn't retain the object that was making the call. For an async call, this caused the caller to be deallocated before the call could be made properly, and that produced the error in question. Using the sync call _appeared_ to fix it because, now that the call was blocking, the caller couldn't get deallocated before the call completed; however, we shouldn't block control flow while we wait for a network call to complete just as a workaround for an object not being retained.

The actual solution is simply to make sure that whichever object is making the call is retained somewhere.

Say you have a class responsible for making the call:

```
class LoginEnpoint {
   func login(request: Request) -> Void) {
      try! client.login(request) { (response, result) in
         // ...
      }
   }
}
```

Then the following will cause the endpoint object to be deallocated immediately, and the call to fail:

```
LoginEndpoint().login(request) // Failed to create subchannel
```

[...]

PR #631, opened 2019-11-13
Retain the client call in async calls until we have received a response.

### LINKED ISSUE #330: A very unstable error
I am having a very strange bug here, my code is as below:

        let request = Rpcpb_GetBlockchainInfoRequest.init()
        _ = try? Rpcpb_RpcServiceServiceClient.init(address: "127.0.0.1:50051", secure: false).rpcGetBlockchainInfo(request, completion: { (response, CallResult) in
            print("RESPONSE:\(String(describing: response?.blockHeight))\(String(describing: response?.tailBlockHash))\(CallResult.statusCode)\(String(describing: CallResult.resultData))")
                print("Result:\(CallResult)")

            })

I sometimes got correct result:
RESPONSE:Optional(861)Optional(32 bytes)okOptional(121 bytes)
Result:successful, status ok: 
resultData: 121 bytes
initialMetadata: [:]
trailingMetadata: [:]

sometimes wrong result:
RESPONSE:nilnilunknownnil
Result:successful, status unknown: Failed to create subchannel
initialMetadata: [:]
trailingMetadata: [:]

Without changing any code, I just re-ran the program, it produced different results, anyone had same issue before, how can I fix it?
Any help would be appreciated!

MrMage:
This means that simply releasing the client will no longer cancel any in-flight requests; instead, the user needs to explicitly call `client.channel.shutdown()` for that to happen.

Fixes #330, #342 (more specifically, fixes unexpected behavior mentioned in these two issues).

### DIFF
--- a/Sources/SwiftGRPC/Runtime/ClientCall.swift
+++ b/Sources/SwiftGRPC/Runtime/ClientCall.swift
@@ -26,10 +26,12 @@ open class ClientCallBase {
   open class var method: String { fatalError("needs to be overridden") }
 
   public let call: Call
+  internal let channel: Channel  // To ensure that the channel is not deallocated before this call has finished.
 
   /// Create a call.
   public init(_ channel: Channel) throws {
     self.call = try channel.makeCall(type(of: self).method)
+    self.channel = channel
   }
 }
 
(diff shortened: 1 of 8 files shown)


MrMage:
No idea why the tests fail...

glbrntt:
> No idea why the tests fail...

Doesn't look related to your change but I'm not so familiar with the CgRPC code so up-to-you whether you want to dig more!

Issue #628, opened 2019-11-13
Should we reduce the amount of generated code?


MrMage:
Currently, we have the following [in our generated code](https://github.com/grpc/grpc-swift/blob/nio/Sources/Examples/Echo/Model/echo.grpc.swift):

```swift
public final class Echo_EchoServiceClient: GRPCServiceClient, Echo_EchoService {
  public let connection: ClientConnection
  public var serviceName: String { return "echo.Echo" }
  public var defaultCallOptions: CallOptions

  /// Creates a client for the echo.Echo service.
  ///
  /// - Parameters:
  ///   - connection: `ClientConnection` to the service host.
  ///   - defaultCallOptions: Options to use for each service call if the user doesn't provide them.
  public init(connection: ClientConnection, defaultCallOptions: CallOptions = CallOptions()) {
    self.connection = connection
    self.defaultCallOptions = defaultCallOptions
  }
```

Should we inline these properties (except for `serviceName`) into `GRPCServiceClient`?

@glbrntt, what do you think?

glbrntt:
`GRPCServiceClient` is a protocol; we're just providing conformance here. (Or do you mean turn `GRPCServiceClient` into a class which we inherit from?)

We could actually do away with service name and just inline it into the generated code. We lose the ability to customise the path but to be honest I don't think we actually need that. We'd also be able to get rid of a protocol as well, since `GRPCServiceClient` extends `GRPCClient` to provide the `serviceName` (this was required so we could have `AnyServiceClient`).

WDYT?

MrMage:
Sorry, I missed that `GRPCServiceClient`. I guess creating a class for that would be overkill.

Not feeling strongly about the service name, either; do you think getting rid of `GRPCServiceClient` through that would provide a substantial benefit? Or should we keep the service name in case a user wants to know the name of the service they are talking to?

glbrntt:
Not substantial, but it tidies things up a little (and removes something from the API). Do you think users need programmatic access to the service name? If not we could just add it to the generated documentation for the client.

I don't think there's a big difference in inlining vs. not inlining the service name so I'm fine with either!

MrMage:
🤷‍♂ I guess we could remove the service name for now — further reduces our API surface for 1.0. Can still add back that later on if needed...

PR #633, opened 2019-11-13
Inline service name into the generated path

glbrntt:
Motivation:

We gave users the option to customise how the path for the RPC was
formatted. The protocol allows this but discourages it.

Modifications:

- Inline the service name into the generated code
- Remove GRPCServiceClient and just use GRPCClient
- Update generated code

Result:

- Users are now discouraged in the strongest possible way to not use
  custom paths.
- Generated code is simpler.


glbrntt:
Fixes #628

Issue #622, opened 2019-11-08
[nio] TSan data race reported in "GRPCTests.ClientConnectionBackoffTests testClientEventuallyConnects"


weissi:
swift-nio: bc661cbb771328e09faa65432628e9ff6fd333d5
swift-nio-http2: 994c319e038e013817a21d4679fa2d73b838a703
grpc-swift: 2cd9ea7008312db794cb62d805f5ace043496ead

command:
```
swift test --sanitize=thread --disable-index-store
```

OS: macOS Catalina

```
Test Case '-[GRPCTests.ClientConnectionBackoffTests testClientEventuallyConnects]' started.
==================
WARNING: ThreadSanitizer: data race (pid=50737)
  Read of size 8 at 0x7b1c0000dc20 by thread T18 (mutexes: write M2633):
    #0 ConnectivityStateMonitor.delegate.getter <compiler-generated> (GRPCPackageTests:x86_64+0x381754)
    #1 ConnectivityStateMonitor.setNewState(to:) ConnectivityState.swift:103 (GRPCPackageTests:x86_64+0x382f7e)
    #2 closure #1 in ConnectivityStateMonitor.state.setter ConnectivityState.swift:82 (GRPCPackageTests:x86_64+0x382897)
    #3 partial apply for closure #1 in ConnectivityStateMonitor.state.setter <compiler-generated> (GRPCPackageTests:x86_64+0x3828eb)
    #4 thunk for @callee_guaranteed () -> (@error @owned Error) <compiler-generated> (GRPCPackageTests:x86_64+0x4b300f)
    #5 partial apply for thunk for @callee_guaranteed () -> (@error @owned Error) <compiler-generated> (GRPCPackageTests:x86_64+0x808197)
    #6 Lock.withLock<A>(_:) lock.swift:79 (GRPCPackageTests:x86_64+0x807ecf)
    #7 Lock.withLockVoid(_:) lock.swift:85 (GRPCPackageTests:x86_64+0x8080be)
    #8 ConnectivityStateMonitor.state.setter ConnectivityState.swift:81 (GRPCPackageTests:x86_64+0x3827c1)
    #9 closure #3 in ClientConnection.willSetChannel(to:) ClientConnection.swift:229 (GRPCPackageTests:x86_64+0x35b00f)
    #10 partial apply for closure #3 in ClientConnection.willSetChannel(to:) <compiler-generated> (GRPCPackageTests:x86_64+0x367f2f)
[...]

PR #634, opened 2019-11-13
Require a lock to access the connectivity state delegate

glbrntt:
Motivation:

- It was not possible to update the connectivity delegate in a thread
  safe way.
- See #622

Modifications:

Require the state lock to get/set the connectivity state delegate.

Result:

TSAN is happy.


glbrntt:
Fixes #622

Issue #642, opened 2019-11-30
Examples/Google/NaturalLanguage does not compile


sroche27r:
### New Issue Checklist

- [x ] Updated grpc-swift to the latest version
- [x ] I read the [Contribution Guidelines](https://github.com/grpc/grpc-swift/blob/master/CONTRIBUTING.md)
- [x ] I searched for [existing GitHub issues](https://github.com/grpc/grpc-swift/issues)(especially, there are many [resolved issues related to SSL](https://github.com/grpc/grpc-swift/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+ssl))

### Issue Description
Trying to compile Examples/Google/NaturalLanguage and I am getting compile errors.  _(I attempted to fix the errors, but could not get a functional example)_ 

Here is what I did:

1. downloaded the latest 'nio' branch
2. ran make plugins _(success)_
3. downloaded the latest protoc compiler 
4. ran Examples/Google/NaturalLanguage/RUNME _(success, compiled Proto files)_   
5. attempted to compile the example by running make _(failed with compile errors)_

<details>
  <pre>
Roche-MBP:NaturalLanguage sroche$ make
swift build -c release
/Volumes/work/DocsAndTutorials/gRPC/test_2019.11.29/grpc-swift-nio/proto_build/grpc-swift/Examples/Google/NaturalLanguage/Sources/main.swift:102:49: error: cannot convert value of type 'HTTPHeaders' to expected argument type 'HPACKHeaders'
  let callOptions = CallOptions(customMetadata: headers, timeout: timeout)
                                                ^~~~~~~
/Volumes/work/DocsAndTutorials/gRPC/test_2019.11.29/grpc-swift-nio/proto_build/grpc-swift/Examples/Google/NaturalLanguage/Sources/main.swift:40:28: error: extra argument 'tlsConfiguration' in call
        tlsConfiguration: .init(sslContext: try makeClientTLS()))
                          ~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
</pre>
</details>

NOTE: I have tried this on the nio branch, 1.0.0-alpha.7, 1.0.0-alpha.6. all with similar issues

I attached a script that will execute every step that I tried. Just copy it into an empty directory and run it.  [build_natural_language_example.sh.txt](https://github.com/grpc/grpc-swift/files/3906805/build_natural_language_example.sh.txt)

### Environment

<details>
  <pre>
  
[...]

PR #643, opened 2019-12-02
Fix compile issues in Google/NaturalLanguage

glbrntt:
Motivation:

The NaturalLanguage example did not compile because of breaking API
changes between pre-release version.

Modifications:

Fix compilation issues.

Result:

NaturalLanguage example compiles.


MrMage:
Fixes #642.

glbrntt:
@MrMage just an FYI: I didn't move this to the `Sources/Examples` directory since it has additional dependencies which aren't required by the library (i.e. google-auth-library-swift).

When I get around to removing the other examples/tidying up I'll add this to the CI.

PR #644, opened 2019-12-02
Percent encoding benchmark

glbrntt:
Motivation:

To measure performance, we need benchmarks. Here's one for percent
encoding and decoding of status messages.

Modifications:

Add a benchmark for encoding and decoding status messages.

Result:

We have a benchmark for encode and decoding status messages.


MrMage:
Never mind, just saw your other PR.

PR #645, opened 2019-12-02
Implement status message encoding and decoding without Foundation

glbrntt:
Motivation:

Percent encoding and decoding go via Foundation so we really get a
bridges CFString.

Modifications:

Implement the percent encoding and decoding without Foundation.

Result:

- Less Obj-C
- A small perf gain when encoding/decoding status messages:
  - before: percent_encode_decode_10k_status_messages: 67,63,62,63,65,62,63,65,63,63
  - after: percent_encode_decode_10k_status_messages: 49,43,43,43,43,43,43,45,44,44

### DIFF
--- a/Sources/GRPC/GRPCStatusMessageMarshaller.swift
+++ b/Sources/GRPC/GRPCStatusMessageMarshaller.swift
@@ -20,7 +20,7 @@ public struct GRPCStatusMessageMarshaller {
   /// - Parameter message: Message to percent encode.
   /// - Returns: Percent encoded string, or `nil` if it could not be encoded.
   public static func marshall(_ message: String) -> String? {
-    return String(decoding: percentEncode(message), as: Unicode.UTF8.self)
+    return percentEncode(message)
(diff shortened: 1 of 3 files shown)


glbrntt:
See also #644

glbrntt:
Thanks for that @MrMage! I added the length-check fast path since I reckon that's _very_ common and a pretty big win:

```
percent_encode_decode_10k_status_messages: 51,48,49,48,49,49,48,47,48,47
percent_encode_decode_10k_ascii_status_messages: 14,14,14,15,14,14,14,14,14,14
```

glbrntt:
Yes: https://github.com/grpc/grpc-swift/blob/nio/Tests/GRPCTests/GRPCStatusMessageMarshallerTests.swift

glbrntt:
Admittedly there aren't many of them though.

MrMage:
Thanks! I must have missed those.

LGTM from my side.

PR #653, opened 2019-12-11
Reduce allocations in the GRPCClientStateMachine

glbrntt:
Motivation:

There were some unnecessary allocations in the client state machine.

Modifications:

- Use `HPACKHeader`s new `first(name:)` method to avoid allocating an
  array of headers before taking the first.
- Turn LengthPrefixedMessageWriter into a struct.
- Turn LengthPrefixedMessageReader into a struct.

Result:

- Reduce allocations, a small perf gain.

### DIFF
--- a/Sources/GRPC/GRPCClientStateMachine.swift
+++ b/Sources/GRPC/GRPCClientStateMachine.swift
@@ -544,9 +544,11 @@ extension GRPCClientStateMachine.State {
     // responses as well as a variety of non-GRPC content-types and to omit Status & Status-Message.
     // Implementations must synthesize a Status & Status-Message to propagate to the application
     // layer when this occurs."
-    let responseStatus = headers.first(name: ":status").flatMap(Int.init).map { code in
-      HTTPResponseStatus(statusCode: code)
-    } ?? .preconditionFailed
+    let responseStatus = headers.first(name: ":status")
+      .flatMap(Int.init)
+      .map { code in
+        HTTPResponseStatus(statusCode: code)
+      } ?? .preconditionFailed
 
     guard responseStatus == .ok else {
       return .failure(.invalidHTTPStatus(responseStatus))
@@ -585,13 +587,14 @@ extension GRPCClientStateMachine.State {
   }
 
   private func readStatusCode(from trailers: HPACKHeaders) -> GRPCStatus.Code? {
-    return trailers.first(name: GRPCHeaderName.statusCode).flatMap(Int.init).flatMap(GRPCStatus.Code.init)
+    return trailers.first(name: GRPCHeaderName.statusCode)
+      .flatMap(Int.init)
+      .flatMap(GRPCStatus.Code.init)
   }
 
   private func readStatusMessage(from trailers: HPACKHeaders) -> String? {
-    return trailers.first(name: GRPCHeaderName.statusMessage).map {
-      GRPCStatusMessageMarshaller.unmarshall($0)
-    }
+    return trailers.first(name: GRPCHeaderName.statusMessage)
+      .map(GRPCStatusMessageMarshaller.unmarshall)
   }
 
   /// Parses a "Trailers-Only" response from the server into a `GRPCStatus`.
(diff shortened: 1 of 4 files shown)


glbrntt:
Before: 
- embedded_client_unary_10k_small_requests: 372,371,354,371,379,371,372,375,385,381

After:
- embedded_client_unary_10k_small_requests: 341,340,343,345,343,342,344,354,352,340

glbrntt:
I also had more targeted benchmark for just the state machine (which involved making a bunch of things public) which saw a decrease from ~60ms to ~50ms.

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -544,21 +544,20 @@ extension GRPCClientStateMachine.State {
     // responses as well as a variety of non-GRPC content-types and to omit Status & Status-Message.
     // Implementations must synthesize a Status & Status-Message to propagate to the application
     // layer when this occurs."
-    let statusHeader = headers[":status"].first
-    let responseStatus = statusHeader.flatMap(Int.init).map { code in
+    let responseStatus = headers.first(name: ":status").flatMap(Int.init).map { code in
       HTTPResponseStatus(statusCode: code)
```
MrMage: `.map(HTTPResponseStatus.init)`?

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -586,14 +585,13 @@ extension GRPCClientStateMachine.State {
   }
 
   private func readStatusCode(from trailers: HPACKHeaders) -> GRPCStatus.Code? {
-    return trailers[GRPCHeaderName.statusCode].first
-      .flatMap(Int.init)
-      .flatMap(GRPCStatus.Code.init)
+    return trailers.first(name: GRPCHeaderName.statusCode).flatMap(Int.init).flatMap(GRPCStatus.Code.init)
```
MrMage: Nit: keep the newlines?

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -586,14 +585,13 @@ extension GRPCClientStateMachine.State {
...
   }
 
   private func readStatusMessage(from trailers: HPACKHeaders) -> String? {
-    return trailers[GRPCHeaderName.statusMessage].first
-      .map(GRPCStatusMessageMarshaller.unmarshall)
+    return trailers.first(name: GRPCHeaderName.statusMessage).map {
```
MrMage: `.map(GRPCStatusMessageMarshaller.unmarshall)`?

Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -609,10 +607,8 @@ extension GRPCClientStateMachine.State {
     // one from the ":status".
     //
     // See: https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md
-    let statusHeader = trailers[":status"].first
-    guard let status = statusHeader.flatMap(Int.init).map({ HTTPResponseStatus(statusCode: $0) })
-      else {
-        return .failure(.invalidHTTPStatus(nil))
+    guard let status = trailers.first(name: ":status").flatMap(Int.init).map({ HTTPResponseStatus(statusCode: $0) }) else {
```
MrMage: `.map(HTTPResponseStatus.init)`?

MrMage:
> I also had more targeted benchmark for just the state machine (which involved making a bunch of things public) which saw a decrease from ~60ms to ~50ms.

We could make those public if we declare them as not part of the public API.

Sources/GRPC/GRPCClientStateMachine.swift
glbrntt: Unfortunately the init has a default second arg: `init(statusCode: Int, reasonPhrase: String = "")` so this isn't possible.

glbrntt:
> > I also had more targeted benchmark for just the state machine (which involved making a bunch of things public) which saw a decrease from ~60ms to ~50ms.
> 
> We could make those public if we declare them as not part of the public API.

True, I think I'll leave that for another PR though!

PR #657, opened 2019-12-16
Make server CallHandlers more internal

glbrntt:
Motivation:

To avoid breaking API changes in the future we should keep our public
API surface as small as possible.

Modifications:

Make the server `BaseCallHandler` underscore-internal.

Result:

Smaller public API.

### DIFF
--- a/Sources/GRPC/CallHandlers/BidirectionalStreamingCallHandler.swift
+++ b/Sources/GRPC/CallHandlers/BidirectionalStreamingCallHandler.swift
@@ -25,7 +25,10 @@ import Logging
 ///   If the framework user wants to return a call error (e.g. in case of authentication failure),
 ///   they can fail the observer block future.
 /// - To close the call and send the status, complete `context.statusPromise`.
-public class BidirectionalStreamingCallHandler<RequestMessage: Message, ResponseMessage: Message>: BaseCallHandler<RequestMessage, ResponseMessage> {
+public class BidirectionalStreamingCallHandler<
+  RequestMessage: Message,
+  ResponseMessage: Message
+>: _BaseCallHandler<RequestMessage, ResponseMessage> {
   public typealias Context = StreamingResponseCallContext<ResponseMessage>
   public typealias EventObserver = (StreamEvent<RequestMessage>) -> Void
   public typealias EventObserverFactory = (Context) -> EventLoopFuture<EventObserver>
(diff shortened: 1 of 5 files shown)


Sources/GRPC/CallHandlers/ClientStreamingCallHandler.swift
```diff
@@ -71,7 +74,7 @@ public class ClientStreamingCallHandler<RequestMessage: Message, ResponseMessage
     }
   }
 
-  public override func handlerAdded(context: ChannelHandlerContext) {
+  override public func handlerAdded(context: ChannelHandlerContext) {
```
MrMage: Why was this order swapped, but not elsewhere? (I prefer visibility first, BTW, unless it violates a style guide.)

Sources/GRPC/CallHandlers/BidirectionalStreamingCallHandler.swift
```diff
@@ -25,7 +25,10 @@ import Logging
 ///   If the framework user wants to return a call error (e.g. in case of authentication failure),
 ///   they can fail the observer block future.
 /// - To close the call and send the status, complete `context.statusPromise`.
-public class BidirectionalStreamingCallHandler<RequestMessage: Message, ResponseMessage: Message>: BaseCallHandler<RequestMessage, ResponseMessage> {
+public class BidirectionalStreamingCallHandler<
```
MrMage: Do we consider these classes a public API or "an implementation detail of the codegen", i.e. could we issue a patch release that breaks compatibility _unless the user re-runs the codegen_?

MrMage:
Responding to https://github.com/grpc/grpc-swift/pull/657#discussion_r358270353:

SGTM. I think the SwiftProtobuf people also try to avoid breaking the generated code through runtime changes, and that should be doable for us, too — especially given that our generated code does not use too much of the runtime. (Speaking of which, is it possible to further reduce the amount of generated code?)

I like the idea of e.g. a "makeHandler" function, though, which _might_ let us hide the concrete implementation.

glbrntt:
> Responding to [#657 (comment)](https://github.com/grpc/grpc-swift/pull/657#discussion_r358270353):
> 
> SGTM. I think the SwiftProtobuf people also try to avoid breaking the generated code through runtime changes, and that should be doable for us, too — especially given that our generated code does not use too much of the runtime. (Speaking of which, is it possible to further reduce the amount of generated code?)

Possibly: also on my list of things-to-look-at related to public API

> I like the idea of e.g. a "makeHandler" function, though, which _might_ let us hide the concrete implementation.

🤞

PR #658, opened 2019-12-16
Make DelegatingErrorHandler underscored-internal

glbrntt:
See #657


glbrntt:
This lives at the end of the client pipeline and closes the channel after any error but also calls any client error delegate.

It's only "public" for testing.

MrMage:
Could we make it `internal` then and use it with `@testable import`?

PR #659, opened 2019-12-16
Make TLSVerificationHandler internal

glbrntt:
See #657

### DIFF
--- a/Sources/GRPC/TLSVerificationHandler.swift
+++ b/Sources/GRPC/TLSVerificationHandler.swift
@@ -38,7 +38,7 @@ public enum GRPCApplicationProtocolIdentifier: String, CaseIterable {
 ///
 /// On fulfillment of the promise this handler is removed from the channel pipeline.
 internal class TLSVerificationHandler: ChannelInboundHandler, RemovableChannelHandler {
-  internal typealias InboundIn = Any
+  typealias InboundIn = Any
 
   private let logger = Logger(subsystem: .clientChannel)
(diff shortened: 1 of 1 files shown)


Sources/GRPC/TLSVerificationHandler.swift
```diff
@@ -37,8 +37,8 @@ public enum GRPCApplicationProtocolIdentifier: String, CaseIterable {
 /// Users of this handler should rely on the `verification` future held by this instance.
 ///
 /// On fulfillment of the promise this handler is removed from the channel pipeline.
-public class TLSVerificationHandler: ChannelInboundHandler, RemovableChannelHandler {
-  public typealias InboundIn = Any
+internal class TLSVerificationHandler: ChannelInboundHandler, RemovableChannelHandler {
+  internal typealias InboundIn = Any
```
MrMage: Given that the entire class is `internal`, shall we drop that modifier on the internal fields and methods?

glbrntt: I prefer the explicit `internal` but fine changing if you prefer?

MrMage: Not feeling too strongly, but it does feel quite redundant to me if the same visibility is already specified on the enclosing scope. I think Swift even warns for these redundant visibilities in some cases.

glbrntt: Haven't seen the warnings but let's remove them anyway.

PR #661, opened 2019-12-16
Make HTTPProtocolSwitcher internal

glbrntt:
See #657


weissi:
@MrMage I've got no real tool but `swift api-digester` can spit out `.json` files that contain a full listing. But I'm not sure how easy that is to consume :).

PR #662, opened 2019-12-16
Make GRPCClientChannelHandler underscore internal

glbrntt:
See #657


glbrntt:
It's used in one of the perf tests.

MrMage:
Ah, thanks. Then we can't `@testable` import it, I guess. But could you document why it is public at all in the comment stating that it is not public API?

glbrntt:
> Ah, thanks. Then we can't `@testable` import it, I guess. But could you document why it is public at all in the comment stating that it is not public API?

Done!

PR #663, opened 2019-12-16
Make Server*ResponseCallContextImpl internal

glbrntt:
See #657

### DIFF
--- a/Sources/GRPC/ServerCallContexts/StreamingResponseCallContext.swift
+++ b/Sources/GRPC/ServerCallContexts/StreamingResponseCallContext.swift
@@ -25,7 +25,7 @@ import Logging
 /// - If `statusPromise` is failed and the error is of type `GRPCStatusTransformable`,
 ///   the result of `error.asGRPCStatus()` will be returned to the client.
 /// - If `error.asGRPCStatus()` is not available, `GRPCStatus.processingError` is returned to the client.
-open class StreamingResponseCallContext<ResponseMessage: Message>: ServerCallContextBase {
+public class StreamingResponseCallContext<ResponseMessage: Message>: ServerCallContextBase {
   public typealias WrappedResponse = GRPCServerResponsePart<ResponseMessage>
 
   public let statusPromise: EventLoopPromise<GRPCStatus>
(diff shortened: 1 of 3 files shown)


Sources/GRPC/ServerCallContexts/StreamingResponseCallContext.swift
```diff
@@ -80,7 +80,7 @@ open class StreamingResponseCallContextImpl<ResponseMessage: Message>: Streaming
 /// Concrete implementation of `StreamingResponseCallContext` used for testing.
 ///
 /// Simply records all sent messages.
-open class StreamingResponseCallContextTestStub<ResponseMessage: Message>: StreamingResponseCallContext<ResponseMessage> {
+public class StreamingResponseCallContextTestStub<ResponseMessage: Message>: StreamingResponseCallContext<ResponseMessage> {
```
MrMage: If you make this public, the members should be, as well. However, I had made this `open` to let library users override individual methods if needed for more complex tests. WDYT?

Sources/GRPC/ServerCallContexts/UnaryResponseCallContext.swift
```diff
@@ -93,4 +93,4 @@ open class UnaryResponseCallContextImpl<ResponseMessage: Message>: UnaryResponse
 /// Concrete implementation of `UnaryResponseCallContext` used for testing.
 ///
 /// Only provided to make it clear in tests that no "real" implementation is used.
-open class UnaryResponseCallContextTestStub<ResponseMessage: Message>: UnaryResponseCallContext<ResponseMessage> { }
+public class UnaryResponseCallContextTestStub<ResponseMessage: Message>: UnaryResponseCallContext<ResponseMessage> { }
```
MrMage: Same as in the other file.

Sources/GRPC/ServerCallContexts/StreamingResponseCallContext.swift
glbrntt: If we make it `open` then any super classes must be `open` as well. Maybe that's not actually a problem here since the API is very small (I'd be okay dropping this PR).

MrMage: Hmm, not sure either way. We could make it `public` for now and add a comment that we would be open (sic!) to making it `open` if requested? I.e. they should file an issue about this if they need it?

glbrntt: Let's hold off on this change for now, given https://github.com/grpc/grpc-swift/pull/657#discussion_r358270353 this might become irrelevant.

PR #669, opened 2019-12-19
Make gRPC error types extensible

glbrntt:
Motivation:

Adding new cases to an enum is a SemVer major change. To allow for
flexibility moving forward we should make our error types extensible so
that we can add cases without having to tag a major release.

Modifications:

- `GRPCError` is now a `struct` backed by the private `GRPCBaseError` `enum`
- `GRPCError` has static members and functions which map to the cases in
  `GRPCBaseError`.
- Updated usages of `GRPCError`

Result:

Our public facing error type, `GRPCError` can now have new 'cases' added
without requiring a SemVer major change.

### DIFF
--- a/Sources/GRPC/CallHandlers/ServerStreamingCallHandler.swift
+++ b/Sources/GRPC/CallHandlers/ServerStreamingCallHandler.swift
@@ -57,7 +57,7 @@ public final class ServerStreamingCallHandler<
     guard let eventObserver = self.eventObserver,
       let callContext = self.callContext else {
         self.logger.error("processMessage(_:) called before the call started or after the call completed")
-        throw GRPCError.server(.tooManyRequests)
+        throw GRPCError.StreamCardinalityViolation(stream: .request).captureContext()
     }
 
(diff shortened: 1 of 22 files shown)


glbrntt:
cc @weissi

MrMage:
@weissi, would you mind taking this review? I'm a bit stretched for time right now, sorry.

glbrntt:
@MrMage fyi weissi gave me some useful feedback offline. I should be able to get to this this week.

glbrntt:
cc @lukasa

glbrntt:
They're only exposed to users via error delegates, and that is without the context. In some cases the information _provided by_ the context is also exposed to the user (but not the context itself).

If we carried that context for all errors, what context should be provided by an error fired by e.g. NIOSSL? At the moment the delegate is just called with placeholder values when context isn't available.

Lukasa:
Ok, if they're only exposed in their unwrapped form then I am happy enough with this design, and I can do a fuller review.

PR #673, opened 2019-12-19
Make LengthPrefixedMessageWriter internal

glbrntt:
See #657

### DIFF
--- a/Sources/GRPC/LengthPrefixedMessageWriter.swift
+++ b/Sources/GRPC/LengthPrefixedMessageWriter.swift
@@ -16,12 +16,12 @@
 import Foundation
 import NIO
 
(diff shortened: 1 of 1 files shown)


Sources/GRPC/LengthPrefixedMessageWriter.swift
```diff
@@ -16,12 +16,12 @@
 import Foundation
 import NIO
 
-public struct LengthPrefixedMessageWriter {
-  public static let metadataLength = 5
+internal struct LengthPrefixedMessageWriter {
+  internal static let metadataLength = 5
```
MrMage: nit: drop the `internal` qualifiers?

Issue #621, opened 2019-11-06
[nio] at least under TSan, flaky test GRPCTests.FunctionalTestsAnonymousClientNIOTS testBidirectionalStreamingLotsOfMessagesBatched


weissi:
### New Issue Checklist

- [x] Updated grpc-swift to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/grpc/grpc-swift/blob/master/CONTRIBUTING.md)
- [x] I searched for [existing GitHub issues](https://github.com/grpc/grpc-swift/issues)(especially, there are many [resolved issues related to SSL](https://github.com/grpc/grpc-swift/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+ssl))

### Issue Description

```
$ swift test --sanitize=thread --disable-index-store
[...]
Test Case '-[GRPCTests.FunctionalTestsAnonymousClientNIOTS testBidirectionalStreamingLotsOfMessagesBatched]' started.
/private/tmp/.nio-release-tools_N0Glvg/grpc-swift/Tests/GRPCTests/FunctionalTests.swift:189: error: -[GRPCTests.FunctionalTestsAnonymousClientNIOTS testBidirectionalStreamingLotsOfMessagesBatched] : Asynchronous wait failed: Exceeded timeout of 15 seconds, with unfulfilled expectations: "Expecting 5000 response(s)", "Expecting status received".
/private/tmp/.nio-release-tools_N0Glvg/grpc-swift/Tests/GRPCTests/FunctionalTests.swift:202: error: -[GRPCTests.FunctionalTestsAnonymousClientNIOTS testBidirectionalStreamingLotsOfMessagesBatched] : XCTAssertEqual failed: ("ok") is not equal to ("unavailable")
Test Case '-[GRPCTests.FunctionalTestsAnonymousClientNIOTS testBidirectionalStreamingLotsOfMessagesBatched]' failed (15.596 seconds).
```

the OS was macOS Catalina

- NIO: fbe0c8a6b8a968e449f1d4facd825e4069120efb
- grpc swift: 2cd9ea7008312db794cb62d805f5ace043496ead

glbrntt:
It _looks_ like this is just timing out. Maybe we finally need to do #437 @MrMage / or increase the timeouts?

MrMage:
I agree that we should probably have an option to exclude time sensitive tests; do you have a suggestion on how to best implement that?

glbrntt:
[...]

Issue #437, opened 2019-04-17
some tests are timing sensitive and they usually fail for me in docker


weissi:
some tests are timing sensitive and they usually fail for me in docker. Likely, there's not much that we can do except for offering a way to disable them. I'd suggest an env var

    ENABLE_TIMING_TESTS=false

that would disable them. Attached the failure trace. Head is

```
$ head /tmp/test-unary-lots.txt
Test Case 'NIOFunctionalTestsAnonymousClient.testUnaryLotsOfRequests' started at 2019-04-17 11:04:39.318



1000 requests sent so far, elapsed time: 17.819645
total time to send 2000 requests: 62.545991
/Users/johannes/.Trash/.nio-release-tools_WPakBo/grpc-swift/Tests/SwiftGRPCNIOTests/NIOFunctionalTests.swift:97: error: NIOFunctionalTestsAnonymousClient.testUnaryLotsOfRequests : failed - Expecteded 'SwiftGRPCNIOTests.Echo_EchoResponse:
text: "Swift echo get: foo 400"
' but received error: GRPCStatus(code: SwiftGRPCNIO.StatusCode.deadlineExceeded, message: Optional("call exceeded timeout of 1 M"), trailingMetadata: [])
/Users/johannes/.Trash/.nio-release-tools_WPakBo/grpc-swift/Tests/SwiftGRPCNIOTests/NIOFunctionalTests.swift:98: error: NIOFunctionalTestsAnonymousClient.testUnaryLotsOfRequests : XCTAssertEqual failed: ("ok") is not equal to ("deadlineExceeded") - 
```

[test-unary-lots.txt](https://github.com/grpc/grpc-swift/files/3089297/test-unary-lots.txt)

environment:

`swift:5.0` docker image with these versions and gprpc-commit 283754e80bb30a630e5e06508af654bc1d68fd7d

```
root@b4a7ae06c069:/Users/johannes/.Trash/.nio-release-tools_WPakBo/grpc-swift# cat Package.resolved 
{
  "object": {
    "pins": [
      {
        "package": "Commander",
        "repositoryURL": "https://github.com/kylef/Commander.git",
        "state": {
          "branch": null,
          "revision": "e5b50ad7b2e91eeb828393e89b03577b16be7db9",
          "version": "0.8.0"
        }
      },
      {
        "package": "Spectre",
        "repositoryURL": "https://github.com/kylef/Spectre.git",
        "state": {
          "branch": null,
[...]

PR #677, opened 2019-12-19
Add toggle for time-sensitive tests.

glbrntt:
Motivation:

Sometimes, like when running under TSan, it's useful to disable long
running tests, or those which are sensitive to time.

Modifications:

Add an option to disable time sensitive tests via an envrionment
variable.

Result:

Certain tests are skipped if `ENABLE_TIMING_TESTS` is set to false.

### DIFF
--- a/Tests/GRPCTests/GRPCTestCase.swift
+++ b/Tests/GRPCTests/GRPCTestCase.swift
@@ -24,7 +24,26 @@ class GRPCTestCase: XCTestCase {
   // locally; conditionally enable it based on the environment.
   //
   // https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
-  private static var isLoggingEnabled = ProcessInfo.processInfo.environment["CI"] != "true"
+  private static let isLoggingEnabled = !Bool(
+      fromTruthLike: ProcessInfo.processInfo.environment["CI"],
+      defaultingTo: false
+  )
+
+  private static let runTimeSensitiveTests = Bool(
+      fromTruthLike: ProcessInfo.processInfo.environment["ENABLE_TIMING_TESTS"],
+      defaultingTo: true
+  )
+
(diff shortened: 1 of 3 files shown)


glbrntt:
Fixes #621

glbrntt:
Fixes #437

Tests/GRPCTests/GRPCTestCase.swift
```diff
@@ -24,7 +24,26 @@ class GRPCTestCase: XCTestCase {
   // locally; conditionally enable it based on the environment.
   //
   // https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
-  private static var isLoggingEnabled = ProcessInfo.processInfo.environment["CI"] != "true"
+  private static let isLoggingEnabled = !Bool(
```
MrMage: `!Bool` looks a bit ugly — can we avoid that?

Tests/GRPCTests/GRPCTestCase.swift
```diff
@@ -46,6 +65,14 @@ class GRPCTestCase: XCTestCase {
     super.setUp()
     XCTAssertTrue(GRPCTestCase.isLoggingConfigured)
   }
+
+  var runTimeSensitiveTests: Bool {
+    let shouldRun = GRPCTestCase.runTimeSensitiveTests
+    if !shouldRun {
+      print("Skipping '\(self.name)' as ENABLE_TIMING_TESTS=false")
```
MrMage: This side effect feels like a code small; at least convert this into a `func`? Also, what will `self.name` evaluate to?

glbrntt: Sure, let's make it a func. I'm okay with the side effect as it's only a print message in tests.

`self.name` looks like "-[FunctionalTestsMutualAuthentication testUnaryLotsOfRequests]"

Issue #680, opened 2019-12-20
[tls] negotiating application protocol invalid


WilliamIzzo83:
Hello

I'm working on an iOS app from some time now, and I've decided to switch to the new nio branch. I've got some problems with tls though. 

Some background first: we use tls for secure connection. The code that is using grpc-swift master branch does something like this

```swift
let client = MyClient(address: configuration.backendHostName, secure: configuration.needsSecureConnection)
```

And this works as intended.

With the new api exposed on the nio branch, things have changed a little. Therefore now I've got:

```swift
let eloop = PlatformSupport.makeEventLoopGroup(
    loopCount: 1,
    networkPreference: .best)

let target = ConnectionTarget.hostAndPort(configuration.backendHostName, configuration.port)
let conf = ClientConnection.Configuration(
      target: target,
      eventLoopGroup: eloop,
      errorDelegate: nil,
      connectivityStateDelegate: nil,
      tls: configuration.needsSecureConnection ? ClientConnection.Configuration.TLS() : nil,
      connectionBackoff: nil)
    
let conn = ClientConnection(configuration: conf)
let callOpts = CallOptions(customMetadata: [:],
    timeout: try! .minutes(10),
    cacheable: false,
    requestIDProvider: .autogenerated, requestIDHeader: nil)

let client = MyClient(connection: conn, defaultCallOptions: callOpts)
```

Unfortunately I'm getting the following

```
2019-12-20T12:34:41+0100 error: negotiated application protocol is invalid: nil
2019-12-20T12:34:41+0100 warning: error=GRPCError(wrappedError: GRPC.GRPCClientError.applicationLevelProtocolNegotiationFailed, origin: GRPC.GRPCError.Origin.client, file: "/Users/William/Library/Developer/Xcode/DerivedData/sphere-dmhvcadfhsxqgbglhgytyjsalwoi/SourcePackages/checkouts/grpc-swift/Sources/GRPC/TLSVerificationHandler.swift", line: 94) connection_id=6EB998B5-81A8-416E-AF24-D472437A012E client connection shutdown failed
```

I guess I need some guidance to understand what is going on

Thank you

glbrntt:
During the TLS handshake the application layer protocol that the client and server will communicate using will also be negotiated. This has the unsurprising name of application layer protocol negotiation (ALPN) https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation.

[...]

PR #681, opened 2019-12-20
Fix typo in the "grpc-exp" ALPN name

glbrntt:
Motivation:

gRPC has an experimental ALPN name "grpc-exp". Unfortunately, we spelled
it as "grpc-ext".

Fortunately if "grpc-exp" is specifid, then so too must "h2", so we
would always still manage to negotiate a protocol.

Modifications:

Fix typo.

Result:

We can now negotiate "grpc-exp" via ALPN.


glbrntt:
See #680

Issue #615, opened 2019-10-28
0.10.0 build fails using Carthage


nakkht:
### New Issue Checklist

- [x] Updated grpc-swift to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/grpc/grpc-swift/blob/master/CONTRIBUTING.md)
- [x] I searched for [existing GitHub issues](https://github.com/grpc/grpc-swift/issues)(especially, there are many [resolved issues related to SSL](https://github.com/grpc/grpc-swift/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+ssl))

### Issue Description

Having `github "grpc/grpc-swift" == "0.10.0"` in `Cartfile` and `github "grpc/grpc-swift" "0.10.0"` in `Cartfile.resolved` and building using `carthage bootstrap --platform iOS` fails and yields the following output:

```
*** Checking out grpc-swift at "0.10.0"
*** xcodebuild output can be found in /var/folders/2y/t66flsfj7dg0bb3zym3d807m0000gn/T/carthage-xcodebuild.qb2r03.log
*** Building scheme "SwiftGRPC-Package" in SwiftGRPC-Carthage.xcodeproj
Build Failed
	Task failed with exit code 65:
	/usr/bin/xcrun xcodebuild -project /Users/.../Carthage/Checkouts/grpc-swift/SwiftGRPC-Carthage.xcodeproj -scheme SwiftGRPC-Package -configuration Release -derivedDataPath /Users/.../Library/Caches/org.carthage.CarthageKit/DerivedData/11.1_11A1027/grpc-swift/0.10.0 -sdk iphoneos ONLY_ACTIVE_ARCH=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES archive -archivePath /var/folders/2y/t66flsfj7dg0bb3zym3d807m0000gn/T/grpc-swift SKIP_INSTALL=YES GCC_INSTRUMENT_PROGRAM_FLOW_ARCS=NO CLANG_ENABLE_CODE_COVERAGE=NO STRIP_INSTALLED_PRODUCT=NO (launched in /Users/.../Carthage/Checkouts/grpc-swift)

This usually indicates that project itself failed to compile. Please check the xcodebuild log for more details: /var/folders/2y/t66flsfj7dg0bb3zym3d807m0000gn/T/carthage-xcodebuild.qb2r03.log
```
[...]

Issue #478, opened 2019-06-05
Xcode11Beta:  Import of C++ module 'BoringSSL' appears within extern "C" language linkage specification


fabstu:
### New Issue Checklist

- [x] Updated grpc-swift to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/grpc/grpc-swift/blob/master/CONTRIBUTING.md)
- [x] I searched for [existing GitHub issues](https://github.com/grpc/grpc-swift/issues)(especially, there are many [resolved issues related to SSL](https://github.com/grpc/grpc-swift/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+ssl))

### Issue Description

Hello! I wanted to see whether grpc-swift works with the spm integrated inside Xcode11 Beta. I created a new project via File -> New -> Swift Package and added `SwiftGRPC` as a dependency. Is it supposed to work differently or an Xcode11Beta bug?

```
Import of C++ module 'BoringSSL' appears within extern "C" language linkage specification
```

Here is an image of it:
https://imgur.com/a/fKcuk65

My `Package.swift` is:

```
// swift-tools-version:5.1
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "SwiftPackageTryout",
    products: [
        // Products define the executables and libraries produced by a package, and make them visible to other packages.
        .library(
            name: "SwiftPackageTryoutLibrary",
            targets: ["SwiftPackageTryout"]),
    ],
    dependencies: [
        .package(url: "https://github.com/grpc/grpc-swift", .branch("master")),
        .package(url: "https://github.com/apple/swift-protobuf", .branch("master")),
        // Dependencies declare other packages that this package depends on.
    ],
    targets: [
        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
        // Targets can depend on other targets in this package, and on products in packages which this package depends on.
        .target(
            name: "SwiftPackageTryout",
            dependencies: ["SwiftGRPC"]),
        .testTarget(
            name: "SwiftPackageTryoutTests",
            dependencies: ["SwiftPackageTryout", "SwiftGRPC"]),
    ]
)
```

##### Complete output when running grpc-swift, including the stack trace and command used

<details>
  <pre>Fails at build already</pre>
</details>

### Environment

<details>
  <pre>
  
| Key                           | Value                                                         |
| ----------------------------- | --------------------------------------------------------------|
[...]

PR #682, opened 2019-12-20
SwiftGrpc Swift Package Manager compile fix

nakkht:
Hi,

Based on #478 issue and suggested solution, here's the fix for CgRPC compilation using SwiftPM.
This also should give an option for people, who has an issue using Carthage #615 and are not able to use CocoaPods.

Notable changes:
- Main `Package.swift` set to use `swift-tools-version:5.0` tools and `Package@swift-4.2.swift` is kept for backwards compatibility
- .gitignore upates to include .swiftpm
- `CFLAGS` removed from Makefile's test script


Package.swift
```diff
@@ -47,54 +48,65 @@ let isLinux = false
...
+                path: "Sources/Examples/Simple"),
+        .testTarget(name: "SwiftGRPCTests", dependencies: ["SwiftGRPC"])
+    ],
+    swiftLanguageVersions: [.v4, .v4_2, .version("5")],
+    cLanguageStandard: .gnu11,
+    cxxLanguageStandard: .cxx11)
+
+if isLinux {
+    package.products.append(
+        .library(name: "SwiftGRPCNIO", targets: ["SwiftGRPCNIO"])
```
glbrntt: @MrMage what do you think about constraining the NIO version to Linux? Given it's a _really_ old version I don't _think_ it should matter. We could even remove it altogether from `master`?

Package.swift
```diff
@@ -47,54 +48,65 @@ let isLinux = false
...
+                    "SwiftGRPC",
+                    "SwiftProtobuf",
+                    "Commander"],
+                path: "Sources/Examples/Echo"),
+        .target(name: "Simple",
+                dependencies: ["SwiftGRPC", "Commander"],
+                path: "Sources/Examples/Simple"),
+        .testTarget(name: "SwiftGRPCTests", dependencies: ["SwiftGRPC"])
+    ],
+    swiftLanguageVersions: [.v4, .v4_2, .version("5")],
```
glbrntt: nit: `.v5` since we now have `swift-tools-version:5.0`

Package.swift
```diff
@@ -18,23 +18,24 @@ import PackageDescription
...
+    .package(url: "https://github.com/kylef/Commander.git", .upToNextMinor(from: "0.8.0")),
+    
+    // SwiftGRPCNIO dependencies:
+    // Transitive dependencies
+    .package(url: "https://github.com/apple/swift-nio-zlib-support.git", .upToNextMajor(from: "1.0.0")),
+    .package(url: "https://github.com/apple/swift-nio-nghttp2-support.git", .upToNextMajor(from: "1.0.0")),
+    // Main SwiftNIO package
+    .package(url: "https://github.com/apple/swift-nio.git", .upToNextMajor(from: "1.12.0")),
+    // HTTP2 via SwiftNIO
+    .package(url: "https://github.com/apple/swift-nio-http2.git", .upToNextMinor(from: "0.2.1"))
```
glbrntt: nit: can you remove the extra space here?

Package.swift
nakkht: Done.

Package.swift
MrMage: I have also been thinking of just removing the old NIO version from `master` because it causes a bunch of problems. Let's go ahead with that, I guess.

nakkht: @MrMage I suppose tests will have to be removed for linux with swift 5.0 as well? Should it be done in a separate PR?

MrMage: I would suggest removing the entire `SwiftGRPCNIO` directories and targets (for all platforms). I think we could do that in this PR. We could temporarily keep them in a separate branch for those who still need it.

nakkht: @MrMage alright, I've removed all NIO related packages/targets. As a consequence I also removed [linux entry](https://github.com/grpc/grpc-swift/pull/682/files#diff-354f30a63fb0907d4ad57269548329e3L29) from `.travis.yml` file for Swift 5.

MrMage: I don't think you should remove the Linux entry from .travis.yml. But you should remove `Sources/GRPCNIO` and `Tests/GRPCNIOTests` and update the test main file accordingly.

nakkht: @MrMage if `Sources/GRPCNIO` and `Tests/GRPCNIOTest` gets removed, `Package@swift-4.2.swift` won't work, does it mean it has to be removed as well?

MrMage: Why would that not work? Why not just remove the NIO packages from all Package.swift files?

glbrntt: The proposed outcome here is to have no `nio`-based version on the master branch at all, so the related targets and dependencies also need to be removed from `Package.swift` and `Package@swift-4.2.swift`.

Makefile
```diff
@@ -28,7 +28,7 @@ project-carthage:
 	@ruby patch-carthage-project.rb SwiftGRPC-Carthage.xcodeproj || echo "xcodeproj ('sudo gem install xcodeproj') is required in order to generate the Carthage-compatible project!"
 
 test: all
-	swift test $(CFLAGS)
+	swift test
```
glbrntt: There are a couple of `nio` related targets in this Makefile, could you delete those as well?

nakkht: Should be done.

Package.swift
nakkht: @MrMage what I was referring to was linux build: if you remove all NIO dependencies (including `swift-nio-ssl-support`) from 4.2 then linux won't compile. Do you expect it to compile?

@glbrntt new change-set should contain no nio related dependencies.

glbrntt:
@nakkht looks like CI is failing for Linux on 4.2 - I suspect the removal of the `CFLAGS` from the `Makefile` is the culprit.

Package.swift
```diff
@@ -18,83 +18,46 @@ import PackageDescription
...
-
-// On Linux, Foundation links with openssl, so we'll need to use that instead of BoringSSL.
-// See https://github.com/apple/swift-nio-ssl/issues/16#issuecomment-392705505 for details.
-// swift build doesn't pass -Xswiftc flags to dependencies, so here using an environment variable
-// is easiest.
-if isLinux || ProcessInfo.processInfo.environment.keys.contains("GRPC_USE_OPENSSL") {
-  packageDependencies.append(.package(url: "https://github.com/apple/swift-nio-ssl-support.git", from: "1.0.0"))
-} else {
-  cGRPCDependencies.append("BoringSSL")
-}
```
glbrntt: @nakkht I think is the source of your build failure: we're now _always_ depending on BoringSSL instead of conditionally depending on the platform.

nakkht: @glbrntt does the same apply when using `Package@swift-4.2`?

glbrntt: I believe so, yes.

nakkht: @glbrntt how about `ProcessInfo.processInfo.environment.keys.contains("GRPC_USE_OPENSSL")`? Is it still have to be included?

glbrntt: I believe so, yes.

Package@swift-4.2.swift
```diff
@@ -0,0 +1,67 @@
...
+  .package(url: "https://github.com/apple/swift-protobuf.git", .upToNextMajor(from: "1.5.0")),
+  
+  // Command line argument parser for our auxiliary command line tools.
+  .package(url: "https://github.com/kylef/Commander.git", .upToNextMinor(from: "0.8.0"))
+]
+
+var cGRPCDependencies: [Target.Dependency] = []
+
+if ProcessInfo.processInfo.environment.keys.contains("GRPC_USE_OPENSSL") {
+  packageDependencies.append(.package(url: "https://github.com/apple/swift-nio-ssl-support.git", from: "1.0.0"))
```
glbrntt: This is actually a NIO dependency so it isn't necessary.

For the CgRPC dependencies, what we want is "if we're using linux or `GRPC_USE_OPENSSL` is set then don't depend on BoringSSL".

nakkht: I didn't manage to make it work without this NIO dependency (which actually is just libressl)
I've found @MrMage PR from some time ago regarding this, but even then this dependency was included: #238 
I've brought back `CFLAGS` used for linux .

glbrntt: Oh sorry, I was completely wrong there, I didn't realise it was the support package.

PR #685, opened 2020-01-07
Updating TLSVerificationHandler

WilliamIzzo83:
Motivation:

When tls handshake is over there is no need to raise a failure if no
protocol has been negotiated.

Modifications:

The ```TLSVerificationHandler.userInboundEventTriggered``` method has
been updated so that it always succeed. However it will notice when a
protocol could not be negotiated

Result:

This change allows a successful connection even when a protocol is not
negotiated. By doing so we are more consistent with other grpc
implementations.

This is related to #680


Sources/GRPC/TLSVerificationHandler.swift
```diff
@@ -85,14 +85,12 @@ internal class TLSVerificationHandler: ChannelInboundHandler, RemovableChannelHa
         return
     }
 
-    self.logger.debug("TLS handshake completed, negotiated protocol: \(String(describing: negotiatedProtocol))")
     if let proto = negotiatedProtocol, GRPCApplicationProtocolIdentifier(rawValue: proto) != nil {
```
glbrntt: Can you remove the check against `GRPCApplicationProtocolIdentifier` since we know that the protocol we agreed upon must be one that we suggested in the first place? 

There's an edge case here (which we didn't handle correctly before) wherein the user could override the suggested protocols to include one that isn't in `GRPCApplicationProtocolIdentifier`, if that was the case now we'd log "unknown protocol" which isn't quite true.

Sources/GRPC/TLSVerificationHandler.swift
```diff
@@ -85,14 +85,12 @@ internal class TLSVerificationHandler: ChannelInboundHandler, RemovableChannelHa
...
-      self.logger.debug("negotiated application protocol is valid")
-      self.verificationPromise.succeed(())
+      self.logger.debug("TLS handshake completed, negotiated protocol: \(proto)")
     } else {
-      self.logger.error("negotiated application protocol is invalid: \(String(describing: negotiatedProtocol))")
-      let error = GRPCError.client(.applicationLevelProtocolNegotiationFailed)
-      self.verificationPromise.fail(error)
+      self.logger.debug("TLS handshake completed, unknown protocol")
```
glbrntt: nit: It's more like "no protocol negotiated" rather than unknown.

Tests/GRPCTests/TLSVerificationHandlerTests.swift
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright 2019, gRPC Authors All rights reserved.
```
glbrntt:
```suggestion
 * Copyright 2020, gRPC Authors All rights reserved.
```

Tests/GRPCTests/TLSVerificationHandlerTests.swift
```diff
@@ -0,0 +1,59 @@
...
+    let tlsVerificationHandler = TLSVerificationHandler()
+    let channel = EmbeddedChannel(handler: tlsVerificationHandler)
+    var expectations = [XCTestExpectation]()
+    
+    GRPCApplicationProtocolIdentifier.allCases.forEach {
+      let exp = self.expectation(description: "tls \(String(describing:$0)) protocol success")
+      expectations.append(exp)
+      let handshakeEvent = TLSUserEvent.handshakeCompleted(negotiatedProtocol: $0.rawValue)
+      channel.pipeline.fireUserInboundEventTriggered(handshakeEvent)
+      tlsVerificationHandler.verification.assertSuccess(fulfill: exp)
```
glbrntt: This won't do _quite_ what you expect: there's only one instance of the `tlsVerificationHandler` here so only one `verification` future. The future will be fulfilled in the first iteration of this loop: subsequent fulfilment of the futures promise in the handler will have no effect here so the `expectation` from each iteration will only depend on the first protocol you try.

Sources/GRPC/TLSVerificationHandler.swift
WilliamIzzo83: Ok!

Sources/GRPC/TLSVerificationHandler.swift
WilliamIzzo83: Yep

Tests/GRPCTests/TLSVerificationHandlerTests.swift
WilliamIzzo83: Oh dear! Sorry, I guess I need to get used to the idea :D

Tests/GRPCTests/TLSVerificationHandlerTests.swift
WilliamIzzo83: Nice catch! I'll have a verification handler for each protocol

WilliamIzzo83:
Glad to be helpful!

PR #688, opened 2020-01-08
Simplify compression representation.

glbrntt:
Motivation:

Since we don't support compression, there's no need for any scaffolding
that we do have to be public. Moreover there's no need to over
complicate something which isn't yet user facing.

Modifications:

Remove `CompressionMechanism` and replace it with
`CompressionAlgorithm`; an enum containing all compression algorithms we
support. At the moment this is limited to `identity`.

Result:

Simpler core, smaller public API.


Sources/GRPC/LengthPrefixedMessageWriter.swift
```diff
@@ -19,10 +19,15 @@ import NIO
 internal struct LengthPrefixedMessageWriter {
   static let metadataLength = 5
 
-  private let compression: CompressionMechanism
+  /// The compression algorithm to use, if one should be used.
+  private let compression: CompressionAlgorithm?
 
-  init(compression: CompressionMechanism) {
-    precondition(compression.supported, "compression mechanism \(compression) is not supported")
+  /// Whether the compression message bit should be written.
```
MrMage: "should be set", and rename field to e.g. `shouldSetCompression{Bit,Flag}`?

Tests/GRPCTests/GRPCClientStateMachineTests.swift
```diff
@@ -738,7 +738,7 @@ extension GRPCClientStateMachineTests {
     headers.add(name: "grpc-encoding", value: "not-a-known-compression-(probably)")
 
     stateMachine.receiveResponseHeaders(headers).assertFailure {
-      XCTAssertEqual($0, .unsupportedMessageEncoding("unknown"))
+      XCTAssertEqual($0, .unsupportedMessageEncoding("not-a-known-compression-(probably)"))
```
MrMage: I think you can remove the "probably" ;-)

PR #691, opened 2020-01-08
Update CI to run Swift 5.2 development snapshot

glbrntt:
Motivation:

We should run our CI against development versions of Swift so that we
can catch any language regressions, bugs, etc. before release.

Modifications:

- Add Swift Development 5.2 to run unit tests for all PRs.
- Allow the job to be failable.
- Update Xcode for CI to 11.3

Result:

Pre-release Swift running as part of CI on Linux.

### DIFF
--- a/.travis-install.sh
+++ b/.travis-install.sh
@@ -78,9 +78,12 @@ install_swift() {
   if [ "$TRAVIS_OS_NAME" != "osx" ]; then
     info "Installing Swift $SWIFT_VERSION"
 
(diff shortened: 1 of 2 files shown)


glbrntt:
Each job has a language, depending on that language additional keys are allowed in the configuration and various dependencies are brought in. If language is omitted it defaults to ruby (which is why the UI would says "Ruby" in a couple of places). Travis thinks that "swift" means xcodeproject / iOS app so brings in things like cocoa pods. Minimal is, well, minimal. Practically speaking I don't think the change will make a noticeable difference.

Issue #693, opened 2020-01-11
[nio] minor issue with project initialization


Kirow:
### New Issue Checklist

- [x] Updated grpc-swift to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/grpc/grpc-swift/blob/master/CONTRIBUTING.md)
- [x] I searched for [existing GitHub issues](https://github.com/grpc/grpc-swift/issues)(especially, there are many [resolved issues related to SSL](https://github.com/grpc/grpc-swift/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+ssl))

### Issue Description

1. Clone and make xcode project file:
```
git clone -b nio https://github.com/grpc/grpc-swift.git
cd grpc-swift
make project
```
Error - `No such file or directory -- fix-project-settings.rb (LoadError)`. This command should be removed from Makefile or `fix-project-settings.rb` should be restored.

2. Open project > Manage Scheme > add GRPC > build on iOS
Error - `Compiling for iOS 8.0, but module 'SwiftProtobuf' has a minimum deployment target of iOS 13.2`. Manual change will helps of course, but it could be that there is a setting for this somewhere and it is possible to make it `8.0` during initialization.

### Environment

<details>
  <pre>
  
| Key                           | Value                                                         |
| ----------------------------- | --------------------------------------------------------------|
| OS Version                    |  10.14.6                                                       |
| Swift Version                 |   5.1.3                                                       |
| Xcode Version                 |   11.3                                                       |
| gRPC-Swift Version            |  [nio](https://github.com/grpc/grpc-swift/commit/be70633c91d722496e5fcb225f822edbd5c36a5a)                                                      |
  </pre>
</details>

glbrntt:
Thanks for filing this @Kirow.

With regards to the second, I think the appropriate fix here is adding platform availability the Package.swift. Just as an FYI, the minimum supported iOS version should be 10.0 (this the minimum version of iOS supported by SwiftNIO).

I'll file a couple of PRs to address these.

glbrntt:
One other thing in case you didn't know, with Xcode 11 you don't need to generate an `.xcodeproj` (if you're only using SwiftPM). You can either open the `grpc-swift` directory from Xcode or you can run `xed .` from the `grpc-swift` directory.

Kirow:
[...]

PR #695, opened 2020-01-13
Remove fix-project-settings.rb references from Makefile

glbrntt:
Motivation:

The Makefile referenced a file which no longer exists.

Modifications:

Remove references to that file.

Result:

More happy project generation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ project: ${XCODEPROJ}
 
 ${XCODEPROJ}:
 	${SWIFT_PACKAGE} generate-xcodeproj --output $@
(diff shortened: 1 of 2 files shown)


glbrntt:
Fixes #693

glbrntt:
It was removed with all the other stale documentation/examples etc. 

To be honest, I removed it without thinking. Happy to put it back though if you think?

MrMage:
Hmm. It can help with some people sending wrongly-formatted PRs, for example. Our indentation settings are fairly non-standard.

I would suggest re-adding it.

glbrntt:
Re-added but to `scripts/`

PR #698, opened 2020-01-15
Update Swift 5.2 development snapshot.

glbrntt:
Motivation:

We should keep our pre-release swift builds up-to-date.

Modifications:

Update development snapshot to latest.

Result:

Up-to-date CI.


glbrntt:
I think it'd be better to keep it to a period update - every week or so - then it's easier to tell if the toolchain is bad vs the PR.

MrMage:
SGTM

PR #702, opened 2020-01-17
Fix bug with BoringSSLError.invalidSNIName exception

esen:
When IP address is given as a host to ClientConnection BoringSSLError.invalidSNIName exception is thrown from [NIOSSLClientHandler](https://github.com/apple/swift-nio-ssl/blob/4f9d64ddd04cd1fe97d74dad7b5a77b882d8bf5d/Sources/NIOSSL/NIOSSLClientHandler.swift#L42)

This pull request fixes this by passing `nil` to `NIOSSLClientHandler` if `serverHostname` is an IP address

### DIFF
--- a/Sources/GRPC/ClientConnection.swift
+++ b/Sources/GRPC/ClientConnection.swift
@@ -361,7 +361,7 @@ extension ClientConnection {
     logger: Logger
   ) -> ClientBootstrapProtocol {
     // Provide a server hostname if we're using TLS. Prefer the override.
-    let serverHostname: String? = configuration.tls.map {
+    var serverHostname: String? = configuration.tls.map {
       if let hostnameOverride = $0.hostnameOverride {
         logger.debug("using hostname override for TLS", metadata: ["server-hostname": "\(hostnameOverride)"])
         return hostnameOverride
@@ -370,6 +370,11 @@ extension ClientConnection {
         logger.debug("using host from connection target for TLS", metadata: ["server-hostname": "\(host)"])
         return host
       }
+    }
+    
+    if let hostname = serverHostname, hostname.isIPAddress {
+      logger.debug("IP address cannot be used for TLS SNI extension. No host used", metadata: ["server-hostname": "nil"])
+      serverHostname = nil
     }
 
     let bootstrap = PlatformSupport.makeClientBootstrap(group: eventLoop)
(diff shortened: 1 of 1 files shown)


Sources/GRPC/ClientConnection.swift
```diff
@@ -576,3 +576,16 @@ extension HTTP2ToHTTP1ClientCodec.HTTPProtocol {
     }
   }
 }
+
+fileprivate extension String {
```
glbrntt: Can you reformat this to match the style used elsewhere?

Sources/GRPC/ClientConnection.swift
```diff
@@ -540,7 +540,7 @@ fileprivate extension Channel {
     do {
       let sslClientHandler = try NIOSSLClientHandler(
         context: try NIOSSLContext(configuration: configuration),
-        serverHostname: serverHostname
+        serverHostname: serverHostname.flatMap { $0.isIPAddress ? nil : $0 }
```
glbrntt: I think this check is better done further up the call stack, specifically in `makeBootstrap` where we check if there is a hostname override. The rationale here is that it's easier for someone to debug if they pass in a hostname override which _is_ an IP address. It would also be useful to have a log message saying that we're not supplying a hostname override because we believe that it is an IP address.

Sources/GRPC/ClientConnection.swift
MrMage: Tip: `make project` will automatically configure the Xcode project to use the correct indentation.

Sources/GRPC/ClientConnection.swift
esen: That makes sense. I moved that check up to `makeBootstrap`. Also added a debug log message.

Sources/GRPC/ClientConnection.swift
```diff
@@ -371,6 +371,11 @@ extension ClientConnection {
         return host
       }
     }
+    
+    if let hostname = serverHostname, hostname.isIPAddress {
+      logger.debug("IP address cannot be used for TLS SNI extension. No host used", metadata: ["server-hostname": "nil"])
```
MrMage: I wonder if the metadata should contain the `hostname` value that has been rejected?

glbrntt: That's probably worth adding @esen

esen: done

MrMage: Thanks! Could you please use just `hostname` instead of `"\(hostname)"`?

esen: A bare string cannot be used here. `"\(hostname)"` is the [preferred](https://github.com/apple/swift-log/blob/master/Sources/Logging/Logging.swift#L289) way

MrMage: Oh, sorry!

Issue #611, opened 2019-10-16
[nio] Provide support for gzip compression?


be4rd:
### Question Checklist

- [x] Updated grpc-swift to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/grpc/grpc-swift/blob/master/CONTRIBUTING.md)
- [x] I read [README](https://github.com/grpc/grpc-swift/blob/master/README.md)
- [x] I searched for [existing GitHub issues](https://github.com/grpc/grpc-swift/issues) (especially, there are many [resolved issues related to SSL](https://github.com/grpc/grpc-swift/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+ssl))

### Question Subject
The question is whether you plan to implement support for __gzip__ compression for _nio_ at an upcoming release 1.0.0 or nearest future?

#### Question Description
Currently, the *nio* implementation [does not support gzip](https://github.com/grpc/grpc-swift/blob/nio/Sources/GRPC/CompressionMechanism.swift#L64) compression mechanism. For applications with intense usage of networking (such as messenger) the lack of support of this compression mechanism can result in high Network and CPU consumption, as well as higher battery draining on iOS devices. This prevents _nio_ implementation from being ready for production usage for such applications.

### Implementation notes
I have my guesses on where to look at first:
- [ ] Include __gzip__ in [list of supported mechanisms](https://github.com/grpc/grpc-swift/blob/nio/Sources/GRPC/CompressionMechanism.swift#L60)
- [ ] Switch over `compression` in [LengthPrefixedMessageWriter's](https://github.com/grpc/grpc-swift/blob/nio/Sources/GRPC/LengthPrefixedMessageWriter.swift#L42) `write` function and compress data properly according to algorithm
- [ ] Switch over `compression` in [LengthPrefixedMessageReader's](https://github.com/grpc/grpc-swift/blob/nio/Sources/GRPC/LengthPrefixedMessageReader.swift#L166) and unzip data properly, than append unzipped chunk to `ByteBuffer`

If you guys do not plan to do it in any nearest future, I would like to write implementation of __gzip__ support myself, if you'll support me with a little guidance.

MrMage:
I guess we would be open to a patch that adds gzip compression, provided that is:

1. Compatible to gRPC's compression spec,
2. Well-tested (including support for non-compressed messages on the same stream),
3. Easy to maintain,
[...]

PR #707, opened 2020-01-24
Add support for compression on the client

glbrntt:
Motivation:

Compression support is an optional part of the gRPC specification but
ideally should be supported. Users have also requested this feature
(#611).

Since introducing compression without breaking API is difficult we
should do this before we tag 1.0.0.

Modifications:

- Add support for gzip and deflate compression on the client.
- Introduce new configuration for supporting compression.
- Add interop tests for the client.
- Compression is enabled at the connection level (compression used for requests and compression methods to advertise to the server which it may use for responses)
- Request compression may be disabled at the RPC level or at the message level for streaming requests.

Result:

Client supports compression.


Sources/GRPC/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -267,14 +267,19 @@ extension HTTP1ToRawGRPCServerCodec: ChannelOutboundHandler {
         // Store the response into an independent buffer. We can't return the message directly as
         // it needs to be aggregated with all the responses plus the trailers, in order to have
         // the base64 response properly encoded in a single byte stream.
-        messageWriter.write(messageBytes, into: &self.responseTextBuffer)
+        //
+        // We can try! here because the server does not support compression at the moment and this
+        // only throws when compression fails.
+        try! messageWriter.write(messageBytes, into: &self.responseTextBuffer)
```
glbrntt: Just a note: I'll add support for the server after this PR is wrapped up, the `try!` will disappear then.

Sources/GRPC/ClientCalls/ServerStreamingCall.swift
```diff
@@ -52,6 +52,7 @@ public final class ServerStreamingCall<RequestMessage: Message, ResponseMessage:
       path: path,
       host: connection.configuration.target.host,
       requestID: requestID,
+      encoding: connection.configuration.compression,
```
MrMage: Can we unify the naming of encoding vs compression?

Sources/GRPC/ClientConnection.swift
```diff
@@ -476,6 +476,21 @@ extension ClientConnection {
...
+    ///
+    /// If compression is enabled for a connection it may be disabled for requests on any RPC by
+    /// setting `CallOptions.disableCompression` to `true`.
+    ///
+    /// Compression may also be disabled at the message-level for streaming requests (i.e. client
+    /// streaming and bidirectional streaming RPCs) by setting `disableCompression` to `true` in
+    /// `sendMessage(_:disableCompression)`, `sendMessage(_:disableCompression:promise)`,
+    /// `sendMessages(_:disableCompression)` or `sendMessages(_:disableCompression:promise)`.
+    ///
+    /// Note that any disabling compression has no effect if compression is disabled on the
```
MrMage: nit: "any disabling compression" sounds weird

Sources/GRPC/ClientOptions.swift
```diff
@@ -46,18 +46,24 @@ public struct CallOptions {
   /// messages associated with the call.
   public var requestIDHeader: String?
 
+  /// Disables request compression on this RPC. Ignored if compression is disabled at the
```
MrMage: nit: "on this call"?

Sources/GRPC/Compression/CompressionAlgorithm.swift
```diff
@@ -0,0 +1,52 @@
...
+ * limitations under the License.
+ */
+
+/// Supported message compression algorithms.
+///
+/// These algorithms are indicated in the "grpc-encoding" header. As such, a lack of "grpc-encoding"
+/// header indicates that there is no message compression.
+public struct CompressionAlgorithm: Equatable {
+  /// Identity compression; "no" compression but indicated via the "grpc-encoding" header.
+  public static let identity = CompressionAlgorithm(.identity)
```
MrMage: This is to avoid enum changes being breaking changes, I presume?

Sources/GRPC/Compression/MessageEncoding.swift
```diff
@@ -0,0 +1,45 @@
...
+    ) {
+      self.outbound = requests
+      self.inbound = responses
+    }
+
+    /// The compression algorithm used for outbound messages.
+    public internal(set) var outbound: CompressionAlgorithm?
+
+    /// The set of compression algorithms advertised to the remote peer that they may use.
+    public internal(set) var inbound: [CompressionAlgorithm]
```
MrMage: Not super happy with these names, neither "requests/responses" nor "outbound/inbound". How about "accepted" and "outbound"? Or something more analogous to the "accept-encoding" headers? Not 100% sure myself. Thinking about this more, maybe the names are actually good.

Also, does gRPC support using different compressions in both directions on one stream? Or is this more like the "preferred" outbound method?

Overall, I'm a bit confused by this so far 😅

Sources/GRPC/Compression/MessageEncoding.swift
```diff
@@ -0,0 +1,45 @@
...
+
+    /// No compression.
+    public static let none = MessageEncoding(requests: nil, responses: [.identity])
+
+    /// Accept all supported compression on responses, do not compress requests.
+    public static let onlyResponses = MessageEncoding(requests: nil, responses: CompressionAlgorithm.all)
```
MrMage: nit: `responsesOnly` (up to you)

Sources/GRPC/Compression/MessageEncoding.swift
```diff
@@ -0,0 +1,45 @@
...
+  }
+}
+
+extension ClientConnection.Configuration.MessageEncoding {
+  var acceptEncoding: String {
```
MrMage: `acceptEncodingHeader`?

Sources/GRPC/LengthPrefixedMessageReader.swift
```diff
@@ -31,13 +31,22 @@ import Logging
 /// - SeeAlso:
 /// [gRPC Protocol](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)
 internal struct LengthPrefixedMessageReader {
-  /// The mechanism that messages will be compressed with.
   var compression: CompressionAlgorithm?
+  var decompressor: Zlib.Inflate?
```
MrMage: Seems a bit hard-coded right now, but we can still refactor this later if we ever need to.

Sources/GRPC/LengthPrefixedMessageWriter.swift
```diff
@@ -39,12 +49,48 @@ internal struct LengthPrefixedMessageWriter {
...
-  func write(_ message: Data, into buffer: inout ByteBuffer) {
+  func write(_ message: Data, into buffer: inout ByteBuffer, disableCompression: Bool = false) throws {
     buffer.reserveCapacity(LengthPrefixedMessageWriter.metadataLength + message.count)
 
-    //! TODO: Add compression support, use the length and compressed content.
-    buffer.writeInteger(Int8(self.shouldSetCompressionFlag ? 1 : 0))
-    buffer.writeInteger(UInt32(message.count))
-    buffer.writeBytes(message)
+    if !disableCompression, let compressor = self.compressor {
+      buffer.writeInteger(UInt8(1))
```
MrMage: Mind adding a comment here as well?

Sources/GRPC/Compression/MessageEncoding.swift
glbrntt: Agh...naming here is tricky. To the user `requests`/`responses` seems more natural for the `init` than `inbound`/`outbound`, but then `responses` doesn't quite fit the "accept-encoding" part. Also `compression.requests`/`compression.responses` isn't as clear as `compression.outbound`, `compression.inbound`.

The `init` could be more explicit:

```swift
init(forRequests outbound: CompressionAlgorithm?, acceptableForResponses inbound: [CompressionAlgorithm]?) {
  // ...
}

...

MessageEncoding(forRequests: .gzip, acceptableForResponses: [.gzip, .identity])
```

I'm still not certain how this API should be spelt! Thoughts?

> Also, does gRPC support using different compressions in both directions on one stream? Or is this more like the "preferred" outbound method?

Compression for the request and response stream are independent and may be different.

Sources/GRPC/LengthPrefixedMessageReader.swift
glbrntt: Agreed. I'm okay with it for now, if we add compression not from zlib then we'll definitely need to change this.

Sources/GRPC/Compression/MessageEncoding.swift
glbrntt: One other thing I've been considering recently which might make things easier for us and users is some sort of client bootstrap: we can be more descriptive with names and add to it without breaking API (with structs we need a new `init` if we add a property).

```swift
let bootstrap = GRPCConnectionBootstrap(group: group)  // hmm sounds familiar...
  .requestCompression(.gzip)
  .acceptableResponseCompression(.gzip, .deflate, .identity)
```

Anyway, that's a story for another time!

MrMage: `MessageEncoding(forRequests: .gzip, acceptableForResponses: [.gzip, .identity])` SGTM. I also like the Bootstrap idea.

Sources/GRPC/Compression/Zlib.swift
```diff
@@ -0,0 +1,422 @@
...
+      self.stream = ZStream()
+      self.format = format
+      self.initialize()
+    }
+
+    deinit {
+      self.end()
+    }
+
+    /// Decompresses the data in `input` into the `output` buffer.
```
MrMage: Wait, shouldn't this method be about *compressing*?

Sources/GRPC/Compression/Zlib.swift
```diff
@@ -0,0 +1,422 @@
...
+            let inflateResult = try self.stream.inflate(
+              outputBuffer: CGRPCZlib_castVoidToBytefPointer(outputPointer.baseAddress!),
+              outputBufferSize: outputPointer.count
+            )
+
+            switch inflateResult.outcome {
+            case .complete:
+              inflated = true
+            case .outputBufferTooSmall:
+              outBufferSize *= 2
```
MrMage: Could we do better than this with an educated guess? Can't Zlib tell you how large the decompressed data is by e.g. reading the compression header?

Sources/GRPC/Compression/Zlib.swift
```diff
@@ -0,0 +1,422 @@
...
+  }
+
+  enum InflateResult {
+    case complete
+    case outputBufferTooSmall
+  }
+
+  // MARK: ZStream
+
+  struct ZStream {
```
MrMage: Mind documenting this type and possibly its properties?

Sources/GRPC/Compression/Zlib.swift
```diff
@@ -0,0 +1,422 @@
...
+        throw GRPCError.ZlibCompressionFailure(code: rc, message: self.lastErrorMessage).captureContext()
+      }
+
+      return InflateResult(
+        bytesWritten: outputBufferSize - self.availableOutputBytes,
+        outcome: outcome
+      )
+    }
+
+    mutating func deflate(
```
MrMage: If I understand this correctly, this method is used both for compressing and decompressing streams? Feels a bit weird to me, but I'm not a Zlib expert.

Sources/GRPC/Compression/Zlib.swift
glbrntt: Good catch, thanks!

Sources/GRPC/Compression/Zlib.swift
glbrntt: I couldn't find anything in the docs about this and there's nothing equivalent to `deflateBound` for `inflate`, it seems. In the gRPC lib they allocate 1KB slices until they have enough space but their API is better suited to dealing with slices anyway.

We could do something similar; decompress into multiple buffers, find how much capacity we need for one, and then allocate enough to write them all into a single buffer. That way we'd only allocate ~2x the total output size. 

I think we'd need to profile this some more first though. IIRC zlib suggests that the compression ratio for many use cases is between 2 and 5 so it's not clear cut to me that allocating multiple buffers would actually be better.

If you're okay with this for the time being I'll add a comment to explain this.

Sources/GRPC/Compression/Zlib.swift
glbrntt: Not quite; it's only used for compressing. "deflate" is a bit overloaded though, it refers to the compression format as well as the zlib method to compress something in that format. The `deflate` method can also be used to compress data in the gzip format (which is based on the deflate format).

The same is true for decompression which is done with `inflate`, it can handle "deflate" and "gzip" formatted data.

Sources/GRPC/Compression/MessageEncoding.swift
glbrntt: I'll leave the bootstrap for another day since that's a fairly large change.

glbrntt:
Missed this before:

> * It looks like the interop tests don't decompress anything; I assume that's expected because the other endpoint would be responsible for that?

They do decompress in the `ServerCompressed{Unary,Streaming}` tests (client sends uncompressed data, server responds with compressed data).

> * You could also add tests for `Zlib`, but I guess that is integration-tested by `CompressorTests`.

That's right. I didn't think there was much value in adding them.

Sources/GRPC/Compression/Zlib.swift
MrMage: This is completely fine then; not worth more effort. I was just hoping for an easy win ;-)

glbrntt:
@MrMage just pushed up a small patch, the compression documentation isn't clear in some places but working on the server part of this made me realise that something wasn't quite right. Here's the change: https://github.com/grpc/grpc-swift/pull/707/commits/ffe5bd4d575593d29cf6d020d3d2616ea45640c7

MrMage:
Still LGTM :-)

PR #714, opened 2020-01-30
Support compression on the server

glbrntt:
Motivation:

We have compression on the client; we should support it on the server as
well!

Modifications:

- Add server configuration to determine a preferred compression used for
  responses, and a list of compression algorithms enabled for requests.
- Add options to disable compression for responses on the RPC context
- Add an option to disable compression for a single response on the
  `sendResponse` method on the RPC context
- Add interop tests.
- Add a handful of other tests.

Result:

Server supports compression


Sources/GRPC/Compression/MessageEncoding.swift
```diff
@@ -49,3 +49,32 @@ extension ClientConnection.Configuration.MessageEncoding {
...
+  public struct MessageEncoding {
+    /// The preferred compression algorithm for response messages. If the client does not support
+    /// this algorithm then a mutually acceptable algorithm will be chosen, or responses will not
+    /// be compressed.
+    public var preferredResponseCompression: CompressionAlgorithm?
+
+    /// The set of compression algorithms advertised that we will accept from clients. Note that
+    /// clients may send us messages compressed with algorithms not included in this list, if we
+    /// support it then we still accept the message.
+    public var acceptableRequestCompression: [CompressionAlgorithm]
```
MrMage: nit: `compressionS`?

Sources/GRPC/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -252,10 +309,21 @@ extension HTTP1ToRawGRPCServerCodec: ChannelOutboundHandler {
         responseTextBuffer = context.channel.allocator.buffer(capacity: 0)
       }
 
+      // Are we compressing responses?
+      if let responseEncoding = self.responseEncodingHeader {
+        headers.add(name: GRPCHeaderName.encoding, value: responseEncoding)
+      }
+
+      // The client may have sent us a message using an encoding we didn't advertise, we'll send
```
MrMage: super-nit: ; instead of ,?

Sources/GRPC/Compression/MessageEncoding.swift
```diff
@@ -49,3 +49,32 @@ extension ClientConnection.Configuration.MessageEncoding {
...
+
+extension Server.Configuration {
+  public struct MessageEncoding {
+    /// The preferred compression algorithm for response messages. If the client does not support
+    /// this algorithm then a mutually acceptable algorithm will be chosen, or responses will not
+    /// be compressed.
+    public var preferredResponseCompression: CompressionAlgorithm?
+
+    /// The set of compression algorithms advertised that we will accept from clients. Note that
+    /// clients may send us messages compressed with algorithms not included in this list, if we
```
MrMage: super-nit: ; instead of ,?

Sources/GRPC/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -270,17 +338,24 @@ extension HTTP1ToRawGRPCServerCodec: ChannelOutboundHandler {
         //
         // We can try! here because the server does not support compression at the moment and this
         // only throws when compression fails.
-        try! messageWriter.write(messageBytes, into: &self.responseTextBuffer)
-
-        // Since we stored the written data, mark the write promise as successful so that the
-        // ServerStreaming provider continues sending the data.
-        promise?.succeed(())
+        do {
```
MrMage: Is this for gRPC-Web? If so, should we support compression for gRPC-Web at all, given that the recipient probably expects a non-binary response?

Sources/GRPC/HTTP1ToRawGRPCServerCodec.swift
```diff
@@ -342,3 +417,73 @@ extension HTTP1ToRawGRPCServerCodec: ChannelOutboundHandler {
...
+      return nil
+    }
+
+    // What do we both support?
+    let mutuallySupported = acceptableEncoding.compactMap {
+      CompressionAlgorithm(rawValue: String($0))
+    }
+
+    // Is our preferred mechanism there?
+    if mutuallySupported.contains(preference) {
```
MrMage: This sounds like _we_ pick the preferred encoding; shouldn't we instead try returning the first supported encoding requested by the client (similar to the `Accept-Language` header)? (I'm not that familiar with the gRPC spec; let me know if I'm mistaken here.)

Sources/GRPC/Server.swift
```diff
@@ -199,18 +211,21 @@ extension Server {
     ///     to handle requests.
     /// - Parameter errorDelegate: The error delegate, defaulting to a logging delegate.
     /// - Parameter tlsConfiguration: TLS configuration, defaulting to `nil`.
+    /// - Parameter messageEncoding: Message compression configuration, defaults to no compression.
```
MrMage: super-nit: either "; defaults" or ", defaulting"?

Sources/GRPCInteroperabilityTestsImplementation/InteroperabilityTestCases.swift
```diff
@@ -429,13 +429,20 @@ class ClientCompressedStreaming: InteroperabilityTest {
     let client = Grpc_Testing_TestServiceServiceClient(connection: connection)
 
     // Does the server support this test?
-    let probe = client.streamingInputCall()
+
+    // To find out we need to send an uncompressed probe. However we need to disable compression at
```
MrMage: nit: to find what out?

Sources/GRPC/HTTP1ToRawGRPCServerCodec.swift
glbrntt: The gRPC-Web spec doesn't mention disallowing compression. I just checked their repo as well, there's some code in there related to gzip so this shouldn't be an issue.

Sources/GRPC/HTTP1ToRawGRPCServerCodec.swift
glbrntt: The gRPC spec is _really_ unclear about compression, most of it relates to the API in libgrpc. I think we should probably do as you suggest, it's easier to understand and simplifies the configuration. Perhaps it would be better to have the configuration decide which compression algorithms are acceptable (defaulting to all). On the call context we could then give the user a choice to enable/disable compression per RPC (and per message on the `send` methods). Obviously if no compression was negotiated then enable/disable will be ignored. WDYT?

In fact there's probably no reason we can't do similar for the client, have a per RPC configuration instead of restricting it to the connection.

Sources/GRPCInteroperabilityTestsImplementation/InteroperabilityTestCases.swift
glbrntt: Whether the server supports this test (line above) -- these lines should be joined.

Sources/GRPC/HTTP1ToRawGRPCServerCodec.swift
MrMage: Sounds good to me, but to be honest I believe most users won't ever need to touch this, neither on the per-connection nor the per-RPC level. I would default to turning everything on and allowing to turn it off (either per-connection or per-RPC, both is fine with me).

Sources/GRPCInteroperabilityTestsImplementation/InteroperabilityTestCases.swift
MrMage: Right, sorry.

PR #717, opened 2020-02-06
Reorder some of the server logic.

glbrntt:
Motivation:

Recently we introduced the `GRPCPayload` protocol to abstract over the
request and response payload types so that we can support message types
other than Protobuf (such as Flatbuffers).

However, in doing so we removed our previous assumption that the payload
would (de)serialize from and to a `Data`. To avoid the needless copying
back and forth the payload writing (i.e. length prefixing) was moved to
the server codec (where serializion happens). This however requires the
type of the message to be known and therefore must happen after the
pipeline has been configured for the RPC in question.

In order to support compression on the server the message readers and
writers must be aware of the encoding being used and require access to
the headers. This happens (/happened) in the HTTP1ToRawGRPCServerCodec
before the RPC is configured. It would be possible to pass this through
with a user-event or extra data in the message type but would result in
fairly difficult to follow code with state spread across multiple
handlers.

Modifications:

Reorder the server channel operations.

- Previously the server channel did:
  1. verify content-type, handle message framing (`HTTP1ToRawGRPCServerCodec`)
  2. requests are routed (`GRPCChannelHandler`)
  3. requests are converted to typed gRPC messages (`GRPCServerCodec`)
  4. requests are passed to user handlers
- Now the server does:
  1. requests are routed and validated as gRPC
     (`GRPCServerRequestRoutingHandler`)
  2. requests are converted from HTTP/1, message framing is handler, messages
     are converted typed gRPC messages (`HTTP1ToGRPCServerCodec`)
  3. requests are passed to user handlers

As such:
- `GRPCChannelHandler` was renamed `GRPCServerRequestRoutingHandler`
- `GRPCServerRequestRoutingHandler` now reads `HTTPServerRequestPart` and
  writes `HTTPServerResponsePart`
- `GRPCServerRequestRoutingHandler` verifies the content-type and then
  constructs the pipeline
- `HTTP1ToRawGRPCServerCodec` is renamed `HTTP1ToGRPCServerCodec` and is
  now generic over the request and response type
- `HTTP1ToGRPCServerCodec` now comes after `GRPCServerRequestRoutingHandler`
- `HTTP1ToGRPCServerCodec` deals with message framing as well as
  (de/)serializion
- The `HTTPRequestHead` is now passed from the `HTTP1ToGRPCServerCodec`
  to the `BaseCallHandler`, which now passes it to the appropriate call
  handler (instead of passing it through the generated code as part of
  the call context).
- Tests updated, dead code removed.

This also paves the way for providing a server state machine, similar to
the client (i.e. it would essentially replace the contents of
`HTTP1ToGRPCServerCodec`).

Result:

Functionally, there should be no change. However this change makes it
possible for compression to be added to the server.


Sources/GRPC/GRPCServerRequestRoutingHandler.swift
```diff
@@ -0,0 +1,241 @@
...
+public struct CallHandlerContext {
+  internal var errorDelegate: ServerErrorDelegate?
+  internal var logger: Logger
+}
+
+/// Attempts to route a request to a user-provided call handler. Also validates that the request has
+/// a suitable 'content-type' for gRPC.
+///
+/// Once the request headers are available, asks the `CallHandlerProvider` corresponding to the request's service name
+/// for an `GRPCCallHandler` object. That object is then forwarded the individual gRPC messages.
```
MrMage: nit: a

Sources/GRPC/GRPCServerRequestRoutingHandler.swift
```diff
@@ -0,0 +1,241 @@
...
+///
+/// After the pipeline has been configured with the `GRPCCallHandler` this handler removes itself
```
MrMage: nit: comma before "this"?

Sources/GRPC/GRPCServerRequestRoutingHandler.swift
```diff
@@ -0,0 +1,241 @@
...
+      let callHandler = providerForServiceName.handleMethod(uriComponents[2], callHandlerContext: context) else {
+        self.logger.notice("could not create handler", metadata: ["path": "\(requestHead.uri)"])
+        return nil
+    }
+    return callHandler
+  }
+
+  private func makeResponseHead(requestHead: HTTPRequestHead, status: GRPCStatus) -> HTTPResponseHead {
+    var headers: HTTPHeaders = [
+      "content-type": "application/grpc",
```
MrMage: should we use string constants for both the key and value?

Sources/GRPC/HTTP1ToGRPCServerCodec.swift
```diff
@@ -157,15 +150,15 @@ extension HTTP1ToRawGRPCServerCodec: ChannelInboundHandler {
       "version": "\(requestHead.version)"
     ])
 
-    if let contentType = requestHead.headers["content-type"].first.flatMap(ContentType.init) {
+    if let contentType = requestHead.headers.first(name: "content-type").flatMap(ContentType.init) {
```
MrMage: magic string here as well

Sources/GRPC/Server.swift
```diff
@@ -45,51 +45,43 @@ import Logging
...
+///                   ByteBuffer│                             │ByteBuffer
+///                           ┌─┴─────────────────────────────▼─┐
+///                           │          NIOSSLHandler          │
+///                           └─▲─────────────────────────────┬─┘
+///                   ByteBuffer│                             │ByteBuffer
+///                             │                             ▼
 ///
-///    The GPRCChannelHandler resolves the request head and configures the rest of the pipeline
-///    based on the RPC call being made.
+///    The GRPCServerRequestRoutingHandler resolves the request head and configures the rest of the
```
MrMage: nit: wrap this class name (and possibly others) in backticks?

Tests/GRPCTests/HTTP1ToGRPCServerCodecTests.swift
```diff
@@ -0,0 +1,99 @@
...
+  func testDeserializationErrorOnInvalidMessageBytes() throws {
+    XCTAssertNoThrow(try self.channel.writeInbound(HTTPServerRequestPart.head(self.makeRequestHead())))
+    var buffer = self.channel.allocator.buffer(capacity: 0)
+    buffer.writeInteger(UInt8(0))  // not compressed
+    buffer.writeInteger(UInt32(3)) // message is 3 bytes
+    buffer.writeBytes([42, 42, 42])
+    XCTAssertThrowsError(try self.channel.writeInbound(HTTPServerRequestPart.body(buffer))) { error in
+      let withContext = error as? GRPCError.WithContext
+      XCTAssertTrue(withContext?.error is GRPCError.DeserializationFailure)
+      XCTAssertEqual(withContext?.error.makeGRPCStatus().code, .internalError)
```
MrMage: nit: I prefer to put the expectation first (also elsewhere I guess) — that's what I learned at Google. I don't want to force my will onto the entire project, though.

Tests/GRPCTests/HTTP1ToGRPCServerCodecTests.swift
```diff
@@ -0,0 +1,99 @@
...
+      let requestPart = try self.channel.readInbound(as: _GRPCServerRequestPart<Echo_EchoRequest>.self)
+      switch requestPart {
+      case .message(let actual):
+        XCTAssertEqual(message, actual)
+      default:
+        XCTFail("Unexpected request part: \(String(describing: requestPart))")
+      }
+    }
+  }
+
```
MrMage: nit: extra newline?

Tests/GRPCTests/HTTP1ToGRPCServerCodecTests.swift
```diff
@@ -0,0 +1,99 @@
...
+    buffer.writeInteger(UInt32(3)) // message is 3 bytes
+    buffer.writeBytes([42, 42, 42])
+    XCTAssertThrowsError(try self.channel.writeInbound(HTTPServerRequestPart.body(buffer))) { error in
+      let withContext = error as? GRPCError.WithContext
+      XCTAssertTrue(withContext?.error is GRPCError.DeserializationFailure)
+      XCTAssertEqual(withContext?.error.makeGRPCStatus().code, .internalError)
+    }
+  }
+
+  func testMultipleMessagesFromSingleBodyPart() throws {
```
MrMage: Would SingleMessageFromMultipleBodyParts also need testing?

Sources/GRPC/CallHandlers/BidirectionalStreamingCallHandler.swift
```diff
@@ -33,62 +33,45 @@ public class BidirectionalStreamingCallHandler<
...
+  private var eventObserver: EventLoopFuture<EventObserver>?
+  private let eventObserverFactory: (StreamingResponseCallContext<ResponsePayload>) -> EventLoopFuture<EventObserver>
 
   // We ask for a future of type `EventObserver` to allow the framework user to e.g. asynchronously authenticate a call.
   // If authentication fails, they can simply fail the observer future, which causes the call to be terminated.
   public init(
     callHandlerContext: CallHandlerContext,
     eventObserverFactory: @escaping (StreamingResponseCallContext<ResponsePayload>) -> EventLoopFuture<EventObserver>
   ) {
-    // Delay the creation of the event observer until `handlerAdded(context:)`, otherwise it is
```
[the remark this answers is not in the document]
glbrntt: I don't think so: it's now no longer possible to make the observer until `processHead` because we don't pass it in the `CallHandlerContext`. The problem we had before was that the status promise could fail before the handler reached the pipeline; now it _must_ be in the pipeline when the observer is created.

Sources/GRPC/CallHandlers/BidirectionalStreamingCallHandler.swift
```diff
@@ -33,62 +33,45 @@ public class BidirectionalStreamingCallHandler<
...
-
-    let eventObserver = factory(callContext)
-    self.observerState = .created(eventObserver)
-
-    // Terminate the call if the future providing an observer fails.
-    // This is being done _after_ we have been added as a handler to ensure that the `GRPCServerCodec` required to
-    // translate our outgoing `GRPCServerResponsePart<ResponsePayload>` message is already present on the channel.
-    // Otherwise, our `OutboundOut` type would not match the `OutboundIn` type of the next handler on the channel.
-    eventObserver.cascadeFailure(to: callContext.statusPromise)
+    context.writeAndFlush(self.wrapOutboundOut(.headers([:])), promise: nil)
```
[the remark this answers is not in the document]
glbrntt: It is, and yes, absolutely we do.

Sources/GRPC/GRPCContentType.swift
```diff
@@ -44,7 +44,7 @@ internal enum ContentType {
   var canonicalValue: String {
     switch self {
     case .protobuf:
-      return "application/grpc+proto"
+      return "application/grpc"
```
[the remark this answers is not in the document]
glbrntt: This was motivated by: https://github.com/grpc/grpc-swift/issues/642#issuecomment-561231923

tl;dr: "application/grpc+proto" is absolutely permitted by the spec, but support for it in the wild isn't all there; seems like "application/grpc" is the real canonical value here.

Sources/GRPC/GRPCServerRequestRoutingHandler.swift
```diff
@@ -0,0 +1,241 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import Foundation
+import SwiftProtobuf
+import NIO
+import NIOHTTP1
+import Logging
+
+/// Processes individual gRPC messages and stream-close events on a HTTP2 channel.
```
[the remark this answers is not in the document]
glbrntt: Hmm... I think I would say "an aytch-tee-tee-pee" instead of "a  haytch-tee-tee-pee". I didn't change this from `GRPCChannelHandler` (I guess git thinks it's a new file since there were a lot of changes?) but I'll change it to "an".

Sources/GRPC/GRPCServerRequestRoutingHandler.swift
glbrntt: Good call.

Tests/GRPCTests/HTTP1ToGRPCServerCodecTests.swift
glbrntt: I do it this way for practical reasons: the type inference comes from the first arg in `XCTAssertEqual`.

There are two useful aspects to this:
- if the left hand side is non-optional, then the right hand side must be also be non-options. If the left hand side is optional, the right hand side doesn't have to be optional.
- the type inference saves some typing; i.e. we can have `.internalError` instead of `GRPCStatus.Code.internalError`

Tests/GRPCTests/HTTP1ToGRPCServerCodecTests.swift
glbrntt: Probably, yes. Will update.

Sources/GRPC/CallHandlers/BidirectionalStreamingCallHandler.swift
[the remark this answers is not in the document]
MrMage: But the delaying part kind of still applies, doesn't it? That's what I was thinking about. I.e. just explain why we need a factory at all.

Sources/GRPC/GRPCServerRequestRoutingHandler.swift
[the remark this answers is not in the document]
MrMage: Yeah, totally depends on whether you are one of the "haytch" Britons or any other English language speaker I guess 😅

Sources/GRPC/GRPCServerRequestRoutingHandler.swift
MrMage: Did you check that these constants are being used everywhere now?

Tests/GRPCTests/HTTP1ToGRPCServerCodecTests.swift
MrMage: I could have sworn that I had both of these in my tests with the other order as well, are you sure?

E.g. I thought had written tests like this in the past:

```
XCTAssertEqual(0, optionalInt)
XCTAssertEqual(.someCase, someEnum)
```

Sources/GRPC/CallHandlers/BidirectionalStreamingCallHandler.swift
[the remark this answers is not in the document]
glbrntt: Oh I see, makes sense to explain that then.

Tests/GRPCTests/HTTP1ToGRPCServerCodecTests.swift
glbrntt: Not so sure now, no. Seems to work. I definitely hit this problem in the past, maybe there was a bug before 🤷‍♂️

Issue #638, opened 2019-11-21
Decode SwiftProtobuf messages with ContiguousBytes


glbrntt:
SwiftProtobuf now supports decoding from `ContiguousBytes`: https://github.com/apple/swift-protobuf/pull/915.

Once this change makes it into a release we should make use of this to decode our `Message`s from `ByteBufferView`s instead of going via `Data`.

PR #719, opened 2020-02-07
Add tests for custom payloads

glbrntt:
Motivation:

We recently added support for arbitrary payloads which conform to
`GRPCPayload`, however, we didn't have any tests demonstrating this.

Modifications:

- Add a test for custom payloads types
- Add a test for payload which skips deserialization
- Update Protobuf to 1.8.0 and deserialize from ContiguousBytes

Result:

Better tests.


glbrntt:
Fixes #638

glbrntt:
It's surprisingly little work! Would be nice to have a more generic codegen to save the boilerplate but not a big deal for now.

PR #727, opened 2020-02-25
Put the client connection behind a protocol

glbrntt:
Motivation:

Generated client stubs may only be initialized using a
`ClientConnection` which makes the assumption that a connection will
only ever be singular, and that all RPCs for that stub will run on a
single event loop.

However, there are scenarios where we may want to drive a stub and have
multiple connections (load balancing), or no real connection at all
(testing).

Modifications:

- Generated client stubs now depend on the implementation of a `GRPCChannel`
- The first argument label for the `init` in generated clients has
  changed from `connection` to `client`
- Added fixits for the above
- Updated generated code
- `GRPCChannel`s contract ensures that RPCs can be made but do not
  specify the transport.
- `GRPCClient` now requires a `GRPCChannel`
- `GRPCClient` provides higher level wrappers for the factory methods on
  `GRPCChannel` (i.e. provide defaults such as for call options)

Result:

- We have a better ability, in the future, to use already-generated
  stubs with different `GRPCChannel`s.
- Cleaner abstraction between call and connection.


Sources/GRPC/ClientCalls/ServerStreamingCall.swift
```diff
@@ -25,32 +24,39 @@ import Logging
 /// - `status`: the status of the gRPC call after it has ended,
 /// - `trailingMetadata`: any metadata returned from the server alongside the `status`.
 public final class ServerStreamingCall<RequestPayload: GRPCPayload, ResponsePayload: GRPCPayload>: BaseClientCall<RequestPayload, ResponsePayload> {
-  public init(
-    connection: ClientConnection,
+
```
MrMage: nit: extra newline

Sources/GRPC/ClientConnection.swift
```diff
@@ -265,7 +296,7 @@ extension ClientConnection {
   /// - Parameter backoffIterator: An `Iterator` for `ConnectionBackoff` providing a sequence of
   ///     connection timeouts and backoff to use when attempting to create a connection.
   private class func makeChannel(
-    configuration: Configuration,
+    configuration: ClientConnection.Configuration,
```
MrMage: Is the `ClientConnection.` prefix necessary?

Sources/GRPC/Shims.swift
```diff
@@ -47,3 +47,15 @@ extension GRPCStatusTransformable {
     return self.makeGRPCStatus()
   }
 }
+
+extension GRPCClient {
+  @available(*, deprecated, renamed: "init(channel:defaultCallOptions:)", message: "You may also need to regenerate your client code")
+  public init(connection: ClientConnection, defaultCallOptions: CallOptions = CallOptions()) {
```
MrMage: I assume we will delete these before 1.0?

glbrntt: That's right, this whole file will be removed (there's a TODO at the top of it)

Sources/GRPC/ClientConnection.swift
glbrntt: Nope, will remove them, that was WIP that didn't get reverted.

glbrntt:
> I didn't find much 😅 Seems like there is less happening here than I initially anticipated ;-)

Afraid so; nothing too exciting going on here!

PR #729, opened 2020-02-27
Add a decompression limit

glbrntt:
Motivation:

Compression can be an attack vector for maclicious users. Specifically,
bad actors may send 'zip-bombs' which have very high compression
ratios which may consume a large amount of memory when decompressed.

Modifications:

- Intoruce a `DecompressionLimit` which has two options, an `absolute`
  (an absolute size limit for a decompressed message), and `ratio`
  (based on the compression ratio of the message).
- Add decompression limits to server and client compression
  configuration
- Explicitly make compression `enabled` or `disabled`

Result:

Easier to protect against zip-bombs


glbrntt:
> Part of this might be technically unrelated refactorings that could just fit well with the PR itself, of course. 

There's definitely been some of that. I'll try to avoid that in the future to make the PRs smaller.

> But in some cases, the amount of changes needed for a particular addition simply feels surprising. That amount of "code churn" also makes it a bit harder to recognize the "essence" of the PR in a few cases. To counter that, though, I have been trying to put that "churn" aside and focus on the parts of the PR where actual stuff happens.

I noticed this too; but I think the changes in functionality have been fairly substantial (compression, support for non-protobuf payloads and support for different channels). We didn't plan on doing these at all (apart from compression) so I'm not too worried about how large they were.

> Also, I am starting to feel that I "lose track" of the overall codebase and the ability to fully evaluate changes (which could also explain the recent decline in comments from me). I guess that can't be fully avoided if one mostly reviews the code without actually working inside the codebase, though.

Totally fair; we need to find another core dev :)

PR #733, opened 2020-03-03
Configure client connections with a builder

glbrntt:
Motivation:

Configuring a client without TLS is easy to do, since no tls is the
default in the client connection configuration. This is bad. The API
should make it clearer that the default is insecure.

The current API for configuring clients falls a bit short in a number of
areas. In no particular order:

1. It can't be easily extended without SemVer major changes (since we'd
   have to provide a new init whenever new properties are added),
2. Nested configuration (i.e. TLS) is unnecessarily verbose,
3. The initializer requires properties to be passed in the correct
   order, and
4. TLS is disabled by default (!) and the API doesn't make this clear that
   this is insecure

Modifications:

- create a `ClientConnectionBuilder` whose public API is limited to
  configuring the private properties; this is similar to NIOs bootstraps
- provide static methods on `ClientConnection` to provide a `secure` or
  `insecure` builder
- update tests in a couple of places to demonstrate usage

Result:

- Configuring a connection is less verbose
- Not using TLS is more obviously 'insecure'

### DIFF
--- a/Sources/GRPC/ClientConnection.swift
+++ b/Sources/GRPC/ClientConnection.swift
@@ -101,6 +101,9 @@ public class ClientConnection {
   }
 
   /// Creates a new connection from the given configuration.
+  ///
+  /// - Important: Users should prefer using `ClientConnection.secure(group:)` to build a connection
+  ///   with TLS, or `ClientConnection.insecure(group:)` to build a connection without TLS.
   public init(configuration: Configuration) {
     self.configuration = configuration
     self.scheme = configuration.tls == nil ? "http" : "https"
--- a/Sources/GRPC/GRPCChannel/GRPCChannelBuilder.swift
+++ b/Sources/GRPC/GRPCChannel/GRPCChannelBuilder.swift
@@ -70,7 +70,7 @@ extension ClientConnection.Builder {
   /// 1 second if not set.
   @discardableResult
   public func withConnectionBackoff(initial amount: TimeAmount) -> Self {
-    self.connectionBackoff.initialBackoff = Double(amount.nanoseconds) / Double(1_000_000_000)
+    self.connectionBackoff.initialBackoff = .seconds(from: amount)
     return self
   }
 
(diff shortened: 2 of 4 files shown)


Sources/GRPC/ClientConnection.swift
```diff
@@ -101,6 +101,9 @@ public class ClientConnection {
   }
 
   /// Creates a new connection from the given configuration.
+  ///
+  /// - Important: Users should prefer using `ClientConnection.secure(group:)` to build a connection
+  ///   with TLS, or `ClientConnection.insecure(group:)` to build a connection without TLS.
   public init(configuration: Configuration) {
```
Davidde94: This is still alpha right? You *could* make an API-breaking change to hide these initialisers and *force* people to use the new `secure` and `insecure`?

Sources/GRPC/GRPCChannel/GRPCChannelBuilder.swift
```diff
@@ -0,0 +1,188 @@
...
+  public func withConnectionTimeout(minimum amount: TimeAmount) -> Self {
+    self.connectionBackoff.minimumConnectionTimeout = Double(amount.nanoseconds) / Double(1_000_000_000)
+    return self
+  }
+
+  /// Sets the initial and maximum backoff to given amount. Disables jitter and sets the backoff
+  /// multiplier to 1.0.
+  @discardableResult
+  public func withConnectionBackoff(fixed amount: TimeAmount) -> Self {
+    let seconds = Double(amount.nanoseconds) / Double(1_000_000_000)
```
Davidde94: Constant 1_000_000_000 is used a couple of times. Perhaps make attached it to `Double` to reduce potential error? (E.g. Double.Nanosecond, or just make it a constant somewhere)

Sources/GRPC/ClientConnection.swift
glbrntt: That's right but it's impossible to provide fix-its for this and it's quite an annoying change to have to make. I'm leaning towards keeping it for now. The longer term plan would be to not add anything to it to gently force people to the builder.

be4rd: Please do not hide this initialiser, as this would disallow precise configuration by hands, which is often needed.

glbrntt: @be4rd what's your use case here? Everything should be representable by the builder as well.

Davidde94: @glbrntt if you'd like to keep it another option to dissuade use would be to deprecate or provide a warning for direct use

be4rd: @glbrntt sorry, did not saw TLS configuration methods in builder when I left a comment 😅

Issue #735, opened 2020-03-06
Service naming


Banck:
Hello!
The google says, that service should named like `DeviceService`
We have proto like in google example:
![telegram-cloud-photo-size-2-5411162737578454437-y](https://user-images.githubusercontent.com/6259254/76085053-d987ec00-5fe3-11ea-98e0-e77c3379d38c.jpg)
`service DevicesService {}`

But swift grpc generate protocol: `internal protocol DevicesServiceService` and class: `internal final class DevicesServiceServiceClient`

Is it possible to somehow override name?

glbrntt:
Hey @Banck, 

I'm afraid it's not possible to override the name of the generated code, however you can create `typealias`es to work around this.

Also regarding the naming the Google Cloud API design guide says [here](https://cloud.google.com/apis/design/naming_convention#interface_names) that:

> In the rare case where an interface name would conflict with another name within the API, a suffix (for example Api or Service) should be used to disambiguate.

In other words, having `Service` in the name of the service is not recommended. However, I do think that we can eliminate the `Service` from `ServiceClient`, seems like bad naming to me!

Banck:
Hm...you are right, API Design guide says another than Proto3  Language guide. 
Thanks, we'll remove `Service` from names. And will be waiting for remove `Service` from `ServiceClient`

PR #736, opened 2020-03-06
Rename generated clients from FooServiceClient to FooClient

glbrntt:
Motivation:

Issue #735 highlighted that the naming of generated clients is clunky.
Let's improve it before the API is locked down.

Note: it's possible to create a shim to create fix-its for this by creating a
typealias from the old name to the new name, and then deprecating the
typealias:

```swift
@available(*, deprecated, renamed: "Echo_EchoClient")
typealias Echo_EchoServiceClient = Echo_EchoClient
```

Modifications:

- Update codegen
- Regenerate clients and update usages

Result:

Generate clients have better names


glbrntt:
cc @Davidde94

Issue #738, opened 2020-03-07
GRPCProtobufPayload issue with import proto in another proto


Banck:
Hello!
Let's imagine we have 2 (or more) protos. And in one proto we imported another one.
In this case, in every *.grpc.swift will be generated extensions for the same messages with conform GRPCProtobufPayload protocol.
For example:
Devices has RegisterRequest and ResponseRequest.
And Video looks like:
```
syntax = "proto3";

option java_package = "*";

import "google/protobuf/empty.proto";
import "devices_services.proto";

service Videos {
    rpc test (RegisterRequest) returns (RegisterResponse);
}
```

Then in `devices_services.grpc` will be 
```
/// Provides conformance to `GRPCPayload` for the request and response messages
extension RegisterRequest: GRPCProtobufPayload {}
extension RegisterResponse: GRPCProtobufPayload {}

```

and in videos_services.grpc will be the same lines:
```
/// Provides conformance to `GRPCPayload` for the request and response messages
extension RegisterRequest: GRPCProtobufPayload {}
extension RegisterResponse: GRPCProtobufPayload {}
```

glbrntt:
Hey @Banck -- thanks for the report; I'm aware of this and will look into it. 

As a precautionary note: as the grpc-swift API hasn't fully stabilised yet you should prefer pinning to a tagged version, i.e. `1.0.0-alpha.9`, and build the protoc plugin at that same tag. (I say this as the issue you mention here hasn't yet been tagged.)

PR #740, opened 2020-03-09
Avoid generating multiple conformances to GRPCProtobufPayload

glbrntt:
Motivation:

- Generated services must also add conformance to the request/response
  types that they require
- If different services rely on the same message type then this
  conformance is generated multiple times; leading to a build error

Modifications:

- Pass the observed message types between each run of the generator

Result:

- Code generated for services defined in different files with the same
  message types does not cause a build error

### DIFF
--- a/Sources/protoc-gen-grpc-swift/Generator-Conformance.swift
+++ b/Sources/protoc-gen-grpc-swift/Generator-Conformance.swift
@@ -17,7 +17,7 @@ import SwiftProtobuf
 import SwiftProtobufPluginLibrary
 
 extension Generator {
-  internal func printProtoBufExtensions() {
+  internal func printProtobufExtensions() {
     println("// Provides conformance to `GRPCPayload` for request and response messages")
     for service in self.file.services {
       self.service = service
--- a/Sources/protoc-gen-grpc-swift/Generator-Conformance.swift
+++ b/Sources/protoc-gen-grpc-swift/Generator-Conformance.swift
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2020, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import SwiftProtobuf
+import SwiftProtobufPluginLibrary
+
+extension Generator {
+  internal func printProtoBufExtensions() {
+    println("// Provides conformance to `GRPCPayload` for request and response messages")
+    for service in self.file.services {
+      self.service = service
+      for method in self.service.methods {
+        self.method = method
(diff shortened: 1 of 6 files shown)


glbrntt:
cc @Davidde94

glbrntt:
Fixes #738

Sources/protoc-gen-grpc-swift/Generator-Conformance.swift
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import SwiftProtobuf
+import SwiftProtobufPluginLibrary
+
+extension Generator {
+  internal func printProtoBufExtensions() {
```
MrMage: Nit: Protobuf

Sources/protoc-gen-grpc-swift/main.swift
```diff
@@ -103,15 +102,20 @@ func main() throws {
   // Build the SwiftProtobufPluginLibrary model of the plugin input
   let descriptorSet = DescriptorSet(protos: request.protoFile)
 
+  // We need to generate conformance to `GRPCPayload` for request/response types. Track which
+  // types we've seen to avoid generating the conformance multiple times.
+  var observedMessages = Set<String>()
+
   // process each .proto file separately
   for fileDescriptor in descriptorSet.files {
```
MrMage: Is this ordering stable? Otherwise the location of the declaration might shift around.

glbrntt: Good question; I honestly have no idea. It's certainly stable when the input is stable so I assume it depends on the order you pass them to protoc. Do you think it matters?

I'm not super happy with this approach but I'm not sure of a better one. The alternative I considered was generating the extensions based on the message descriptions (instead of the input/output for each RPC), that would make the location stable but it means you can generate conformance for messages which aren't used for RPCs. We'd also need to provide conformance for the "built in" messages (e.g. Any, Timestamp, etc) within `GRPC` so we'd be forced to update if SwiftProtobuf adds new messages.

glbrntt:
> Is there a way we could test that no double conformance are generated?

I ran this on a case I ran into the other day to double check.

I think we probably need some integration style tests to generate code and build it to validate but we need all the infrastructure first. I'll have a think about it because it would also be useful to have  things like allocation counting and perf tests as part of CI.

Sources/protoc-gen-grpc-swift/main.swift
MrMage: The problem with an unstable ordering would be that every `protoc` run could create different files, resulting in a lot of churn. How about simply sorting the list ourselves in some stable fashion to be sure? Just an idea, though.

MrMage:
> I ran this on a case I ran into the other day to double check.
> 
> I think we probably need some integration style tests to generate code and build it to validate but we need all the infrastructure first. I'll have a think about it because it would also be useful to have things like allocation counting and perf tests as part of CI.

SGTM; let's not worry about that in this PR.

Sources/protoc-gen-grpc-swift/main.swift
glbrntt: Sorting by filename is straightforward to add, so might as well.

kmarcell:
Hi @glbrntt,

I updated to alpha-11 and still run into this issue in some cases where redundant GRPCProtobufPayload conformance was generated for structs.

Unfortunately I can't share the original proto files and the generated swift files, but pls let me know how I could help debug this issue.

Issue #597, opened 2019-09-27
[feature] Generate RPC documentation from .proto source comments


glbrntt:
[SwiftProtobuf](https://github.com/apple/swift-protobuf) generates useful documentation on fields by extracting documentation from the .proto source: it would be useful to do the same for generated clients and servers.

PR #743, opened 2020-03-11
Generate documentation from proto comments

glbrntt:
Motivation:

We have access to the underlying comments associated with RPCs in the
protobuf definition. These often explain what each RPC does and can have
great value to the user.

Modifications:

- Generate documentation for client stubs and server provider protocol
  based on the source comments in the proto file.
- For clients: fall back to some basic documentation when no comments
  are present
- Re-generate code.

Result:

- The generate code has more helpful documentation


glbrntt:
Resolves #597

PR #745, opened 2020-03-12
Add a notice to the nio branch

glbrntt:


glbrntt:
Let's avoid merging this until the branches are sorted. Same goes for #744

PR #761, opened 2020-03-24
Remove required platforms from SPM

glbrntt:
Motivation:

We recently specified which platforms we supported in Package.swift
based on what SwiftNIO supported. These minimum versions recently
decreased making the platform requirement specification in Package.swift
unnecessary.

Modifications:

- Remove platforms from Package.swift
- Fix #file passed to function argument

Result:

- Dependents of gRPC Swift don't require their Package.swift to declare
  their supported platforms

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -3,7 +3,7 @@
     "pins": [
       {
         "package": "swift-log",
-        "repositoryURL": "https://github.com/apple/swift-log",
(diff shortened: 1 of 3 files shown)


MrMage:
What are the new platform requirements for SwiftNIO? I'm curious.

Lukasa:
They align with the defaults: macOS 10.10, oldest SDKs available in all Swift 5-supporting Xcodes for everything else.

Issue #636, opened 2019-11-16
[nio] Provide option to return additional trailing metadata in ServerErrorDelegate.transformRequestHandlerError


dinhhungle:
### Question Checklist

- [x] Updated grpc-swift to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/grpc/grpc-swift/blob/master/CONTRIBUTING.md)
- [x] I read [README](https://github.com/grpc/grpc-swift/blob/master/README.md)
- [x] I searched for [existing GitHub issues](https://github.com/grpc/grpc-swift/issues) (especially, there are many [resolved issues related to SSL](https://github.com/grpc/grpc-swift/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+ssl))

### Question Subject
Unable to return additional trailing metadata in ServerErrorDelegate.transformRequestHandlerError

#### Question Description
We would like to use swift services along with our golang services. We handle our errors by using 
https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto
which uses the `grpc-status-details-bin` key  in metadata

It would be nice being able to handle this error transformation in the error delegate.
 
The current signature for transformRequestHandlerError
https://github.com/grpc/grpc-swift/blob/062b341d63a324699ac7b4d5f2eb6656d61a5158/Sources/GRPC/ServerErrorDelegate.swift#L64-L70
I believe it was possible to add trailing metadata prior to #572 and would suggest to restore that functionality.

glbrntt:
Thanks for raising this @ledinhhu  -- I wasn't aware of `grpc-status-details-bin`, it's not mentioned in the spec but it is implemented as an experimental feature in a number of languages (e.g. [Python](https://github.com/grpc/grpc/blob/b0c9849393da9ec4d4e6e1910f67e94e6a1834b0/src/python/grpcio_status/grpc_status/rpc_status.py#L40), [Go](https://github.com/grpc/grpc-go/blob/40ed2eb467471df2bd3c59e66cc5357159062d48/internal/transport/http_util.go#L366), etc.). 

This seems like a useful feature that we should support. I think the correct way to achieve this would be to add `var details: [Any]` (the Protobuf `Any`, not the Swift `Any`) to our `GRPCStatus` and internally do the first level of encoding/decoding, the user would then have to decode the `Any` into the appropriate type on the client side.

Would this solve your problem @ledinhhu?

@MrMage how do you feel about supporting an experimental feature? Any other suggestions?

[...]

PR #762, opened 2020-03-24
Added trailing HTTPHeaders for custom error reporting

fabianfett:
Issue: #636

### DIFF
--- a/Sources/GRPC/CallHandlers/_BaseCallHandler.swift
+++ b/Sources/GRPC/CallHandlers/_BaseCallHandler.swift
@@ -52,7 +52,7 @@ public class _BaseCallHandler<RequestPayload: GRPCPayload, ResponsePayload: GRPC
 
   /// Sends an error status to the client while ensuring that all call context promises are fulfilled.
   /// Because only the concrete call subclass knows which promises need to be fulfilled, this method needs to be overridden.
-  internal func sendErrorStatus(_ status: GRPCStatus) {
+  internal func sendErrorStatus(_ status: GRPCStatus, trailers: HTTPHeaders?) {
     fatalError("needs to be overridden")
   }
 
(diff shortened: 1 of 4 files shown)


MrMage:
I think this would be a breaking change; @fabianfett do you think we could get this in before 1.0?

The overall intention of the PR makes sense to me; I wonder whether we could find a more elegant way than returning a tuple, though. @glbrntt, what do you think?

glbrntt:
> The overall intention of the PR makes sense to me; I wonder whether we could find a more elegant way than returning a tuple, though. @glbrntt, what do you think?

Agreed. I'd be fine with something like a `StatusAndMetadata` struct to be honest.

glbrntt:
@fabianfett do you have any plans to pick this work back up?

FranzBusch:
@glbrntt I would also happily pick this up. I was playing around with the implementation of `grpc-status-details-bin` and noticed that returning both `status` and in the `trailingMetadata` in one future makes handling this a lot easier. I am open to suggestions here and happy to contribute :)

fabianfett:
@FranzBusch It’s all yours. 🚀 I sadly don’t have the time right now.

glbrntt:
@FranzBusch you're welcome to pick this up; I started looking at it last week but didn't get much further than this draft. Let me know if you'd like to take this one on.

FranzBusch:
@glbrntt Happy to take this. Would it also be okay to add [this](https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto) proto to the repo and vend it? Or would you leave this to the clients?

glbrntt:
> @glbrntt Happy to take this. 

Awesome, thank you! Let me know if you need anything 😃 

> Would it also be okay to add [this](https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto) proto to the repo and vend it? Or would you leave this to the clients?

I think I'd rather leave this to clients since it wouldn't be directly used by the library.

glbrntt:
hey @FranzBusch -- any updates on this?

glbrntt:
Closing in favour of #873

PR #772, opened 2020-04-08
Update gRPC-Swift macOS deployment target to match SwiftNIOHTTP2 & SwiftNIOTransportServices

drcapulet:


drcapulet:
Updated the script, though it also impacts `CGRPCZlib.podspec`. I actually realized the compile error I was getting was from the import of `NIOHPACK` from `SwiftNIOHTTP2`.

MrMage:
I still don't understand why we need to increase our deployment target to 10.12, though? (See my comment above.)

MrMage:
@drcapulet could you please comment on why this PR is necessary?

drcapulet:
I'm not sure why the project wasn't originally working, but it is now generating with the correct minimum version, so I'm going to close this out.

PR #784, opened 2020-05-07
Add a retry limit to ConnectionBackoff

### LINKED ISSUE #783: Limit retry count on connection transientFailure
### Question Checklist

### Question Subject
I use gRPC to connect my Swift daemon and client on macOS through Unix Domain Sockets. 

Is there a way to limit retry count? There is ConnectionBackoff structure that has `maximumBackoff` property but it does not stop reconnections ever, just limits maximum time between attempts

#### Question Description
Expected behavior: set maximum retry count, receive up to maxRetry state changes to `transientFailure` and next state change is to `shutdown`

Observed behaviour: with any ConnectionBackoff parameters state changes connecting -> transientFailure -> connecting indefinitely and does not stop. I even tried to set maximumBackoff to 0.01 but it only changes timeout between reconnections.

There is open issue  for NIO https://github.com/apple/swift-nio/issues/599 regarding calling `setsockopt` on UDS and warnings coming out of it, but I don't think it matters here.

glbrntt:
Motivation:

See #783

Modifications:

- Add a `retries` option to `ConnectionBackoff`

Result:

Connection attempts may now be made a limited number of times. By
default the behaviour is unchanged.

Resolves #783

### DIFF
--- a/Sources/GRPC/ConnectionBackoff.swift
+++ b/Sources/GRPC/ConnectionBackoff.swift
@@ -41,8 +41,8 @@ public struct ConnectionBackoff: Sequence {
   /// A limit on the number of times to attempt reconnection.
   public var retries: Retries
 
-  public struct Retries {
-    fileprivate enum Limit {
+  public struct Retries: Hashable {
+    fileprivate enum Limit: Hashable {
       case limited(Int)
       case unlimited
     }
@@ -60,7 +60,7 @@ public struct ConnectionBackoff: Sequence {
 
     /// A limited number of retry attempts. `limit` must be positive. Note that a limit of zero is
     /// identical to `.none`.
-    public static func limited(to limit: Int) -> Retries {
+    public static func upTo(_ limit: Int) -> Retries {
       precondition(limit >= 0)
       return Retries(.limited(limit))
     }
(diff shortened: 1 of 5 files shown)


Sources/GRPC/ConnectionBackoff.swift
```diff
@@ -38,6 +38,34 @@ public struct ConnectionBackoff: Sequence {
   /// The minimum amount of time in seconds to try connecting.
   public var minimumConnectionTimeout: TimeInterval
 
+  /// A limit on the number of times to attempt reconnection.
+  public var retries: Retries
+
+  public struct Retries {
```
Lukasa, line 44: Any reason not to just make this `Hashable`?

Sources/GRPC/ConnectionBackoff.swift
```diff
@@ -38,6 +38,34 @@ public struct ConnectionBackoff: Sequence {
...
+
+    /// An unlimited number of retry attempts.
+    public static let unlimited = Retries(.unlimited)
+
+    /// No retry attempts will be made.
+    public static let none = Retries(.limited(0))
+
+    /// A limited number of retry attempts. `limit` must be positive. Note that a limit of zero is
+    /// identical to `.none`.
+    public static func limited(to limit: Int) -> Retries {
```
weissi, line 63: maybe `upTo(Int)`.

glbrntt: Good suggestion!

Sources/GRPC/ConnectionBackoff.swift
glbrntt: Done

PR #786, opened 2020-05-08
Make HTTP/2 flow control window size configurable for clients and servers

johnkassebaum:
Make HTTP/2 flow control target window size configurable in a client or server HTTP2 connection. 
Motivation:

Higher streaming thruput between server and client is available by the either specifying a HTTP/2 flow control window size higher than the initial value of 2^16-1. The target value needs to be passed either using `SETTINGS_INITIAL_WINDOW_SIZE` in the connection preface, or via a `WINDOW_UPDATE`. As implemented here, the `HTTP2FlowControlWindow` passes the value configured in a `windowUpdate`.

Modifications:

For the client:
Added property `targetWindowSize` to `ClientConnection.Configuration`.
In the existing `ClientConnection.Configuration.init`, sets the new `targetWindowSize` property to the initial flow-control window size (65535).
Duplicated `ClientConnection.Configuration.init` and added a `targetWindowSize` arg.
Added property and setter for `targetWindowSize` to `ClientConnection.Builder`.
During client bootstrap, pass `configuration.targetWindowSize` to the channel initializer.
Duplicated channel initializer `configureGRPCClient` and added a `targetWindowSize` arg.
Duplicated all tests that directly invoke the original `ClientConnection.Configuration.init` to explicitly pass `targetWindowSize`to the new initializer.

For the server:
Added property `httpTargetWindowSize` to `Server.Configuration`.
In `Server.Configuration.init`, added httpTargetWindowSize arg with default value 2^16-1.
Added property and setter for `httpTargetWindowSize` to `Server.Builder`.
During server bootstrap in `Server.start`, pass `configuration.httpTargetWindowSize` to the `HTTPProtocolSwitcher`.
Configure http/2 flow control target window size on the http2 pipeline in `HTTPProxySwitcher`.
Result:

One can build a `ClientConnection` or `Server`with a target window size value that is configured via a WINDOW_UPDATE after the connection is established.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -29,7 +29,7 @@ let package = Package(
     // Main SwiftNIO package
     .package(url: "https://github.com/apple/swift-nio.git", from: "2.14.0"),
     // HTTP2 via SwiftNIO
-    .package(url: "https://github.com/apple/swift-nio-http2.git", from: "1.8.0"),
+    .package(url: "https://github.com/apple/swift-nio-http2.git", from: "1.12.1"),
     // TLS via SwiftNIO
     .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.6.2"),
     // Support for Network.framework where possible.
--- a/Sources/GRPC/ClientConnection.swift
+++ b/Sources/GRPC/ClientConnection.swift
@@ -240,6 +240,9 @@ extension ClientConnection {
     /// The connection backoff configuration. If no connection retrying is required then this should
     /// be `nil`.
     public var connectionBackoff: ConnectionBackoff?
+    
+    ///The HTTP/2 flow control target window size.
+    public var httpTargetWindowSize: Int
 
     /// The HTTP protocol used for this connection.
     public var httpProtocol: HTTP2ToHTTP1ClientCodec.HTTPProtocol {
@@ -258,13 +261,15 @@ extension ClientConnection {
     /// - Parameter tlsConfiguration: TLS configuration, defaulting to `nil`.
     /// - Parameter connectionBackoff: The connection backoff configuration to use.
     /// - Parameter messageEncoding: Message compression configuration, defaults to no compression.
+    /// - Parameter targetWindowSize: The HTTP/2 flow control target window size.
     public init(
       target: ConnectionTarget,
       eventLoopGroup: EventLoopGroup,
       errorDelegate: ClientErrorDelegate? = LoggingClientErrorDelegate(),
       connectivityStateDelegate: ConnectivityStateDelegate? = nil,
       tls: Configuration.TLS? = nil,
-      connectionBackoff: ConnectionBackoff? = ConnectionBackoff()
+      connectionBackoff: ConnectionBackoff? = ConnectionBackoff(),
+      httpTargetWindowSize: Int = 65535
     ) {
       self.target = target
       self.eventLoopGroup = eventLoopGroup
@@ -272,6 +277,7 @@ extension ClientConnection {
       self.connectivityStateDelegate = connectivityStateDelegate
       self.tls = tls
       self.connectionBackoff = connectionBackoff
+      self.httpTargetWindowSize = httpTargetWindowSize
     }
   }
 }
@@ -324,6 +330,7 @@ extension Channel {
   }
 
   func configureGRPCClient(
+    httpTargetWindowSize: Int,
     tlsConfiguration: TLSConfiguration?,
     tlsServerHostname: String?,
     connectionManager: ConnectionManager,
@@ -335,7 +342,7 @@ extension Channel {
     }
 
     return (tlsConfigured ?? self.eventLoop.makeSucceededFuture(())).flatMap {
-      self.configureHTTP2Pipeline(mode: .client)
+      self.configureHTTP2Pipeline(mode: .client, targetWindowSize: httpTargetWindowSize)
     }.flatMap { _ in
       return self.pipeline.handler(type: NIOHTTP2Handler.self).flatMap { http2Handler in
         self.pipeline.addHandler(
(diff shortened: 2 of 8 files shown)


Sources/GRPC/ClientConnection.swift
```diff
@@ -467,6 +468,9 @@ extension ClientConnection {
   public struct Configuration {
     /// The target to connect to.
     public var target: ConnectionTarget
+    
+    ///The HTTP/2 flow control target window size.
+    public var targetWindowSize: Int
```
glbrntt, line 473: nit: can you name this `httpTargetWindowSize` and move it down to where `httpProtocol` is?

glbrntt:
> If you confirm that the correct place is where `HTTPProtocolSwitcher.channelRead(context:data:)`, in the case of `.http2`, invokes `Channel.configureHTTP2Pipeline(mode:)` then I would add the same `targetWindowSize` to `Server.Configuration` and to `Server.Builder`, and would then pass it to the `HTTPProtocolSwitcher` in `Server.makeBootstrap(configuration:)`

That's the right place 👍

glbrntt:
@johnkassebaum FYI your NIO HTTP/2 change was [tagged](https://github.com/apple/swift-nio-http2/releases/tag/1.12.0) a little while ago

johnkassebaum:
Thanks @glbrntt I saw that. So sorry for the delay on this, was able to get back to it last night and will address comments and add server configuration today.

Sources/GRPC/ClientConnection.swift
```diff
@@ -324,6 +330,7 @@ extension Channel {
   }
 
   func configureGRPCClient(
+    httpTargetWindowSize: Int = 65535,
```
glbrntt, line 333: I don't think we need a default here: we should have always have a value from the configuration.

Sources/GRPC/GRPCChannel/GRPCChannelBuilder.swift
```diff
@@ -195,6 +197,15 @@ extension ClientConnection.Builder.Secure {
   }
 }
 
+extension ClientConnection.Builder {
+  /// Sets the HTTP/2 flow control target window size.
```
glbrntt, line 201:
```suggestion
  /// Sets the HTTP/2 flow control target window size. Defaults to 65,535 if not explicitly set.
```

Sources/GRPC/Server.swift
```diff
@@ -173,7 +176,7 @@ extension Server {
   public struct Configuration {
     /// The target to bind to.
     public var target: BindTarget
-
+
```
glbrntt, line 179:
```suggestion

```

Sources/GRPC/ServerBuilder.swift
```diff
@@ -102,6 +104,15 @@ extension Server.Builder.Secure {
   }
 }
 
+extension Server.Builder {
+  /// Sets the HTTP/2 flow control target window size.
```
glbrntt, line 108:
```suggestion
  /// Sets the HTTP/2 flow control target window size. Defaults to 65,535 if not explicitly set.
```

Sources/GRPC/ServerBuilder.swift
```diff
@@ -102,6 +104,15 @@ extension Server.Builder.Secure {
...
+  @discardableResult
+  public func withHttpTargetWindowSize(_ httpTargetWindowSize: Int) -> Self {
```
glbrntt, line 110:
```suggestion
  public func withHTTPTargetWindowSize(_ httpTargetWindowSize: Int) -> Self {
```

gRPC-Swift.podspec
```diff
@@ -18,7 +18,7 @@ Pod::Spec.new do |s|
 
     s.dependency 'Logging', '1.2.0'
     s.dependency 'SwiftNIO', '2.15.0'
-    s.dependency 'SwiftNIOHTTP2', '1.11.0'
+    s.dependency 'SwiftNIOHTTP2', '1.12.0'
```
glbrntt, line 21: No need to update this; it's a generated file.

Tests/GRPCTests/ClientTLSFailureTests.swift
```diff
@@ -135,6 +149,43 @@ class ClientTLSFailureTests: GRPCTestCase {
       XCTFail("Expected NIOSSLError.handshakeFailed(BoringSSL.sslError)")
     }
   }
+  
+  func testClientConnectionFailsWhenServerIsUnknownWithHTTPTargetWindowSize() throws {
```
glbrntt, line 153: Could you remove the extra tests since we're not actually validating a change in behaviour here? I think I'm okay just deferring to NIO HTTP/2 for the tests.

Package.swift
```diff
@@ -29,7 +29,7 @@ let package = Package(
     // Main SwiftNIO package
     .package(url: "https://github.com/apple/swift-nio.git", from: "2.14.0"),
     // HTTP2 via SwiftNIO
-    .package(url: "https://github.com/apple/swift-nio-http2.git", from: "1.8.0"),
+    .package(url: "https://github.com/apple/swift-nio-http2.git", from: "1.12.0"),
```
glbrntt, line 32: Mind bumping this to `1.12.1`? We missed something the first time with the window size: https://github.com/apple/swift-nio-http2/pull/205

Sources/GRPC/ClientConnection.swift
```diff
@@ -240,6 +240,9 @@ extension ClientConnection {
     /// The connection backoff configuration. If no connection retrying is required then this should
     /// be `nil`.
     public var connectionBackoff: ConnectionBackoff?
+    
+    ///The HTTP/2 flow control target window size.
```
glbrntt, line 244:
```suggestion
    /// The HTTP/2 flow control target window size.
```

PR #796, opened 2020-05-14
do not add own module import in generated code

### LINKED ISSUE #795: gen-grpc-swift: allow generating code in different modules
The protobuf swift generator has an option ([ProtoPathModuleMappings](https://github.com/apple/swift-protobuf/blob/master/Documentation/PLUGIN.md#generation-option-protopathmodulemappings---swift-module-names-for-proto-paths)) that lets us define the module for each proto in a single file. This automatically adds the required imports for every generated swift class.

We have a decent amount of protos, and we are splitting our generated code into modules. **Are there any plans to add a similar mechanism into `grpc-swift`?**

I'm aware that we can use `ExtraModuleImports` to add extra imports, and we could somehow automate the calculation of the extra modules, but I still think the current mechanism is not powerful enough. It would be much easier to use the same module mapping file as `swift` generator does.

jagobagascon:
When generating a file using the option `ProtoPathModuleMappings` the module of the generated proto is being added to the header of the file.

swift mappings file:
```
mapping {
    module_name: "A"
    proto_file_path: "a.proto"
}
mapping {
    module_name: "B"
    proto_file_path: "b.proto"
}
```

a.grpc.swift (generated from a.proto that depends on b.proto)
```swift
...
import A // this is not needed
import B
...
```


Sources/protoc-gen-grpc-swift/Generator.swift
```diff
@@ -99,9 +99,6 @@ class Generator {
     // mappings to import the service protos as well as and any other proto modules that the file
     // imports.
     let moduleMappings = options.protoToModuleMappings
-    if let serviceProtoModuleName = moduleMappings.moduleName(forFile: file) {
```
glbrntt, line 102: cc @allevato who added this originally (https://github.com/grpc/grpc-swift/pull/347)

My understanding of the comment above is that if we have a foo.proto containing:

```proto
service FooService { ... }

message FooMessage { ... }
```

Then Bazel would produce two Swift modules, say `FooServices` and `FooMessages`; and this line is to import the `FooMessages` module. 

If my understanding of this is correct this seems like a misuse of module mappings since the output generated from a single proto is split across multiple modules. Using it in this way also leads to an unnecessary warning if you want to use module mappings in the intended way. Can this be better expressed in the Bazel rules as an additional import instead of module mappings?

allevato: You're correct about the original motivation; the Bazel rule `swift_proto_library` generates just the code for the messages, the rule `swift_grpc_library` generates just the code for the services, and they're in separate modules so the latter needs to import the former.

The `ExtraModuleImports` flag was added after my change in https://github.com/grpc/grpc-swift/pull/402 to support getting the right imports for client stubs, so that wasn't an option when we first spun up Bazel support. But looking at the rule implementation again (it's been a while), I think it should be a straightforward change to use `ExtraModuleImports` in the same way here, so I don't foresee a problem with this change; we'll just need to update the rules when we update to this version.

cc @mpetrov for awareness as well.

glbrntt: Awesome, thanks very much.

As a heads up: this is change is for `protoc-gen-grpc-swift` (gRPC Swift 1.x);  `protoc-gen-swiftgrpc` (gRPC Swift 0.x) won't be affected.

Issue #756, opened 2020-03-19
threading issue in/shown by GRPCTests.ClientConnectionBackoffTests testClientEventuallyConnects


weissi:
### New Issue Checklist

- [x] Updated grpc-swift to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/grpc/grpc-swift/blob/master/CONTRIBUTING.md)
- [x] I searched for [existing GitHub issues](https://github.com/grpc/grpc-swift/issues)(especially, there are many [resolved issues related to SSL](https://github.com/grpc/grpc-swift/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+ssl))

### Issue Description

I ran

```
swift test --sanitize=thread --disable-index-store
```

and got

##### Complete output when running grpc-swift, including the stack trace and command used

```
[...]
Test Case '-[GRPCTests.ClientConnectionBackoffTests testClientEventuallyConnects]' started.
==================
WARNING: ThreadSanitizer: data race (pid=16913)
  Write of size 8 at 0x7b4c00005298 by thread T18:
    #0 ClientConnection.channel.setter <compiler-generated> (grpc-swiftPackageTests:x86_64+0x33036a)
    #1 closure #3 in ClientConnection.willSetChannel(to:) ClientConnection.swift:191 (grpc-swiftPackageTests:x86_64+0x336ab7)
    #2 partial apply for closure #3 in ClientConnection.willSetChannel(to:) <compiler-generated> (grpc-swiftPackageTests:x86_64+0x3477df)
    #3 thunk for @escaping @callee_guaranteed (@guaranteed Result<(), Error>) -> () <compiler-generated> (grpc-swiftPackageTests:x86_64+0x337037)
    #4 partial apply for thunk for @escaping @callee_guaranteed (@guaranteed Result<(), Error>) -> () <compiler-generated> (grpc-swiftPackageTests:x86_64+0x347880)
    #5 closure #1 in EventLoopFuture.whenComplete(_:) EventLoopFuture.swift:738 (grpc-swiftPackageTests:x86_64+0x73f475)
    #6 partial apply for closure #1 in EventLoopFuture.whenComplete(_:) <compiler-generated> (grpc-swiftPackageTests:x86_64+0x74366f)
    #7 CallbackList._run() EventLoopFuture.swift:88 (grpc-swiftPackageTests:x86_64+0x733560)
[...]

Issue #706, opened 2020-01-23
Close client connection when it becomes idle


glbrntt:
At the moment the client will persist a connection to the server even if there is no active RPCs. Ideally the client should disconnect when the connection becomes idle after 5 minutes of inactivity, for example. On starting an RPC the client should re-establish a connection to the server.

MrMage:
We could also consider adding on option for this to the _server_, to have it enforce conserving of server-side resources.

audio-algorithms:
this is so stupid. managing a connection is the responsibility of the application.

PR #798, opened 2020-05-15
Rewrite client connection management

glbrntt:
Motivation:

The client connection management code is pretty bad. It's hard to follow
and test reliably. There were some thread safety issues and it's also
possible to give a call a multiplexer from a future channel which will
fail but may succeed after retrying the connection. It also lacked the
abilitiy to close an idle channel (i.e. one with no open HTTP streams).

Modifications:

- Rewrite the client connection management logic: this is handled by
  `ConnectionManager` and `ClientConnectivityHandler`.
- Channels will idle after 5 minutes of having no open streams
  (configurability will come in a later PR)
- Connectivity state changes are async'd onto a dispatch queue
  (configurable queue will come later)

Result:

- Connection management code is much easier to follow and can be tested
  with an EmbeddedChannel
- Fewer bugs
- We no longer hold a lock when calling out to the connectivity delegate

---
This change is [Reviewable](https://reviewable.io/reviews/grpc/grpc-swift/798)


glbrntt:
Resolves #756 and another issue mentioned in #783. Resolves the client half of #706.

Sources/GRPC/ClientConnectivityHandler.swift
```diff
@@ -0,0 +1,100 @@
...
+      default:
+        ()
+      }
+    }
+
+    context.fireChannelRead(data)
+  }
+
+  private func idle(context: ChannelHandlerContext) {
+    assert(self.activeStreams == 0)
```
Lukasa, line 96: This assertion can fail. In particular, `Scheduled.cancel` may fail to cancel the task if it has already been de-queued to be run. It should be a `guard`.

Sources/GRPC/ClientConnectivityHandler.swift
```diff
@@ -0,0 +1,100 @@
...
+        let remoteAddressDescription = context.channel.remoteAddress.map { "\($0)" } ?? "n/a"
+        self.connectionManager.logger.info("gRPC connection ready", metadata: [
+          "remote_address": "\(remoteAddressDescription)",
+          "event_loop": "\(context.eventLoop)"
+        ])
+
+        // Start the idle timeout.
+        self.scheduledIdle = context.eventLoop.scheduleTask(in: self.idleTimeout) {
+          self.idle(context: context)
+        }
```
Lukasa, line 80: This code is a bit duplicated: maybe make it a function?

Sources/GRPC/ClientConnectivityHandler.swift
```diff
@@ -0,0 +1,100 @@
...
+      }
+    }
+
+    context.fireChannelRead(data)
+  }
+
+  private func idle(context: ChannelHandlerContext) {
+    assert(self.activeStreams == 0)
+    self.connectionManager.idle()
+    context.close(mode: .all, promise: nil)
```
Lukasa, line 98: You may need state to determine when you've already fired a `close`, and so from `connectionManager`'s perspective this connection is dead. That's because until `close` completes it is possible that more "things" may happen on this connection, including stream creation.

Sources/GRPC/ConnectionManager.swift
```diff
@@ -0,0 +1,586 @@
...
+
+        // We have a channel, close it.
+        state.channel.close(mode: .all, promise: nil)
+
+      // Like `.connecting` and `.active` the application does not have a `.ready` channel. We'll
+      // do the same but also cancel any scheduled connection attempts and deal with any fallout
+      // if we cancelled too late.
+      case .transientFailure(let state):
+        // Stop the creation of a new channel, if we can.
+        state.scheduled.cancel()
```
Lukasa, line 306: Given that this `cancel` may fail, do we need to handle the possibility the connection attempt goes ahead anyway?

Sources/GRPC/ConnectionManager.swift
```diff
@@ -0,0 +1,586 @@
...
+      }
+
+      return shutdown.closeFuture
+    }
+  }
+
+  // MARK: - State changes from the channel handler.
+
+  /// The connecting channel became `active`. Must be called on the `EventLoop`.
+  internal func channelActive(channel: Channel) {
```
Lukasa, line 326: Aren't we supposed to close the channel here?

Tests/GRPCTests/CompressionTests.swift
```diff
@@ -24,7 +24,7 @@ class MessageCompressionTests: GRPCTestCase {
   var group: EventLoopGroup!
   var server: Server!
   var client: ClientConnection!
-  var defaultTimeout: TimeInterval = 0.1
+  var defaultTimeout: TimeInterval = 1.0
```
Lukasa, line 27: Why did this need to be changed?

glbrntt: One of the tests was occasionally hitting the timeout.

Sources/GRPC/ConnectionManager.swift
glbrntt: Good question: I think the comment may actually be wrong here; I don't think the cancel can fail: the scheduled task is `startConnecting()`: if that gets run the actual connection attempt will start _after_ the transition from `transientFailure` to `.connecting` (because we submit the connect onto the EventLoop).

Sources/GRPC/ConnectionManager.swift
glbrntt: Why?

Lukasa: If we're in `shutdown` and we get `channelActive`, who is closing it?

Sources/GRPC/ConnectionManager.swift
Lukasa: I mean, `cancel` can always fail. The question is whether we tolerate its failure. I think the answer now is yes, but we should say what we expect to happen here.

Sources/GRPC/ConnectionManager.swift
glbrntt: Ah, yes, we should!

glbrntt:
> _[Sources/GRPC/ClientConnectivityHandler.swift, line 49 at r3](https://reviewable.io/reviews/grpc/grpc-swift/798#-M7S8zfMCTAoYaPkSwhS:-M7S8zfMCTAoYaPkSwhT:bohb1ae) ([raw file](https://github.com/grpc/grpc-swift/blob/7df117edb367e4f0007267039607eed370374b60/Sources/GRPC/ClientConnectivityHandler.swift#L49)):_
> 
> > ```swift
> >       // We have a stream: don't go idle
> >       self.scheduledIdle?.cancel()
> >       self.activeStreams += 1
> > ```
> 
> I assume these operations do not need to be thread-safe or atomic?

No: we're only ever accessing/modifying from the event loop.

---

> _[Sources/GRPC/ClientConnectivityHandler.swift, line 53 at r3](https://reviewable.io/reviews/grpc/grpc-swift/798#-M7S9SYsBH7qP-juFD4I:-M7S9SYsBH7qP-juFD4J:brfcrw3) ([raw file](https://github.com/grpc/grpc-swift/blob/7df117edb367e4f0007267039607eed370374b60/Sources/GRPC/ClientConnectivityHandler.swift#L53)):_
> 
> > ```swift
> >       self.activeStreams -= 1
> >       // No active streams: go idle soon.
> >       if self.activeStreams == 0 {
> > ```
> 
> I wonder whether cancelling/scheduling could be done as a side effect on the `activeStreams` setter instead?

We can if we hold on to the `context` as well; I'm not sure it actually makes things any better though.

---

> _[Sources/GRPC/ClientConnectivityHandler.swift, line 123 at r3](https://reviewable.io/reviews/grpc/grpc-swift/798#-M7S9Izk4mJj4a6Amyy0:-M7S9Izk4mJj4a6Amyy1:b-wne6ez) ([raw file](https://github.com/grpc/grpc-swift/blob/7df117edb367e4f0007267039607eed370374b60/Sources/GRPC/ClientConnectivityHandler.swift#L123)):_
> 
> > ```swift
> >     }
> > 
> >     self.scheduledIdle = context.eventLoop.scheduleTask(in: self.idleTimeout) {
> > ```
> 
> Should any previous scheduled idle be cancelled, or can we 100% guarantee that that won't be necessary?

It won't be necessary: the `scheduleIdleTimeout` from `channelRead` can only happen once and this happens _before_ the channel is accessible to the application (i.e. it's not possible to create a stream before it's scheduled). The other `scheduleIdleTimeout` is hopefully clear as it is.

---

> _[Sources/GRPC/ClientConnectivityHandler.swift, line 124 at r3](https://reviewable.io/reviews/grpc/grpc-swift/798#-M7S9aRJBnRc7vCwK8a0:-M7S9aRJBnRc7vCwK8a1:bu1c69) ([raw file](https://github.com/grpc/grpc-swift/blob/7df117edb367e4f0007267039607eed370374b60/Sources/GRPC/ClientConnectivityHandler.swift#L124)):_
> 
> > ```swift
> >     self.scheduledIdle = context.eventLoop.scheduleTask(in: self.idleTimeout) {
> >       self.idle(context: context)
> > ```
> 
> should we reference `self` weakly here?

No this should be fine: NIO will drop the callback as soon as it's executed/cancelled.

---

> _[Sources/GRPC/ConnectionManager.swift, line 152 at r3](https://reviewable.io/reviews/grpc/grpc-swift/798#-M7SA-Zj4YVkuGZANTmu:-M7SA-Zj4YVkuGZANTmv:bvhhy13) ([raw file](https://github.com/grpc/grpc-swift/blob/7df117edb367e4f0007267039607eed370374b60/Sources/GRPC/ConnectionManager.swift#L152)):_
> 
> > ```swift
> >         // Create a new id; it'll be used for the *next* channel we create.
> >         self.channelNumber &+= 1
> > ```
> 
> I assume HTTP2 allows for channel numbers >= 2^63?

This *isn't* related to HTTP/2 streams; it's just a count of separate HTTP/2 connections that have existed.

---

> _[Sources/GRPC/ConnectionManager.swift, line 262 at r3](https://reviewable.io/reviews/grpc/grpc-swift/798#-M7SAXO62rIIbe8tbx0Z:-M7SAXO62rIIbe8tbx0_:bqs1ync) ([raw file](https://github.com/grpc/grpc-swift/blob/7df117edb367e4f0007267039607eed370374b60/Sources/GRPC/ConnectionManager.swift#L262)):_
> 
> > ```swift
> >       // We don't have a channel and we don't want one, easy!
> >       case .idle:
> >         shutdown = ShutdownState(closeFuture: self.eventLoop.makeSucceededFuture(()))
> > ```
> 
> Sorry, but what is `closeFuture for`?

Why do we hold on to it? In case `shutdown()` is called twice it should return the same value.

glbrntt:
> > _[Reviewable](https://reviewable.io/reviews/grpc/grpc-swift/798)_ status: 15 of 16 files reviewed, 6 unresolved discussions (waiting on @glbrntt, @Lukasa, and @MrMage)
> 
> _[Sources/GRPC/ClientConnectivityHandler.swift, line 90 at r3](https://reviewable.io/reviews/grpc/grpc-swift/798#-M7S98dX5kdMB8R9VZfM:-M7b4Wmk4FKvNDB5c4_o:b-37daoq) ([raw file](https://github.com/grpc/grpc-swift/blob/7df117edb367e4f0007267039607eed370374b60/Sources/GRPC/ClientConnectivityHandler.swift#L90)):_
> 
> Previously, MrMage (Daniel Alm) wrote…
> I'd prefer an "as long as", IMHO that's clearer (unless that's actually wrong)

I think "as long as we're not `.ready`" suggests `closed` is also okay (which it isn't).

> _[Sources/GRPC/ConnectionManager.swift, line 152 at r3](https://reviewable.io/reviews/grpc/grpc-swift/798#-M7SA-Zj4YVkuGZANTmu:-M7b7-CDBdH9-0Aba4-1:by54yn5) ([raw file](https://github.com/grpc/grpc-swift/blob/7df117edb367e4f0007267039607eed370374b60/Sources/GRPC/ConnectionManager.swift#L152)):_
>
> Previously, glbrntt (George Barnett) wrote…
> Alright, but are we okay with "overflows"?

Yes: it's only used to disambiguate between different `Channel`s in logging, and the likelihood of cycling through enough connection to reach an overflow here is practically zero anyway.

> _[Sources/GRPC/ConnectionManager.swift, line 262 at r3](https://reviewable.io/reviews/grpc/grpc-swift/798#-M7SAXO62rIIbe8tbx0Z:-M7b78Qp-M3bSxy9-h-c:b-ve7e2d) ([raw file](https://github.com/grpc/grpc-swift/blob/7df117edb367e4f0007267039607eed370374b60/Sources/GRPC/ConnectionManager.swift#L262)):_
> 
> Previously, glbrntt (George Barnett) wrote…
> Sorry, I don't quite understand what the future means exactly.

The `Channel` will hold a promise (not accessible to the user) which gets fulfilled when it is eventually closed: this is the `futureResult` for that promise. (I.e. the same future returned by `channel.close()`)

glbrntt:
> How about this? "// We only care about SETTINGS as long as we are in state `.notReady`." IMHO that has the least possible ambiguity. At least the original comment had confused me a bit.

Fine by me; I've taken it verbatim.

Sources/GRPC/ClientConnectivityHandler.swift
```diff
@@ -0,0 +1,140 @@
...
+    // We called `close` on the channel.
+    case closed
+  }
+
+  init(connectionManager: ConnectionManager, idleTimeout: TimeAmount = .minutes(5)) {
+    self.connectionManager = connectionManager
+    self.idleTimeout = idleTimeout
+  }
+
+  func userInboundEventTriggered(context: ChannelHandlerContext, event: Any) {
```
Lukasa, line 45: This should probably also guard on whether you've called `close`.

glbrntt: done

PR #800, opened 2020-05-18
[draft] client test stubs

glbrntt:
@MrMage this is a draft for the client test stubs. The relevant parts here are:

- `Sources/Examples/Echo/Model/echo.grpc.swift`, and
- `Tests/GRPCTests/TestClientExample.swift`

Would like your feedback on the general shape of this before I take it any further!

---
This change is [Reviewable](https://reviewable.io/reviews/grpc/grpc-swift/800)


glbrntt:
I ended up taking a slightly different approach. The changes will come over a couple of PRs to make it easier to review, the first of which is #855. I'm closing this PR in favour of that.

PR #804, opened 2020-05-21
Update CgRPC to v1.24.3

glbrntt:


MrMage:
Is there a good way to view only the non-vendored code you had to change to make this work? If not, how about splitting the vendoring scripts and the actual upgrade into two separate PRs?

scripts/vendor-grpc.sh
```diff
@@ -42,41 +43,60 @@ echo "COPYING public gRPC headers"
...
-perl -pi -e 's/\/\* #define PB_FIELD_16BIT 1 \*\//#define PB_FIELD_16BIT 1/' $DSTROOT/CgRPC/third_party/nanopb/pb.h
-
-echo "MOVING nanopb headers to CgRPC/include"
-mv $DSTROOT/CgRPC/third_party/nanopb/*.h $DSTROOT/CgRPC/include/
-
 echo "ADDING additional compiler flags to tsi/ssl_transport_security.cc"
 perl -pi -e 's/#define TSI_OPENSSL_ALPN_SUPPORT 1/#define TSI_OPENSSL_ALPN_SUPPORT 0/' $DSTROOT/CgRPC/src/core/tsi/ssl_transport_security.cc
 
+echo "COPYING CgRPC/src to CgRPC/include"
+cp -rp $DSTROOT/CgRPC/src $DSTROOT/CgRPC/include/src
```
MrMage, line 73: Does this need to be copied or would a move suffice? Also, why is this necessary?

glbrntt: `src/` contains implementation _and_ headers: we're copying here to maintain the directory structure, we remove all the `.c` and `.cc` files below. There was a symlink from `include/src` to `../src` which I removed because that didn't work with the structure of the `upb-generated` stuff below.

glbrntt:
> I was able to scroll to the bottom and find the changed scripts. Looks good overall; are there any relevant learnings you could share to make this process easier in case it needs to be repeated in the future?

Sorry, I should've mentioned that changes to the scripts were in [one commit](https://github.com/grpc/grpc-swift/pull/804/commits/6c2f0cb0213d88e4bad7a0fbb682b355a3e24a0e) and the output in the other!

I think the main takeaway is that the divide between shim code and vendored code needs to be much clearer. Ideally the script should produce the whole `CgRPC` module by copying in the vendored code and then copying in any required shim code. I ran into problems where it wasn't clear if code was required or leftover from a previous time the script was run. The few bugs in the vendoring scripts also didn't help! That said, I don't think it's worth our time doing this because this is effectively EOL anyway.

MrMage:
> I think the main takeaway is that the divide between shim code and vendored code needs to be much clearer. Ideally the script should produce the whole `CgRPC` module by copying in the vendored code and then copying in any required shim code. I ran into problems where it wasn't clear if code was required or leftover from a previous time the script was run. The few bugs in the vendoring scripts also didn't help! That said, I don't think it's worth our time doing this because this is effectively EOL anyway.

👍

LGTM from my side!

PR #806, opened 2020-05-22
Update our issue templates

MrMage:
Motivation: A bunch of recent issues due to version mismatches between library and protoc plugin could have become obvious when entering the version list..

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.md
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.md
@@ -8,7 +8,7 @@ about: If something isn't working as expected 🤔
 
 ### New Issue Checklist
 
-- [ ] Updated grpc-swift to the latest version
+- [ ] Updated grpc-swift *and protoc-gen-swiftgrpc* to the latest version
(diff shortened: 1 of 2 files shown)


glbrntt:
@MrMage I'm fine with this for now but I think we should simplify our issue templates; there's a wall of text when you create an issue and I don't think it provides _that_ much, the information it does say can certainly be expressed in less text. WDYT?

I also think it'd be worth having a separate template for feature requests (bug report / question doesn't really cover it).

MrMage:
> @MrMage I'm fine with this for now but I think we should simplify our issue templates; there's a wall of text when you create an issue and I don't think it provides _that_ much, the information it does say can certainly be expressed in less text. WDYT?
> 
> I also think it'd be worth having a separate template for feature requests (bug report / question doesn't really cover it).

👍 To all of this. This was just a quick fix; it doesn't solve the underlying problem.

glbrntt:
Awesome, I actually threw together some ideas in a local branch a while ago, never got around to finishing them up though. I'll try to get to it soon!

PR #807, opened 2020-05-22
Update Cocoapods to 0.11.0.

MrMage:
Has already been pushed successfully.

### DIFF
--- a/SwiftGRPC.podspec
+++ b/SwiftGRPC.podspec
@@ -11,7 +11,7 @@
 
 Pod::Spec.new do |s|
   s.name = 'SwiftGRPC'
-  s.version = '0.10.0'
+  s.version = '0.11.0'
   s.license     = { :type => 'Apache License, Version 2.0',
                     :text => <<-LICENSE
                       Copyright 2018, gRPC Authors. All rights reserved.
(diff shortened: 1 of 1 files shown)


SwiftGRPC.podspec
```diff
@@ -41,6 +41,6 @@ Pod::Spec.new do |s|
   s.source_files = 'Sources/SwiftGRPC/*.swift', 'Sources/SwiftGRPC/**/*.swift', 'Sources/CgRPC/shim/*.[ch]'
   s.public_header_files = 'Sources/CgRPC/shim/cgrpc.h'
 
-  s.dependency 'gRPC-Core', '~> 1.23.0'
+  s.dependency 'gRPC-Core', '~> 1.24.1'
```
glbrntt, line 44: Should this be 1.24.3?

MrMage: Sorry, I mistook .1 for .3. The .1 has already been pushed; should we keep that (given that `pod update` would install the .3, anyway), or push an 0.11.1 version?

glbrntt: I think it's probably fine

Issue #801, opened 2020-05-20
Conformance to GRPCProtobufPayload generates wrong code when using `ProtoPathModuleMappings` option


jagobagascon:
### New Issue Checklist
- [x] Updated grpc-swift to the latest version
- [x] I read the [Contribution Guidelines](https://github.com/grpc/grpc-swift/blob/master/CONTRIBUTING.md)
- [x] I searched for [existing GitHub issues](https://github.com/grpc/grpc-swift/issues)(especially, there are many [resolved issues related to SSL](https://github.com/grpc/grpc-swift/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+ssl))

### Issue Description
We are currently generating our swift code using `ProtoPathModuleMappings` option to split it into different modules. As each module needs to be in its own folder we were generating the code by calling `protoc-gen-grpc-swift` plugin once per module with a different out folder.
```shell
protoc ... --grpc-swift_out="/module-a" a.proto
protoc ... --grpc-swift_out="/module-b" b.proto
```
This causes the following error during swift build:
```shell
error: redundant conformance of 'Google_Protobuf_Empty' to protocol 'GRPCProtobufPayload'
```

This issue was fixed in [this commit](https://github.com/grpc/grpc-swift/commit/c4dc6a3c1e1246f9b4bfee7e633142624c7be86f) by avoiding the definition of multiple conformances to GRPCProtobufPayload for the same class. But for it to work we had to generate all our protos in the same plugin call and then move the generated code to its module folder:
```shell
protoc ... --grpc-swift_out=".tmp" a.proto b.proto
mv a.*.swift /module-a/
mv b.*.swift /module-b/
```

This is not ideal, but it "fixes" our previous problem. After this change, the `GRPCProtobufPayload` is added only to the first class that needs it. In our example this would be `a.grpc.swift` and when trying to compile the code in `/module-b` it fails with the following error:
```shell
b.grpc.swift:32:74: error: type 'Google_Protobuf_Empty' does not conform to protocol 'GRPCPayload'
```

I'm not sure what the fix should be. I'm kind of new to swift. **I managed to compile the code by adding a protocol that extends `GRPCProtobufPayload` per module and using that**.
```swift
protocol ModuleAPayload : GRPCProtobufPayload {}
extension SwiftProtobuf.Google_Protobuf_Empty: ModuleAPayload {}
```
**I don't know if this is good practice or not**, it feels kinda hacky.

[...]

PR #811, opened 2020-05-26
Provide conformance for messages provided by SwiftProtobuf

glbrntt:
Motivation:

We currently generate conformance to `GRPCProtobufPayload` based on the
input and output types of `rpc`s defined in .proto files passed to the
plugin. This works in some situations, including when the .proto file
for the message in question is not present (such as those provided by
the SwiftProtobuf library; `Google_Protobuf_*`). However this is
problematic if the plugin is invoked on a file-by-file basis where each
file contains the message since the conformance would be erroneously
generated mulutple times.

Modifications:

- Provide conformance to all `Google_Protobuf_*` messages currently
  vendored by `SwiftProtobuf`
- Generaate conformance only based on the `message`s defined in each
  .proto file
- Add a few generation tests

Result:

Multiple conformance generation is avoided in the scenario set out
above.


glbrntt:
Resolves #801

Issue #808, opened 2020-05-22
Should we document "moving from CgRPC to NIO" in the README?


MrMage:
Maybe we should have a brief section on the differences. For example:

- different threading model might require a substantial rewrite (referring to the new examples)
- the CocoaPod was renamed from `SwiftGRPC` to `grpc-Swift` (this one tripped a few people up)
- the protoc plugin needs to be updated and re-run to get the new generated code
- etc.

Lutzifer:
having done this migration last week: yes, documentation on this would be much of a help

MrMage:
> having done this migration last week: yes, documentation on this would be much of a help

Makes sense! Were there any things you stumbled upon besides the ones mentioned in the list above? It would be good to know about them so that we can include them in the documentation :-)

Lutzifer:
- I found it hard to find the examples (as there is a second folder)
- I did not find an easy to understand example on how to use NIO based grpc in an iOS app (I found the SpeechToText example, but even now I find this one a bit confusing). It took me some time to use the group correctly (the defer block from the example triggered too early in the iOS app, thus destroying the connection before I could use it) [https://github.com/grpc/grpc-swift/blob/9a41c8efdad71274e676ec735565dd16b241efb9/Sources/Examples/RouteGuide/Client/main.swift#L206].
- Having examples of "in cGRPC you might have done X, so do Y now" would be cool
- It took me some time to update the protoc plugins correctly and make sure the right version is used
- overall, I think, having a simpler iOS example App (Bonus Points if it is Rx 😂) would have been a lot of help. But maybe I had a different case to implement (chunking and sending a binaray file using streaming) than most of the people.
- I think we need a better overview over the examples and how to run them.
- https://github.com/grpc/grpc-swift/blob/9a41c8efdad71274e676ec735565dd16b241efb9/Examples/Google/SpeechToText/Sources/SpeechService.swift#L41 is what I would have needed most in hindsight, but I landed at [https://github.com/grpc/grpc-swift/blob/9a41c8efdad71274e676ec735565dd16b241efb9/Sources/Examples/RouteGuide/Client/main.swift#L196]. A combination of both did it for me (with some trial and error).

Hope this helps a bit.

glbrntt:
Thanks @Lutzifer - that's really helpful!

Looks like we have a few things to do:

- [x] Explain difference between examples (#814)
- [x] Explain high-level difference between versions  (#814)
- [ ] Simple iOS example
[...]

PR #814, opened 2020-05-27
Add high level differences between 1.x and 0.y to README

glbrntt:
Motivation:

- There's been some confusion recently around the two versions of gRPC.
- See also: https://github.com/grpc/grpc-swift/issues/808

Modifications:

- Add high level differences between two versions to README
- Add current SemVer version build badge to make version more obvious
- Explain difference between two example directories in README

Result:

Better documentation

---
This change is [Reviewable](https://reviewable.io/reviews/grpc/grpc-swift/814)


glbrntt:
> Ideally, we would also have a dedicated guide on how to switch, but I guess that'll have to wait due to bandwidth concerns... (I also haven't experienced the transition on the client side myself yet, so can't report on that.)

Agreed, I didn't have time to do all of that, but this felt like a lot of benefit for little effort.

PR #818, opened 2020-06-01
Add a server-idle handler

### LINKED ISSUE #706: Close client connection when it becomes idle

glbrntt:
Motivation:

Connections should be dropped to save resources if there are not RPCs
for an idle timeout. This was added for clients in #798; this adds
similart functionality to the server.

Modifications:

- Add a `GRPCServerIdleHandler` to drop connections after 5 minutes of
  no RPCs

Note: timeout configuration will be added for client and server in a
followup PR.

Result:

- Server's will drop connections if the client isn't doing anything.
- Resolves #706


glbrntt:
ping @Lukasa

Issue #822, opened 2020-06-05
Archive Failing on Xcode 11.4 / 11.5


mRs-:
### Describe the bug

When using Xcode 11.4 or 11.5 and archiving the grpc-swift implementation in a Framework or a static lib xcode is crashing with an Libtool failed command without an exit code. All grpc-swift version are **1.0.0-alpha.1** to **1.0.0-alpha.12**.

```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: object: /Users/mariuslandwehr/Library/Developer/Xcode/DerivedData/Networking-cskqpnmvfnmzorctpckvxmsmisys/Build/Intermediates.noindex/ArchiveIntermediates/Networking/BuildProductsPath/Release/CGRPCZlib.o malformed object (string table at offset 0 with a size of 8, overlaps Mach-O headers at offset 0 with a size of 160)
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: object: /Users/mariuslandwehr/Library/Developer/Xcode/DerivedData/Networking-cskqpnmvfnmzorctpckvxmsmisys/Build/Intermediates.noindex/ArchiveIntermediates/Networking/BuildProductsPath/Release/CNIOLinux.o malformed object (string table at offset 0 with a size of 8, overlaps Mach-O headers at offset 0 with a size of 160)
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: object: /Users/mariuslandwehr/Library/Developer/Xcode/DerivedData/Networking-cskqpnmvfnmzorctpckvxmsmisys/Build/Intermediates.noindex/ArchiveIntermediates/Networking/BuildProductsPath/Release/CGRPCZlib.o malformed object (string table at offset 0 with a size of 8, overlaps Mach-O headers at offset 0 with a size of 160)
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: object: /Users/mariuslandwehr/Library/Developer/Xcode/DerivedData/Networking-cskqpnmvfnmzorctpckvxmsmisys/Build/Intermediates.noindex/ArchiveIntermediates/Networking/BuildProductsPath/Release/CNIOLinux.o malformed object (string table at offset 0 with a size of 8, overlaps Mach-O headers at offset 0 with a size of 160)
[...]

PR #828, opened 2020-06-08
Updated Dependencies to use latest

mRs-:
This also adds the workaround in https://github.com/grpc/grpc-swift/issues/822

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -15,8 +15,8 @@
         "repositoryURL": "https://github.com/apple/swift-nio.git",
         "state": {
           "branch": null,
-          "revision": "a27a07719ca785bcaca019a5b9fe1814b981b4a2",
-          "version": "2.15.0"
+          "revision": "120acb15c39aa3217e9888e515de160378fbcc1e",
+          "version": "2.18.0"
         }
       },
       {
@@ -33,8 +33,8 @@
         "repositoryURL": "https://github.com/apple/swift-nio-ssl.git",
         "state": {
           "branch": null,
-          "revision": "ae213938e151964aa691f0e902462fbe06baeeb6",
-          "version": "2.7.1"
+          "revision": "9cb641037babeda5031b6ee6078d579c86fd67cb",
+          "version": "2.7.4"
         }
       },
       {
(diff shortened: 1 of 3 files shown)


Jake-Prickett:
I believe this PR will be required for us to release another Pod. With the update to NIOHTTP2, there's a version conflict between dependencies giving a build error for `1.0.0-alpha.13` Podspec validation.

mRs-:
Yeah and the Swift 5.0 Compiler is not happy about the changes... (https://travis-ci.org/github/grpc/grpc-swift/jobs/696149912)

Lukasa:
Huh, that's interesting.

The errors are around the establishment and data transfer report accessing fields in the new NIO Transport Services. Those are guarded behind OS availability guards (`#available` and `@available`), so this ought not to be a problem.

_However_, I think that in this case the CI build is testing 5.0 compiles not by using a recent Xcode in 5.0 mode but by using an _old_ Xcode. In this case, 10.3, which shipped before macOS 10.15. This means it doesn't have the macOS 10.15 SDK and so literally doesn't know these symbols exist. Because availability guards don't affect the parse, just the notion of availability, we error out.

@glbrntt is this testing environment reflective of an intended use case? Are we really saying that we'll support not just "compile in 5.0 mode" but also "compile with the old compiler"?

glbrntt:
> @glbrntt is this testing environment reflective of an intended use case? Are we really saying that we'll support not just "compile in 5.0 mode" but also "compile with the old compiler"?

No, 10.3 was the latest Xcode when we setup the CI, we just haven't changed it yet.

Lukasa:
Cool, we should do that then. :grin:

glbrntt:
hey @mRs-, could you rebase your changes? I've fixed the flaky test and updated the CI.

mRs-:
Yep will do it 👍

mRs-:
The Tests are passing 🎉 All ready to go now

glbrntt:
> I believe this PR will be required for us to release another Pod. With the update to NIOHTTP2, there's a version conflict between dependencies giving a build error for `1.0.0-alpha.13` Podspec validation.

@Jake-Prickett would you mind doing the cocoapods release?

glbrntt:
@Jake-Prickett @MrMage has was updating the pods (CGRPCZlib is done) but is now blocked on NIOSSL which hasn't been updated yet; could you update when you a get a chance?

Issue #829, opened 2020-06-08
Automate Podspec deployment with Github action


SebastianThiebaud:
### Is your feature request related to a problem? Please describe it.
Everytime a new release is created, it can take hours before the associated podspec is published to Cocoapods.

### Describe the solution you'd like
Instead of manually pushing the podspec to Cocoapods, a Github action could take care of running any required scripts, stamp / bump the version of the podspec to match the new release tag and publish the new podspec to Cocoapods.

Jake-Prickett:
I really like this idea! I’ve been trying to work through how to do this as well, but haven’t thought of an elegant solution. (Also at my work project) I’d love some help on this one! 🙂

The biggest challenge for this is that the Pods (`CGRPCZlib` and `gRPC-Swift`) have to be pushed in a specific order for them to pass validation. The time required for `CGRPCZlib` to become available after pushing to trunk can vary from seconds to minutes. This requires the script to sometimes be ran multiple times manually.

It could be retried with some sort of delay, but there’s no guarantee that it’s available that second run (could even be  a build issue or _actual_ validation failure)

SebastianThiebaud:
I can definitely help on this one as I've implemented a similar solution at my work. The only difference is that the Github action I wrote is pushing to a private specs repository rather than the main Cocoapods trunk. There are plenty of actions available already doing the job of pushing the podspec to Cocoapods, we can see if using them works for `grpc-swift`.

Concerning `CGRPCZlib`, it seems like we would have to implement a similar action on that repository. But then, we might run into an issue when a `grpc-swift` release is created before the new `CGRPCZlib` podspec is published and available. @Jake-Prickett Do you know why the podspec in not immediately available? I've never seen this happening for my own projects.

Jake-Prickett:
Both the `CGRPCZlib` and `gRPC-Swift` are part of this repo and are pushed sequentially to Trunk. Trunk defaults to use the CDN instead of the Git repository, which takes some time to reflect updates to Podspecs. (Up to 10 minutes)

The issue is properly described here: https://github.com/CocoaPods/CocoaPods/issues/9497

But it looks like last month a fix was merged in last month!! 🎉 https://github.com/CocoaPods/cocoapods-trunk/pull/147 We could add `—synchronous` to `build_podspec.py` when we push to Trunk to resolve that issue. 

Now this problem just got much easier! 😂

SebastianThiebaud:
[...]

PR #830, opened 2020-06-09
Add workflow to deploy podspecs once a release is created

SebastianThiebaud:
Related to https://github.com/grpc/grpc-swift/issues/829

Things to consider:
- The workflow will not commit the updated podspecs in `master` (what are your thoughts about this? we could have the github runner to commit in `master` but then the actual tagged commit would still have the old podspecs...)
- A Github secret named `COCOAPODS_TRUNK_TOKEN` needs to be set with the value available in the `Authorization` header by running `pod trunk me --verbose` (don't include the `Token` prefix of the header value).
- Do we need to exclude this action/workflow from 0.y.z versions?
- I'm not sure that this works for pre-release too (I haven't tested this scenario). I will check that in a test repo and make any changes if needed.

### DIFF
--- a/Sources/CGRPCZlib/empty.c
+++ b/Sources/CGRPCZlib/empty.c
@@ -13,3 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+// Xcode's Archive builds with Xcode's Package support struggle with empty .c files
+// (https://bugs.swift.org/browse/SR-12939).
+void CGRPCZlib_i_do_nothing_just_working_around_a_darwin_toolchain_bug(void) {}


SebastianThiebaud:
@Jake-Prickett (I can't add reviewers / assignees, just making sure you get notified)

MrMage:
> * The workflow will not commit the updated podspecs in `master` (what are your thoughts about this? we could have the github runner to commit in `master` but then the actual tagged commit would still have the old podspecs...)

I find it both options weird:

a) to commit the updated Podspec before the tag even exists
b) the tag not to contain the corresponding Podspec.

So I don't feel too strongly about this.

> * A Github secret named `COCOAPODS_TRUNK_TOKEN` needs to be set with the value available in the `Authorization` header by running `pod trunk me --verbose` (don't include the `Token` prefix of the header value).

I think these tokens expire after a while; that might be a problem.

> * Do we need to exclude this action/workflow from 0.y.z versions?

Probably; we should definitely keep pushing Pods for updates to the `cgrpc` branch, but I'm not sure if that would be more effort to automate because we don't have a script for that right now.

.github/workflows/cocoapods-deploy.yml
```diff
@@ -0,0 +1,17 @@
...
+  build:
+    runs-on: macOS-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install Cocoapods
+      run: gem install cocoapods
+    - name: Deploy to Cocoapods
+      run: |
+        set -eo pipefail
+        python ./scripts/build_podspecs.py -u $(git describe --tags `git rev-list --tags --max-count=1`)
```
glbrntt, line 15: Can we just get the tag from the event? https://help.github.com/en/actions/reference/events-that-trigger-workflows#release-event-release

.github/workflows/cocoapods-deploy.yml
```diff
@@ -0,0 +1,17 @@
+name: Deploy to Cocoapods
+on:
+  release:
+    types: [published]
```
glbrntt, line 4: It looks like we _might_ need "prereleased" as well, the documentation isn't that clear.

SebastianThiebaud:
> I think these tokens expire after a while; that might be a problem.

Yeah, mine expires in 4 months. Let me explore options and get back to you.

.github/workflows/cocoapods-deploy.yml
SebastianThiebaud: Yeah, I hate this documentation.. It gives a list of options but doesn't explain their meaning.

PR #840, opened 2020-06-11
Add 'isOk' to GRPCStatus

### LINKED ISSUE #821: Consider internalising the convenience "commonly used" statuses
### Is your feature request related to a problem? Please describe it.

It's too easy to accidentally compare a `GRPCStatus` instead of a status code.

Specifically, `GRPCStatus` has public static `.ok` [here](https://github.com/grpc/grpc-swift/blob/4dd5ee2b6e0fd74dffda2096407c186c477de04f/Sources/GRPC/GRPCStatus.swift#L42), however this is distinct from the `.ok` status code. 

This makes it really easy to write code like:

```swift
if status == .ok {
  celebrate(with: "🍺")
} else {
  commiserate(with: "🍺")
}
```

However `GRPCStatus` also contains a message so the equality check can fail when the status code is `.ok` and the messages are different.

### Describe the solution you'd like

- `GRPCStatus.ok` should be made `internal`
- Add an `isOk: Bool` computed property which checks only the `code`

### Describe alternatives you've considered

None.

### Additional context

None.

glbrntt:
Motivation:

`GRPCStatus` has a static `.ok` status which is useful when implementing
service providers. However, since the status also includes a message it's
too easy for someone receiving a status to check `status == .ok` where
they really mean 'status.code == .ok'.

Modifications:

- Add a convenience `.isOk` to `GRPCStatus`
- Add a note to `.ok`

Result:

- Slightly reduced chance of user error.
- Resolves #821

### DIFF
--- a/Sources/GRPC/GRPCStatus.swift
+++ b/Sources/GRPC/GRPCStatus.swift
@@ -31,6 +31,11 @@ public final class GRPCStatus: Error {
   /// The status message of the RPC.
   public let message: String?
 
+  /// Whether the status is '.ok'.
+  public var isOk: Bool {
+    return self.code == .ok
+  }
+
   public init(code: Code, message: String?) {
     self.code = code
     self.message = message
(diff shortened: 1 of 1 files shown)


MrMage:
LGTM; we could still consider to make `.ok` internal, which is a bit more controversial.

glbrntt:
I _think_ this is probably sufficient.

PR #842, opened 2020-06-16
Allow deadlines or timeouts to be set on RPCs

glbrntt:
Motivation:

Sometimes timeouts are useful, other times deadlines are useful. We
should support both.

Modifications:

- Add a `timeLimit` to `CallOptions`, which is essentially a deadline or
  a timeout. When we start a call we generate a deadline from the time
  limit and forward that timeline through the pipeline and then compute
  a timeout to send to the server.
- For timeouts, `GRPCTimeout` has been replaced by `TimeAmount`, the
  user doesn't care how the timeout is serialized over the wire so it
  makes more sense to use a type which provides better interoperability
  with the rest of the community. Most of the `GRPCTimeout` code has
  been relocated to shims and marked as deprecated.
- Shims for deprecating `CallOptions(timeout:)`

Result:

- Users can set timeouts or deadlines
- Better deadline/timeout interoperability with other NIO applications


glbrntt:
Hmm GitHub says CI is pending but Travis says it's completed and passed. I'll believe Travis here; merging.

PR #845, opened 2020-06-17
Add default user-agent to header if customMetadata doesn't have one

ikait:
This pull request adds a userAgent option to CallOptions struct. It makes clients possible to request with customized user-agent to the gRPC servers which require the particular format of user-agent. 

Currently it partially supports customize user-agent by setting `customMetadata` with "user-agent" key but it always produces string starts with "grpc-swift-nio, ". 

In a project I belong to have a log parser that requires a particular user-agent and couldn't accept string having that prefix. I think other gRPC clients can be set custom user-agent so that it's also useful for grpc-swift.

### DIFF
--- a/Sources/GRPC/GRPCClientStateMachine.swift
+++ b/Sources/GRPC/GRPCClientStateMachine.swift
@@ -519,7 +519,6 @@ extension GRPCClientStateMachine.State {
       ":scheme": scheme,
       "content-type": "application/grpc",
       "te": "trailers",  // Used to detect incompatible proxies, part of the gRPC specification.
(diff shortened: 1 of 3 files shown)


ikait:
Thank you for the comment, @glbrntt. I saw your points and rewrote my proposal with much simpler changes. The idea that setting the user-agent at the connection level is looks good but does it need much more changes than ones with RPC level? At this pull request I would like to address my original problem with RPC level fix.

glbrntt:
> The idea that setting the user-agent at the connection level is looks good but does it need much more changes than ones with RPC level? At this pull request I would like to address my original problem with RPC level fix.

That's fine with me, I'll create an issue to capture this information, there's no need to do it now :)

PR #848, opened 2020-06-18
Stop sending requests in call init

glbrntt:
Motivation:

Previously @lukasa noted that we did some networking in the call init's
which isn't ideal. I forgot that calls are made though an internal
class, so it is indeed possible to move the networking calls.

Modifications:

- Move networking calls to just after the call object is created
- Shuffling code about, no real functional changes

Result:

- No networking in the *Call inits
- Cleaner inits
- Nothing functional


glbrntt:
> You made `transport` public on your 4 call objects so that the connection objects can invoke it. Would it be better to keep it private and instead have the call objects implement that method, and then just call it on their private object?

100%. Will update, thanks.

Issue #843, opened 2020-06-16
Keep-Alive and HTTP2 pings


SebastianThiebaud:
### What are you trying to achieve? 

The `grpc-swift` library built on top of the C library (0.x) used to expose a few channel arguments like `http2MinSentPingInterval` or `keepAliveTimeout`
https://github.com/grpc/grpc-swift/blob/cgrpc/Sources/SwiftGRPC/Core/ChannelArgument.swift#L46

I'm struggling to find the equivalent for the new library (1.x) and based on initial investigation, it doesn't seem that SwiftNIOHTTP2 implements any logic to ping. Am I missing something?

Lukasa:
Right now the new library does not auto-ping, but it could! The infrastructure is there for it to do so, the code just isn't written yet.

SebastianThiebaud:
@Lukasa Looking quickly at the codebase and how Swift NIO works (I'm still not 100% familiar with it), it seems like the best strategy would be to add a new handler (or even two: one for client and one for server) to the existing channel's pipeline that will handle keep alive (ping and pong + logic documented [here](https://github.com/grpc/grpc/blob/master/doc/keepalive.md)).

Is that correct? If so, I can start looking at it and maybe submitting a pull request in the near future.

Lukasa:
Yeah, that would work. To implement the `WITHOUT_DATA` versions is a bit trickier and requires that the handler live in a very specific spot in the pipeline (between the `HTTP2Handler` and the `HTTP2StreamMultiplexer`).

SebastianThiebaud:
I don't think I encountered `WITHOUT_DATA` yet. Is that the same thing as `GRPC_ARG_KEEPALIVE_PERMIT_WITHOUT_CALLS` aka no active RPCs?

SebastianThiebaud:
Messing with it quickly, that's how I added the handlers to the pipelines.

```
                        [I] ↓↑ [O]
            NIOHTTP2Handler ↓↑ NIOHTTP2Handler            [handler0]
            GRPCIdleHandler ↓↑                            [handler3]
 GRPCClientKeepaliveHandler ↓↑ GRPCClientKeepaliveHandler [handler2]
     HTTP2StreamMultiplexer ↓↑ HTTP2StreamMultiplexer     [handler1]

                        [I] ↓↑ [O]
 GRPCServerKeepaliveHandler ↓↑ GRPCServerKeepaliveHandler [handler0]
[...]

PR #850, opened 2020-06-18
Add keepalive handlers

SebastianThiebaud:
This is a work-in-progress pull request for https://github.com/grpc/grpc-swift/issues/843

I just wanted to submit a quick draft so I can get early feedback while I'm working on this implementation of this feature.

Based on the behavior described here https://github.com/grpc/grpc/blob/master/doc/keepalive.md


Sources/GRPC/ClientConnection.swift
```diff
@@ -264,17 +267,19 @@ extension ClientConnection {
     /// - Parameter errorDelegate: The error delegate, defaulting to a delegate which will log only
     ///     on debug builds.
     /// - Parameter connectivityStateDelegate: A connectivity state delegate, defaulting to `nil`.
-    /// - Parameter tlsConfiguration: TLS configuration, defaulting to `nil`.
+    /// - Parameter tls: TLS configuration, defaulting to `nil`.
     /// - Parameter connectionBackoff: The connection backoff configuration to use.
-    /// - Parameter messageEncoding: Message compression configuration, defaults to no compression.
-    /// - Parameter targetWindowSize: The HTTP/2 flow control target window size.
+    /// - Parameter connectionKeepalive: The keepalive configuration to use.
+    /// - Parameter connectionIdleTimeout: The amount of time to wait before closing the connection, defaulting to 5 minutes..
```
Lukasa, line 273:
```suggestion
    /// - Parameter connectionIdleTimeout: The amount of time to wait before closing the connection, defaulting to 5 minutes.
```

Sources/GRPC/ConnectionKeepalive.swift
```diff
@@ -0,0 +1,100 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import NIO
+
+/// Provides keepalive pings.
+///
+/// The defaults are determined by the gRPC keepalive
+/// [documentation] (https://github.com/grpc/grpc/blob/master/doc/keepalive.md).
+public struct ClientConnectionKeepalive {
```
Lukasa, line 22: These two structures should be conformed to `Hashable`, as they're just simple aggregates.

Sources/GRPC/ConnectionKeepalive.swift
```diff
@@ -0,0 +1,100 @@
...
+  /// The amount of time to wait before sending a keepalive ping.
+  public var time: TimeAmount
```
Lukasa, line 24: I don't love this name: maybe `interval`?

Sources/GRPC/ConnectionKeepalive.swift
```diff
@@ -0,0 +1,100 @@
...
+    self.timeout = timeout
+    self.permitWithoutCalls = permitWithoutCalls
+    self.maximumPingsWithoutData = maximumPingsWithoutData
+    self.minimumSentPingIntervalWithoutData = minimumSentPingIntervalWithoutData
+  }
+}
+
+public struct ServerConnectionKeepalive {
+  /// The amount of time to wait before sending a keepalive ping.
+  public var time: TimeAmount
```
Lukasa, line 57: Same note here, maybe `interval`?

Sources/GRPC/Server.swift
```diff
@@ -122,7 +123,6 @@ public final class Server {
           return channel.pipeline.addHandler(protocolSwitcher)
         }
       }
-
```
Lukasa, line 125: No need for this change.

Sources/GRPC/Server.swift
```diff
@@ -214,14 +217,18 @@ extension Server {
     /// - Parameter serviceProviders: An array of `CallHandlerProvider`s which the server should use
     ///     to handle requests.
     /// - Parameter errorDelegate: The error delegate, defaulting to a logging delegate.
-    /// - Parameter tlsConfiguration: TLS configuration, defaulting to `nil`.
+    /// - Parameter tls: TLS configuration, defaulting to `nil`.
+    /// - Parameter connectionKeepalive: The keepalive configuration to use.
+    /// - Parameter connectionIdleTimeout: The amount of time to wait before closing the connection, defaulting to 5 minutes..
```
Lukasa, line 222:
```suggestion
    /// - Parameter connectionIdleTimeout: The amount of time to wait before closing the connection, defaulting to 5 minutes.
```

docs/keepalive.md
```diff
@@ -0,0 +1,2 @@
+# Keepalive
```
Lukasa, line 1: I assume there are some plans to add to this?

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,281 @@
...
+    context.close(mode: .all, promise: nil)
+  }
+}
+
+protocol ChannelKeepaliveHandler: AnyObject {
+  func send(payload: HTTP2Frame.FramePayload, context: ChannelHandlerContext)
+  func close(context: ChannelHandlerContext)
+}
+
+private final class PingHandler {
```
Lukasa, line 125: I think this factoring out of the shared logic is really useful, but I think we should reconsider this exact approach.

The biggest downside of this approach, as noted, is that the keep alive handler needs to hold a reference to the ping handler, and the ping handler needs ta reference to the keep alive handler. This strong cycle needs to be broken. You've done this here using `weak`, which is fine, but it's arguably the worst of the options we have (the other two options are `unowned` which is semantically just fine here and performs better, or even better still to have a direct strong reference that is simply optional, and nil the ref out in `handlerAdded`).

However, an even better design is to flip this class to a struct and drop the reference to the parent object altogether. This object will then be driven by the parent, and will ask the parent to act on its behalf using return values. That is `func read(pingData: HTTP2PingData, ack: Bool, context: ChannelHandlerContext)` becomes `func read(pingData: HTTP2PingData, ack: Bool, context: ChannelHandlerContext) -> Action`, where `Action` is some enum that describes the possible results of this operation:

```swift
struct PingHandler {
    enum Action {
        case cancelScheduledTimeout
        case goawayAndClose
        case reply(HTTP2Frame)
    }
}
```

and so on.

The advantage of this design (and of putting all the state in this struct but all the side effects elsewhere) is twofold. Firstly, it makes testing _way_ easier: you can make the `PingHandler` struct `internal` but prefix it with an underscore, and then you can simply drive it directly in testing, which allows you to easily test any complex logic in isolation.

Secondly, Swift's law of exclusivity will force us into a useful programming pattern. A common danger in NIO-based programs is to only partially perform a state transition before you make outcalls (i.e. calls to other code you don't control). For example, succeeding a promise before you've moved all your state. These outcalls are potentially re-entrant, and so can lead you to nasty state management bugs where your code observes impossible states.

However, because Swift forbids two mutating calls to the same object at the same time, if this ever happens when you are using mutating functions on a struct, either the compiler will catch it or the runtime will force it to crash, depending on the complexity of the pattern. This is really great! Swift will police us and prevent us from making this mistake, which is really nice.

While we're here, you've defined a `ChannelKeepaliveHandler` protocol, which is great. That protocol can be rewritten to be a useful place for code sharing. For example, you could do:

```swift
protocol _ChannelKeepaliveHandler {
    var pingHandler: PingHandler { get set }
}

extension _ChannelKeepaliveHandler {
    func read(pingData: HTTP2PingData, ack: Bool, context: ChannelHandlerContext) {
        let action = self.pingHandler.read(pingData: pingData, ack: ack)
        switch action {
            // Do whatever
        }
    }
}
```

In essence your protocol can define the stored properties necessary to perform the computation, and then define all the computation in protocol extensions to achieve the code re-use. This allows us to really trim down the code we need.

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,281 @@
...
+    let payload = HTTP2Frame.FramePayload.goAway(lastStreamID: .rootStream, errorCode: .enhanceYourCalm, opaqueData: nil)
+    self.channelHandler?.send(payload: payload, context: context)
+  }
+
+  private func acknowledgePong() {
+    self.scheduledClose?.cancel()
+    self.scheduledClose = nil
+  }
+
+  private func shouldBlockPing() -> Bool {
```
Lukasa, line 264: There's no reason for this not to just be a `var`.

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,281 @@
...
+    // There is no active call on the transport and pings should not be sent
+    if self.activeStreams == 0 && !self.permitWithoutCalls { return true }
```
glbrntt, line 266: This might be slightly easier to follow if we `guard self.activeStreams == 0 else { return false }` and then deal with `permitWithoutCalls`.

nit: generally we don't have one-liners like this, please move the return to a new line.

Tests/GRPCTests/GRPCKeepaliveTests.swift
```diff
@@ -0,0 +1,122 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+@testable import GRPC
+import NIO
+import EchoModel
+import EchoImplementation
+import XCTest
+
+class GRPCKeepaliveTests: GRPCTestCase {
```
glbrntt, line 22: Could you run `swift test --generate-linuxmain` to generate the test manifest for Linux?

Tests/GRPCTests/GRPCKeepaliveTests.swift
```diff
@@ -0,0 +1,122 @@
...
+  func testClientKeepaliveTimeout() {
+    XCTAssertNoThrow(try self.doTestKeepalive(
+      clientKeepalive: ClientConnectionKeepalive(time: .seconds(3), timeout: .seconds(1)),
+      expectedFinalChange: .shutdown)
+    )
+  }
+
+  func doTestKeepalive(clientKeepalive: ClientConnectionKeepalive = ClientConnectionKeepalive(),
+                       serverKeepalive: ServerConnectionKeepalive = ServerConnectionKeepalive(),
+                       expectedFinalChange: ConnectivityState) throws {
```
glbrntt, line 32: This isn't used?

Tests/GRPCTests/GRPCKeepaliveTests.swift
```diff
@@ -0,0 +1,122 @@
...
+      XCTAssertNoThrow(try server.close().wait())
+    }
+
+    // Setup a state change recorder for the client.
+    let stateRecorder = RecordingConnectivityDelegate()
+    stateRecorder.expectChanges(3) { changes in
+      XCTAssertEqual(changes, [
+        Change(from: .idle, to: .connecting),
+        Change(from: .connecting, to: .ready),
+        Change(from: .ready, to: .shutdown)
```
glbrntt, line 54: Hmm, `.shutdown` shouldn't be the final state. It should be `.idle`. I.e. "we don't have a connection now but we could have one in the future" whereas shutdown is "we don't have a connection now and we'll never have one in the future".

docs/keepalive.md
SebastianThiebaud: Yes, I was waiting to be done with the implementation first. Just a reminder for me to create some documentation before this gets merged.

Tests/GRPCTests/GRPCKeepaliveTests.swift
SebastianThiebaud: Those tests are not final at all. I committed this one just in case you have any feedback to guide me but they are just playgrounds for me at this point.

Lukasa:
Please give us a ping when you want this re-reviewed.

SebastianThiebaud:
Far from perfect, but this should be in a OK state for review.

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    return nil
+  }
+
+  func read(pingData: HTTP2PingData, ack: Bool) -> Action? {
+    let isPong = ack && pingData.integer == self.pingCode
+    let isIllegalWithoutCalls = !ack && self.activeStreams == 0 && !self.permitWithoutCalls
+    let isInvalidPing = !ack && isPingStrike
+    let isValidPing = !ack && !isPingStrike
```
glbrntt, line 238:
```suggestion
    let isInvalidPing = !ack && self.isPingStrike
    let isValidPing = !ack && !self.isPingStrike
```

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,297 @@
...
+
+    if isPong {
+      return .cancelScheduledTimeout
+    } else if isIllegalWithoutCalls {
+      return .reply(self.goAwayFrame)
+    } else if isInvalidPing, let maximumPingStrikes = self.maximumPingStrikes {
+      if self.pingStrikes > maximumPingStrikes {
```
glbrntt, line 245: If the maximum ping strikes is zero we should tolerate any number of strikes, that doesn't seem to be the case here?

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,297 @@
...
+    } else if isValidPing {
+      self.lastReceivedPingDate = .now()
+      return .reply(self.generatePingFrame(code: pingData.integer, ack: true))
+    } else {
+      return nil
+    }
+  }
+
+  func pingFired() -> Action? {
+    guard !self.shouldBlockPing else { return nil }
```
glbrntt, line 260: IMHO `guard !predicate` is less clear then `if predicate`

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,297 @@
...
+    return .reply(self.generatePingFrame(code: pingCode, ack: false))
+  }
+
+  private func generatePingFrame(code: UInt64, ack: Bool) -> HTTP2Frame.FramePayload {
+    if self.activeStreams == 0 {
+      self.sentPingsWithoutData += 1
+    }
+
+    lastSentPingDate = .now()
```
glbrntt, line 269:
```suggestion
    self.lastSentPingDate = .now()
```

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,297 @@
...
+    return HTTP2Frame.FramePayload.ping(HTTP2PingData(withInteger: code), ack: ack)
+  }
+
+  private var isPingStrike: Bool {
+    guard self.activeStreams == 0 && self.permitWithoutCalls else { return false }
+    guard let lastReceivedPingDate = self.lastReceivedPingDate else { return false }
+    guard let minimumReceivedPingIntervalWithoutData = self.minimumReceivedPingIntervalWithoutData else { return false }
```
glbrntt, line 276: nit: this could just be a single `guard`, also could you reformat to:

```swift
... else {
  return false
}
```

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,297 @@
...
+
+  private let goAwayFrame = HTTP2Frame.FramePayload.goAway(lastStreamID: .rootStream, errorCode: .enhanceYourCalm, opaqueData: nil)
+
+  enum Action {
+    case schedulePing(delay: TimeAmount, timeout: TimeAmount)
+    case cancelScheduledTimeout
+    case reply(HTTP2Frame.FramePayload)
+  }
+
+  init(
```
glbrntt, line 203: The testability of this is really good at the moment, except for having to deal with time. Passing in a mechanism to provide `NIODeadline`s would allow you to control time without `sleep`ing in tests.

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,297 @@
...
+    didSet {
+      if activeStreams > 0 {
+        self.sentPingsWithoutData = 0
+      }
+    }
+  }
+
+  private let goAwayFrame = HTTP2Frame.FramePayload.goAway(lastStreamID: .rootStream, errorCode: .enhanceYourCalm, opaqueData: nil)
+
+  enum Action {
```
glbrntt, line 197: nit: a `none` or `noAction` case would be clearer and neater than having to handle `Action?`

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,297 @@
...
+
+  /// If there are no data/header frames being sent:
+  /// The minimum amount of time expected between receiving successive pings.
+  /// If the time between successive pings is less than this value, then the ping will be considered a bad ping from the peer.
+  /// Such a ping counts as a "ping strike".
+  private let minimumReceivedPingIntervalWithoutData: TimeAmount?
+
+  /// Maximum number of bad pings that the server will tolerate before sending an HTTP2 GOAWAY frame and closing the connection.
+  /// Setting it to `0` allows the server to accept any number of bad pings.
+  private let maximumPingStrikes: UInt?
```
glbrntt, line 169: `nil` means we're a client and ping strikes aren't applicable: we should document that here. I'd also be open to having an enum here to codify this.

Sources/GRPC/ConnectionKeepalive.swift
```diff
@@ -0,0 +1,106 @@
...
+    self.timeout = timeout
+    self.permitWithoutCalls = permitWithoutCalls
+    self.maximumPingsWithoutData = maximumPingsWithoutData
+    self.minimumSentPingIntervalWithoutData = minimumSentPingIntervalWithoutData
+    self.minimumReceivedPingIntervalWithoutData = minimumReceivedPingIntervalWithoutData
+    self.maximumPingStrikes = maximumPingStrikes
+  }
+}
+
+extension TimeAmount: Hashable {
```
glbrntt, line 102: You should _only_ conform types to protocols when you own either the type or the protocol.

I submitted this PR https://github.com/apple/swift-nio/pull/1571 to directly add `Hashable` conformance to `TimeAmount`.

We can either wait for the release or drop our requirement to `Equatable` and then add `Hashable` once it becomes available in NIO.

SebastianThiebaud: any idea when the next release will be available?

Sources/GRPC/GRPCKeepaliveHandlers.swift
SebastianThiebaud: Agreed, but I'm just not sure how to implement that.

SebastianThiebaud: I tried something, let me know if that is good enough. (Pretty much copied what was in `GRPCTimeout`).

SebastianThiebaud:
@glbrntt @Lukasa It should be in a pretty good state now.

docs/keepalive.md
```diff
@@ -0,0 +1,50 @@
...
+
+gRPC sends HTTP2 pings on the transport to detect if the connection is down.
+If the ping is not acknowledged by the other side within a certain period, the connection
+will be closed. Note that pings are only necessary when there is no activity on the connection. 
+
+## What should I set?
```
glbrntt, line 7: This reads as if we're suggesting these value are used instead of the defaults. I think it might be clearer to show the default values first and then follow with how users can configure them (i.e. providing `{Client,Server}ConnectionKeepalive` to the relevant builder).

docs/keepalive.md
```diff
@@ -0,0 +1,50 @@
...
+timeout|.seconds(20)|.seconds(20)|The amount of time to wait for an acknowledgment.
+permitWithoutCalls|false|false|Send keepalive pings even if there are no calls in flight.
+maximumPingsWithoutData|2|2|Maximum number of pings that can be sent when there is no data/header frame to be sent/
+minimumSentPingIntervalWithoutData|.minutes(5)|.minutes(5)|If there are no data/header frames being received: the minimum amount of time to wait between successive pings.
+minimumReceivedPingIntervalWithoutData|N/A|.minutes(5)|If there are no data/header frames being sent: the minimum amount of time expected between receiving successive pings. If the time between successive pings is less than this value, then the ping will be considered a bad ping from the peer. Such a ping counts as a "ping strike".
+maximumPingStrikes|N/A|2|Maximum number of bad pings that the server will tolerate before sending an HTTP2 GOAWAY frame and closing the connection. Setting it to `0` allows the server to accept any number of bad pings.
+
+Fore more information, please visit the [gRPC Core documentation for keepalive](https://raw.githubusercontent.com/grpc/grpc/master/doc/keepalive.md)
```
glbrntt, line 50: Should this just be https://github.com/grpc/grpc/blob/master/doc/keepalive.md rather than the raw content?

glbrntt:
Failing CI should be fixed by https://github.com/grpc/grpc-swift/pull/858

docs/keepalive.md
SebastianThiebaud: I'm not sure my latest commit satisfies your comment. Feel free to update this doc on your own.

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,354 @@
...
+}
+
+protocol _ChannelKeepaliveHandler: ChannelInboundHandler {
+  var pingHandler: PingHandler { get }
+  var scheduledPing: Scheduled<Void>? { get set }
+  var scheduledClose: Scheduled<Void>? { get set }
+  func close(context: ChannelHandlerContext)
+}
+
+extension _ChannelKeepaliveHandler where OutboundOut == HTTP2Frame, InboundIn == HTTP2Frame {
```
Lukasa, line 95: Rather than put these constraints on the extension, let's require these constraints as part of the protocol.

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,354 @@
...
+
+  private func send(payload: HTTP2Frame.FramePayload, context: ChannelHandlerContext) {
+    let frame = self.wrapOutboundOut(.init(streamID: .rootStream, payload: payload))
+    context.writeAndFlush(frame, promise: nil)
+  }
+
+  private func schedulePing(delay: TimeAmount, timeout: TimeAmount, context: ChannelHandlerContext) {
+    guard delay != .nanoseconds(Int64.max) else { return }
+
+    self.scheduledPing = context.eventLoop.scheduleTask(in: delay) { [unowned self] in
```
Lukasa, line 140: I don't think we gain anything from `unowned self` here: we'll crash if we get it wrong, so instead let's just eat the strong reference. The cycle isn't a problem anyway because this task will either execute or be cancelled on straightforward timelines, so let's just do away with `unowned`.

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,354 @@
...
+      self.perform(action: self.pingHandler.pingFired(), context: context)
+      self.scheduleClose(timeout: timeout, context: context)
+    }
+  }
+
+  private func scheduleClose(timeout: TimeAmount, context: ChannelHandlerContext) {
+    self.scheduledClose = context.eventLoop.scheduleTask(in: timeout) { [unowned self] in
```
Lukasa, line 147: Same note here, let's remove `unowned`.

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,354 @@
...
+  private func cancelScheduledPing() {
+    self.scheduledPing?.cancel()
+    self.scheduledPing = nil
+  }
+
+  private func cancelScheduledTimeout() {
+    self.scheduledClose?.cancel()
+    self.scheduledClose = nil
+  }
+}
```
Lukasa, line 161: Can we add an implementation of `handlerRemoved` that also cancels the two timeouts? That'll help with ensuring that any strong reference cycles really are broken.

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,354 @@
...
+
+final class PingHandler {
```
Lukasa, line 163: Is there any reason for this not to be a `struct`?

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,354 @@
...
+  /// Number of active streams
+  private var activeStreams = 0 {
+    didSet {
+      if activeStreams > 0 {
+        self.sentPingsWithoutData = 0
+      }
+    }
+  }
+
+  private let goAwayFrame = HTTP2Frame.FramePayload.goAway(lastStreamID: .rootStream, errorCode: .enhanceYourCalm, opaqueData: nil)
```
Lukasa, line 220: This can be a `static`.

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,354 @@
...
+
+  // For testing only
+  var testingOnlyNow: NIODeadline?
```
Lukasa, line 223: Nit: worth prefixing this with an underscore to keep it out of autocomplete?

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,354 @@
...
+    self.maximumPingsWithoutData = maximumPingsWithoutData
+    self.minimumSentPingIntervalWithoutData = minimumSentPingIntervalWithoutData
+    self.minimumReceivedPingIntervalWithoutData = minimumReceivedPingIntervalWithoutData
+    self.maximumPingStrikes = maximumPingStrikes
+  }
+
+  func streamCreated() -> Action {
+    self.activeStreams += 1
+
+    if self.activeStreams == 1 {
```
Lukasa, line 256: The choice of `1` here is interesting: can you elaborate on why we schedule the ping that way?

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,354 @@
...
+    } else {
+      return .none
+    }
+  }
+
+  func streamClosed() -> Action {
+    self.activeStreams -= 1
+
+    if self.activeStreams == 0 && !self.permitWithoutCalls {
+      return .cancelScheduledPing
```
Lukasa, line 267: This diverges slightly from the implied documented behaviour of grpc, which is that the keep alive timer fires cyclically, once per `GRPC_ARG_KEEPALIVE_TIME_MS` from connection completion. The ping is only _emitted_ based on some filtering criteria, but the timer still fires on this regular interval.

Given that you have this filtering criteria in place anyway, should we use a repeating call instead of a one-off scheduled call, and just cancel the repeating call in handlerRemoved?

Sources/GRPC/GRPCKeepaliveHandlers.swift
SebastianThiebaud: Changed.

Sources/GRPC/GRPCKeepaliveHandlers.swift
SebastianThiebaud: You're right, I already check this in `shouldBlockPing` so no need to add a check here.

Sources/GRPC/GRPCKeepaliveHandlers.swift
SebastianThiebaud: Changed and now based on `startedAt`. If it's `nil`, I start the ping schedule task.

SebastianThiebaud:
@Lukasa @glbrntt Last round (🤞) of review?

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,355 @@
...
+    let frame = self.wrapOutboundOut(.init(streamID: .rootStream, payload: payload))
+    context.writeAndFlush(frame, promise: nil)
+  }
+
+  private func schedulePing(delay: TimeAmount, timeout: TimeAmount, context: ChannelHandlerContext) {
+    guard delay != .nanoseconds(Int64.max) else { return }
+
+    self.scheduledPing = context.eventLoop.scheduleRepeatedTask(initialDelay: delay, delay: delay) { _ in
+      self.perform(action: self.pingHandler.pingFired(), context: context)
+      self.scheduleClose(timeout: timeout, context: context)
```
Lukasa, line 145: Hmm, this seems strange. We'll unconditionally schedule a close here, but that does not include cancelling any previously scheduled close. Is it our expectation that no previously scheduled close exists? What's the intended behaviour?

I find this line of code surprising enough that I think we should at least comment what the intended behaviour is.

SebastianThiebaud: The scheduled close is getting cancelled as soon as we get a pong back. I think it makes sense that `timeout` value is smaller than `interval` so we shouldn't have a situation where a new ping is generated before either a pong is received or the timeout is executed. I see that this is not documented and I realize now that we might have situations where a `timeout` might be greater than `interval`...

Lukasa: Yeah, if we want that to be a criterion that's fine, but we should a) add code to enforce that invariant, and b) comment here that that invariant is why we don't need to worry about this, and c) add an assertion for debug mode that ensures that invariant is upheld.

SebastianThiebaud: Would you rather have a failable initializer for both `ClientConnectionKeepalive` and `ServerConnectionKeepalive` which would guarantee a valid configuration?

SebastianThiebaud: Nevermind, carrying optional isn't fun.

SebastianThiebaud: Looking at the go implementation, `timeout` can be greater than `interval`, so we should respect that too. 
https://godoc.org/google.golang.org/grpc/keepalive

glbrntt: > Looking at the go implementation, `timeout` can be greater than `interval`, so we should respect that too.
> https://godoc.org/google.golang.org/grpc/keepalive

Unless I missed something, it's not clear to me whether that is or is not the case from the linked docs.

Are there cases where it would make sense for the `timeout` to be larger than the `interval`?

SebastianThiebaud:
@glbrntt I agree that a `timeout` larger than `interval` doesn't make sense. I'm fine making sure this doesn't happen. What's the best way to validate the configuration?

Sources/GRPC/GRPCKeepaliveHandlers.swift
SebastianThiebaud: I agree that a timeout larger than interval doesn't make sense. I'm fine making sure this doesn't happen. What's the best way to validate the configuration?

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,355 @@
...
+  var pingHandler: PingHandler
+
+  /// The scheduled task which will ping.
+  var scheduledPing: RepeatedTask? = nil
+
+  /// The scheduled task which will close the connection.
+  var scheduledClose: Scheduled<Void>? = nil
+
+  func close(context: ChannelHandlerContext) {
+    self.connectionManager?.idle()
```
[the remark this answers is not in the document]
SebastianThiebaud: Submitted something for this. Let me know.

Sources/GRPC/GRPCKeepaliveHandlers.swift
Lukasa: The answer there is usually to `precondition` on a valid configuration and express what the validity criteria are in the docs. As these values are unlikely to be chosen based on user-supplied data, a `precondition` is rarely going to be an issue here.

Sources/GRPC/ConnectionKeepalive.swift
```diff
@@ -0,0 +1,104 @@
...
+  public var minimumSentPingIntervalWithoutData: TimeAmount
+
+  public init(
+    interval: TimeAmount = .nanoseconds(Int64.max),
+    timeout: TimeAmount = .seconds(20),
+    permitWithoutCalls: Bool = false,
+    maximumPingsWithoutData: UInt = 2,
+    minimumSentPingIntervalWithoutData: TimeAmount = .minutes(5)
+  ) {
+    precondition(timeout < interval, "`timeout` should be less than `interval`")
```
glbrntt, line 48:
```suggestion
    precondition(timeout < interval, "`timeout` must be less than `interval`")
```

Sources/GRPC/ConnectionKeepalive.swift
```diff
@@ -0,0 +1,104 @@
...
+///
+/// The defaults are determined by the gRPC keepalive
+/// [documentation] (https://github.com/grpc/grpc/blob/master/doc/keepalive.md).
+public struct ClientConnectionKeepalive: Hashable {
+  /// The amount of time to wait before sending a keepalive ping.
+  public var interval: TimeAmount
+
+  /// The amount of time to wait for an acknowledgment.
+  /// If it does not receive an acknowledgment within this time, it will close the connection
+  /// This value should be less than `interval`
```
glbrntt, line 28:
```suggestion
  /// This value must be less than `interval`
```

Sources/GRPC/ConnectionKeepalive.swift
```diff
@@ -0,0 +1,104 @@
...
+  public init(
+    interval: TimeAmount = .hours(2),
+    timeout: TimeAmount = .seconds(20),
+    permitWithoutCalls: Bool = false,
+    maximumPingsWithoutData: UInt = 2,
+    minimumSentPingIntervalWithoutData: TimeAmount = .minutes(5),
+    minimumReceivedPingIntervalWithoutData: TimeAmount = .minutes(5),
+    maximumPingStrikes: UInt = 2
+  ) {
+    precondition(timeout < interval, "`timeout` should be less than `interval`")
```
glbrntt, line 95:
```suggestion
    precondition(timeout < interval, "`timeout` must be less than `interval`")
```

Sources/GRPC/ConnectionKeepalive.swift
```diff
@@ -0,0 +1,104 @@
...
+  }
+}
+
+public struct ServerConnectionKeepalive: Hashable {
+  /// The amount of time to wait before sending a keepalive ping.
+  public var interval: TimeAmount
+
+  /// The amount of time to wait for an acknowledgment.
+  /// If it does not receive an acknowledgment within this time, it will close the connection
+  /// This value should be less than `interval`
```
glbrntt, line 63:
```suggestion
  /// This value must be less than `interval`
```

docs/keepalive.md
```diff
@@ -0,0 +1,61 @@
...
+
+## What should I set?
+
+It should be sufficient for most users to only change `interval` and `timeout` properties, but the 
+following properties can also be useful in certain use cases.
+
+Property | Client | Server | Description
+---------|--------|--------|------------
+interval|Int64.max (disabled)|.hours(2)|The amount of time to wait before sending a keepalive ping.
+timeout|.seconds(20)|.seconds(20)|The amount of time to wait for an acknowledgment. This value should be less than `interval`.
```
glbrntt, line 15:
```suggestion
timeout|.seconds(20)|.seconds(20)|The amount of time to wait for an acknowledgment. This value must be less than `interval`.
```

Sources/GRPC/GRPCIdleHandler.swift
```diff
@@ -72,6 +72,9 @@ internal class GRPCIdleHandler: ChannelInboundHandler {
         if self.activeStreams == 0 {
           self.scheduleIdleTimeout(context: context)
         }
+      } else if event is StreamIdledEvent {
```
glbrntt, line 75:
```suggestion
      } else if event is ConnectionIdledEvent {
```

Sources/GRPC/StreamIdledEvent.swift
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import NIOHTTP2
+
+/// A `StreamIdledEvent` is fired whenever a connection has expired (keepalive).
```
glbrntt, line 18:
```suggestion
/// A `ConnectionIdledEvent` is fired whenever a connection has expired (keepalive).
```

Sources/GRPC/StreamIdledEvent.swift
```diff
@@ -0,0 +1,20 @@
...
+struct StreamIdledEvent: Hashable {
```
glbrntt, line 19:
```suggestion
struct ConnectionIdledEvent: Hashable {
```

Sources/GRPC/GRPCIdleHandler.swift
```diff
@@ -72,6 +72,9 @@ internal class GRPCIdleHandler: ChannelInboundHandler {
...
+        // Force idle (closing) because we received a `StreamIdledEvent` from a keepalive handler
```
glbrntt, line 76:
```suggestion
        // Force idle (closing) because we received a `ConnectionIdledEvent` from a keepalive handler
```

Sources/GRPC/GRPCKeepaliveHandlers.swift
```diff
@@ -0,0 +1,341 @@
...
+
+    self.scheduledPing = context.eventLoop.scheduleRepeatedTask(initialDelay: delay, delay: delay) { _ in
+      self.perform(action: self.pingHandler.pingFired(), context: context)
+      self.scheduleClose(timeout: timeout, context: context)
+    }
+  }
+
+  private func scheduleClose(timeout: TimeAmount, context: ChannelHandlerContext) {
+    self.scheduledClose = context.eventLoop.scheduleTask(in: timeout) {
+      context.fireUserInboundEventTriggered(StreamIdledEvent())
```
glbrntt, line 137:
```suggestion
      context.fireUserInboundEventTriggered(ConnectionIdledEvent())
```

Sources/GRPC/GRPCIdleHandler.swift
```diff
@@ -169,9 +172,10 @@ internal class GRPCIdleHandler: ChannelInboundHandler {
     }
   }
 
-  private func idle(context: ChannelHandlerContext) {
-    // Don't idle if there are active streams.
-    guard self.activeStreams == 0 else {
+  private func idle(context: ChannelHandlerContext, force: Bool = false) {
+    // Don't idle if there are active streams unless we manually request
+    // example: keepalive handler sends a `StreamIdledEvent` event
```
glbrntt, line 177:
```suggestion
    // example: keepalive handler sends a `ConnectionIdledEvent` event
```

Sources/GRPC/GRPCKeepaliveHandlers.swift
glbrntt: > Yeah, if we want that to be a criterion that's fine, but we should a) add code to enforce that invariant, and b) comment here that that invariant is why we don't need to worry about this, and c) add an assertion for debug mode that ensures that invariant is upheld.

b) and c) still need addressing.

PR #851, opened 2020-06-19
fix Examples/Google pod install failed

zhungxd:
when i build the example app
**pod install**  ->
Analyzing dependencies
[!] CocoaPods could not find compatible versions for pod "SwiftNIOSSL":
  In snapshot (Podfile.lock):
    SwiftNIOSSL (= 2.7.0)

  In Podfile:
    gRPC-Swift (= 1.0.0-alpha.11) was resolved to 1.0.0-alpha.11, which depends on
      SwiftNIOSSL (= 2.7.1)

Specs satisfying the `SwiftNIOSSL (= 2.7.0), SwiftNIOSSL (= 2.7.1)` dependency were found, but they required a higher minimum deployment target.

update the Podfile

### DIFF
--- a/Examples/Google/SpeechToText/Podfile
+++ b/Examples/Google/SpeechToText/Podfile
@@ -6,5 +6,5 @@ target 'SpeechToText-gRPC-iOS' do
   use_frameworks!
 
   pod 'SnapKit'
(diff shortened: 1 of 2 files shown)


zhungxd:
EasyCLA check passed

PR #855, opened 2020-06-24
Provide 'fake' responses for generated test clients

glbrntt:
Motivation:

When consuming gRPC it is often helpful to be able to write tests that
ensure the client is integrated correctly. At the moment this is only
possible by running a local gRPC server with a custom service handler to
return the responses you would like to test.

Modifications:

This pull request is the first of a series to introduce generated test
clients. It adds 'fake response streams', which allow the user to act as
the server for a single RPC. The stream comes in unary and streaming
varieties and is backed by an EmbeddedChannel.

Later PRs will connect these streaming objects to call objects and
provide code generation so that users can more readily make fake streams
for their test clients.

Result:

We're on the road to generated test clients.

---
This change is [Reviewable](https://reviewable.io/reviews/grpc/grpc-swift/855)


Sources/GRPC/FakeResponseStream.swift
```diff
@@ -0,0 +1,321 @@
...
+/// - `sendMessage(_:initialMetadata:trailingMetadata:status)`, or
+/// - `sendError(status:trailingMetadata)`
+///
+/// `sendMessage` sends a normal unary response with the provided message and allows the caller to
+/// also specify initial metadata, trailing metadata and the status. Both metadata arguments are
+/// empty by default and the status defaults to one with an 'ok' status code.
+///
+/// `sendError` may be used to terminate an RPC without providing a response. As for `sendMessage`,
+/// the `trailingMetadata` defaults to being empty.
+public class FakeUnaryResponse<Request: GRPCPayload, Response: GRPCPayload>: FakeResponseStream<Request, Response> {
```
davidpasztor, line 219: Making this type public means that this can be used in prod frameworks/apps depending on gRPC as well, not only their test targets. Is that the intended behaviour?
Do correct me if I'm wrong, but I would expect the test-only types to be internal and hence only accessible from (unit) test targets with `@testable import GRPC`.

Sources/GRPC/FakeResponseStream.swift
```diff
@@ -0,0 +1,321 @@
...
+ * limitations under the License.
+ */
+import NIO
+import NIOHPACK
+
+/// A fake response stream into which users may inject response parts for use in unit tests.
+///
+/// Users may not interact with this class directly but may do so via one of its subclasses
+/// `FakeUnaryResponse` and `FakeStreamingResponse`.
+public class FakeResponseStream<Request: GRPCPayload, Response: GRPCPayload> {
```
Lukasa, line 23: If we don't want users interacting with this directly, should its name be prefixed with an underscore?

Sources/GRPC/FakeResponseStream.swift
```diff
@@ -0,0 +1,321 @@
...
+    super.init()
+  }
+
+  /// Send a response message to the client.
+  ///
+  /// - Parameters:
+  ///   - response: The message to send.
+  ///   - initialMetadata: The initial metadata to send. By default the metadata will be empty.
+  ///   - trailingMetadata: The trailing metadata to send. By default the metadata will be empty.
+  ///   - status: The status to send. By default this have status code '.ok'.
```
Lukasa, line 230: Nit

```suggestion
  ///   - status: The status to send. By default this is status code '.ok'.
```

Sources/GRPC/FakeResponseStream.swift
```diff
@@ -0,0 +1,321 @@
...
+    switch self.filter(part) {
+    case .include:
+      self.writeOrBuffer(part)
+
+    case .includeAndPrepend(let extraPart):
+      self.writeOrBuffer(extraPart)
+      self.writeOrBuffer(part)
+
+    case .exclude:
+      ()
```
Lukasa, line 108: What's the reasoning for tolerating these errors? Should the user not deal with the crash, given that this is a test helper?

glbrntt: It's marginally easier to use in the happy case – which isn't exactly a strong reason – I'll change this.

PR #856, opened 2020-06-24
Refactor some of the client code generation

glbrntt:
Motivation:

In the future it might be useful to generate some of the client code in
multiple places.

Modifications:

- Split up the client code gen to print such that there are separate
  methods to print RPC methods
- The style of the generated code changed slightly too, to match the
  rest of the style
- Regenerate code (separate commit)

Result:

- We can avoid repeating ourselve in the future
- No functional change


glbrntt:
The actual changes are in: https://github.com/grpc/grpc-swift/pull/856/commits/b4b474977e46d8ce48c603691158286100cdee54

The other commit is just codegen.

glbrntt:
1000% agree; I've already filed #857 for this.

PR #860, opened 2020-06-25
Allow users to choose a policy for starting RPCs

glbrntt:
Motivation:

In most cases RPCs should be started once a connection has become
'ready', this allows RPCs to wait for transient network failures to be
recovered from instead of just failing. However in some contexts this is
undesirable, and failing RPCs quickly when it does not have a connection
is the right thing to do.

Modifications:

- Add a 'CallStartPolicy' which allows users RPCs to configure their
  connection such that RPCs will wait for the connection to be ready
  before starting the RPC or fail quickly if the most recent connection
  attempt failed.

Result:

Users can fail RPCs quickly if they do not wish to wait for a connection
to be eventually established.


Sources/GRPC/ClientConnection.swift
```diff
@@ -247,6 +257,40 @@ public enum ConnectionTarget {
   }
 }
 
+/// The policy used to determine how an RPC should be started in relation to the connection it
+/// will use.
+public struct CallStartPolicy: Hashable {
+  internal enum Policy: Hashable {
+    case waitForReadyConnection
+    case fastFailure
+  }
```
glbrntt, line 266: Very much open to naming suggestions here.

Lukasa: The prior art from Apple platforms is `waitsForConnectivity`, so we could investigate something in this vein.

glbrntt: I was concerned about overloading waiting for connectivity: in this context we might try to establish a connection a number of times during which – if adopted – we could see many `WaitingForConnectivity` events. If we later surface an option to disable waiting via NIOTS then we _could_ end up with a third policy which is somewhere between the two. 

However, thinking some more, disabling waiting for connectivity via NIOTS should just be done when applicable if the fast failure policy is being used. Observing the event in the connectivity delegate serves as extra information which _could_ have been used to cancel an RPC: however there would be no need to action on it to cancel RPCs since `fastFailure` would give you that behaviour anyway.

So, naming:

- `waitsForConnectivity`
- `fastFailure`

I'm also not super keen on `CallStartPolicy`, `Behaviour` seems like a better fit than `Policy`. WDYT?

Lukasa: Yeah, I like `Behaviour` better. The rest of your proposal sounds great

PR #865, opened 2020-06-30
Require generated client protocols to accept `CallOptions`

glbrntt:
Motivation:

The generated protocol for clients allows the caller to optionally pass
in some `CallOptions`. The generated client conforming to this protocol
defaults this argument to `nil` and falls back to default options set on
that client in that case.

It is therefore possible to call, for example,
`theClient.someUnaryRPC(someRequest)` but not
`theProtocol.someUnaryRPC(someRequest)`.

Without explicitly passing the call options, the protocol offers little
value when coding to the protocol rather than to a concrete
implementation.

Modifications:

- Generated client protocols now extend `GRPCClient` (previously the
  generated client implementations conformed to `GRPCClient` and the
  protocol)
- Generated client protocol requires `CallOptions`
- Generated client protocols now have a generated extension for each RPC
  which does not require call options whose implementation forwards the
  default call options courtesy of `GRPCClient`.

Result:

- It is possible to reasonably implement code against the generated
  protocol rather than a concrete implementation
- We lose the ability to call an RPC with `callOptions: nil`

### DIFF
--- a/Sources/protoc-gen-grpc-swift/Generator-Client.swift
+++ b/Sources/protoc-gen-grpc-swift/Generator-Client.swift
@@ -22,35 +22,106 @@ extension Generator {
     println()
     printServiceClientProtocol()
     println()
+    printClientProtocolExtension()
+    println()
     printServiceClientImplementation()
   }
 
+  private func printFunction(name: String, arguments: [String], returnType: String, access: String? = nil, bodyBuilder: (() -> ())?) {
+    // Add a space after access, if it exists.
+    let accessOrEmpty = access.map { $0 + " " } ?? ""
+
+    let hasBody = bodyBuilder != nil
+
+    if arguments.isEmpty {
+      // Don't bother splitting across multiple lines if there are no arguments.
+      self.println("\(accessOrEmpty)func \(name)() -> \(returnType)", newline: !hasBody)
+    } else {
+      self.println("\(accessOrEmpty)func \(name)(")
+      self.withIndentation {
+        // Add a comma after each argument except the last.
+        arguments.forEach(beforeLast: {
+          self.println($0 + ",")
+        }, onLast: {
+          self.println($0)
+        })
+      }
+      self.println(") -> \(returnType)", newline: !hasBody)
+    }
+
+    if let bodyBuilder = bodyBuilder {
+      self.println(" {")
+      self.withIndentation {
+        bodyBuilder()
+      }
+      self.println("}")
+    }
+  }
+
   private func printServiceClientProtocol() {
-    println("/// Usage: instantiate \(clientClassName), then call methods of this protocol to make API calls.")
-    println("\(options.visibility.sourceSnippet) protocol \(clientProtocolName) {")
-    indent()
-    for method in service.methods {
-      self.method = method
-      switch streamingType(method) {
-      case .unary:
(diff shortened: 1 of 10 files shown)


glbrntt:
Note: https://github.com/grpc/grpc-swift/pull/865/commits/3c6fe84c7081f09175add493735d711b4acdd5cd contains the actual change and a regenerated Echo client. The other commits are just other clients regenerated.

Sources/protoc-gen-grpc-swift/Generator-Client.swift
```diff
@@ -22,35 +22,106 @@ extension Generator {
...
+      self.withIndentation {
+        bodyBuilder()
+      }
+      self.println("}")
+    }
+  }
+
+  private func printServiceClientProtocol() {
+    self.println("/// Usage: instantiate \(self.clientClassName), then call methods of this protocol to make API calls.")
+    self.println("\(self.access) protocol \(self.clientProtocolName): GRPCClient {")
```
Lukasa, line 63: Do we want to continue putting the access level of this protocol in the generated code? Or would it be better just to put it on all the methods? This gets us a bit closer to NIO, but I don't know if we want to do that.

Lukasa: It would also avoid the need for `accessOrEmpty` in `printFunction`, as all other call sites have the access passed.

Sources/protoc-gen-grpc-swift/Generator-Client.swift
```diff
@@ -226,6 +289,73 @@ fileprivate extension Generator {
...
+    case .bidirectionalStreaming:
+      return [
+        "callOptions: CallOptions",
+        "handler: @escaping (\(methodOutputName)) -> Void"
+      ]
+    }
+  }
+
+
+  fileprivate var methodArgumentsWithoutCallOptions: [String] {
```
Lukasa, line 319: To avoid the risk of getting things wrong here, should this be defined programmatically in terms of `methodArguments` by delegating to `methodArguments` and stripping the one that starts with `callOptions:`?

Sources/protoc-gen-grpc-swift/Generator-Client.swift
```diff
@@ -257,3 +387,20 @@ extension MethodDescriptor {
...
+    var index = self.startIndex
+    while index != endIndex {
+      let element = self[index]
+      self.formIndex(after: &index)
+      if index != endIndex {
+        body(element)
+      } else {
+        last(element)
+      }
+    }
```
Lukasa, line 404: Might be less difficult to write this in terms of sequence ops:

```swift
for element in self.dropLast() {
    body(element)
}
if let last = self.last {
    last(element)
}
```

Sources/protoc-gen-grpc-swift/Generator-Client.swift
glbrntt: For protocol declaration it _has_ to be on the `protocol` rather than the method sadly :(

Sources/protoc-gen-grpc-swift/Generator-Client.swift
glbrntt: Yeah, good call.

Sources/protoc-gen-grpc-swift/Generator-Client.swift
glbrntt: 🤦‍♂️ not sure why that didn't occur to me, thanks!

PR #866, opened 2020-06-30
Turn Network{Implementation,Preference} enums into structs

glbrntt:
Motivation:

Public enums are a pain for API evolution and there's no need for users to
be able to exhaustively switch over `NetworkImplementation` and
`NetworkPreference`.

Modifications:

- Turn both into structs backed by enums

Result:

More evolvable API.

### DIFF
--- a/Sources/GRPC/PlatformSupport.swift
+++ b/Sources/GRPC/PlatformSupport.swift
@@ -18,27 +18,50 @@ import NIOTransportServices
 import Logging
 
 /// How a network implementation should be chosen.
-public enum NetworkPreference {
+public struct NetworkPreference: Hashable {
+  private enum Wrapped: Hashable {
+    case best
+    case userDefined(NetworkImplementation)
+  }
+
+  private var wrapped: Wrapped
+  private init(_ wrapped: Wrapped) {
+    self.wrapped = wrapped
+  }
+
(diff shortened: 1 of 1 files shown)


Sources/GRPC/PlatformSupport.swift
```diff
@@ -18,27 +18,50 @@ import NIOTransportServices
 import Logging
 
 /// How a network implementation should be chosen.
-public enum NetworkPreference {
+public struct NetworkPreference {
```
Lukasa, line 21: Any reason for this not to be `Hashable`?

Sources/GRPC/PlatformSupport.swift
```diff
@@ -18,27 +18,50 @@ import NIOTransportServices
...
-  case userDefined(NetworkImplementation)
+  public static func userDefined(_ implementation: NetworkImplementation) -> NetworkPreference {
+    return NetworkPreference(.userDefined(implementation))
+  }
 }
 
 /// The network implementation to use: POSIX sockets or Network.framework. This also determines
 /// which variant of NIO to use; NIO or NIOTransportServices, respectively.
-public enum NetworkImplementation {
+public struct NetworkImplementation {
```
Lukasa, line 46: Any reason for this not to be `Hashable`?

glbrntt:
ping @Lukasa

Issue #802, opened 2020-05-20
Include Plugins in Cocoapods checkout


Lutzifer:
Enhancement Proposal:

How about including precompiled versions of the plugins in the cocoapods checkout similar to how google does it with the crashlytics support tools 'run' and 'upload-symbols'? (see   https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCrashlytics.podspec)

Compiling the protofiles could then be something like

```
for PROTO_FILE in $BASE_PATH/ProtobufDefinitions/*.proto; do
  protoc $PROTO_FILE \
		--proto_path=$BASE_PATH/ProtobufDefinitions \
		--plugin=./Pods/gRPC-Swift/bin/protoc-gen-swift \
		--swift_opt=Visibility=Public \
		--swift_out=$BASE_PATH/ProtobufGeneratedCode/
  protoc $PROTO_FILE \
  		--proto_path=$BASE_PATH/ProtobufDefinitions \
  		--plugin=./Pods/gRPC-Swift/bin/protoc-gen-grpc-swift \
  		--grpc-swift_opt=Visibility=Public \
      --grpc-swift_out=$BASE_PATH/GRPCGeneratedCode/
done;
```
(the most interesting part here is the use of 'Pods/gRPC-Swift/bin')
and the plugins and grpc-swift would always use versions that fit together.

MrMage:
Given the amount of issues we have been encountering due to version mismatches between plugin and library, I like the idea. I am not an expert in how to best package that up, though, and I'm not sure if we have the bandwidth to tackle this right now 😅

But if you want to look into this, we would really appreciate it :-)

@glbrntt, what do you think?

glbrntt:
I think this is a great idea; I also don't have bandwidth at the moment. @Lutzifer is this something you'd like to contribute?

Lutzifer:
I will try to find the time to look into this.

Some Ideas:
- add a podspec for the plugins, which references gRPC-Swift.podspec as a dependency
- include the binary through this new plugin podspec, similar to how https://github.com/SwiftGen/SwiftGen/blob/master/SwiftGen.podspec does it
- missing part: how do we upload the compiled plugins and add them to a release? Code for this can be found at https://github.com/SwiftGen/SwiftGen/blob/c6477a6950d313c94c964c6c8ec960e3aeccb8f6/rakelib/release.rake#L132. Does this repo already have a script to make a release? In which order do you release what and how?

[...]

PR #869, opened 2020-07-01
Add Generation of gRPC-Swift.podspec

Lutzifer:
This implements the automatic Podspec Generation part for plugins as discussed in #802 

As there was some discussion about Automating Releases (see #830 ) via Github Actions, I left the uploading part of the plugins to Github manual for now.


scripts/build_podspecs.py
```diff
@@ -75,29 +77,39 @@ def as_podspec(self):
...
+        podspec += indent + "s.version = '%s'\n" % self.version
+        podspec += indent + "s.license = { :type => 'Apache 2.0', :file => 'LICENSE' }\n"
+        podspec += indent + "s.summary = '%s'\n" % self.description
+        podspec += indent + "s.homepage = 'https://www.grpc.io'\n"
+        podspec += indent + "s.authors  = { 'The gRPC contributors' => \'grpc-packages@google.com' }\n\n"
+
+        if self.module_name:
+            podspec += indent + "s.source = { :git => \"https://github.com/grpc/grpc-swift.git\", :tag => s.version }\n\n"
+        else:
+            podspec += indent + "s.source = { :http => \"https://github.com/grpc/grpc-swift/releases/download/#{s.version}/protoc-swift-plugins-#{s.version}.zip\"}\n\n"
```
glbrntt, line 95: I think we need to be more specific with the naming here: `protoc-grpc-swift-plugins-VERSION-macOS-x86_64.zip`

scripts/build_podspecs.py
```diff
@@ -75,29 +77,39 @@ def as_podspec(self):
...
+            podspec += indent + "s.source_files = 'Sources/%s/**/*.{swift,c,h}'\n" % (self.module_name)
+
+            podspec += "\n" if len(self.dependencies) > 0 else ""
 
         for dep in self.dependencies:
             podspec += dep.as_podspec()
+                
+        if not self.module_name:
+            podspec += indent + "s.preserve_paths = '*'\n"
+            podspec += indent + "s.exclude_files = '**/file.zip'\n"
```
glbrntt, line 112: This doesn't seem necessary?

scripts/build_podspecs.py
```diff
@@ -75,29 +77,39 @@ def as_podspec(self):
...
+            podspec += indent + "s.tvos.deployment_target = '10.0'\n"
+            podspec += indent + "s.source_files = 'Sources/%s/**/*.{swift,c,h}'\n" % (self.module_name)
+
+            podspec += "\n" if len(self.dependencies) > 0 else ""
 
         for dep in self.dependencies:
             podspec += dep.as_podspec()
+                
+        if not self.module_name:
+            podspec += indent + "s.preserve_paths = '*'\n"
```
glbrntt, line 111: What does this do?

scripts/build_podspecs.py
```diff
@@ -75,29 +77,39 @@ def as_podspec(self):
...
+        podspec += indent + "s.name = '%s'\n" % self.name
+        if self.module_name:
+            podspec += indent + "s.module_name = '%s'\n" % self.module_name
+        podspec += indent + "s.version = '%s'\n" % self.version
+        podspec += indent + "s.license = { :type => 'Apache 2.0', :file => 'LICENSE' }\n"
+        podspec += indent + "s.summary = '%s'\n" % self.description
+        podspec += indent + "s.homepage = 'https://www.grpc.io'\n"
+        podspec += indent + "s.authors  = { 'The gRPC contributors' => \'grpc-packages@google.com' }\n\n"
+
+        if self.module_name:
```
glbrntt, line 92: I don't really like using the `module_name` as a "are we doing the plugins podspec or not"; can you make this more explicit?

scripts/build_podspecs.py
Lutzifer: leftover from https://github.com/SwiftGen/SwiftGen/blob/stable/SwiftGen.podspec

scripts/build_podspecs.py
Lutzifer: Added -macOS-x86_64.zip for now, but it is not yet dynamic

scripts/build_podspecs.py
Lutzifer: change it to use a new flag called is_plugins_pod

scripts/build_podspecs.py
Lutzifer: Without this line, the checkout is simply empty.

If I understood https://stackoverflow.com/questions/18373760/cocoapods-podspec-preserve-paths correctly, as we only add files from the zip, we need to specify all files from inside the zip to not be deleted.

Lutzifer:
Caveat: In this configuration, the Licence File must be included in the zip file along with the two plugins.

README.md
```diff
@@ -117,6 +117,29 @@ To install these plugins, just copy the two executables (`protoc-gen-swift` and
...
+		--swift_out=ProtobufGeneratedCode/
+```
+
+and
+
+```
+protoc <Protobuf Definition File> \
+ --plugin=Pods/gRPC-Swift-Plugins/protoc-gen-grpc-swift \
+ --grpc-swift_out=GRPCGeneratedCode/
+```
```
glbrntt, line 141: I don't think we need examples of running `protoc` here; saying that the precompiled binaries are available via that pod is fine.

scripts/build_podspecs.py
```diff
@@ -75,29 +78,38 @@ def as_podspec(self):
...
+        podspec += indent + "s.summary = '%s'\n" % self.description
+        podspec += indent + "s.homepage = 'https://www.grpc.io'\n"
+        podspec += indent + "s.authors  = { 'The gRPC contributors' => \'grpc-packages@google.com' }\n\n"
+
+        if self.is_plugins_pod:
+            podspec += indent + "s.source = { :http => \"https://github.com/grpc/grpc-swift/releases/download/#{s.version}/protoc-swift-plugins-#{s.version}-macOS-x86_64.zip\"}\n\n"
+        else:
+            podspec += indent + "s.source = { :git => \"https://github.com/grpc/grpc-swift.git\", :tag => s.version }\n\n"
+        
+        if not self.is_plugins_pod:
```
glbrntt, line 98: We can just merge this into the `else` block.

gRPC-Swift-Plugins.podspec
```diff
@@ -0,0 +1,15 @@
+Pod::Spec.new do |s|
+
+    s.name = 'gRPC-Swift-Plugins'
+    s.version = '1.0.0-alpha.14'
+    s.license = { :type => 'Apache 2.0', :file => 'LICENSE' }
+    s.summary = 'Swift gRPC code generator plugin binaries'
+    s.homepage = 'https://www.grpc.io'
+    s.authors  = { 'The gRPC contributors' => 'grpc-packages@google.com' }
+
+    s.source = { :http => "https://github.com/grpc/grpc-swift/releases/download/#{s.version}/protoc-swift-plugins-#{s.version}-macOS-x86_64.zip"}
```
glbrntt, line 10: Sorry I think I've changed my position on this slightly: we need to wary of the fact that Macs on Apple Silicon are a thing. Either we need separate pods: or we should structure the zip appropriately and bundle them up together. What do you think @MrMage?

Lutzifer: I think including binaries for both architectures in the zip either in folders or as fat binaries if possible would be better than having separate pods.

Lutzifer: I will change the name of the zip again, then.

glbrntt: (Unresolving since I'd like to hear @MrMage's opinion here)

gRPC-Swift-Plugins.podspec
```diff
@@ -0,0 +1,15 @@
+Pod::Spec.new do |s|
```
glbrntt, line 1: This has been useful to see in the review but I don't think we should publish it to `master` yet (since the zip doesn't exist and no pod has been pushed).

README.md
```diff
@@ -117,13 +117,22 @@ To install these plugins, just copy the two executables (`protoc-gen-swift` and
 that is part of your `PATH` environment variable. Alternatively the full path to
 the plugins can be specified when using `protoc`.
 
+Alternatively, you can get the latest precompiled version of the plugins by adding
```
glbrntt, line 120: Perfect! (Thanks for fixing the typos below as well.)

scripts/build_podspecs.py
```diff
@@ -135,10 +142,20 @@ def build_pods(self):
...
             get_grpc_deps()
         )
+        
+        grpc_plugins_pod = Pod(
+            'gRPC-Swift-Plugins',
+            '',
+            self.version,
+            'Swift gRPC code generator plugin binaries',
+            [],
+            True
```
glbrntt, line 152:
```suggestion
            is_plugins_pod=True
```

gRPC-Swift-Plugins.podspec
MrMage: Full-fledged Mac app bundles can be distributed as a "fat" binary containing both architectures. Would that be a possibility here as well?

Lutzifer: When looking at the project generated by "make project" I would dare to say that this should be possible. Have not tried it in xcode12 though.

glbrntt: Not sure since if that would work or not. I don't think the details are important at the moment though; the relevant question is do we distribute them in a single zip? It seems like we have consensus that we should.

scripts/build_podspecs.py
glbrntt: This one also hasn't been resolved yet.

scripts/build_podspecs.py
```diff
@@ -91,7 +91,7 @@ def as_podspec(self):
         podspec += indent + "s.authors  = { 'The gRPC contributors' => \'grpc-packages@google.com' }\n\n"
 
         if self.is_plugins_pod:
-            podspec += indent + "s.source = { :http => \"https://github.com/grpc/grpc-swift/releases/download/#{s.version}/protoc-swift-plugins-#{s.version}-macOS-x86_64.zip\"}\n\n"
+            podspec += indent + "s.source = { :http => \"https://github.com/grpc/grpc-swift/releases/download/#{s.version}/protoc-swift-plugins-#{s.version}.zip\"}\n\n"
```
Lutzifer, line 94:
```suggestion
            podspec += indent + "s.source = { :http => \"https://github.com/grpc/grpc-swift/releases/download/#{s.version}/protoc-grpc-swift-plugins-#{s.version}.zip\"}\n\n"
```

PR #870, opened 2020-07-01
Provide the codegen with an option to generate test clients

glbrntt:
Motivation:

When consuming gRPC it is often helpful to be able to write tests that
ensure the client is integrated correctly. At the moment this is only
possible by running a local gRPC server with a custom service handler to
return the responses you would like to test.

Modifications:

This builds on work in #855, #864, and #865.

This pull request introduces code generation for the test clients. It
provides type-safe wrappers and convenience methods on top of
`FakeChannel` and a code-gen option to enable 'TestClient' generation.

It also removes an `init` requirement on the `GRPCClient` protocol.

Result:

Users can generate test clients.

---
This change is [Reviewable](https://reviewable.io/reviews/grpc/grpc-swift/870)


glbrntt:
@MrMage this is the relevant commit to look at (the other is regenerating code): https://github.com/grpc/grpc-swift/pull/870/commits/50464e07e5a2243126987ddeea2015e56c9a3078

PR #872, opened 2020-07-03
Add a script to generate a bundle of plugins for CocoaPods

glbrntt:
Motivation:

https://github.com/grpc/grpc-swift/pull/869 adds a plugin containing the
protoc binaries for CocoaPods. It relies on having a zip of the binaries
in the tagged release.

Modifications:

- Add a script to generate a zip of the protoc binaries.

Result:

We can generate a bundle of plugins which may be pulled in by a
CocoaPod.

---
This change is [Reviewable](https://reviewable.io/reviews/grpc/grpc-swift/872)

### DIFF
--- a/scripts/bundle-plugins-for-release.sh
+++ b/scripts/bundle-plugins-for-release.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+
+# Copyright 2020, gRPC Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


glbrntt:
cc @Lutzifer

Lutzifer:
Should this be another Makefile target?

glbrntt:
> Should this be another Makefile target?

A script is fine; it'll only be run by me or @MrMage when we tag a release.

Lutzifer:
Ah, thanks to the new Github design, I thought this was only an issue, not a PR 🙈

Good job!

Issue #877, opened 2020-07-06
Nested messages don't have conformance to `GRPCProtobufPayload` generated


davidpasztor:
### Describe the bug

After updating to 1.0.0-alpha.15, updating protobuf to the latest version and regenerating the proto files, the generated code doesn't compile anymore, due to missing `extension ProtoType: GRPCProtobufPayload {}` conformances for all nested types.

Dependency versions:
grpc-swift 1.0.0-alpha.15
protobuf 3.12.3
swift-protobuf 1.10.0

Xcode 11.5

### To reproduce

1) Create a proto file with messages that contain nested messages.

All.pb
```
message ProtoMarketData {
    oneof marketData {
        List list = 1;
        Detail detail = 2;
    }

    message List {
        int32 id = 1;
    }

    message Detail {
        int32 id = 1;
    }
}
```

2) Generate the swift files
3) Try to build the swift files
4) Compiler error
>Type 'ProtoMarketData.Detail' does not conform to protocol 'GRPCPayload'

### Expected behaviour

The generated code should compile

### Additional information
Looking at the All.grpc.swift file, it contains a line `extension ProtoMarketData: GRPCProtobufPayload {}`, but no extensions for its nested messages, `List` and `Detail`.

glbrntt:
I don't think we ever explicitly supported this; I think you were just lucky if it worked before. Nevertheless, definitely a bug. I'll have a patch up fixing this shortly.

davidpasztor:
@glbrntt thanks for quickly jumping on this! I wasn't aware that this was never explicitly supported, since it was working fine before.

glbrntt:
No problem. I don't like it when I break working code!

Feel free to checkout my branch with the fix here to unblock yourself: https://github.com/glbrntt/grpc-swift/tree/gb-conformance-nested-message

glbrntt:
This was fixed by #879

PR #879, opened 2020-07-06
Generate conformance when message are nested

glbrntt:
Motivation:

Protobuf message definitions may be nested yet we only generated payload
conformance for the top-level message.

Modifications:

- Generate conformance for messages defined within of other messages
- Update test

Result:

Messages which are defined within another message will have conformance
for GRPCProtobufPalyoad generated.

---
This change is [Reviewable](https://reviewable.io/reviews/grpc/grpc-swift/879)


glbrntt:
Fixes #877

Issue #778, opened 2020-04-30
Redundant GRPCProtobufPayload conformance


glbrntt:
I updated to alpha-11 and still run into this issue in some cases where redundant GRPCProtobufPayload conformance was generated for structs.

Unfortunately I can't share the original proto files and the generated swift files, but pls let me know how I could help debug this issue.

_Originally posted by @kmarcell in https://github.com/grpc/grpc-swift/pull/740#issuecomment-621776043_

glbrntt:
@kmarcell, please just create an issue in the future instead of commenting on closed PRs; it's easier for others who run in to the same problem to find the issue.

I just want to clarify the question: do you have redundancy conformances created (i.e. conformance is generated for types you don't need) or do you have multiple conformances created (i.e. one message has conformance generated for it multiple times)?

glbrntt:
@kmarcell are you still having this issue?

glbrntt:
Closing this because of inactivity. Please re-open if you're still facing this issue.

PR #885, opened 2020-07-10
Remove the requirement that messages conform to GRPCPayload on the server

glbrntt:
Motivation:

To support payloads other than `SwiftProtobuf.Message` we required that
all messages conform to `GRPCPayload`. For protobuf messages we added
`GRPCProtobufPayload` which provides a default implemenation of
`GRPCPayload` for protobuf messages. We generated this conformance for
all protobuf messages we saw. This lead to a number issues and
workarounds including: #738, #778, #801, #837, #877, #881.

The intention is to continue to support `GRPCPayload` in addition to
protobuf, however, support for protobuf will not be via the
`GRPCProtobufPayload` protocol.

This PR adjust the server components such they only support
SwiftProtobuf. Once the client side has had the same treatment (and
`GRPCProtobufPayload` no longer inherits from `SwiftProtobuf.Message`),
support for `GRPCPayload` will be added back.

Modifications:

- The `HTTP1ToGRPCServerCodec` has had the message encoding and decoding
  removed. It now deals in `ByteBuffer`s rather than request/response
  messages.
- An additional `GRPCServerCodecHandler` which sits between the
  `HTTP1ToGRPCServerCodec` and `_BaseCallHandler` has been added which
  serializes/deserializes messages.
- Custom payload tests have been commented out. They will return when
  the transition has completed.

Result:

- Servers only support SwiftProtobuf
- Genertic constraints on the server have been removed; the constraints
  are place on the `init` of public handlers instead.
- `GRPCProtobufPayload` is no longer required on the server.


glbrntt:
Sigh... this works for the 5.3 compiler targeting Swift version 5.0 but not before the 5.2 compiler. Since that's not released yet we'll have to take a different route.

PR #886, opened 2020-07-10
Remove the requirement that messages conform to GRPCPayload on the server

glbrntt:
Motivation:

To support payloads other than `SwiftProtobuf.Message` we required that
all messages conform to `GRPCPayload`. For protobuf messages we added
`GRPCProtobufPayload` which provides a default implemenation of
`GRPCPayload` for protobuf messages. We generated this conformance for
all protobuf messages we saw. This lead to a number issues and
workarounds including: #738, #778, #801, #837, #877, #881.

The intention is to continue to support `GRPCPayload` in addition to
protobuf, however, support for protobuf will not be via the
`GRPCProtobufPayload` protocol.

This PR adjusts server components such that they are not constrained to
`GRPCPayload`. At the moment only `SwiftProtobuf.Message` is supported.
Once the client side has had the same treatment and
`GRPCProtobufPayload` no longer inherits from `SwiftProtobuf.Message`,
support for `GRPCPayload` will be added back.

Modifications:

- The `HTTP1ToGRPCServerCodec` has had the message encoding and decoding
  removed. It now deals in `ByteBuffer`s rather than request/response
  messages.
- An additional `GRPCServerCodecHandler` which sits between the
  `HTTP1ToGRPCServerCodec` and `_BaseCallHandler` has been added which
  serializes/deserializes messages.
- Custom payload tests have been commented out. They will return when
  the transition has completed.

Result:

- Servers only support SwiftProtobuf
- Genertic constraints on the server have been removed; the constraints
  are place on the `init` of public handlers instead.
- `GRPCProtobufPayload` is no longer required on the server.


Sources/GRPC/CallHandlers/_BaseCallHandler.swift
```diff
@@ -23,12 +23,11 @@ import Logging
...
-public class _BaseCallHandler<RequestPayload: GRPCPayload, ResponsePayload: GRPCPayload>: GRPCCallHandler {
+public class _BaseCallHandler<Request, Response>: GRPCCallHandler {
+  private let codec: ChannelHandler
+
   public func makeGRPCServerCodec() -> ChannelHandler {
-    return HTTP1ToGRPCServerCodec<RequestPayload, ResponsePayload>(
-      encoding: self.callHandlerContext.encoding,
-      logger: self.logger
-    )
+    return self.codec
```
Lukasa, line 30: This has very subtly changed behaviour: it now returns the _same_ codec everywhere, while before it returned different ones. I believe it's ok, but we need to make sure we document clearly on the codecs that they are _required_ to be stateless.

Lukasa: In fact this is so unlikely to be achieved (as those codecs are generic over other things) that it would maybe be better to require that this have a codec _factory_, not a codec.

Sources/GRPC/LengthPrefixedMessageWriter.swift
```diff
@@ -41,6 +41,63 @@ internal struct LengthPrefixedMessageWriter {
...
+
+    // Leave a gap for the length, we'll set it in a moment.
+    let payloadSizeIndex = output.writerIndex
+    output.moveWriterIndex(forwardBy: MemoryLayout<UInt32>.size)
+
+    // Compress the message. We know that we need to drop the first 5 bytes, and we know that these
+    // bytes must exist.
+    var buffer = buffer
+    buffer.moveReaderIndex(forwardBy: 5)
+    let bytesWritten = try compressor.deflate(&buffer, into: &output)
```
Lukasa, line 60: This scrambles `output` if it throws: is that intentional?

Sources/GRPC/Serialization.swift
```diff
@@ -0,0 +1,108 @@
...
+import SwiftProtobuf
+
+internal protocol MessageSerializer {
+  associatedtype Input
+
+  /// Serializes `input` into a `ByteBuffer` allocated using the provided `allocator`.
+  ///
+  /// The serialized buffer should have 5 leading bytes: the first must be zero, the following
+  /// four bytes are the `UInt32` encoded length of the serialized message. The bytes of the
+  /// serialized message follow.
```
Lukasa, line 27: Woof, that's not a great API. Is there any reason not to instead update this to require a `serializeInto` method instead, and put the bytes we want there ourselves instead?

Sources/GRPC/Shims.swift
```diff
@@ -288,3 +289,44 @@ public struct GRPCTimeoutError: Error, Equatable, CustomStringConvertible {
   /// the gRPC wire-format.
   public static let tooManyDigits = GRPCTimeoutError(.tooManyDigits)
 }
+
+
+extension UnaryCallHandler {
+  @available(*, deprecated, message: "Please regenerate your code or use 'CallHandler.makeUnary'")
```
Lukasa, line 295: Maybe it's better to just mark this `unavailable` instead of `deprecated`, given that it explodes anyway?

Sources/GRPC/Serialization.swift
glbrntt: I agree; it's a bit janky. I'd feel much worse about this if it were public API. The tradeoff with doing as you suggest is that you need to guess how large to allocate the buffer. 

Although if you just allocate with a minimum capacity of 5 bytes then it won't be _too_ bad since the copy with any required change in capacity will be small. Not sure about the tradeoff here.

Sources/GRPC/Shims.swift
glbrntt: Didn't realise that was err.. available. Thanks for the tip.

Sources/GRPC/LengthPrefixedMessageWriter.swift
glbrntt: It is not. Good catch!

Sources/GRPC/CallHandlers/_BaseCallHandler.swift
glbrntt: True; this is only ever called once per RPC. We should just change the API to reflect that.

Sources/GRPC/Serialization.swift
glbrntt: Or we could also break this up:

```swift
protocol MessageSerializer {
  associatedtype Input

  // returns the serialized size, or an estimate, if known.
  //
  // For `Message`: serialize and hold on to the `Data`, returns the exact size
  // For `GRPCPayload`: hold on to the input, return `nil`
  func willSerialize(_ input: Input) throws -> Int?
  
  // returns the actual number of bytes written.
  func writeSerializedBytes(into buffer: inout ByteBuffer) -> Int
}
```

Lukasa: Yeah, I dunno. Perhaps it's better to leave it for now.

Sources/GRPC/CallHandlers/_BaseCallHandler.swift
```diff
@@ -23,13 +23,8 @@ import Logging
...
 /// - Important: This is **NOT** part of the public API.
-public class _BaseCallHandler<RequestPayload: GRPCPayload, ResponsePayload: GRPCPayload>: GRPCCallHandler {
-  public func makeGRPCServerCodec() -> ChannelHandler {
-    return HTTP1ToGRPCServerCodec<RequestPayload, ResponsePayload>(
-      encoding: self.callHandlerContext.encoding,
-      logger: self.logger
-    )
-  }
+public class _BaseCallHandler<Request, Response>: GRPCCallHandler {
+  public let codec: ChannelHandler
```
Lukasa, line 27: Should this also be underscored, for consistency? We don't actually want users touching it.

PR #888, opened 2020-07-13
Alter the documented requirements of `MessageSerializer`

glbrntt:
Motivation:

requirements that implementations must prefix their serialized messages
with a compression byte and `UInt32` length. The motivation behind this
was to avoid a reallocation later on when the message was being framed.
The result was that in some cases the framing would be done by the
serialier and in other cases by the framer (or
`LengthPrefixedMessageWriter`).

Modifications:

- Remove the requirement for the serialier to frame the message
- Optimize the `LengthPrefixedMessageWriter` so that it checks whether
  there is enough space _before_ the serialized bytes in which the
  compression byte and length could be written.
- gRPC-Web: Replace the single response buffer with a circular buffer of
  buffers; this avoids writing into a buffer and simplifies the
  `LengthPrefixedMessageWriter`.

Result:

- Better separation of concerns.


Sources/GRPC/HTTP1ToGRPCServerCodec.swift
```diff
@@ -383,7 +379,14 @@ extension HTTP1ToGRPCServerCodec: ChannelOutboundHandler {
       }
 
       if contentType == .webTextProtobuf {
-        precondition(responseTextBuffer != nil)
+        var responseTextBuffer = self.responseTextBuffers.popFirst() ?? context.channel.allocator.buffer(capacity: 5)
+
+        // Append the rest of the buffers.
+        let requiredExtraCapacity = self.responseTextBuffers.map { $0.readableBytes }.reduce(0, +)
```
Lukasa, line 385: May be worth using `.lazy` here to avoid the intermediate array of ints.

Sources/GRPC/HTTP1ToGRPCServerCodec.swift
```diff
@@ -394,14 +397,13 @@ extension HTTP1ToGRPCServerCodec: ChannelOutboundHandler {
...
+        // client decoding. Initial results say that they are innocuous, but we might have to keep
         // an eye on this in case something trips up.
-        if let binaryData = responseTextBuffer.readData(length: responseTextBuffer.readableBytes) {
-          let encodedData = binaryData.base64EncodedString()
-          responseTextBuffer.clear()
-          responseTextBuffer.reserveCapacity(encodedData.utf8.count)
-          responseTextBuffer.writeString(encodedData)
-        }
+        let binaryData = responseTextBuffer.readData(length: responseTextBuffer.readableBytes)!
+        let encodedData = binaryData.base64EncodedString()
```
Lukasa, line 403: I was wondering why we had to do all that compaction of `ByteBuffer`s into a linear data structure, but now I see why. This is actually quite a lot of intermediate byte copying to only have to (potentially) copy those bytes again. At the absolute minimum we should use the flag to `readData` that prevents the copy, and hope that we will dodge the CoW operation that we hit here.

But the other option is to just do all of this work in `Data`-space. Given that we need a `Data` _anyway_, should we just copy all the bytes into a `Data` via `ByteBufferView`?

This will probably be marginally slower than doing this in `ByteBuffer`-land, but I suspect that slowness will be marginal at best, and we'll avoid a lot of complexity in _this_ code. We can still keep hold of the one buffer for clearing.

glbrntt: Yeah, the gRPC Web path is a bit unfortunate... The `Data`-land approach makes sense, I'll do that.

Sources/GRPC/HTTP1ToGRPCServerCodec.swift
glbrntt: Good suggestion, thanks.

PR #902, opened 2020-07-20
Allow loggers to be passed in

glbrntt:
Motivation:

The community recently discussed how logging should be done for the
`AsyncHTTPClient` project. One of the outcomes was that frameworks
should not log unless they are explicitly passed a logger. This allows
applications to decide whether they care about logging for that
component and have more control over the level at which that component
logs at.

Modifications:

- Client allows for loggers in two places: one for background activity
  (connectivity state management and so on) and another for RPCs
- Server allows a single logger to be specificed
- By default all loggers are no-op loggers
- Added 'none' option for 'CallOptions.RequestIDProvider' for cases
  where the request ID is already present on the logger

Result:

- gRPC does not log by default
- Note: this includes not logging in tests, this will be addressed in a separate PR to
  keep PR size down


Sources/GRPC/ClientConnection.swift
```diff
@@ -536,7 +549,8 @@ extension ClientConnection {
       connectionKeepalive: ClientConnectionKeepalive = ClientConnectionKeepalive(),
       connectionIdleTimeout: TimeAmount = .minutes(5),
       callStartBehavior: CallStartBehavior = .waitsForConnectivity,
-      httpTargetWindowSize: Int = 65535
+      httpTargetWindowSize: Int = 65535,
+      backgroundActivityLogger: Logger = Logger(label: "io.grpc", factory: { _ in SwiftLogNoOpLogHandler() })
```
Lukasa, line 553: Is it worth encapsulating this in a `static func` on `Logger` we define internally, just so we have one-and-only-one place to write this code?

Sources/GRPC/ClientOptions.swift
```diff
@@ -60,24 +61,33 @@ public struct CallOptions {
...
+  /// If a `requestIDProvider` exists then a request ID will automatically attached to the logger's
+  /// metadata using the 'grpc-request-id' key.
+  public var logger: Logger
+
   public init(
     customMetadata: HPACKHeaders = HPACKHeaders(),
     timeLimit: TimeLimit = .none,
     messageEncoding: ClientMessageEncoding = .disabled,
-    requestIDProvider: RequestIDProvider = .autogenerated,
+    requestIDProvider: RequestIDProvider = .none,
```
Lukasa, line 74: What's the reason for this change?

Sources/GRPC/PlatformSupport.swift
```diff
@@ -155,23 +153,22 @@ public enum PlatformSupport {
     networkPreference: NetworkPreference = .best,
     logger: Logger? = nil
```
Lukasa, line 154: Why are we defaulting the loggers in this file to `nil` as opposed to the no-op logger?

Sources/GRPC/ClientConnection.swift
glbrntt: It has to be `public` as it's used in a `public` signature. I didn't think making it underscored-internal was worth it in this case.

Lukasa: Ok, fair enough. :smile:

Sources/GRPC/PlatformSupport.swift
glbrntt: I don't think there's a good reason to use an optional logger here; I just didn't think to change it.

Sources/GRPC/ClientOptions.swift
glbrntt: It seemed more natural to opt-in to having request IDs generated rather than opting out. Not sure how much I agree with that thinking now. WDYT?

Lukasa: I'm inclined to want to go to "fall into a pit of success": if users don't express a preference, giving them nice UUIDs makes their lives easier, and also our own during bug triage.

PR #910, opened 2020-07-27
Update QUESTION.md

glbrntt:


MrMage:
How about a table (like we used to have) explicitly listing the versions we want and how they can be obtained? In particular for the library and the protoc plugin.

Also, should we ask for the input/output files and command invocations right away? At least if they have "an issue generating the SwiftGRPC code or running the `protoc` compiler"?

PR #911, opened 2020-07-27
Add a server debug channel initializer

glbrntt:
Motivation:

Sometimes it's helpful to be able to add additional handlers to the
channel pipeline. We let clients to do this, so we should let servers do
it too.

Modifications:

- Add a `debugChannelInitializer` to server configuration and the server
  builder

Result:

- Users can add additional channel handlers to servers


Sources/GRPC/Server.swift
```diff
@@ -216,19 +227,29 @@ extension Server {
     /// available to service providers via `context`. Defaults to a no-op logger.
     public var logger: Logger
 
+    /// A channel initializer which will be run after gRPC has initialized each accepted channel.
+    /// This may be used to add additional handlers to the pipeline and is intended for debugging.
+    /// This is analogous to `NIO.ServerBootstrap.childChannelInitializer`.
+    ///
+    /// - Warning: The initializer closure may be invoked *multiple times*.
```
Lukasa, line 234: Strictly this one is invoked no more than once per connection.

Sources/GRPC/ServerBuilder.swift
```diff
@@ -142,6 +142,21 @@ extension Server.Builder {
   }
 }
 
+extension Server.Builder {
+  /// A channel initializer which will be run after gRPC has initialized each accepted channel.
+  /// This may be used to add additional handlers to the pipeline and is intended for debugging.
+  /// This is analogous to `NIO.ServerBootstrap.childChannelInitializer`.
+  ///
+  /// - Warning: The initializer closure may be invoked *multiple times*.
```
Lukasa, line 150: Same note here.

Sources/GRPC/Server.swift
glbrntt: Good point; I added clarification.

PR #914, opened 2020-07-31
Fix an ObjC memory consumption increase as the app runs.

MrMage:
This reflects https://github.com/grpc/grpc/pull/21798 in gRPC-ObjC.


MrMage:
@glbrntt, sorry, I should have checked whether this actually compiles 😅 Please take another look.

glbrntt:
> @glbrntt, sorry, I should have checked whether this actually compiles 😅 Please take another look.

That's why we have CI :D

PR #917, opened 2020-08-04
Mitigate zero-length writes issue.

Lukasa:
Motivation:

On some iOS releases, zero-length writes cause problems for
Network.framework. These problems can lead to difficult to debug
catastrophic failure modes, so they ought to be avoided.

NIOTransportServices provides a workaround channel handler, which we
ought to use.

Modifications:

- Add a PlatformSupport check for whether to apply the workaround.
- Add the workaround code to the pipeline configuration callbacks.
- Tests

Results:

Users can work around the zero length writes problem.


Tests/GRPCTests/ZeroLengthWriteTests.swift
```diff
@@ -0,0 +1,167 @@
...
+  func zeroLengthWriteExpectation() -> XCTestExpectation {
+    let expectation = self.expectation(description: "Expecting zero length write workaround")
+    expectation.expectedFulfillmentCount = 1
+    expectation.assertForOverFulfill = true
+    return expectation
+  }
+
+  func noZeroLengthWriteExpectation() -> XCTestExpectation {
+    let expectation = self.expectation(description: "Not expecting zero length write workaround")
+    expectation.isInverted = true
```
glbrntt, line 85: Rather than inverting the expectation, can we fulfil our expectation when we see the error from not finding the handler instead?

Sources/GRPC/PlatformSupport.swift
```diff
@@ -227,4 +227,21 @@ public enum PlatformSupport {
     logger.debug("creating a ServerBootstrap")
     return ServerBootstrap(group: group)
   }
+
+  static func requiresZeroLengthWriteWorkaround(group: EventLoopGroup, hasTLS: Bool) -> Bool {
```
glbrntt, line 231: Can you add a reference back to the NIOTS issue here?

Tests/GRPCTests/ZeroLengthWriteTests.swift
Lukasa: We can do, but it breaks the test symmetry: I have to write two test flows instead. Is that your preference?

glbrntt: Why two test flows? Couldn't we just pass two `(Result<NIOFilterEmptyWritesHandler, Error>) -> Void` closures through to `_runTest` and `debugPipelineExpectation` and run them run in an `always` block rather than passing through the expectation?

Lukasa: Oh, yeah, we can do that. It's a bit less clean at the test site, but works fine.

PR #929, opened 2020-08-10
Add .swiftformat and apply a lint check in CI

glbrntt:
Motivation:

The code style used by gRPC Swift is quite different from what most
contributors are used to, indenting with two spaces instead of Xcode's default
of four, for example. This is often irritating for contributors since
they need to adjust their Xcode settings and reformat their code to
match our style, and also for reviewers since they need to highlight
these differences.

Modifications:

- Add a .swiftformat configuration file and ensure it's validated as
  part of the CI preflight checks
- Apply swiftformat to baseline the formatting

Result:

- CI will check fail before running test if there are formatting
  violations
- Users can run fix their formatting by running 'swiftformat .' in the
  root of the repository


glbrntt:
The first commit in this PR is the configuration and enabling a link check in CI. The second commit is the result of baselining the formatting.

PR #938, opened 2020-08-14
Log some HTTP/2 information

glbrntt:
Motivation:

Sometimes it's useful to know what's going on at the HTTP/2 layer, none
of our logs currently surface this information.

Modifications:

- Add logs for stream created/closed events (includes open stream count)
- Add logs for initial settings
- Add more logging for connection lifecycle
- Add trace logs to the stream for read/written frames
- Prefix request and connection ID metadata keys with "grpc_"
- Remove duplicated log messages

Result:

- Better insight for HTTP/2

### DIFF
--- a/Sources/GRPC/ClientConnection.swift
+++ b/Sources/GRPC/ClientConnection.swift
@@ -166,7 +166,6 @@ extension ClientConnection: GRPCChannel {
     callOptions: CallOptions
   ) -> UnaryCall<Serializer.Input, Deserializer.Output> {
     let (logger, requestID) = self.populatedLoggerAndRequestID(from: callOptions)
-    logger.debug("starting rpc", metadata: ["path": "\(path)"])
 
     let call = UnaryCall<Serializer.Input, Deserializer.Output>.makeOnHTTP2Stream(
       multiplexer: self.multiplexer,
@@ -229,7 +228,6 @@ extension ClientConnection {
     callOptions: CallOptions
   ) -> ClientStreamingCall<Serializer.Input, Deserializer.Output> {
     let (logger, requestID) = self.populatedLoggerAndRequestID(from: callOptions)
-    logger.debug("starting rpc", metadata: ["path": "\(path)"])
 
     let call = ClientStreamingCall<Serializer.Input, Deserializer.Output>.makeOnHTTP2Stream(
       multiplexer: self.multiplexer,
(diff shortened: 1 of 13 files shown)


Sources/GRPC/Server.swift
```diff
@@ -103,12 +103,16 @@ public final class Server {
       )
       // Set the handlers that are applied to the accepted Channels
       .childChannelInitializer { channel in
+        var logger = configuration.logger
+        logger[metadataKey: MetadataKey.connectionID] = "\(UUID().uuidString)"
+        logger[metadataKey: "remote_address"] = channel.remoteAddress.map { "\($0)" } ?? "n/a"
```
Lukasa, line 108: Should we add a metadata key for this?

Sources/GRPC/_GRPCClientChannelHandler.swift
```diff
@@ -396,6 +402,12 @@ extension _GRPCClientChannelHandler: ChannelInboundHandler {
       preconditionFailure("Received DATA frame with non-ByteBuffer IOData")
     }
 
+    self.logger.trace("received HTTP2 frame", metadata: [
+      "h2_payload": "DATA",
+      "h2_data_bytes": "\(content.data.readableBytes)",
+      "h2_end_stream": "\(content.endStream)",
```
Lukasa, line 408: Do we need metadata keys for these too?

Sources/GRPC/GRPCIdleHandler.swift
```diff
@@ -52,22 +56,33 @@ internal class GRPCIdleHandler: ChannelInboundHandler {
...
         // We have a stream: don't go idle
         self.scheduledIdle?.cancel()
         self.scheduledIdle = nil
-
         self.activeStreams += 1
-      } else if event is StreamClosedEvent {
+
+        self.logger.debug("HTTP2 stream created", metadata: [
+          MetadataKey.streamID: "\(created.streamID)",
+          "h2_active_streams": "\(self.activeStreams)",
```
Lukasa, line 76: And do we need a metadata key for this?

Sources/GRPC/ClientConnection.swift
```diff
@@ -362,7 +359,6 @@ extension ClientConnection {
     handler: @escaping (Deserializer.Output) -> Void
   ) -> BidirectionalStreamingCall<Serializer.Input, Deserializer.Output> {
     let (logger, requestID) = self.populatedLoggerAndRequestID(from: callOptions)
-    logger.debug("starting rpc", metadata: ["path": "\(path)"])
```
Lukasa, line 365: Is the theory we don't need these anymore?

glbrntt: These are duplicates.

PR #941, opened 2020-08-16
[DO NOT MERGE] Add tracing instrumentation

slashmo:
This is a WIP for tracing gRPC services. It's meant as a playground for experimenting with how such instrumentation will look like.


Sources/GRPC/SpanStatus+GRPCStatus.swift
```diff
@@ -0,0 +1,66 @@
...
+      canonicalCode = .unimplemented
+    case .internalError:
+      canonicalCode = .internal
+    case .unavailable:
+      canonicalCode = .unavailable
+    case .dataLoss:
+      canonicalCode = .dataLoss
+    case .unauthenticated:
+      canonicalCode = .unauthenticated
+    case .doNotUse:
```
glbrntt, line 59: FYI: I just filed #942 -- `doNotUse` made it's way over from the old gRPC Swift implementation which in turn came from [here](https://github.com/grpc/grpc/blob/c79bef55ee348c298cbdcc60448d8e70e4972a19/include/grpcpp/impl/codegen/status_code_enum.h#L138). Its purpose was to force a `default` branch, which we should do here.

slashmo: Cool 👍 Updated it to use `default` instead.

glbrntt: This is a great start by the way; please @ me if there's anything you want me to look at specifically :)

slashmo: Thanks, I appreciate that! 😎

slashmo:
I'm currently finishing up the server span attributes. Per Otel spec, we should either add HTTP attributes to the same Span or create a child Span of the gRPC span for the HTTP transport:

> HTTP calls can generally be represented using just HTTP spans. If they address a particular remote service and method known to the caller, i.e., when it is a remote procedure call transported over HTTP, the rpc.* attributes might be added additionally on that span, or in a separate RPC span that is a parent of the transporting HTTP call. Note that method in this context is about the called remote procedure and not the HTTP verb (GET, POST, etc.).

@glbrntt / @ktoso
Which option would you prefer here?

ktoso:
Try what is simpler for now (so just one span I guess) and let's try doing the client as well so we can show two grpc services talking to one another, and then we continue improving the span quality perhaps?

slashmo:
> Try what is simpler for now (so just one span I guess) and let's try doing the client as well so we can show two grpc services talking to one another, and then we continue improving the span quality perhaps?

Sounds good 👌

Sources/GRPC/_GRPCClientChannelHandler.swift
```diff
@@ -552,3 +586,13 @@ extension _GRPCClientChannelHandler: ChannelOutboundHandler {
     }
   }
 }
+
+// TODO: Extract type
+public struct HPACKHeadersInjector: InjectorProtocol {
+  public init() {}
+
+  public func inject(_ value: String, forKey key: String, into headers: inout HPACKHeaders) {
+    // TODO: Replace or add?
```
ktoso, line 595: hmm good question though I guess replace here.
technically the same header could be added a few times (is that also true in h2, i dont remember?, but i think for our purposes it's a set()

Sources/GRPC/_GRPCClientChannelHandler.swift
```diff
@@ -460,16 +466,44 @@ extension _GRPCClientChannelHandler: ChannelOutboundHandler {
                     promise: EventLoopPromise<Void>?) {
     switch self.unwrapOutboundIn(data) {
     case let .head(requestHead):
+      // TODO: Don't force unwrap request info
+      let requestInfo = GRPCRequestInfo(parsing: requestHead.path)!
+      var span = InstrumentationSystem.tracingInstrument.startSpan(
+        named: String(requestInfo.uri.dropFirst()),
+        context: context.baggage,
+        ofKind: .client
+      )
```
ktoso, line 475: 👍

glbrntt:
We won't adopt tracing out-of-the-box in v1, instead this will come in v2 via interceptors.

PR #943, opened 2020-08-19
Update README for the latest Pod version, add support for auto-updating.

MrMage:


README.md
```diff
@@ -96,11 +96,11 @@ gRPC Swift is currently available [from CocoaPods][pod-new]. To integrate, add
 the following line to your `Podfile`:
 
 ```ruby
-    pod 'gRPC-Swift', '1.0.0-alpha.12'
+    pod 'gRPC-Swift', '~> 1.0.0-alpha.18'
```
glbrntt, line 99: hmm. Do we actually want to suggest this now given there may be breaking changes between pre-releases?

MrMage: I would say so. As long as we are not at 1.0, we'd want people to adopt the latest and greatest quickly, and to quickly adopt breaking changes.

It seems like we haven't had that many breaking changes recently anymore, anyway.

glbrntt: Fair enough. Yeah, that's sorta true, at least not source breaking.

glbrntt (approved):

Issue #915, opened 2020-07-31
Conflict using Firestore and gRPC-Swift pods


hugobgranja:
### Describe the bug

The following error results from using both Firestore and gRPC-Swift pods:
[!] The 'Pods-SwiftGRPCTest' target has frameworks with conflicting names: grpc.

Podfile:
```
target 'SwiftGRPCTest' do
  use_frameworks!

  pod 'Firebase/Firestore'
  pod 'gRPC-Swift', '1.0.0-alpha.17'
end
```

glbrntt:
This was resolved by #917 - the next release should work fine with Firestore.

glbrntt:
The fix here was actually in error and has been reverted: unfortunately I don't see a way to work around this properly without renaming gRPC Swift entirely.

glbrntt:
Sorry for the churn on this one, it's a bit of an awkward issue.

The current workaround – once #945 lands – is copy our [Podfile](https://github.com/grpc/grpc-swift/blob/main/gRPC-Swift.podspec) and replace the module name with something that doesn't have conflicts, "GRPCSwift", for example.

When you generate your code from your proto file you'll have to set the "GRPCModuleName" option to whatever name you chose above.

PR #945, opened 2020-08-19
Allow the gRPC module name to be customised when generating code

glbrntt:
Motivation:

Some Cocoapods may find themselves in a situation where their dependency
graph contains both gRPC Swift and the gRPC core library. This is
problematic since both have the same module name ("GRPC"). Changing our
own module name is a very large breaking change and one we'd like to avoid.
As an escape valve, users who take a direct dependency on us may rename
our module via their Podfile; doing so would require their generated
code to depend on that module, rather than the currently hardcoded
'GRPC' module.

Modifications:

- Add an option to the codegen to specify the gRPC module name
- Remove a couple of imports which were previously used but are now no
  longer required
- Regenerate

Result:

It's possible for Cocoapods users to take a direct dependency on gRPC
Swift and libgrpc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,9 @@ the following line to your `Podfile`:
 Then, run `pod install` from command line and use your project's generated
 `.xcworkspace` file. You might also need to add `use_frameworks!` to your `Podfile`.
 
+*⚠️ If you have conficting modules as a result of having a transitive
+dependency on '[gRPC-Core][grpc-core-pod]' see [grpc/grpc-swift#945][grpc-swift-945].*
+
 ### Getting the `protoc` Plugins
 
 Binary releases of `protoc`, the Protocol Buffer Compiler, are available on
@@ -171,6 +174,8 @@ Please get involved! See our [guidelines for contributing](CONTRIBUTING.md).
 [docs-keepalive]: ./docs/keepalive.md
 [docs-tutorial]: ./docs/basic-tutorial.md
 [grpc]: https://github.com/grpc/grpc
+[grpc-core-pod]: https://cocoapods.org/pods/gRPC-Core
+[grpc-swift-945]: https://github.com/grpc/grpc-swift/pull/945
 [protobuf-releases]: https://github.com/protocolbuffers/protobuf/releases
 [swift-nio-platforms]: https://github.com/apple/swift-nio#supported-platforms
 [swift-nio]: https://github.com/apple/swift-nio
--- a/Sources/Examples/Echo/Model/echo.grpc.swift
+++ b/Sources/Examples/Echo/Model/echo.grpc.swift
@@ -20,11 +20,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-import Foundation
 import GRPC
 import NIO
-import NIOHTTP1
-import SwiftProtobuf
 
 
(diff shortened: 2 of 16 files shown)


MrMage (approved):

MrMage (changes requested):
Could you please also document this in https://github.com/grpc/grpc-swift/blob/main/docs/plugin.md?

MrMage:
This "fixes #915." If you encounter conflicting module names, please follow this process:

1. Download our current Podspec from https://github.com/grpc/grpc-swift/raw/main/gRPC-Swift.podspec and place it next to your `Podfile`.
2. In the downloaded `gRPC-Swift.podspec`, replace `s.module_name = 'GRPC'` with `s.module_name = 'GRPC_Swift'`.
3. Include the dependency in your Podfile as follows: `pod 'gRPC-Swift', :podspec => 'gRPC-Swift.podspec'`
4. Change your protobuf codegen script to include `--grpc-swift_out=GRPCModuleName_Swift` in your options (or append it to the end of an existing config string).

Whenever we tag a new release, I'm afraid you will need to repeat steps 1 and 2 (or manually increment `s.version` in `gRPC-Swift.podspec`).

MrMage (approved):
Looks good. I wonder whether we should also include the steps from my comment, WDYT?

glbrntt:
> Looks good. I wonder whether we should also include the steps from my comment, WDYT?

If anything I'd think they should be included in the README, I suspect If someone hits this they _might_ search the README but will be unlikely to look in the plugin docs.

MrMage:
> If anything I'd think they should be included in the README, I suspect If someone hits this they _might_ search the README but will be unlikely to look in the plugin docs.

Fine with me as well. Maybe just mention in the README to look here when encountering this particular issue?

vlopezferrando:
> This "fixes #915." If you encounter conflicting module names, please follow this process:
> 
>     1. Download our current Podspec from https://github.com/grpc/grpc-swift/raw/main/gRPC-Swift.podspec and place it next to your `Podfile`.
> 
>     2. In the downloaded `gRPC-Swift.podspec`, replace `s.module_name = 'GRPC'` with `s.module_name = 'GRPC_Swift'`.
> 
>     3. Include the dependency in your Podfile as follows: `pod 'gRPC-Swift', :podspec => 'gRPC-Swift.podspec'`
> 
>     4. Change your protobuf codegen script to include `--grpc-swift_out=GRPCModuleName=GRPC_Swift` in your options (or append it to the end of an existing config string).
> 
> 
> Whenever we tag a new release, I'm afraid you will need to repeat steps 1 and 2 (or manually increment `s.version` in `gRPC-Swift.podspec`).

Please note that the option for protobuf codegen is `--grpc-swift_opt=...` and not `--grpc-swift_out`.

nshoura:
Hello everyone,

How can the [mentioned solution](https://github.com/grpc/grpc-swift/pull/945#issuecomment-676051018) work when `gRPC-Swift` is used as a dependency for another pod since, in `podspec` files, it's only allowed to define the name and the version specifier of a Pod? 
Basically, the following isn't valid:
`s.dependency 'gRPC-Swift', :podspec => 'gRPC-Swift.podspec'`

Artfire:
@nshoura 
Hi, 
Did you solve this issue?

PR #967, opened 2020-09-17
Idle client connection on 'channelInactive' if there are no active RPCs

glbrntt:
Motivation:

Connections are idled if there are no RPCs for a given timeout. A
connection may also be idled if the remote sends it a GO_AWAY frame.
However, If the remote fails to send a GO_AWAY then from the perspective
of the client, the connection has been dropped and it will attempt
re-establish a connection. As a result a connection may bounce from
'ready' to 'transientFailure' to 'connecting' and back to 'ready' even
though there are no active RPCs.

Modifications:

- Idle the client connection on 'channelInactive' when there are no
  active RPCs and we know that the connection was previously up-and-running.
- Update tests.

Result:

We avoid creating unnecessary connections when the connection is dropped
and there are no active RPCs.

### DIFF
--- a/Sources/GRPC/ConnectionManager.swift
+++ b/Sources/GRPC/ConnectionManager.swift
@@ -423,7 +423,19 @@ internal class ConnectionManager {
     case .shutdown:
       channel.close(mode: .all, promise: nil)
 
-    case .idle, .active, .ready, .transientFailure:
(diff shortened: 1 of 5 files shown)


Lukasa (approved):
LGTM, nice.

PR #974, opened 2020-09-25
Log when active streams reaches and drops below max concurrent streams

glbrntt:
Motivation:

Reaching max concurrent streams will cause frames on additional streams
to be buffered, and can therefore see increased latency. Observing this
is difficult.

Modifications:

- Log any changes in HTTP/2 settings, not just the initial settings
- Add a warning log when our active stream count matches max concurrent
  streams
- Add a notice log when our stream count drops below max concurrent
  streams

Result:

Better observability when we get close to max concurrent streams.

### DIFF
--- a/Sources/GRPC/GRPCIdleHandler.swift
+++ b/Sources/GRPC/GRPCIdleHandler.swift
@@ -26,6 +26,9 @@ internal class GRPCIdleHandler: ChannelInboundHandler {
   /// The number of active streams.
   private var activeStreams = 0
 
+  /// The maximum number of streams we may create concurrently on this connection.
+  private var maxConcurrentStreams: Int?
+
   /// The scheduled task which will close the channel.
   private var scheduledIdle: Scheduled<Void>?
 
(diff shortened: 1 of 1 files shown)


Sources/GRPC/GRPCIdleHandler.swift
```diff
@@ -154,43 +171,58 @@ internal class GRPCIdleHandler: ChannelInboundHandler {
...
+      case let .settings(.settings(settings)):
+        // Log any changes to HTTP/2 settings.
+        self.logger.debug(
+          "HTTP2 settings update",
+          metadata: Dictionary(settings.map {
+            ("\($0.parameter.loggingMetadataKey)", "\($0.value)")
+          }, uniquingKeysWith: { a, _ in a })
+        )
+
+        let maxConcurrentStreams = settings.first(where: { $0.parameter == .maxConcurrentStreams })
```
Lukasa, line 184: A quick FYI: if there are multiple instances of a setting in a HTTP/2 settings frame, the _last_ instance of the setting is the one that counts. `HTTP2Settings` is a `RandomAccessCollection` so you can do `reversed()` for free.

Lukasa (commented):
One note about calculating max concurrent streams.

Sources/GRPC/GRPCIdleHandler.swift
glbrntt: Oh, good catch. Thanks!

Lukasa (approved):

PR #980, opened 2020-09-30
Add an option to set 'certificateVerification' to the client builder

glbrntt:
Motivation:

We allow users to set the `certificateVerification` if they configure
their client directly using `ClientConnection.Configuration` but not via
the builder API.

Modifications:

- Add `withTLS(certificateVerification:)` to the client connection builder.
  (The same option is already available on the server builder)

Result:

Users can set the certificate verification mode on the client builder.


Lukasa (approved):

Lukasa:
Worth adding a test for this?

glbrntt:
> Worth adding a test for this?

Done. I was just being lazy 🙈

Lukasa (approved):

PR #984, opened 2020-10-05
Provide more specific errors messages for failing channels

glbrntt:
Motivation:

If a connection attempt fails, the reason for the failure should be
propagated to the RPCs.

Modifications:

- Store errors in the active and ready states so that they can be used
  as contextual information when requesting a channel from the transient
  failure or shutdown states

Result:

- Connection level errors get propagated to RPCs made on that same
  Channel


Lukasa (approved):

glbrntt:
Hmmm the Travis CI run [passed](https://travis-ci.org/github/grpc/grpc-swift/builds/732905530), not sure why it isn't listed here though.

PR #986, opened 2020-10-09
Add scaffolding for client interceptors

glbrntt:
Motivation:

To support client interceptors we first need to define a few types to
work with.

Modifications:

- Add client request and response parts.
- Add the client interceptor protocol.
- Add the client interceptor context, which is passed to functions on
  the interceptor protocol in order to allow implementations to invoke
  functions on interceptors either side of it in the pipeline.
- Add a stubbed out client interceptor pipeline.

Result:

We have some base types for client side interceptors.


Sources/GRPC/Array+BoundsCheck.swift
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+extension Array {
+  internal subscript(checkingIndices index: Index) -> Element? {
```
Lukasa, line 18: Minor nit: `checked` instead of `checkingIndices`? Might read a little better.

Sources/GRPC/Interceptor/ClientInterceptorContext.swift
```diff
@@ -0,0 +1,145 @@
...
+  /// interceptor pipeline.
+  internal init(pipeline: ClientInterceptorPipeline<Request, Response>, index: Int) {
+    self.pipeline = pipeline
+    self.index = index
+  }
+
+  /// Forwards the response part to the next inbound interceptor in the pipeline, if there is one.
+  ///
+  /// - Parameter part: The response part to forward.
+  /// - Note: This does *not* have to be called from the `EventLoop`.
```
Lukasa, line 56: Is that worthwhile? NIO doesn't bother with this.

Sources/GRPC/Interceptor/ClientInterceptorContext.swift
```diff
@@ -0,0 +1,145 @@
...
+      }
+    }
+  }
+
+  /// Forwards the request part to the next outbound interceptor in the pipeline, if there is one.
+  ///
+  /// - Parameters:
+  ///   - part: The request part to forward.
+  ///   - promise: The promise the complete when the part has been written.
+  /// - Note: This does *not* have to be called from the `EventLoop`.
```
Lukasa, line 72: Same note here.

Sources/GRPC/Interceptor/ClientInterceptorContext.swift
```diff
@@ -0,0 +1,145 @@
...
+        self._write(part, promise: promise)
+      }
+    }
+  }
+
+  /// Forwards a request to cancel the RPC to the next outbound interceptor in the pipeline.
+  ///
+  /// - Parameter promise: The promise to complete with the outcome of the cancellation request.
+  /// - Note: This does *not* have to be called from the `EventLoop`.
+  public func cancel(promise: EventLoopPromise<Void>?) {
```
Lukasa, line 90: Same note here.

Sources/GRPC/Interceptor/ClientInterceptorContext.swift
```diff
@@ -0,0 +1,145 @@
...
+  private func _write(
+    _ part: ClientRequestPart<Request>,
+    promise: EventLoopPromise<Void>?
+  ) {
+    self.eventLoop.assertInEventLoop()
+
+    if let outbound = self.nextOutbound {
+      outbound.invokeWrite(part, promise: promise)
+    } else {
+      promise?.fail(GRPCStatus(code: .unavailable, message: "The RPC has already completed"))
```
Lukasa, line 116: Is this right? There seems to be some inconsistency between `_read` and `_write` that I don't fully understand about when the next interceptor might be `nil`.

Sources/GRPC/Interceptor/ClientInterceptorPipeline.swift
```diff
@@ -0,0 +1,113 @@
...
+    return self.contexts.last
+  }
+
+  internal init(
+    eventLoop: EventLoop,
+    logger: Logger
+  ) {
+    self.eventLoop = eventLoop
+    self.logger = logger
+    self.contexts = []
```
Lukasa, line 54: This doesn't seem useful given that `self.contexts` is `let`. Presumably we'll want to take them as an argument?

Sources/GRPC/Interceptor/ClientInterceptorPipeline.swift
```diff
@@ -0,0 +1,113 @@
...
+    if self.eventLoop.inEventLoop {
+      self._read(part)
+    } else {
+      self.eventLoop.execute {
+        self._read(part)
+      }
+    }
+  }
+
+  internal func write(_ part: ClientRequestPart<Request>, promise: EventLoopPromise<Void>?) {
```
Lukasa, line 67: These functions need some doc comments to explain who calls them and when.

Sources/GRPC/Interceptor/ClientParts.swift
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+import NIOHPACK
+
+public enum ClientRequestPart<Request> {
+  /// User provided metadata sent at the start of the request stream.
+  case metadata(HPACKHeaders)
+
+  /// A message to send to the server.
+  case message(Request, compress: Bool, flush: Bool)
```
Lukasa, line 23: `compress` and `flush` seem like two of a possibly-extendable set of things we might want to attach to a message. I can in principle believe that `flush` should be promoted to top-level, but should this have a `Metadata` struct instead of `compress`?

Sources/GRPC/Interceptor/ClientInterceptorContext.swift
glbrntt: Hmm I figured it'd be better to avoid failure since this could be called from off of the event loop, e.g. after deferring to e.g. UI for authentication or something. Perhaps it'd be better to document that it must be called from the event loop and keep the relevant assertions. WDYT?

Sources/GRPC/Interceptor/ClientInterceptorContext.swift
glbrntt: So `_read` is more like NIOs `fireChannelRead` rather than NIOs `read`.

What's not implemented here is removing the contexts when the RPC completes, that's when we expect them to be `nil`.

Sources/GRPC/Interceptor/ClientInterceptorPipeline.swift
glbrntt: Eventually, yes. This the pipeline is mostly just placeholder at the moment.

Sources/GRPC/Interceptor/ClientParts.swift
glbrntt: Yeah, that's a good call I think.

Sources/GRPC/Interceptor/ClientInterceptorContext.swift
Lukasa: Yeah, I think that that's probably better? It's a pretty significant perf win to avoid the check.

glbrntt: I gave the pipeline the same treatment: it'll be internal and the inbound side will already be on the event loop. The outbound side will be invoked from the call object, so we can push the check there instead.

Lukasa (approved):

PR #989, opened 2020-10-12
Make 'GRPCStatus' a struct

glbrntt:
Motivation:

`GRPCStatus` is a `class` because it was too wide to fit within the
inline storage of an existential container and would incur unnecessary
heap allocations as a result.

In #942 the implementation of the `GRPCStatus.Code` stored by the status
was changed from an `enum` to a `struct`. As a result, the width of
`GRPCStatus` (as a `struct`) is now 24 and may therefore be stored
inline in an existential container. Unfortunately, this now makes
`_GRPCClientResponsePart` too wide!

We could make the relevant case `indirect` on `_GRPClientResponsePart`,
however since we transform between two specializations of that type per
RPC (before and after deserialization) we would incur a heap allocation
each time.

Since the `message` property is optional, we can box that instead, only
incurring the heap allocation when we construct a `GRPCStatus` with a
non-nil status.

Modifications:

- Add `Refernce` and `Box`
- Store the optional message in a `Box`
- Make `GRPCStatus` a `struct` instead of a `class` (note: we already
  have a test in place to validate it may be stored inline in an
  existential container).

Result:

- `GRPCStatus` only allocates when the `message` isn't `nil`

### DIFF
--- a/Sources/GRPC/GRPCStatus.swift
+++ b/Sources/GRPC/GRPCStatus.swift
@@ -19,17 +19,12 @@ import NIOHTTP1
 import NIOHTTP2
 
 /// Encapsulates the result of a gRPC call.
-///
-/// We use a `class` here for a couple of reasons:
-/// - The size of the equivalent `struct` is larger than the value buffer in an existential
-///   container so would incur a heap allocation each time a `GRPCStatus` is passed to a function
-///   taking an `Error`.
-/// - We aren't using value semantics (since all properties are constant).
-public final class GRPCStatus: Error {
-  /// The status code of the RPC.
-  public let code: Code
+public struct GRPCStatus: Error {
   /// The status message of the RPC.
-  public let message: String?
+  public var message: String?
+
+  /// The status code of the RPC.
+  public var code: Code
 
   /// Whether the status is '.ok'.
   public var isOk: Bool {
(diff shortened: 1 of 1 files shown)


Lukasa (approved):

glbrntt:
Oh, sad, I only ran the test for `GRPCStatus` but the `_GRPCClientResponsePart` type has `GRPCStatus` as an associated value in one of its cases, this has now pushed that over the limit.

Sources/GRPC/GRPCStatus.swift
```diff
@@ -75,19 +70,25 @@ extension GRPCStatus {
...
+
+    public var rawValue: Int {
+      return Int(self._rawValue)
+    }
 
     public init?(rawValue: Int) {
       switch rawValue {
       case 0 ... 16:
-        self.rawValue = rawValue
+        self._rawValue = UInt8(rawValue)
```
Lukasa, line 84: I mean, this is a really silly performance optimisation that we totally don't need, but it's nice to have: given that we have the check above, we know the value is in-bounds for UInt8. So we can use `UInt8(truncatingIfNeeded: rawValue)` and save ourselves some branches.

Lukasa: https://swift.godbolt.org/z/jG1YYv

glbrntt: Silly, but no point leaving it on the table. Plus, I think it's generally useful to be thinking about these optimisations.

Sources/GRPC/GRPCStatus.swift
```diff
@@ -19,18 +19,13 @@ import NIOHTTP1
...
-/// - The size of the equivalent `struct` is larger than the value buffer in an existential
-///   container so would incur a heap allocation each time a `GRPCStatus` is passed to a function
-///   taking an `Error`.
-/// - We aren't using value semantics (since all properties are constant).
-public final class GRPCStatus: Error {
-  /// The status code of the RPC.
-  public let code: Code
+public struct GRPCStatus: Error {
   /// The status message of the RPC.
   public let message: String?
```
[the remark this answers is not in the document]
glbrntt: Sure, no harm in changing it.

Lukasa (approved):

PR #990, opened 2020-10-13
Basic Async Unary implementation of QPS benchmark

PeterAdams-A:
Motivation:

The main grpc repo contains a benchmarking suite which can be implemented in any language.
This is an implementation of the async unary test and the framework surrounding it.

Notes:
  * Benchmark so far only run on OSX, not linux
  * Benchmark is not optimised - there are currently allocations within the benchmark which
    could be removed.

Modifications:

Add framework for QPS style testing.
Add implementation of Async Unary benchmark (without TLS)
Add brief README describing how to run the benchmark

Result:

It's possible to run a benchmark of gRPC-swift which is in some way comparible to other
languages.


PeterAdams-A:
Note:  All the proto files in the Model directory are taken from grpc/grpc repo - all the swift to match is generated.

Performance/QPSBenchmark/Tests/BenchmarkUtilsTests/HistogramTests.swift
```diff
@@ -0,0 +1,103 @@
...
+            case 0:
+                break
+            case 1:
+                XCTAssertFalse(oneSeen)
+                oneSeen = true
+            case 2:
+                XCTAssertFalse(twoSeen)
+                twoSeen = true
+            default:
+                XCTAssertTrue(false)
```
glbrntt: You can use `XCTFail` here

Performance/QPSBenchmark/Makefile
```diff
@@ -0,0 +1,74 @@
+# Which Swift to use.
+SWIFT:=swift
+# Where products will be built; this is the SPM default.
+GRPC_SWIFT_PATH:=../..
+SWIFT_BUILD_PATH:=./.build
+SWIFT_BUILD_CONFIGURATION=debug
```
glbrntt, line 6: let's make this release

Performance/QPSBenchmark/Makefile
```diff
@@ -0,0 +1,74 @@
...
+SWIFT_FLAGS_RELEASE=$(patsubst --configuration=%,--configuration=release,$(SWIFT_FLAGS))
+
+# protoc plugins.
+PROTOC_GEN_SWIFT=${GRPC_SWIFT_PATH}/${SWIFT_BUILD_PATH}/release/protoc-gen-swift
+PROTOC_GEN_GRPC_SWIFT=${GRPC_SWIFT_PATH}/${SWIFT_BUILD_PATH}/release/protoc-gen-grpc-swift
+
+SWIFT_BUILD:=${SWIFT} build ${SWIFT_FLAGS}
+SWIFT_BUILD_RELEASE:=${SWIFT} build ${SWIFT_FLAGS_RELEASE}
+SWIFT_TEST:=${SWIFT} test ${SWIFT_FLAGS}
+SWIFT_PACKAGE:=${SWIFT} package ${SWIFT_FLAGS}
```
glbrntt, line 18: Let's get rid of these too

Performance/QPSBenchmark/Makefile
```diff
@@ -0,0 +1,74 @@
+# Which Swift to use.
+SWIFT:=swift
+# Where products will be built; this is the SPM default.
+GRPC_SWIFT_PATH:=../..
+SWIFT_BUILD_PATH:=./.build
+SWIFT_BUILD_CONFIGURATION=debug
+SWIFT_FLAGS=--build-path=${SWIFT_BUILD_PATH} --configuration=${SWIFT_BUILD_CONFIGURATION}
+# Force release configuration (for plugins)
+SWIFT_FLAGS_RELEASE=$(patsubst --configuration=%,--configuration=release,$(SWIFT_FLAGS))
```
glbrntt, line 9: Let's get rid of this

Performance/QPSBenchmark/Makefile
```diff
@@ -0,0 +1,74 @@
...
+PROTOC_GEN_SWIFT=${GRPC_SWIFT_PATH}/${SWIFT_BUILD_PATH}/release/protoc-gen-swift
+PROTOC_GEN_GRPC_SWIFT=${GRPC_SWIFT_PATH}/${SWIFT_BUILD_PATH}/release/protoc-gen-grpc-swift
+
+SWIFT_BUILD:=${SWIFT} build ${SWIFT_FLAGS}
+SWIFT_BUILD_RELEASE:=${SWIFT} build ${SWIFT_FLAGS_RELEASE}
+SWIFT_TEST:=${SWIFT} test ${SWIFT_FLAGS}
+SWIFT_PACKAGE:=${SWIFT} package ${SWIFT_FLAGS}
+
+# Name of generated xcodeproj
+XCODEPROJ:=GRPC.xcodeproj
```
glbrntt, line 21: And this one

Performance/QPSBenchmark/Makefile
```diff
@@ -0,0 +1,74 @@
...
+
+### Package and plugin build targets ###########################################
+
+all:
+	${SWIFT_BUILD}
+
+.PHONY:
+plugins: ${PROTOC_GEN_SWIFT} ${PROTOC_GEN_GRPC_SWIFT}
+	cp $^ .
```
glbrntt, line 30: we can get rid of these too, I think

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/ClientUtilities.swift
```diff
@@ -0,0 +1,50 @@
...
+        let serverTargets = self.serverTargets
+        return try serverTargets.map { target in
+            if let splitIndex = target.lastIndex(of: ":") {
+                let host = target[..<splitIndex]
+                let portString = target[(target.index(after: splitIndex))...]
+                if let port = Int(portString) {
+                    return HostAndPort(host: String(host), port: port)
+                }
+            }
+            throw GRPCStatus(code: .invalidArgument, message: "Server targets could not be parsed")
```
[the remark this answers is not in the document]
PeterAdams-A: We could do.  But then I'd probably want to convert it back to an invalidArgument for sending back to the client as it was an argument which came from the client.  For discussion.

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/AsyncClient.swift
```diff
@@ -0,0 +1,275 @@
...
+
+        // Start the requested number of channels.
+        precondition(serverTargets.count > 0)
+        var channelRepeaters : [ChannelRepeater] = []
+        for channelNumber in 0..<Int(config.clientChannels) {
+            channelRepeaters.append(ChannelRepeater(target: serverTargets[channelNumber % serverTargets.count],
+                                                    config: config,
+                                                    eventLoopGroup: eventLoopGroup))
+        }
+        self.channelRepeaters = channelRepeaters
```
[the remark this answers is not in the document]
PeterAdams-A: Yes, sadly means I need  a local for eventLoopGroup to avoid self capture but still feels better.

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/BenchmarkServiceImpl.swift
```diff
@@ -0,0 +1,91 @@
...
+        context.logger.warning("streamingBothWays not implemented yet")
+        return context.eventLoop.makeFailedFuture(GRPCStatus(code: GRPCStatus.Code.unimplemented,
+                                                             message: "Not implemented"))
+    }
+
+    /// Make a payload for sending back to the client.
+    private static func makePayload(type: Grpc_Testing_PayloadType, size: Int) throws -> Grpc_Testing_Payload {
+        if type != .compressable {
+            throw GRPCStatus(code: .internalError, message: "Failed to make payload")
+        }
```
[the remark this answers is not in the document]
PeterAdams-A: I'll comment - I think you have to make data which doesn't compress which is hard (and not implemented in the other implementations I looked at)

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/QPSWorker.swift
```diff
@@ -0,0 +1,70 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import NIO
+import GRPC
+import Logging
+
+/// Sets up and runs a worker service which listens for instructions on what tests to run.
+/// Currently doesn't understand TLS.
```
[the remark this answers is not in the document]
PeterAdams-A: Probably not worth adding - when I wrote this I was actually looking at the code which does TLS to the driver which we may never want to do.  I'll change this command and add TODOs in client and server code for the other.

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/AsyncServer.swift
```diff
@@ -0,0 +1,104 @@
...
+        }.always { result in
+            return self.eventLoopGroup.shutdownGracefully { error in
+                if let error = error {
+                    promise.fail(error)
+                } else {
+                    promise.succeed(())
+                }
+            }
+        }
+        return promise.futureResult
```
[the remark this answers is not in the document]
PeterAdams-A: After much confusion I've realised that flatMap change the event loop your future is on.  Now I understand this I'll try and see if I can come up with a better answer.

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/AsyncClient.swift
```diff
@@ -0,0 +1,275 @@
...
+    /// - parameters:
+    ///     - callbackLoop: Which eventloop should be called back on completion.
+    /// - returns: A future on the `callbackLoop` which will succeed on completion of shutdown.
+    func shutdown(callbackLoop: EventLoop) -> EventLoopFuture<Void> {
+        let promise: EventLoopPromise<Void> = callbackLoop.makePromise()
+        let stoppedFutures = self.channelRepeaters.map { repeater in repeater.stop() }
+        let allStopped = EventLoopFuture<Void>.reduce((),
+                                                      stoppedFutures,
+                                                      on: callbackLoop, { (_, _) -> Void in return () } )
+        _ = allStopped.always { result in
```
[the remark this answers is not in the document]
PeterAdams-A: I doubt this is safe as I think it will convert the future to another event loop - I'll see if I can come up with something that works.

PeterAdams-A:
I've addressed most of your code review issues apart from the few things I've commented on above indicating I will not.  I have added a tiny amount of extra docs as part of breaking up functions.  Is there anything in particular you feel is definitely needing more doc?

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/AsyncServer.swift
```diff
@@ -0,0 +1,106 @@
...
+    let currentResourceUsage = getResourceUsage()
+    var result = Grpc_Testing_ServerStatus()
+    result.stats.timeElapsed = (currentTime - self.statsPeriodStart).asSeconds()
+    result.stats.timeSystem = currentResourceUsage.systemTime - self.cpuStatsPeriodStart
+      .systemTime
+    result.stats.timeUser = currentResourceUsage.userTime - self.cpuStatsPeriodStart.userTime
+    result.stats.totalCpuTime = 0
+    result.stats.idleCpuTime = 0
+    result.stats.cqPollCount = 0
+    self.logger.info("Sending response")
```
glbrntt, line 79: more useful to say "Sending server status" here, "response" is a little ambiguous

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/AsyncServer.swift
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  /// Shutdown the service.
+  /// - parameters:
+  ///     - callbackLoop: Which eventloop should be called back on completion.
+  /// - returns: A future on the `callbackLoop` which will succeed on completion of shutdown.
+  func shutdown(callbackLoop: EventLoop) -> EventLoopFuture<Void> {
+    return self.server.flatMap { server in
+      server.close()
+    }.hop(to: callbackLoop).flatMap { _ in
```
glbrntt, line 94: We should probably stick a `recover` in here too in case the server errors on close.

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/BenchmarkServiceImpl.swift
```diff
@@ -0,0 +1,111 @@
...
+    } catch {
+      return context.eventLoop.makeFailedFuture(error)
+    }
+  }
+
+  /// Repeated sequence of one request followed by one response.
+  /// Should be called streaming ping-pong
+  /// The server returns the client payload as-is on each response
+  func streamingCall(context: StreamingResponseCallContext<Grpc_Testing_SimpleResponse>)
+    -> EventLoopFuture<(StreamEvent<Grpc_Testing_SimpleRequest>) -> Void> {
```
glbrntt, line 39: This is probably the work of swift format, but preferred style in gRPC is:

```swift
func streamingCall(
  context: StreamingResponseCallContext<Grpc_Testing_SimpleResponse>
) -> EventLoopFuture<(StreamEvent<Grpc_Testing_SimpleRequest>) -> Void> {
```

i.e. one argument per line and closing `)` on a new line (since it breaks visually splits signature a body a little better.

I'm fine with these as-is given they're not implemented so just an FYI.

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/BenchmarkServiceImpl.swift
```diff
@@ -0,0 +1,111 @@
...
+      message: "Not implemented"
+    ))
+  }
+
+  /// Make a payload for sending back to the client.
+  private static func makePayload(type: Grpc_Testing_PayloadType,
+                                  size: Int) throws -> Grpc_Testing_Payload {
+    if type != .compressable {
+      // Making a payload which is not compressable is hard - and not implemented in
+      // other implementations too.
```
glbrntt, line 87: I guess they planned on doing this but never did and didn't want to break API 🤷‍♂️

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/WorkerServiceImpl.swift
```diff
@@ -0,0 +1,325 @@
...
+        context: context,
+        clientConfig: clientConfig
+      )
+    } catch {
+      context.statusPromise.fail(error)
+    }
+  }
+
+  /// Create a client of the requested type.
+  private static func createClient(
```
glbrntt, line 296: nit: `makeClient`

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/grpcTime.swift
```diff
@@ -0,0 +1,50 @@
...
+  /// Convert to a potentially fractional number of seconds.
+  func asSeconds() -> Double {
+    return Double(self.value) * 1e-9
+  }
+}
+
+extension Nanoseconds: CustomStringConvertible {
+  /// Description to aid debugging.
+  var description: String {
+    return String(self.value)
```
glbrntt, line 48: Should we include the units too?

```suggestion
    return "\(self.value) ns"
```

Performance/QPSBenchmark/Makefile
glbrntt: This one hasn't been addressed.

Although I'm also wondering whether we should just piggyback on top of the main `Makefile` in the root of the repo to generate the protos. WDYT?

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/main.swift
```diff
@@ -0,0 +1,73 @@
...
+        logger.info("\(self) started successfully 🚀")
+      }
+    }
+
+    lifecycle.wait()
+
+    logger.info("Worker has finished.")
+  }
+}
+
```
glbrntt, line 72: Can we add in the following?

```swift
assert({
  print("⚠️ WARNING: YOU ARE RUNNING IN DEBUG MODE ⚠️")
  return true
}())
```

glbrntt (changes requested):
Thanks for adding some more docs and splitting things up; the smaller functions help a lot.

This looks pretty good, I've left some comments mostly in `AsyncClient` which I think we can simplify a little.

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/AsyncClient.swift
```diff
@@ -0,0 +1,293 @@
...
+
+    /// Get stats for sending to the driver.
+    /// - parameters:
+    ///     - reset: Should the stats reset after copying.
+    /// - returns: The statistics for this channel.
+    func getStats(reset: Bool) -> Stats {
+      return self.stats.copyData(reset: reset)
+    }
+
+    private static func makeClientRequest(payloadConfig: Grpc_Testing_PayloadConfig) throws
```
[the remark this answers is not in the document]
PeterAdams-A: I think having looked at it, that this would make us more similar to the C version too.

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/AsyncClient.swift
```diff
@@ -192,7 +238,7 @@ final class AsyncUnaryQPSClient: QPSClient {
         self.stopIsComplete()
       } else {
         // Try scheduling another request.
-        try! self.launchRequests()
+        try! self.makeRequestAndRepeat()
```
glbrntt, line 241: This function only `throws` because of this `try` on itself, let's drop both!

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/AsyncClient.swift
```diff
@@ -150,23 +191,28 @@ final class AsyncUnaryQPSClient: QPSClient {
...
       precondition(self.connection.eventLoop.inEventLoop)
-      while !self.stopRequested,
-        self.numberOfOutstandingRequests < self.maxPermittedOutstandingRequests {
+      while self.canMakeRequest() {
         try self.makeRequestAndRepeat()
       }
     }
 
+    /// Returns if it is permissible to make another request - ie we've not been asked to stop, and we're not at the limit of outstanding requests.
+    private func canMakeRequest() -> Bool {
```
glbrntt, line 200: This should be a computed property: 

```swift
private var canMakeRequest: Bool {
  return !(self.stopRequested || self.numberOfOutstandingRequests >= self.maxPermittedOutstandingRequests)
}
```

glbrntt (changes requested):
Looks great, just two nits remaining

glbrntt (approved):
LGTM thanks @PeterAdams-A!

PR #993, opened 2020-10-14
Add head and tail interceptors.

glbrntt:
Motivation:

An interceptor pipeline needs interceptors, and users need to be able to
create their own interceptors.

Modifications:

- Rename 'ClientInterceptor' to 'ClientInterceptorProtocol'
- Add a 'ClientInterceptor' base class for users to extend
- Add head and tail interceptors
- Add 'AnyClientInterceptor' which wraps the above.
- Build out a little more of the pipeline; note that this is by no means
  complete yet but it gets us to a stage where we can fire events
  through it.
- Move next inbound/outbound context fetching from context to pipeline
  because the context doesn't need to know the ordering of interceptors.
- Add a couple of simple pipeline tests.

Result:

We have a basic interceptor pipeline and a way for users to define
interceptors.


Lukasa (approved):
Nice patch, LGTM.

PR #1000, opened 2020-10-19
Pin formatter (#995)

glbrntt:
Motivation:

We added SwiftFormat as part our CI to in order to keep formatting
consistent. However, new rules are often introduced and new behaviours
are added and the formatting can change quite significantly as a result.
In order to get stable formatting we need to pin the version we use or
include the tool as part of the project. This is somewhat annoying: in
the first case it requires the user to get the right version, in the
second case we add a depdendency which isn't required to consume the
library, that's a little silly! We can make the first case a little
nicer by providing a script to build the right version from source.

Modifications:

- Add scripts/format.sh which builds a predefined version of SwiftFormat
  and runs it
- Pin the version in .travis-install.sh to the same as in format.sh

Result:

- SwiftFormat is pinned to a version

(cherry picked from commit cdb46f4d24fc4aef5c6f4ef5ebf50215354c28bb)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ Examples/EchoWeb/dist
 Examples/EchoWeb/node_modules
 Examples/EchoWeb/package-lock.json
 dev/codegen-tests/**/generated/*
+/scripts/.swiftformat-source/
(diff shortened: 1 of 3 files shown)


glbrntt:
(This was cherry picked from `main`)

PR #1001, opened 2020-10-19
Add a client transport

glbrntt:
Motivation:

To support interceptors we need some glue between the interceptor
pipeline and any `NIO.Channel` that provides the transport.

Modifications:

- Add 'ClientTransport' which holds an interceptor pipeline and provides
  the glue between that pipeline and the `NIO.Channel`. It is also a
  `ChannelHandler` and provides a way for itself to be added to a
  `ChannelPipeline`.
- Added tests

Result:

More pieces in places for interceptors.


Sources/GRPC/Interceptor/ClientTransport.swift
```diff
@@ -0,0 +1,752 @@
...
+    }
+  }
+
+  // Our `NIO.Channel` will fire trailers and the `GRPCStatus` to us separately. It's more
+  // convenient to have both at the same time when intercepting response parts. We'll hold on to the
+  // trailers here and only forward them when we receive the status.
+  private var trailers: HPACKHeaders?
+
+  /// The interceptor pipeline connected to this transport. This must be a `var` and an IOU because
+  /// it requires `self` to be
```
Lukasa, line 79: Truncated comment.

Sources/GRPC/Interceptor/ClientTransport.swift
```diff
@@ -0,0 +1,752 @@
...
+  internal init(
+    details: CallDetails,
+    eventLoop: EventLoop,
+    interceptors: [ClientInterceptor<Request, Response>],
+    errorDelegate: ClientErrorDelegate?,
+    _ onResponsePart: @escaping (ClientResponsePart<Response>) -> Void
+  ) {
+    self.eventLoop = eventLoop
+    self.callDetails = details
+    self.pipeline = ClientInterceptorPipeline(
```
Lukasa, line 100: Do we need to `nil` this out on `handlerRemoved` to break the reference cycle?

glbrntt: Yes, we should. Good catch!

Lukasa (approved):

PR #1002, opened 2020-10-20
Provide a "call" object for clients

glbrntt:
Motivation:

To invoke an RPC you must know the type of RPC you are calling (unary,
client streaming, etc.), its path (e.g. "/echo.Echo/Get"), and the
request and response type it expects as well as some other ancillary
objects to provide the transport for the call. At the moment the type of
the RPC is encoded as the object encapsulating it, e.g. we have a
`UnaryCall` for unary calls and `ClientStreamingCall` for client
streaming calls, and so on. We need something below this: a `Call`
object which can be told what type of call it is. This requirement is
particularly important for interceptors which may retry RPCs, since
within an interceptor we are only aware of the type of a call rather
than the type of the call object.

An additional advantange of this is that it enables us to greatly
simplify the requirements of `GRPCChannel`; instead of implementors
being required to have methods which create a call of each type for each
payload type (`SwiftProtobuf.Message` and `GRPCPayload`) - 8 methods in
total - they may instead have a requirement of making a `Call` for each
of the two payload types (2 methods). The higher level types can be
implemented as default extensions on `GRPCChannel` although that is not
done in this change (to keep it a resonable size).

Modifications:

- Add a `Call` object
- Add a `ClientTransportFactory` for making and configure `ClientTransport`
- Add call making functions to `GRPCChannel` and each implementation
- Add the 'type' and 'path' of the RPC to the interceptor context

Result:

Users can create a `Call` from an interceptor pipeline.


Sources/GRPC/ClientCalls/Call.swift
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  /// Starts the call and provides a callback which is invoked on every response part received from
+  /// the server.
+  ///
+  /// This must be called prior to `send(_:promise:)` or `cancel(promise:)`.
+  ///
+  /// - Parameter onResponsePart: A callback which is invoked on every response part.
+  /// - Important: This function should only be called once. Subsequent calls will be ignored.
+  public func invoke(_ onResponsePart: @escaping (ClientResponsePart<Response>) -> Void) {
```
Lukasa, line 97: There are lots of generic functions here: should we consider making them inlinable?

Sources/GRPC/GRPCChannel/GRPCChannel.swift
```diff
@@ -19,6 +19,53 @@ import NIOSSL
...
+  /// for each response part (or error) received. Any call to `send(_:promise:)` prior to calling
+  /// `invoke` will fail and not be sent. Users are also responsible for closing the request stream
+  /// by sending the `.end` request part.
+  ///
+  /// - Parameters:
+  ///   - path: The path of the RPC, e.g. "/echo.Echo/get".
+  ///   - type: The type of the RPC, e.g. `.unary`.
+  ///   - callOptions: Options for the RPC.
+  ///   - interceptors: A list of interceptors to intercept the request and response stream with.
+  func makeCall<Request: SwiftProtobuf.Message, Response: SwiftProtobuf.Message>(
```
Lukasa, line 39: Are we trying to wave users away from this? Do we want underscores in the names, if so?

Sources/GRPC/Interceptor/ClientTransportFactory.swift
```diff
@@ -0,0 +1,275 @@
...
+      type: type,
+      path: path,
+      authority: self.authority,
+      scheme: self.scheme,
+      options: options
+    )
+  }
+}
+
+private struct Fake<Request, Response> {
```
Lukasa, line 220: Should we nest this and `HTTP2` inside extensions on `ClientTransportFactory`, just so their names are easier to identify in tracebacks and stuff?

Sources/GRPC/GRPCChannel/GRPCChannel.swift
glbrntt: Not quite, they should prefer the wrapped calls but there are situations where using this is preferred (e.g. retrying interceptors) so I don't think we should include underscores.

Sources/GRPC/Interceptor/ClientTransportFactory.swift
glbrntt: Hmm I had some weird compile errors when they were nested (with pre 5.3 compilers). I don't recall exactly what the error was but I'll make the names more obvious; `HTTP2ClientTransportFactory`.

Sources/GRPC/ClientCalls/Call.swift
glbrntt: Yes, good thinking.

Lukasa (approved):

PR #1003, opened 2020-10-21
Ping pong benchmark server

PeterAdams-A:
Motivation:

This is the first part of having a method of benchmarking
which can be compared to other languages.

Modifications:

Add a trivial server side implementation.

Result:

QPSBenchmark will respond to requests to be a ping pong streaming
server.


glbrntt (approved):
LGTM

PR #1004, opened 2020-10-21
Add an Async Ping Pong streaming benchmark

PeterAdams-A:
Motivation:

Unary benchmark already exists - adding a streaming benchmark
completes very high level coverage.

Modifications:

Refactor Unary benchmaark to break out what's specific to
that and add a second implementation of that broken out
class which runs a streaming test.

Result:

Both Unary and Streaming benchmarks exist.


Performance/QPSBenchmark/README.md
```diff
@@ -16,7 +16,9 @@ You can easily run the tests locally using the C++ driver program from gRPC - no
 of running the C++ tests which can be done in a gRPC checkout with 
 `./tools/run_tests/run_performance_tests.py -l c++ -r cpp_protobuf_async_unary_qps_unconstrained_insecure`
 
-For an example of running a benchmarking tests proceed as follows
+For a examples of running benchmarking tests proceed as follows
```
glbrntt, line 19:
```suggestion
For examples of running benchmarking tests proceed as follows
```

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/AsyncPingPongRequestMaker.swift
```diff
@@ -0,0 +1,93 @@
...
+
+    /// Handle a response from the server - potentially triggers making another request.
+    func handleResponse(response: Grpc_Testing_SimpleResponse) {
+      let endTime = grpcTimeNow()
+      self.stats.add(latency: endTime - startTime)
+      if !self.stopRequested,
+        self.messagesPerStream == 0 || messagesSent < self.messagesPerStream {
+        messagesSent += 1
+        startTime = endTime // Use end of previous request as the start of the next.
+        _ = streamingCall!.sendMessage(self.requestMessage)
```
glbrntt, line 74: We can avoid allocating the promise here: `sendMessage(self.requestMessage, promise: nil)`

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/AsyncPingPongRequestMaker.swift
```diff
@@ -0,0 +1,93 @@
...
+  func makeRequest() -> EventLoopFuture<GRPCStatus> {
+    var startTime = grpcTimeNow()
+    var messagesSent = 1
+    var streamingCall: BidirectionalStreamingCall<
+      Grpc_Testing_SimpleRequest,
+      Grpc_Testing_SimpleResponse
+    >?
+
+    /// Handle a response from the server - potentially triggers making another request.
+    func handleResponse(response: Grpc_Testing_SimpleResponse) {
```
glbrntt, line 67: This will be expected on the event loop; I _think_ we're okay because `ChannelRepeater` has to execute on the event loop as well, but documenting these thread safety requirements and asserting we're on the event loop in the right places would be very helpful.

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/AsyncPingPongRequestMaker.swift
```diff
@@ -0,0 +1,93 @@
...
+        _ = streamingCall!.sendMessage(self.requestMessage)
+      } else {
+        streamingCall!.sendEnd(promise: nil)
+      }
+    }
+
+    // Setup the call.
+    streamingCall = self.client.streamingCall(handler: handleResponse)
+    // Kick start with initial request
+    _ = streamingCall!.sendMessage(self.requestMessage)
```
glbrntt, line 83: Avoid the promise: `promise: nil`

glbrntt (changes requested):
Looks pretty good: we can avoid allocating promises in a couple of places so we should do that.

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/AsyncClient.swift
```diff
@@ -207,12 +207,14 @@ final class AsyncQPSClient<RequestMakerType: RequestMaker>: QPSClient {
 
     /// Returns if it is permissible to make another request - ie we've not been asked to stop, and we're not at the limit of outstanding requests.
     private var canMakeRequest: Bool {
+      assert(self.connection.eventLoop.inEventLoop)
```
glbrntt, line 210: As per Cory's comment offline: let's use `assertInEventLoop()` and `preconditionInEventLoop()`

glbrntt (approved):

PR #1010, opened 2020-10-23
Wire up the Call objects

glbrntt:
Motivation:

We have the new `Call` object and a way to make calls. We should
refactory the existing `UnaryCall`, `ClientStreamingCall`, etc. to be
wrappers around `Call`.

Modifications:

- Reimplement the four call types on top of `Call`; these are now
  implemented as structs.
- This required:
  - some additional awkward logic in `Call` if the users requests the
    stream `Channel` future
  - removing requirements from `GRPCChannel` for constructing each call
    type, and their concrete implementations. This is now done as an
    extension.

Result:

- The four call objects are wired up so that they can use interceptors.
- Much less boilerplate


Sources/GRPC/ClientCalls/BidirectionalStreamingCall.swift
```diff
@@ -22,63 +22,60 @@ import NIOHTTP2
 ///
 /// Messages should be sent via the `sendMessage` and `sendMessages` methods; the stream of messages
 /// must be terminated by calling `sendEnd` to indicate the final message has been sent.
-public final class BidirectionalStreamingCall<
+public struct BidirectionalStreamingCall<
```
Lukasa, line 25: This struct still has reference semantics. It'd be good to clarify that in the doc comment. The same applies to the others.

Sources/GRPC/ClientCalls/Call.swift
```diff
@@ -162,6 +177,84 @@ extension Call {
...
+
+    var iterator = messages.makeIterator()
+    var maybeNext = iterator.next()
+    while let current = maybeNext {
+      let next = iterator.next()
+      // If there's no next message, then we'll flush.
+      let flush = next == nil
+
+      // Make a promise, store the future.
+      let promise = self.eventLoop.makePromise(of: Void.self)
```
Lukasa, line 245: Do we need a promise on every message? It seems sufficient to attach one to the _last_ message, doesn't it? For streaming protocols like this you'd expect all these writes to fail with the same error.

Sources/GRPC/Interceptor/ClientTransport.swift
```diff
@@ -43,6 +43,10 @@ internal final class ClientTransport<Request, Response> {
   /// The current state of the transport.
   private var state: State = .idle
 
+  /// A lazy promise for the underlying `Channel`. We'll succeed this when we transition to `active`
```
Lukasa, line 46: Mild nit, but this promise isn't lazy.

Sources/GRPC/ClientCalls/Call.swift
glbrntt: Hmm, I guess not? I suppose it's technically possible for one write to fail if a handler does something weirdo but practically speaking that's highly unlikely to happen. Let's save some allocations.

Sources/GRPC/ClientCalls/BidirectionalStreamingCall.swift
glbrntt: Ah, yes, that's certainly worth documenting.

Lukasa (approved):

PR #1011, opened 2020-10-23
Reduce excessive server logging.

PeterAdams-A:
Motivation:

When running a unary server, sending the response can put the
HTTP1ToGRPCServerCodec into ignore state.  This happens before
the end of stream is processed.  The 2 end stream messages
(empty buffer of data and end) then trigger logging.

Modifications:

Ignore end and empty data without logging when codec is in
ignore state.

Result:

Less logging and hence faster performance.

### DIFF
--- a/Sources/GRPC/HTTP1ToGRPCServerCodec.swift
+++ b/Sources/GRPC/HTTP1ToGRPCServerCodec.swift
@@ -128,13 +128,19 @@ extension HTTP1ToGRPCServerCodec: ChannelInboundHandler {
   public typealias InboundOut = _RawGRPCServerRequestPart
 
   public func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+    let unwrappedData = self.unwrapInboundIn(data)
     if case .ignore = self.inboundState {
+        if case let .body(body) = unwrappedData, body.readableBytes == 0 {
+            return // Ignoring a read of zero bytes is always fine.
+        } else if case .end = unwrappedData {
+            return // Ignoring an end stream is not an event worth reporting.
+        }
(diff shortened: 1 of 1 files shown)


Sources/GRPC/HTTP1ToGRPCServerCodec.swift
```diff
@@ -128,13 +128,19 @@ extension HTTP1ToGRPCServerCodec: ChannelInboundHandler {
...
   public func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+    let unwrappedData = self.unwrapInboundIn(data)
     if case .ignore = self.inboundState {
+        if case let .body(body) = unwrappedData, body.readableBytes == 0 {
+            return // Ignoring a read of zero bytes is always fine.
+        } else if case .end = unwrappedData {
+            return // Ignoring an end stream is not an event worth reporting.
+        }
       self.logger.notice("ignoring read data", metadata: ["data": "\(data)"])
       return
```
glbrntt, line 139: Can we `switch` on the `unwrappedData` here instead? We can drop our `return` count to 1 and neaten it up a bit. 

Something like:

```swift
switch unwrappedData {
  case let .body(body) where body.readableBytes == 0:
    ()
  case .end:
    ()
  default:
    self.logger.notice(...)
  }
  return
}
```

glbrntt: Also we may as well log the `unwrappedData` instead of the `data` now.

glbrntt: Also you'll need to run `./scripts/format.sh`.

glbrntt (changes requested):

glbrntt (approved):

PR #1013, opened 2020-10-24
Remove duplicated code.

PeterAdams-A:
Remove unwrapping of data on two consecutive lines.


Lukasa (approved):
seems good, no risk of CoW here

PR #1016, opened 2020-10-25
Tolerate identity compression as no compression

### LINKED ISSUE #1005: grpc-swift servers reject requests with "grpc-encoding: identity" when compression is disabled, even though such a request is not compressed.
If a gRPC server is set up without compression support, and it receives an rpc containing a `grpc-encoding: identity` header, [that request will be rejected](https://github.com/grpc/grpc-swift/blob/a374528d108008795bc587549cda5c21dca970ee/Sources/GRPC/HTTP1ToGRPCServerCodec.swift#L569) despite the rpc not actually being compressed. We should add an escape hatch for this case to continue to accept the request.

Lukasa:
Motivation:

The GRPC specification acknowledges that "identity" is a way to say "no
compression", but when the user has disabled compression support we
refuse it as a compression format. That's not right.

Modifications:

- Add code to validate that the compression format isn't identity.
- Add some tests for newly added code.

Results:

Resolves #1005.

### DIFF
--- a/Sources/GRPC/HTTP1ToGRPCServerCodec.swift
+++ b/Sources/GRPC/HTTP1ToGRPCServerCodec.swift
@@ -563,11 +563,22 @@ struct MessageEncodingHeaderValidator {
         )
       }
 
-    // Compression is disabled and the client sent a message encoding header. We clearly don't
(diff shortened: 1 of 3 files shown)


glbrntt (approved):
Great, thanks @Lukasa

PR #1017, opened 2020-10-26
Remove HTTP1 types from the public server API

glbrntt:
Motivation:

Currently, on the server, we always operate on HTTP/1 types. The
advantage of this is that supporting gRPC-Web is trivial. However, the
vast majority of connections are HTTP/2 and this transformation
incurs a non-trivial performance cost.

In the future we'd like to change how we support gRPC-Web by always
operating on HTTP/2 types and converting to HTTP/1 types only when
necessary.

The first step in achieving this is removing HTTP/1 types from the
public server API.

Modifications:

- Remove 'request' (i.e. 'HTTPRequestHead') from the server call
  contexts and add 'headers' (as 'HPACKHeaders') in their place
- Modify the server request and response parts to use 'HPACKHeaders'
  rather than 'HTTPHeaders'.
- Renaming 'GRPCStatusAndMetadata' to 'GRPCStatusAndTrailers' adding a
  deprecated typealias to help
- Add 'trailers' as 'HPACKHeaders' to 'GRPCStatusAndTrailers' and
  deprecate 'metadata'
- Change the relevant 'ServerErrorDelegate' methods. Unforunately these
  all have default implementations, so deprecating them isn't useful
  (since the implementor will be unaware of the change).

Result:

Server API is free of `HTTPHeaders` and `HTTPRequestHead`

### DIFF
--- a/Sources/GRPC/CallHandlers/BidirectionalStreamingCallHandler.swift
+++ b/Sources/GRPC/CallHandlers/BidirectionalStreamingCallHandler.swift
@@ -16,7 +16,7 @@
 import Foundation
 import Logging
 import NIO
-import NIOHTTP1
+import NIOHPACK
 import SwiftProtobuf
 
 /// Handles bidirectional streaming calls. Forwards incoming messages and end-of-stream events to the observer block.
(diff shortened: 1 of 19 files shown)


Lukasa (approved):
LGTM, nice start!

PR #1018, opened 2020-10-26
Tidy up some of the logic in the server call contexts

glbrntt:
Motivation:

The server call contexts rely on the service provider completing a
promise, either for the response or status depending on the streaming
type. Doing so then writes the relevant parts into the pipeline.
However, we also have to deal with the promises being failed, and since
the logic there is a little circuitous the callback chains were a little
hard to follow and had some redundancy.

Modifications:

- Add a 'processError(_:delegate:)` to the base server call context to
  transform an error into a status and trailers.
- Refactor the promise completion chains for the unary and streaming
  response call contexts.

Result:

- Code is easier to follow. Less duplication.


glbrntt:
Could you review this @Davidde94? (I can't tag you as a reviewer.)

Davidde94 (approved):
I think I get it, looks good to me

Lukasa (approved):
Nice, LGTM

PR #1019, opened 2020-10-27
Rename the client interceptor methods

glbrntt:
Motivation:

The gRPC API usually refers to sending and receiving messages. The
interceptor API uses 'write' and 'read' which are a little closer to the
NIO API. We should bias to being more consistent with terms already used
in the API.

Modifications:

- Rename 'write' to 'send', 'read' to 'receive'
- Make the 'ClientInterceptor' methods 'open'
- 'ClientInterceptorContext' methods delegate to a private
  implementation, this was done as the public methods ensured we were on
  the 'EventLoop' first. This is no longer the case so the indirection
  has been removed.

Result:

- Better naming


Lukasa (approved):
Agree, this makes sense.

PR #1020, opened 2020-10-27
Allow responses to be sent without a promise

### LINKED ISSUE #539: [nio] Provide more flexibility for server streaming contexts to send messages
Similar to #537, we should allows users to have more flexibility in how they send messages. At the moment they only have:
- `sendResponse(message:) -> EventLoopFuture<Void>`

Ideally there should be options to:
- provide an optional promise
- choose whether to flush or not
- send multiple messages
- manually flush

This would bring it into alignment to the client with changes made in #538.

glbrntt:
Motivation:

For server streming RPCs, users must send each response one-at-a-time,
there is also no option for callers to avoid allocating the promise
associated with writing that response.

Modifications:

- Add methods to send a single response or sequence of responses  with
  an optional promise
- Have the old implementations which return a future delegate to these
  methods
- Update the echo service provider so that we hit these paths
- Update the QPS streaming benchmark to avoid allocating the future.
- Update the docs for the QPS benchmark to be a little easier to follow
  and move the scenarios from an inline JSON string to JSON files.

Result:

- Users can send responses on streaming RPCs without allocating a future
- Users can send multple responses at a time (avoiding excessive
  flushing)
- A small (~1%) improvement in QPS in the streaming benchmark
- Resolves #539

### DIFF
--- a/Sources/GRPC/ServerCallContexts/StreamingResponseCallContext.swift
+++ b/Sources/GRPC/ServerCallContexts/StreamingResponseCallContext.swift
@@ -44,13 +44,64 @@ open class StreamingResponseCallContext<ResponsePayload>: ServerCallContextBase 
 
   /// Send a response to the client.
   ///
-  /// - Parameter message: The message to send to the client.
-  /// - Parameter compression: Whether compression should be used for this response. If compression
-  ///   is enabled in the call context, the value passed here takes precedence. Defaults to deferring
-  ///   to the value set on the call context.
-  open func sendResponse(_ message: ResponsePayload,
-                         compression: Compression = .deferToCallDefault) -> EventLoopFuture<Void> {
+  /// - Parameters:
+  ///   - message: The message to send to the client.
+  ///   - compression: Whether compression should be used for this response. If compression is
+  ///     is enabled in the call context, the value passed here takes precedence. Defaults to
+  ///     deferring to the value set on the call context.
+  ///   - promise: A promise to complete once the message has been sent.
+  open func sendResponse(
+    _ message: ResponsePayload,
+    compression: Compression = .deferToCallDefault,
+    promise: EventLoopPromise<Void>?
+  ) {
     fatalError("needs to be overridden")
+  }
+
+  /// Send a response to the client.
+  ///
+  /// - Parameters:
+  ///   - message: The message to send to the client.
+  ///   - compression: Whether compression should be used for this response. If compression is
+  ///     is enabled in the call context, the value passed here takes precedence. Defaults to
+  ///     deferring to the value set on the call context.
+  open func sendResponse(
+    _ message: ResponsePayload,
+    compression: Compression = .deferToCallDefault
+  ) -> EventLoopFuture<Void> {
+    let promise = self.eventLoop.makePromise(of: Void.self)
+    self.sendResponse(message, compression: compression, promise: promise)
+    return promise.futureResult
+  }
+
+  /// Sends a sequence of responses to the client.
+  /// - Parameters:
+  ///   - messages: The messages to send to the client.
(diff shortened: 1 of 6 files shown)


glbrntt:
cc @Davidde94

Sources/GRPC/ServerCallContexts/StreamingResponseCallContext.swift
```diff
@@ -44,14 +44,65 @@ open class StreamingResponseCallContext<ResponsePayload>: ServerCallContextBase
...
+  ) -> EventLoopFuture<Void> {
+    let promise = self.eventLoop.makePromise(of: Void.self)
+    self.sendResponse(message, compression: compression, promise: promise)
+    return promise.futureResult
+  }
+
+  /// Sends a sequence of responses to the client.
+  /// - Parameters:
+  ///   - messages: The messages to send to the client.
+  ///   - compression: Whether compression should be used for this response. If compression is
```
Davidde94, line 80: Nite: double "is"

```suggestion
  ///   - compression: Whether compression should be used for this response. If compression
```

Sources/GRPC/ServerCallContexts/StreamingResponseCallContext.swift
```diff
@@ -44,14 +44,65 @@ open class StreamingResponseCallContext<ResponsePayload>: ServerCallContextBase
...
+    compression: Compression = .deferToCallDefault,
+    promise: EventLoopPromise<Void>?
+  ) where Messages.Element == ResponsePayload {
+    fatalError("needs to be overridden")
+  }
+
+  /// Sends a sequence of responses to the client.
+  /// - Parameters:
+  ///   - messages: The messages to send to the client.
+  ///   - compression: Whether compression should be used for this response. If compression is
```
Davidde94, line 95: Same double is 

```suggestion
  ///   - compression: Whether compression should be used for this response. If compression
```

Davidde94 (approved):

Sources/GRPC/ServerCallContexts/StreamingResponseCallContext.swift
```diff
@@ -44,14 +44,65 @@ open class StreamingResponseCallContext<ResponsePayload>: ServerCallContextBase
...
+    promise: EventLoopPromise<Void>?
+  ) {
     fatalError("needs to be overridden")
   }
+
+  /// Send a response to the client.
+  ///
+  /// - Parameters:
+  ///   - message: The message to send to the client.
+  ///   - compression: Whether compression should be used for this response. If compression is
```
glbrntt, line 65:
```suggestion
  ///   - compression: Whether compression should be used for this response. If compression
```

Sources/GRPC/ServerCallContexts/StreamingResponseCallContext.swift
```diff
@@ -44,14 +44,65 @@ open class StreamingResponseCallContext<ResponsePayload>: ServerCallContextBase
...
   ///
-  /// - Parameter message: The message to send to the client.
-  /// - Parameter compression: Whether compression should be used for this response. If compression
-  ///   is enabled in the call context, the value passed here takes precedence. Defaults to deferring
-  ///   to the value set on the call context.
-  open func sendResponse(_ message: ResponsePayload,
-                         compression: Compression = .deferToCallDefault) -> EventLoopFuture<Void> {
+  /// - Parameters:
+  ///   - message: The message to send to the client.
+  ///   - compression: Whether compression should be used for this response. If compression is
```
glbrntt, line 49:
```suggestion
  ///   - compression: Whether compression should be used for this response. If compression
```

Lukasa (approved):
LGTM, nice change.

PR #1021, opened 2020-10-27
Fix typo protoc plugin name in README

chigichan24:
I found a typo in README so fix it.


glbrntt (approved):
Thank you @chigichan24!

PR #1022, opened 2020-10-28
Codegen for client interceptors

glbrntt:
Motivation:

We have client interceptors wired all the way through but no convnenient
way for them to be used.

Modifications:

- Generate an client interceptor factory which has a
  `make<Method>Interceptors()` function for each method on the service
- The interceptor factory is an optional requirement on the generated
  client protocol, RPC invocations will pull interceptors from the
  factory, if present.

Result:

Users can provide interceptor factory implementations to their generated
clients to use interceptors.


glbrntt:
The changes are in `ea1e537`, the other commit is just regenerated code.

Lukasa (approved):
Nice! Smooth change, I like it.

PR #1027, opened 2020-11-02
Split path strings using UTF8

PeterAdams-A:
Motivation:

For unary calls there can be a lot of path manipulation.
Splitting in utf8 space is approximately 10x faster.

Modifications:

Hand craft a two part split using utf8.

Result:

Very small overall speed improvement (less than 1%)

### DIFF
--- a/Sources/GRPC/GRPCServerRequestRoutingHandler.swift
+++ b/Sources/GRPC/GRPCServerRequestRoutingHandler.swift
@@ -291,7 +291,7 @@ extension GRPCServerRequestRoutingHandler: ChannelInboundHandler, RemovableChann
 extension Collection where Self == Self.SubSequence, Self.Element: Equatable {
   /// Trims out the prefix up to `separator`, and returns it.
   /// Sets self to the subsequence after the separator, and returns the subsequence before the separator.
-  /// If the self is emtpy returns `nil`
+  /// If self is emtpy returns `nil`
   /// - parameters:
   ///     - separator : The Element between the head which is returned and the rest which is left in self.
   /// - returns: SubSequence containing everything between the beginnning and the first occurance of
--- a/Sources/GRPC/GRPCServerRequestRoutingHandler.swift
+++ b/Sources/GRPC/GRPCServerRequestRoutingHandler.swift
@@ -231,13 +231,14 @@ extension GRPCServerRequestRoutingHandler: ChannelInboundHandler, RemovableChann
     /// URI format: "/package.Servicename/MethodName"
     init?(requestURI: String) {
       var utf8View = requestURI.utf8[...]
-      guard utf8View.splitFirst(separator: self.pathSplitDelimiter)?.count == 0 else {
+      // Check and remove the split character at the beginning.
+      guard utf8View.trimPrefix(to: self.pathSplitDelimiter)?.isEmpty == true else {
         return nil
       }
-      guard let service = utf8View.splitFirst(separator: pathSplitDelimiter) else {
+      guard let service = utf8View.trimPrefix(to: pathSplitDelimiter) else {
         return nil
       }
-      guard let method = utf8View.splitFirst(separator: pathSplitDelimiter) else {
+      guard let method = utf8View.trimPrefix(to: pathSplitDelimiter) else {
         return nil
       }
 
@@ -297,7 +298,7 @@ extension Collection where Self == Self.SubSequence, Self.Element: Equatable {
   /// - returns: SubSequence containing everything between the beginnning and the first occurance of
   /// `separator`.  If `separator` is not found this will be the entire Collection. If the collection is empty
   /// returns `nil`
-  mutating func splitFirst(separator: Element) -> SubSequence? {
+  mutating func trimPrefix(to separator: Element) -> SubSequence? {
     guard !self.isEmpty else {
       return nil
     }
--- a/Sources/GRPC/GRPCServerRequestRoutingHandler.swift
+++ b/Sources/GRPC/GRPCServerRequestRoutingHandler.swift
@@ -216,6 +216,40 @@ extension GRPCServerRequestRoutingHandler: ChannelInboundHandler, RemovableChann
     }
   }
 
+    /// A call URI split into components.
+    struct CallPath {
+        /// The name of the service to call.
+        var service: String.UTF8View.SubSequence
+        /// The name of the method to call.
+        var method: String.UTF8View.SubSequence
+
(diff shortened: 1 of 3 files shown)


PeterAdams-A:
For discussion on if this is worth it - makes things very slightly faster at the cost of more complex code.  

I'll not be upset if we decide not to do this as it ended up more complex than I originally thought it would be and the overall gains are quite small.

Sources/GRPC/GRPCServerRequestRoutingHandler.swift
```diff
@@ -216,25 +216,59 @@ extension GRPCServerRequestRoutingHandler: ChannelInboundHandler, RemovableChann
...
+            let requestUriUTF8 = requestUri.utf8
+            let maybeFirstIndex = requestUriUTF8.firstIndex(of: pathSplitDelimiter)
+            guard let firstIndex = maybeFirstIndex else {
+                return nil
+            }
+            let afterFirstDelimiter = requestUriUTF8[requestUriUTF8.index(after:firstIndex)...]
+            if let secondIndex = afterFirstDelimiter.firstIndex(of: pathSplitDelimiter) {
+                self.service = afterFirstDelimiter[..<secondIndex]
+                let afterSecondDelimiter = afterFirstDelimiter[afterFirstDelimiter.index(after: secondIndex)...]
+                if let thirdIndex = afterSecondDelimiter.firstIndex(of: pathSplitDelimiter) {
```
Lukasa, line 242: All of this seems like it might be nicer wrapped up in a helpful extension. Perhaps:

```swift
extension Collection where SubSequence == Self {
    /// Trims out the prefix up to separator, and returns it. Sets self to the subsequence
    /// after the separator, and returns the subsequence before the separator. Returns nil
    /// and does not mutate self if `separator` is not present.
    mutating func trim(through separator: Element) -> SubSequence? {
        if let separatorIndex = self.firstIndex(of: separator) {
            let indexAfterSeparator = self.index(after: separatorIndex)
            defer { self = self[indexAfterSeparator...] }
            return self[..<separatorIndex]
        } else {
            return nil
        }
    }
}
```

This will compose pretty nicely, I think, and greatly clean up this call site.

glbrntt (changes requested):
I'm generally on board with this, it's not that complex and has a useful performance gain. That said, it needs a little tidying up. You'll also need to run `./scripts/format.sh` and `make generate-linuxmain`

Sources/GRPC/GRPCServerRequestRoutingHandler.swift
```diff
@@ -216,25 +216,55 @@ extension GRPCServerRequestRoutingHandler: ChannelInboundHandler, RemovableChann
...
+
+    /// Charater used to split the path into components.
+    private let pathSplitDelimiter = UInt8(ascii: "/")
+
+    /// Split a path into service and method.
+    /// Split is done in UTF8 as this turns out to be approximately 10x faster than a simple split.
+    /// URI format: "/package.Servicename/MethodName"
+    init?(requestURI: String) {
+      var utf8View = requestURI.utf8[...]
+      guard utf8View.splitFirst(separator: self.pathSplitDelimiter)?.count == 0 else {
```
glbrntt, line 234: nit: `isEmpty` over `count == 0`

glbrntt (changes requested):
You still need to run` make generate-linuxmain`. I have a couple of naming nits but looks great otherwise.

Lukasa (approved):
No notes other than @glbrntt's.

Sources/GRPC/GRPCServerRequestRoutingHandler.swift
```diff
@@ -257,3 +287,27 @@ extension GRPCServerRequestRoutingHandler: ChannelInboundHandler, RemovableChann
...
+extension Collection where Self == Self.SubSequence, Self.Element: Equatable {
+  /// Trims out the prefix up to `separator`, and returns it.
+  /// Sets self to the subsequence after the separator, and returns the subsequence before the separator.
+  /// If self is emtpy returns `nil`
+  /// - parameters:
+  ///     - separator : The Element between the head which is returned and the rest which is left in self.
+  /// - returns: SubSequence containing everything between the beginnning and the first occurance of
+  /// `separator`.  If `separator` is not found this will be the entire Collection. If the collection is empty
+  /// returns `nil`
+  mutating func splitFirst(separator: Element) -> SubSequence? {
```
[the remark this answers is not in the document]
PeterAdams-A: OK - I'll go with `trimPrefix(to:`.  I'd originally chosen splitFirst as the function performed is more like the current `split` and I more associate `trim` with throwing away whitespace at either end.  You are right that regular `split` doesn't mutate though making that unclear.

Sources/GRPC/GRPCServerRequestRoutingHandler.swift
PeterAdams-A: I'm not sure you'll like my change here - I'm not sure I like it either but however you express this it's a little ugly.

Sources/GRPC/GRPCServerRequestRoutingHandler.swift
```diff
@@ -216,25 +216,56 @@ extension GRPCServerRequestRoutingHandler: ChannelInboundHandler, RemovableChann
...
+    /// Charater used to split the path into components.
+    private let pathSplitDelimiter = UInt8(ascii: "/")
+
+    /// Split a path into service and method.
+    /// Split is done in UTF8 as this turns out to be approximately 10x faster than a simple split.
+    /// URI format: "/package.Servicename/MethodName"
+    init?(requestURI: String) {
+      var utf8View = requestURI.utf8[...]
+      // Check and remove the split character at the beginning.
+      guard utf8View.trimPrefix(to: self.pathSplitDelimiter)?.isEmpty == true else {
```
glbrntt, line 235: This is the least ugly, I think.

```suggestion
      guard let prefix = utf8View.trimPrefix(to: self.pathSplitDelimiter), prefix.isEmpty else {
```

PeterAdams-A: Slightly less dark magic but adds another name `prefix` into scope below - no strong feelings really.

glbrntt: Agreed that all ways are a bit ugly but the extra `prefix` in the scope below is better than the `== true`, I think.

glbrntt (approved):
LGTM, thanks @PeterAdams-A!

PR #1028, opened 2020-11-03
Restructure `_BaseCallHandler` et al. for interceptors

glbrntt:
Motivation:

The state management in `_BaseCallHandler` and its associated subclasses
is a little handwavy and somewhat implicit. To support server
interceptors we need a stronger idea of the state of the call.

Modifications:

- Refactor `_BaseCallHandler` to have an explicit state machine and hold
  an interceptor pipeline. This is the 'main' state machine. Each
  subclass also has its own down state machine holding state relating to
  the event observer and observer context. Most of the logic shared between
  the state machines is very similar, it may be beneficial to abstract
  some of this away in the future.
- Add head and tail interceptors.
- Wire through interceptors from the call handler factory (much like the
  client, generated server code will have its own interceptor factory
  which will pass interceptors to the call handler factory).

Result:

- Server interceptors (minus codegen)


Lukasa (approved):
LGTM.

PR #1030, opened 2020-11-04
Codegen for server interceptors

glbrntt:
Motivation:

We need codegen to use server interceptors. This PR adds it.

Modifications:

- Add codegen for server interceptors.
- Also removes the default implementation of generated client
  interceptor factories, this makes the requirements clearer to the
  user.
- Adds a handful of tests using interceptors as well.

Result:

Codegen for server interceptors.


glbrntt:
Changes are in c0b9b3b; 6bd7f75 is just regenerated code.

Lukasa (approved):
LGTM.

PR #1033, opened 2020-11-04
Use the server hostname override as the :authority, if present

### LINKED ISSUE #1029: How to change HTTP authority Header.
# What are you trying to achieve?

I am trying to change the HTTP "authority" header to a different one from the host specified. I can see that on "ClientConnection"class the authority is specified as an internal var and it takes the value from the host specified.

For my project my server requires a different authority than the host name. I was wondering if there is any way I can set this. 

I was trying to make a connection and I run into a Error "401 Unauthorized" discussed in  #978 After doing some research and trying to change the authority using a callback and setting the HPACK header I was still getting the error. So I decrypted the HTTP2 packet using what was suggested on #1026. After decrypting the packet I saw that I had 2 authority headers set, one was equal to the host and the second one was the one set in the callback so I decided to check grpc-swift code and found out that in "ClientConnection" class the authority takes the value of the host. As a temporary solution I modified the class and made authority a public var then changed its value to the one I needed and removed the callback with the HPACKHeader. After doing that I successfully connected to my server.

gRPC-swift version: 1.0.0-alpha.19
TLS connection made using a custom configuration:

```
var configTLS = TLSConfiguration.forClient()
            configTLS.trustRoots = .certificates(pemCert)
            configTLS.certificateChain = [.certificate(serverSignedCert)]
            configTLS.privateKey = .privateKey(privKey)
            configTLS.minimumTLSVersion = TLSVersion.tlsv12
            configTLS.maximumTLSVersion = TLSVersion.tlsv12
            configTLS.certificateVerification = .fullVerification
            configTLS.applicationProtocols = ["grpc-exp", "h2"]
            configTLS.signingSignatureAlgorithms = [SignatureAlgorithm.rsaPkcs1Sha256]
            configTLS.verifySignatureAlgorithms = [SignatureAlgorithm.rsaPkcs1Sha256]
            configTLS.keyLogCallback = keyLogCallback

let config = ClientConnection.Configuration(target: ConnectionTarget.hostAndPort(self.CONTROLLER_ADDRESS, self.CONTROLLER_PORT), eventLoopGroup: group, errorDelegate: LoggingClientErrorDelegate(), connectivityStateDelegate: nil, connectivityStateDelegateQueue: nil, tls: ClientConnection.Configuration.TLS.init(configuration: configTLS, hostnameOverride: self.METRICS_AUTHORITY_HEADER), connectionBackoff: ConnectionBackoff(), connectionKeepalive: ClientConnectionKeepalive(), connectionIdleTimeout: .minutes(30), callStartBehavior: .waitsForConnectivity, httpTargetWindowSize: 65535, backgroundActivityLogger: logger, debugChannelInitializer: nil)

let connection = ClientConnection.init(configuration: config)
connection.authority = self.METRICS_AUTHORITY_HEADER

```

glbrntt:
Motivation:

When using a secured connection to a server, the client may present a
server hostname as part of the TLS SNI extension. However, the
':authority' header sent by the client will still just be the host the
caller specified, rather than the server hostname.

Modifications:

- User the TLS server hostname override, if present, in prefernce to the
  hostname when setting the authority.

Result:

- Resolves #1029

### DIFF
--- a/Sources/GRPC/ClientConnection.swift
+++ b/Sources/GRPC/ClientConnection.swift
@@ -116,7 +116,7 @@ public class ClientConnection {
   public init(configuration: Configuration) {
     self.configuration = configuration
     self.scheme = configuration.tls == nil ? "http" : "https"
-    self.authority = configuration.target.host
+    self.authority = configuration.tls?.hostnameOverride ?? configuration.target.host
     self.connectionManager = ConnectionManager(
       configuration: configuration,
       logger: configuration.backgroundActivityLogger
--- a/Tests/GRPCTests/ClientTLSTests.swift
+++ b/Tests/GRPCTests/ClientTLSTests.swift
@@ -136,4 +136,76 @@ class ClientTLSHostnameOverrideTests: GRPCTestCase {
 
     try self.doTestUnary()
   }
+
+  func testAuthorityUsesTLSHostnameOverride() throws {
+    // This test validates that when suppled with a server hostname override, the client uses it
+    // as the ":authority" pseudo-header.
+
+    self.server = try Server.secure(
+      group: self.eventLoopGroup,
+      certificateChain: [SampleCertificate.exampleServer.certificate],
+      privateKey: SamplePrivateKey.exampleServer
+    )
+    .withTLS(trustRoots: .certificates([SampleCertificate.ca.certificate]))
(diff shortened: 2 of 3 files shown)


Davidde94 (approved):

Lukasa (approved):
Good clean patch, nothing to say about it.

PR #1035, opened 2020-11-05
Remove the 'error' case from the 'GRPCClientResponsePart'

glbrntt:
Motivation:

When implementing a bunch of interceptors I found it a little irksome
having to deal with errors along with response parts. It feels more
natural to deal with them separately.

Modifications:

- Remove 'error' from 'GRPCClientResponsePart'
- Add an 'errorCaught' client interceptor function

Result:

Errors are handled separately to response parts.

### DIFF
--- a/Sources/GRPC/ClientCalls/BidirectionalStreamingCall.swift
+++ b/Sources/GRPC/ClientCalls/BidirectionalStreamingCall.swift
@@ -78,7 +78,10 @@ public struct BidirectionalStreamingCall<
   }
 
   internal func invoke() {
-    self.call.invokeStreamingRequests(self.responseParts.handle(_:))
+    self.call.invokeStreamingRequests(
+      onError: self.responseParts.handleError(_:),
+      onResponsePart: self.responseParts.handle(_:)
+    )
   }
 
   // MARK: - Requests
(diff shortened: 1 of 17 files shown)


glbrntt:
cc @PeterAdams-A @Davidde94

Davidde94 (approved):
LGTM, but wouldn't have using `Result<Error, Success>` saved you some effort?

glbrntt:
> LGTM, but wouldn't have using `Result<Error, Success>` saved you some effort?

How do you mean?

PeterAdams-A (commented):
Have you tried running this through an allocation test?

glbrntt:
@PeterAdams-A we don't have allocation counting tests yet

Davidde94:
@glbrntt `onResult: (Result<Error, SuccessType>) -> Void`. Anyway, just a thought, the code looks good

glbrntt:
> @glbrntt `onResult: (Result<Error, SuccessType>) -> Void`. Anyway, just a thought, the code looks good

Oh I see, that becomes a but ugly since the success type is also an enum. It's also against the motivation of the PR!

PeterAdams-A (approved):
Seems ok.  I did wonder if a little struct to hold error and response functions might make sense as they keep getting passed about together but no really strong feelings on the matter.

PR #1037, opened 2020-11-06
Fix incompatible code between main and the interceptors feature branch

glbrntt:
Motivation:

The generated code in the interceptors feature branch includes new
protocol requirements. Some tests added recently in #1033 to main don't
meet these requirements.

Modifications:

- Add the missing requirement

Result:

Feature branch compiles when rebased on 'main'.


PeterAdams-A (approved):
Looks good.

Davidde94 (approved):

PR #1041, opened 2020-11-12
Add HTTP2 to gRPC server state machine

glbrntt:
Motivation:

We use the HTTP 2-to-1 codecs on the server to simplify the
implementation. However this comes at a performance cost. We removed
the HTTP/1 types from surface API in #1017; this change introduces a
state machine for a translating between HTTP2 frame payloads and gRPC
server request parts. Future PRs will build on top of this to replace
both the 'HTTP1ToGRPCServerCodec' and the 'GRPCServerRequestRoutingHandler'.

Modifications:

Add an 'HTTP2ToRawGRPCStateMachine' and tests.

Result:

We have a state machine in place on which we can build a channel
handler.


Sources/GRPC/HTTP2ToRawGRPCStateMachine.swift
```diff
@@ -0,0 +1,1277 @@
...
+  }
+
+  /// Extract and validate the request message encoding header.
+  /// - Parameters:
+  ///   - headers: The headers to extract the message encoding header from.
+  /// - Returns: `RequestEncodingValidation`, either a message reader suitable for decoding requests
+  ///   and an accept encoding response header if the request encoding was valid, or a pair of
+  ///     `GRPCStatus` and trailers to close the RPC with.
+  private func extractRequestEncoding(from headers: HPACKHeaders) -> RequestEncodingValidation {
+    let encodingHeader = headers.first(name: GRPCHeaderName.encoding)
```
Lukasa, line 460: Is the use of `first` safe here? Are we choosing to ignore other appearances of this header on purpose?

Sources/GRPC/HTTP2ToRawGRPCStateMachine.swift
```diff
@@ -0,0 +1,1277 @@
...
+    contentType: ContentType,
+    responseEncoding: String?,
+    acceptableRequestEncoding: String?,
+    userProvidedHeaders: HPACKHeaders,
+    normalizeUserProvidedHeaders: Bool
+  ) -> HPACKHeaders {
+    let capacity = 2 // :status, content-type
+      + (responseEncoding == nil ? 0 : 1)
+      + (acceptableRequestEncoding == nil ? 0 : 1)
+      + userProvidedHeaders.count
```
Lukasa, line 1182: Is the math worthwhile here, or should we just do 4 + count? Seems like excessive branching for minimal savings.

Sources/GRPC/HTTP2ToRawGRPCStateMachine.swift
```diff
@@ -0,0 +1,1277 @@
...
+    for status: GRPCStatus,
+    contentType: ContentType,
+    acceptableRequestEncoding: String?,
+    userProvidedHeaders: HPACKHeaders?,
+    normalizeUserProvidedHeaders: Bool
+  ) -> HPACKHeaders {
+    let capacity = 3 // :status, status-code, content-type
+      + (status.message == nil ? 0 : 1)
+      + (acceptableRequestEncoding == nil ? 0 : 1)
+      + (userProvidedHeaders.map { $0.count } ?? 0)
```
Lukasa, line 1214: Same question here about the math

Sources/GRPC/HTTP2ToRawGRPCStateMachine.swift
```diff
@@ -0,0 +1,1277 @@
...
+    return headers
+  }
+
+  static func makeResponseTrailers(
+    for status: GRPCStatus,
+    userProvidedHeaders: HPACKHeaders,
+    normalizeUserProvidedHeaders: Bool
+  ) -> HPACKHeaders {
+    let capacity = 1 // status-code
+      + (status.message == nil ? 0 : 1)
```
Lukasa, line 1246: Same question here about math.

Sources/GRPC/HTTP2ToRawGRPCStateMachine.swift
glbrntt: Good question. We don't expect other occurrences of this header which is why I used `first`. Do you think we should use `canonicalForm` instead and close if there's more than one value?

Sources/GRPC/HTTP2ToRawGRPCStateMachine.swift
glbrntt: You're probably right tbh.

Sources/GRPC/HTTP2ToRawGRPCStateMachine.swift
Lukasa: I'd be inclined to do that, yeah.

Lukasa (approved):

PR #1045, opened 2020-11-19
Add single connection performance scenarios

glbrntt:
Motivation:

Running the unconstrained performance benchmark on a development
machines yields results with a large amount of variance (I have seen
over 20% between runs in some cases) since all cores become fully
loaded. As such it's much harder to determine whether a change was
actually beneficial.

Modifications:

- Add a single connection unary scenario, a copy of the unconstrained
  unary scenario but uses a single client such that only two threads are
  used: one for the server and one for the client.

Result:

- Results with less variance between runs.


glbrntt:
cc @PeterAdams-A

Lukasa (approved):

PR #1046, opened 2020-11-19
Move server (de)serialization to the base call handler

glbrntt:
Motivation:

Recently, when experimenting I found that found that doing
serialization/deserializion in the server call handler (rather than a
separate codec) led to a non-trivial performance gain.

Modifications:

- Make the `_BaseCallHandler` and its subclasses generic over the
  serializer and deserializer and do (de)serialization there rather than
  in a separate codec handler.
- Update the call handler factory - which provides convenience methods
  used by generated code. While this is a breaking change, it remains
  source compatible, so code is not required to be regenerated.
- Remove the requirement that a 'GRPCCallHandler' provide a codec
- Remove the server codec handler and its tests
- Make MessageSerializer and MessageDeserializer and their
  implementations public

Result:

A 6% gain in QPS benchmarks (5.8% for unary, 6.4% for bidirectional
streaming, average across 3 runs)


Sources/GRPC/CallHandlers/_BaseCallHandler.swift
```diff
@@ -41,6 +45,12 @@ public class _BaseCallHandler<Request, Response>: GRPCCallHandler, ChannelInboun
   /// Some context provided to us from the routing handler.
   private let callHandlerContext: CallHandlerContext
 
+  /// A request deserializer.
+  private let requestDeserializr: RequestDeserializer
```
Lukasa, line 49: Nit: missing `e`

Lukasa (approved):
Basically looks good, one typo note

Sources/GRPC/CallHandlers/_BaseCallHandler.swift
glbrntt: Oof, thanks.

PR #1047, opened 2020-11-19
Add a gRPC Web to HTTP/2 codec

glbrntt:
Motivation:

Following from #1041, we'd like to switch our currency type on the
server from HTTP/1 to HTTP/2. This means that to retain support for gRPC
Web we need a codec to convert between HTTP/1 and HTTP/2 in the opposite
direction to the codecs provided by NIO HTTP/2.

Modifications:

- Add a gRPC Web to HTTP2 codec

Result:

- Nothing yet, as this remains unused.


Sources/GRPC/GRPCWebToHTTP2ServerCodec.swift
```diff
@@ -0,0 +1,482 @@
...
+// MARK: - Inbound
+
+extension GRPCWebToHTTP2ServerCodec.StateMachine.State {
+  private mutating func processRequestHead(
+    _ head: HTTPRequestHead,
+    allocator: ByteBufferAllocator
+  ) -> GRPCWebToHTTP2ServerCodec.StateMachine.Action {
+    switch self {
+    case .idle:
+      var headers = HPACKHeaders(httpHeaders: head.headers, normalizeHTTPHeaders: true)
```
Lukasa, line 216: Should we reserve capacity here?

Sources/GRPC/GRPCWebToHTTP2ServerCodec.swift
```diff
@@ -0,0 +1,482 @@
...
+      headers.add(name: ":path", value: head.uri)
+      headers.add(name: ":method", value: head.method.rawValue)
```
Lukasa, line 218: These pseudo-headers are surely not going to be in the right place in the block, right? And I think we're missing two: `:authority` and `:scheme` are both expected as well.

Sources/GRPC/GRPCWebToHTTP2ServerCodec.swift
```diff
@@ -0,0 +1,482 @@
...
+    return buffer
+  }
+}
+
+extension HTTPHeaders {
+  fileprivate init(hpackHeaders headers: HPACKHeaders) {
+    self.init()
+    self.reserveCapacity(headers.count)
+    // Don't add any pseudo-headers.
+    for (name, value, _) in headers where name.utf8.first != .some(UInt8(ascii: ":")) {
```
Lukasa, line 478: There's a mild perf improvement available here: well-formatted HTTP/2 header blocks must have all their pseudo-headers _first_. You can therefore do a `dropFirst(where:)` to only look at the minimal number of headers and then just do `add(contentsOf:)` or similar to reduce the amount of unnecessary checking you do.

Sources/GRPC/GRPCWebToHTTP2ServerCodec.swift
glbrntt: I didn't think the order mattered given these will only be read by the HTTP/2 to gRPC codec? Easy enough to change though.

I wasn't sure about including `:authority` and `:scheme`; they're also not required but I suppose they _might_ be expected by a user in e.g. an interceptor. WDYT?

Sources/GRPC/GRPCWebToHTTP2ServerCodec.swift
glbrntt: Oh, smart thinking, I like it!

Sources/GRPC/GRPCWebToHTTP2ServerCodec.swift
Lukasa: I'm inclined to suggest that we should do the right thing here. It reduces the risk that we might accidentally introduce a component that doesn't expect this and get issues down the road. The same applies to adding the extra headers.

glbrntt: That seems wise. Done!

Lukasa (approved):

PR #1050, opened 2020-11-20
Plug in and switch on the 'HTTP/2 first' server handlers

glbrntt:
Motivation:

We have all the parts in place to use HTTP/2 concepts by default on the
server, so let's use them!

Modifications:

- Wire in the server error delegate to the 'HTTP2ToRawGRPCServerCodec'
  since this was missing
- Modify the 'HTTPProtocolSwitcher' and relevant pipeline initializer to
  configure the pipeline correctly
- Update various tests which relied on previous types
- Move 'CallPath' out of the routing handler
- Remove the routing handler and tests
- Move the 'MessageEncodingHeaderValidator' out of the HTTP1 to GRPC
  server handler
- Remove the 'HTTP1ToGRPCServerCodec'

Result:

When handling requests on HTTP/2 connections we no longer go translate
to HTTP1 types first.

A ~30-35% increase in QPS in unary benchmarks.


Sources/GRPC/MessageEncodingHeaderValidator.swift
```diff
@@ -0,0 +1,92 @@
+/*
```
glbrntt, line 1: This whole file was copied verbatim from the (now deleted) HTTP1ToGRPCServerCodec.swift

Tests/GRPCTests/CallPathTests.swift
```diff
@@ -0,0 +1,67 @@
+/*
```
glbrntt, line 1: This whole file was copied from the routing handler tests (which is now deleted)

Lukasa (approved):
This is a great patch. Performance wins are great, negative diffs are great, and it's really good to see this work culminate in such a big win. Nice one @glbrntt!

Issue #1042, opened 2020-11-16
Consider enforcing ALPN negotiation


Lukasa:
Right now both the server and the client ignore the result of the ALPN negotiation. The client does log the outcome, but otherwise does nothing: the server largely ignores it and sniffs the initial content. This is likely suboptimal: the ALPN is normative and we should probably pay attention to it. In some cases we likely want to skip that, but we can make that configuration.

glbrntt:
Resolved by #1055

PR #1055, opened 2020-11-24
Use ALPN to determine what sort of connection we've accepted

glbrntt:
Motivation:

To determine whether we're dealing with gRPC or gRPC Web (i.e. HTTP/2 or
HTTP/1) the first bytes received on a new connection are parsed. When
TLS is enabled and ALPN is supported there's no need to do this: we
will be informed about the protocol we negotiated with our peer and
therefore how to configure the channel. We also never enforced that the
protocol negotiated via ALPN was used!

Modifications:

- Add a 'GRPCServerPipelineConfigurator' handler to supersede the
  'HTTPProtocolSwitcher'. The new handler will use either ALPN and fall
  back to parsing the incoming bytes.
- The parsing behavior is slightly different to that of
  'HTTPProtocolSwitcher' in that we only read off enough bytes for the
  HTTP/2 connection preface, or enough bytes to parse an HTTP/1 request
  line (rather than reading the whole buffer as a string and processing
  that).
- Added a server configuration option to disable ALPN being required.
- Added tests.

Result:

- Server pipeline configuration is driven by ALPN, falling back to
  parsing.
- ALPN is enforced on the server (partially resolving #1042)


Sources/GRPC/GRPCServerPipelineConfigurator.swift
```diff
@@ -0,0 +1,350 @@
...
+    UInt8(ascii: "\r"),
+    UInt8(ascii: "\n"),
+  ]
+
+  /// Determines whether the bytes in the `ByteBuffer` are prefixed with the HTTP/2 client
+  /// connection preface.
+  static func prefixedWithHTTP2ConnectionPreface(_ buffer: ByteBuffer) -> Bool {
+    var buffer = buffer
+
+    guard let bytes = buffer.readBytes(length: HTTPVersionParser.http2ClientMagic.count) else {
```
Lukasa, line 304: It's a shame to allocate an array here. You can use `ByteBufferView.elementsEqual` to get something pretty close, I think, that skips the alloc.

Sources/GRPC/GRPCServerPipelineConfigurator.swift
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    // Check that the LF followed the CR.
+    guard readableBytesView.first == UInt8(ascii: "\n") else {
+      return false
+    }
+
+    // Now check the HTTP version.
+    var version = ByteBuffer(versionView)
+    let versionBytes = version.readBytes(length: version.readableBytes)!
```
Lukasa, line 347: Again, it's a shame to alloc here.

Sources/GRPC/GRPCServerPipelineConfigurator.swift
```diff
@@ -0,0 +1,350 @@
...
+  /// The server configuration.
+  private let configuration: Server.Configuration
+
+  /// Reads which we're holding on to before the pipeline is configured.
+  private var bufferedReads = CircularBuffer<NIOAny>()
+
+  /// The current state.
+  private var state: State = .notConfigured
+
+  private enum State {
```
Lukasa, line 41: This doesn't seem to know what it's _expecting_ to see out of ALPN or text. That means in principle if the timing gets weird you can be tricked by the ALPN arriving late. This is unlikely in a proper NIO pipeline, but it seems like it might be better to know which thing we're expecting?

glbrntt: Yeah that's fair I think, we should be more explicit about this.

Sources/GRPC/GRPCServerPipelineConfigurator.swift
glbrntt: Oh neat, I wasn't aware of that one.

Lukasa (approved):
Nice one, looks great.

Issue #1036, opened 2020-11-06
Service name in client 


mengxiangyue:
Hello team, I am do the investigation how to integrate with `client load balance` strategy, but I find I can't get the service name which I want to request in client code. 

In provider side, I can use service name to register instance, but in client side, I can't get the same service name, because I want to use `service name` to find right instance to create `channel` to pass into client.

Can we add `serviceName` property in client side code?

I also check the code, it is easy.

glbrntt:
Yes, I think we can add a similar `var serviceName: String { get }` requirement to the generated client protocol.

glbrntt:
Closed via #1059

PR #1059, opened 2020-11-30
Generate client service name and proto service comments 

### LINKED ISSUE #962: Support .proto file comments for Services
Looks like Methods use the helper `protoSourceComments()` to get comments, but it doesn't look like the Service comments are brought over into the generated sources, likely useful to still have them.

glbrntt:
Motivation:

It can be useful to know the service path a client is for, particularly
when constructing calls manually.

Additionally, we include the methods comments for RPCs but not the
service comments, they may be helpful in some cases.

Modifications:

- Generate the 'serviceName' for the client
- Generate the proto source comments for the client and server

Result:

- Generated code is a touch more useful
- Resolves #962, #1036

### DIFF
--- a/Sources/Examples/Echo/Model/echo.grpc.swift
+++ b/Sources/Examples/Echo/Model/echo.grpc.swift
@@ -25,8 +25,9 @@ import NIO
 import SwiftProtobuf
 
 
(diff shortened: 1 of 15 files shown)


glbrntt:
Changes in 34b78ff

Lukasa (approved):

PR #1061, opened 2020-12-02
Add an interceptors tutorial

### LINKED ISSUE #983: Unfriendly quick start examples: Guide needed to run them

I want to run an example very quickly according these `README` files. However, there is no clear guidance for me.

1. Enter `Sources/Examples/Echo`
2. Go into `client`
3. Run `swift main.swift`

The output looks like:
```
➜  client git:(main) swift main.swift
main.swift:16:8: error: no such module 'GRPC'
import GRPC
       ^
```

### LINKED ISSUE #997: Missing Instruction for RouteGuideModel?
New to Swift here so may be my lack of knowledge but in the example for RouteGuide the client file (and other files) have a "import RouteGuideModel" but other than the import line, nothing else creates or is named "RouteGuideModel"

Where is "RouteGuideModel" or how is it created to import?

ref: https://github.com/grpc/grpc-swift/blob/main/Sources/Examples/RouteGuide/Client/main.swift

glbrntt:
Motivation:

Using interceptors isn't obvious without digging through generated
source code and reading gRPC source. Users shouldn't have to do this to
use a new feature.

Modifications:

- Add an interceptors tutorial
- Add a logging interceptor to the Echo example
- Add an option to the Echo CLI to enable the interceptors
- Add other docs to make running other examples more obvious

Result:

- It's easier to create and use interceptors
- Resolves #983
- Resolves #997

### DIFF
--- a/docs/interceptors-tutorial.md
+++ b/docs/interceptors-tutorial.md
@@ -1,0 +1,298 @@
+# Interceptors Tutorial
+
+This tutorial provides an introduction to interceptors in gRPC Swift. It assumes
+you are familiar with gRPC Swift (if you aren't, try the
+[quick-start guide][quick-start] or [basic tutorial][basic-tutorial] first).
+
+### What are Interceptors?
+
+Interceptors are a mechanism which allows users to, as the name suggests,
+intercept the request and response streams of RPCs. They may be used on the
+client and the server, and any number of interceptors may be used for a single
+RPC. They are often used to provide cross-cutting functionality such as logging,
+metrics, and authentication.
+
+### Interceptor API
+
+Interceptors are created by implementing a subclass of `ClientInterceptor` or
+`ServerInterceptor` depending on which peer the interceptor is intended for.
+Each type is interceptor base class is generic over the request and response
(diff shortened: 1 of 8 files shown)


docs/interceptors-tutorial.md
```diff
@@ -0,0 +1,298 @@
...
+This tutorial provides an introduction to interceptors in gRPC Swift. It assumes
+you are familiar with gRPC Swift (if you aren't, try the
+[quick-start guide][quick-start] or [basic tutorial][basic-tutorial] first).
+
+### What are Interceptors?
+
+Interceptors are a mechanism which allows users to, as the name suggests,
+intercept the request and response streams of RPCs. They may be used on the
+client and the server, and any number of interceptors may be used for a single
+RPC. They are often used to provide cross-cutting such as logging, metrics, and
```
Lukasa, line 12: Should this say "cross-cutting functionality"?

Lukasa (approved):
One minor grammar nit, otherwise looks great.

docs/interceptors-tutorial.md
glbrntt: Indeed it should, thanks for spotting.

PR #1062, opened 2020-12-02
Pass multiplexer from ConnectionManager rather than channel

PeterAdams-A:
Motivation:

The multiplexer is what's actually required and is expensive to
look up from the channel.

Modifications:

Pass the multiplexor through the Idle Handler and the ConnectionManager
to the requester.

Update tests to match.

Result:

Approximately 3% faster on unary calls.

### DIFF
--- a/Sources/GRPC/ConnectionManager.swift
+++ b/Sources/GRPC/ConnectionManager.swift
@@ -556,7 +556,6 @@ internal class ConnectionManager {
         }
         self.logger.debug("scheduling connection attempt", metadata: ["delay_secs": "\(delay)"])
         self.state = .transientFailure(TransientFailureState(from: active, scheduled: scheduled))
-        // candidate mux needs to fail as they didn't sign up to reconnect.
       }
 
     // The channel was ready and working fine but something went wrong. Should we try to replace
--- a/Sources/GRPC/ConnectionManager.swift
+++ b/Sources/GRPC/ConnectionManager.swift
@@ -17,11 +17,10 @@ import Foundation
 import Logging
 import NIO
 import NIOConcurrencyHelpers
+import NIOHTTP2
 
 internal class ConnectionManager {
-  internal struct IdleState {
-    var configuration: ClientConnection.Configuration
-  }
+  internal struct IdleState {}
 
   internal enum Reconnect {
     case none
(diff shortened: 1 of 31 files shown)


Sources/GRPC/ConnectionManager.swift
```diff
@@ -282,51 +281,86 @@ internal class ConnectionManager {
...
+  /// Returns a future for the multiplexer which succeeded when the channel is connected.
+  /// Reconnects are handled if necessary.
+  private func getHTTP2MultiplexerPatient() -> EventLoopFuture<HTTP2StreamMultiplexer> {
+    let multiplexer: EventLoopFuture<HTTP2StreamMultiplexer>
 
-      case let .active(state):
-        channel = state.readyChannelPromise.futureResult
+    switch self.state {
+    case .idle:
+      self.startConnecting()
```
Lukasa, line 303: Would it be easier for `startConnecting` to just return the future you want?

PeterAdams-A: Probably not as in `getHTTP2MultiplexerOptimistic` the one you want is different.  I guess the logic could be pushed in but then the decision point wouldn't match for the other cases.

glbrntt (changes requested):
I have a small thread safety concern and a few nits but this generally looks good.

PeterAdams-A:
I believe I've fixed all your comments George.

PeterAdams-A:
@glbrntt Notice the sneaky extra commit.

glbrntt:
@PeterAdams-A you're hitting an invalid state:

```
Test Case 'GRPCClientKeepaliveTests.testKeepaliveTimeoutFiresBeforeConnectionIsReady' started at 2020-12-03 08:33:59.119
Fatal error: Invalid state connecting(GRPC.ConnectionManager.ConnectingState(backoffIterator: Optional(GRPC.ConnectionBackoffIterator), reconnect: GRPC.ConnectionManager.Reconnect.after(1.0), candidate: NIO.EventLoopFuture<NIO.Channel>, readyChannelMuxPromise: NIO.EventLoopPromise<NIOHTTP2.HTTP2StreamMultiplexer>(futureResult: NIO.EventLoopFuture<NIOHTTP2.HTTP2StreamMultiplexer>))) for fulfillMuxPromiseGivenState(promise:): file /home/travis/build/grpc/grpc-swift/Sources/GRPC/ConnectionManager.swift, line 355
```

Sources/GRPC/ConnectionManager.swift
```diff
@@ -282,51 +279,98 @@ internal class ConnectionManager {
...
+          code: .unavailable,
+          message: "Connection multiplexer requested while backing off"
+        )
+        promise.fail(error)
+      case let .shutdown(state):
+        promise.fail(state.reason)
+      }
+    }
+
     return self.eventLoop.flatSubmit {
```
glbrntt, line 372: Since `getHTTP2Multiplexer` ensures we're on the event loop we no longer need to submit the work to it

Sources/GRPC/ConnectionManager.swift
```diff
@@ -335,34 +379,21 @@ internal class ConnectionManager {
         guard case let .connecting(connecting) = self.state else {
           self.invalidState()
         }
-        channel = connecting.candidate
+        connecting.candidate
+          .whenComplete { _ in fulfillMuxPromiseGivenState(promise: muxPromise) }
```
glbrntt, line 383: I'm not sure if this is correct: the channel future can complete before we receive channel active, i.e. we can still be in the connecting state when fulfilling the promise.

PeterAdams-A: This turned out to be quite a big change involving adding an extra promise to state.

Sources/GRPC/ConnectionManager.swift
```diff
@@ -387,9 +401,10 @@ internal class ConnectionManager {
         shutdown = .shutdownByUser(closeFuture: self.eventLoop.makeSucceededFuture(()))
         self.state = .shutdown(shutdown)
 
-        // Fail the ready channel promise: we're shutting down so even if we manage to successfully
-        // connect the application shouldn't should have access to the channel.
-        state.readyChannelPromise.fail(GRPCStatus(code: .unavailable, message: nil))
+        // Fail the ready channel mux promise: we're shutting down so even if we manage to successfully
+        // connect the application shouldn't should have access to the channel or multiplexer.
```
glbrntt, line 405:
```suggestion
        // connect the application shouldn't have access to the channel or multiplexer.
```

Sources/GRPC/ConnectionManager.swift
```diff
@@ -401,9 +416,9 @@ internal class ConnectionManager {
...
```
glbrntt, line 420:
```suggestion
        // connect the application shouldn't have access to the channel or multiplexer.
```

glbrntt (approved):
This looks good to me, thanks @PeterAdams-A! I added a couple of suggestions to fix pre-existing typos.

glbrntt (approved):

PR #1065, opened 2020-12-04
Special case 0 interceptors construction

PeterAdams-A:
Enumerating empty arrays forces the creation of a new array which isn't necessary.

We can help the compiler by special casing the empty array case.

### DIFF
--- a/Sources/GRPC/Interceptor/ClientInterceptorPipeline.swift
+++ b/Sources/GRPC/Interceptor/ClientInterceptorPipeline.swift
@@ -270,13 +270,19 @@ private extension InterceptorContextList {
     onRequestPart: @escaping (GRPCClientRequestPart<Request>, EventLoopPromise<Void>?) -> Void,
     onResponsePart: @escaping (GRPCClientResponsePart<Response>) -> Void
   ) where Element == ClientInterceptorContext<Request, Response> {
-    let middle = interceptors.enumerated().map { index, interceptor in
-      ClientInterceptorContext(
-        for: .userProvided(interceptor),
-        atIndex: index,
-        in: pipeline
-      )
-    }
+    let middle = { () -> [Element] in
+      if interceptors.count == 0 {
+        return [] // Special case 0 to reduce cost.
+      } else {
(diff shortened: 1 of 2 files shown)


Sources/GRPC/Interceptor/ClientInterceptorPipeline.swift
```diff
@@ -270,13 +270,19 @@ private extension InterceptorContextList {
...
+      } else {
+        return interceptors.enumerated().map { index, interceptor in
+          ClientInterceptorContext(
+            for: .userProvided(interceptor),
+            atIndex: index,
+            in: pipeline
+          )
+        }
+      }
+    }()
```
Lukasa, line 285: Why is this a self-invoking block?

PeterAdams-A: Because I keep forgetting how smart swift is - changed it.

Sources/GRPC/Interceptor/ClientInterceptorPipeline.swift
```diff
@@ -270,12 +270,17 @@ private extension InterceptorContextList {
...
     onResponsePart: @escaping (GRPCClientResponsePart<Response>) -> Void
   ) where Element == ClientInterceptorContext<Request, Response> {
-    let middle = interceptors.enumerated().map { index, interceptor in
-      ClientInterceptorContext(
-        for: .userProvided(interceptor),
-        atIndex: index,
-        in: pipeline
-      )
+    let middle: [Element]
+    if interceptors.count == 0 {
```
glbrntt, line 274: nit: use `isEmpty` rather than comparing the count

glbrntt (changes requested):
I'm a little surprised there's a noticeable difference here. Curious whether you tried with a plain for-loop and reserving capacity directly?

Lukasa:
The difference in the case of 0 comes from the fact that `map` over an empty array bafflingly allocates a new one.

glbrntt:
This change is now redundant; this type no longer exists.

PR #1068, opened 2020-12-08
Extract the logic from the GRPCIdleHandler into a state machine

glbrntt:
Motivation:

Shutting down gracefully is generally a better option than slamming the
door in the remote peer's face. To that end we need to do a better job
of handling GOAWAY frames (both receiving and sending them). At the
moment these are received by the 'GRPCIdleHandler' and used as an
indication to close the connection. Since we also have to deal with
various client connection management state the handler has grown a
little out of hand.

Modifications:

- Pull the state machine out of the idle handler, add support for
  quiescing (to the state machine, not to the handler)

Result:

- Idle handler is way easier to follow


glbrntt:
I should add that PR doesn't add support for shutting down gracefully: it just paves the way for it.

Lukasa (approved):
LGTM, seems like a good refactor.

PR #1073, opened 2020-12-09
Remove a bunch of (unspecialized) generics

glbrntt:
Motivation:

The state machine code in the BaseCallHandler works returns an 'Action'
to perform as a result of prodding the state machine. These actions are
generic over the request and response type of the RPC as the often
include instructions like 'forward this request to the interceptors' and
so on. When these generics are unspecialized we burn a bunch of
instructions in the runtime, and in many cases they can be represented
without generics entirely.

Modifications:

- Shuffle some of the state machine code around in BaseCallHandler such
  that neither State or Action are generic

Result:

- Instructions executed in the unary server benchmark decreases by ~20%


Lukasa (approved):
LGTM, nice patch!

PR #1074, opened 2020-12-09
Add a unary server benchmark

glbrntt:
Motivation:

Benchmarks are useful for measuring performance. We don't currently have
any benchmarks looking at just the server.

Modifications:

- Add a benchmark to measure unary RPCs on an EmbededChannel on the
  server.

Result:

- More benchmarks.


Lukasa (approved):
Shiny

PR #1076, opened 2020-12-10
Allow the server to be gracefully shutdown.

glbrntt:
Motivation:

It should be possible to shutdown a server in a way which allows
existing connection to run to completion.

Modifications:

- Add new API to allow servers to be initiate a graceful shutdown.
- Add new API to the client connectivity delegate so that users can be
  informed that the connection is going away
- Fixed a bug in the idle state machine that sent a GOAWAY frame too soon
  in response to receiving GOAWAY frame

Result:

Servers can be shutdown gracefully


Lukasa (approved):
Really nice patch, I like this a lot.

PR #1078, opened 2020-12-11
Remove dead code

glbrntt:
Motivation:

There's no point carrying dead code.

Modification:

- Remove dead code

Result:

Less code.


Lukasa (approved):
-1300 lines? This is a good patch.

PR #1079, opened 2020-12-15
Merge idle and keepalive handlers

glbrntt:
Motivation:

Both client and the server have two handlers between the two HTTP/2
handlers which inspect various HTTP/2 frames and stream creation/close
events. Each handler has to unwrap NIOAny, which isn't special-cased for
HTTP2Frame, and cast any user inbound events to check whether they are
events they're interested in. This has is duplicated work which we can easily
avoid on such a hot path.

Modifications:

- Move the ping handler into the idle handler
- Update tests

Result:

A 4% drop in instructions in the unary_10k_small_requests benchmark.


Lukasa (approved):
LGTM.

PR #1080, opened 2020-12-15
Simplify client transport actions and state representation

glbrntt:
Motivation:

The representation of state and any outcomes as a result of changing
state in 'ClientTransport' is generic over the request and response
types. This turns out to be a non-trivial performance hit.

Modifications:

- Split up state actions into per-function actions
- Move state out of 'ClientTransport'

Result:

- A 5% decrease in instruction for the 'unary_10k_small_requests'
  benchmark (note that this change is client only and yet the benchmark
  includes instructions for client and server).


Lukasa (approved):
Straightforward refactor, looks good.

PR #1081, opened 2020-12-16
Make the remote address available on the server interceptor context

glbrntt:
Motivation:

Sometimes, if you're an interceptor, it's useful to know who you're
talking too.

Modifications:

- Make the 'remoteAddress' available on the 'ServerInterceptorContext'

Result:

Users can access the remote address from server interceptors.

### DIFF
--- a/Sources/GRPC/HTTP2ToRawGRPCServerCodec.swift
+++ b/Sources/GRPC/HTTP2ToRawGRPCServerCodec.swift
@@ -111,6 +111,7 @@ internal final class HTTP2ToRawGRPCServerCodec: ChannelDuplexHandler {
         headers: payload.headers,
         eventLoop: context.eventLoop,
         errorDelegate: self.errorDelegate,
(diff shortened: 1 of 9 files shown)


Sources/GRPC/HTTP2ToRawGRPCServerCodec.swift
```diff
@@ -111,6 +111,7 @@ internal final class HTTP2ToRawGRPCServerCodec: ChannelDuplexHandler {
         headers: payload.headers,
         eventLoop: context.eventLoop,
         errorDelegate: self.errorDelegate,
+        remoteAddress: context.channel.remoteAddress,
```
glbrntt, line 114: @Lukasa given this is in `channelRead` is it safe to force unwrap this optional?

Lukasa: No, not all channels can answer this question.

Lukasa (approved):

PR #1090, opened 2021-01-06
Avoid copying server interceptor context

glbrntt:
Motivation:

The server interceptor pipeline holds an array of interceptor contexts.
Each context is a struct and also holds a reference to the pipeline.
Each time the context asks the pipeline for the next inbound/outbound
context we incur a copy of the context which is surprisingly costly due
to retain and release traffic for the pipeline ref as well as the cost
of initializing the unspecialized contexts.

Modifications:

- interceptor contexts now ask the pipeline to perform an operation for
  them rather than asking for a context on which to perform the
  operation, this avoids copying contexts
- as the pipeline is now responsible for performing the operations a
  further optimisation is opened up: head and tail interceptors and
  contexts are no longer required, the same action can be done directly
  by the pipeline

Result:

- An 11.2% reduction in instructions in the
  embedded_server_unary_10k_small_requests benchmark


Sources/GRPC/Interceptor/ServerInterceptorPipeline.swift
```diff
@@ -42,56 +42,41 @@ internal final class ServerInterceptorPipeline<Request, Response> {
...
+  /// part has traversed the pipeline.
+  @usableFromInline
+  internal var _isOpen = true
+
+  /// The index of the next context on the inbound side of the context at the given index.
   @inlinable
-  internal var head: ServerInterceptorContext<Request, Response>? {
-    return self._contexts?.first
+  internal func _nextInboundIndex(after index: Int) -> Int {
+    return index &+ 1
```
Lukasa, line 73: Is it worth putting asserts on this unchecked arithmetic? I agree that this _practically speaking_ cannot overflow, but it might be nice to guard against it in debug mode.

Sources/GRPC/Interceptor/ServerInterceptorPipeline.swift
```diff
@@ -128,8 +122,57 @@ internal final class ServerInterceptorPipeline<Request, Response> {
...
+      self._invokeReceive(
+        part,
+        onContextAtUncheckedIndex: self._nextInboundIndex(after: self._headIndex)
+      )
+
+    case self._tailIndex:
+      self._onRequestPart(part)
+
+    default:
+      self._userContexts[index].interceptor.receive(part, context: self._userContexts[index])
```
Lukasa, line 174: This seems like a slightly tricky spelling: I don't love that we have to subscript the array twice. Can this be pushed into a method on the context?

Sources/GRPC/Interceptor/ServerInterceptorPipeline.swift
```diff
@@ -140,50 +183,80 @@ internal final class ServerInterceptorPipeline<Request, Response> {
...
-      for: .tail(onRequestPart),
-      atIndex: middle.endIndex,
-      in: pipeline
-    )
+    default:
+      self._userContexts[index].interceptor.send(
+        part,
+        promise: promise,
+        context: self._userContexts[index]
+      )
```
Lukasa, line 251: Same question here.

Sources/GRPC/Interceptor/ServerInterceptorPipeline.swift
glbrntt: No harm in it, I added a comment on why it's okay as well.

Lukasa (approved):
Shiny, I love it.

PR #1093, opened 2021-01-13
Scaffolding for better server codegen

glbrntt:
Motivation:

Most of the RPC handling code on the server is generic but not
specialized; as a result we pay a fairly large performance cost. This is
the first in a series of PRs which changes how the shape of the
generated server code.

The plan is to have a new protocol between the generated code and gRPC,
'GRPCServerHandlerProtocol', which operates on concrete types
('ByteBuffer' as opposed to typed request/responses). As such the
protocol for asking for a handler ('CallHandlerProvider') will also
change to return an instance of this new protocol. In generated code the
returned handler will be generic over the request deserializer and
response serializer. The responsibility of the call handler will not
differ much from now (it will receive headers, serialized bytes, etc.)
but will be implemented as a class per call type (this is close to the
current situation where the same is true but a common base class is
shared). In addition the new handlers will not channel handlers: they
will eventually be held by the routing handler.

Modifications:

- Add 'GRPCServerHandlerProtocol', the inbound protocol that the new
  handlers will conform to
- Add 'GRPCServerResponseWriter', the internal outbound counterpart to
  'GRPCServerHandlerProtocol' used by the handlers to communicate back
  to gRPC and add an unimplemented conformance of this to
  'HTTP2ToRawGRPCServerCodec'.
- Add 'UnaryServerHandler' (and tests), a handler for unary requests
- Add 'ServerErrorProcessor'; cargo culted from '_BaseCallHandler'
- Modify the 'CallHandlerProvider' to add a 'handle(method:context:)'

Result:

We have a rough outline of how the new server handlers will look.

### DIFF
--- a/Sources/GRPC/CallHandlers/ServerHandlerProtocol.swift
+++ b/Sources/GRPC/CallHandlers/ServerHandlerProtocol.swift
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2021, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


Lukasa (approved):
Cool, this seems like a really good start on this work.

PR #1098, opened 2021-01-15
Add a bidirectional streaming server handler.

glbrntt:
Motivation:

Following from #1093, #1095 and #1097; we need a bidirectional streaming
server handler.

Modifications:

- Add a bidirectional streaming server handler and tests.

Result:

We'll be able to do bidirectional streaming RPCs with new codegen.


Lukasa (approved):
Shiny, LGTM.

PR #1100, opened 2021-01-15
Simplify some of the interaction between HTTP2 to GRPC state machine …

glbrntt:
…and codec

Motivation:

The HTTP2ToRawGRPCStateMachine returns a somewhat generic 'Action' which
the called is meant to act on. This pattern was used wholesale for every
interaction with the statemachine. This is a little silly: in some cases
the cases any action can be specified just fine with a `Bool`, in others
a `Result` is fine. The other downside to this approach is that it makes
it harder to see at the callsite what could happen as a result of
interacting with the state machine.

Modifications:

- Specify an action to take for each different interaction with the state
  machine.
- Update tests.

Result:

A handful of reductions in instructions in benchmarks:
- unary_10k_small_requests: -0.5%
- embedded_server_unary_10k_small_requests: -2.6%
- embedded_server_client_streaming_1_rpc_10k_small_requests: -4.0%
- embedded_server_client_streaming_10k_rpcs_1_small_requests: -2.6%
- embedded_server_server_streaming_1_rpc_10k_small_responses: -1.7%
- embedded_server_server_streaming_10k_rpcs_1_small_response: -2.6%
- embedded_server_bidi_1_rpc_10k_small_requests: -3.1%
- embedded_server_bidi_10k_rpcs_1_small_request: -2.4%

### DIFF
--- a/Sources/GRPC/HTTP2ToRawGRPCServerCodec.swift
+++ b/Sources/GRPC/HTTP2ToRawGRPCServerCodec.swift
@@ -28,6 +28,7 @@ internal final class HTTP2ToRawGRPCServerCodec: ChannelDuplexHandler, GRPCServer
   private var logger: Logger
   private var state: HTTP2ToRawGRPCStateMachine
   private let errorDelegate: ServerErrorDelegate?
+  private var context: ChannelHandlerContext!
 
   init(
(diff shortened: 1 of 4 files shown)


Lukasa (approved):
Shiny, go for it.

PR #1107, opened 2021-01-19
Added support for NIOSSLCustomVerificationCallback for client connection

### LINKED ISSUE #723: Certificate pinning with NIO?
### Question Checklist

### Question Subject

Does the NIO implementation of Swift GRPC support certificate pinning? If so, how do I use it?

#### Question Description

I should preface this by saying I only have a basic grasp on certificate pinning and TSL in general. My understanding of it is as follows: instead of trusting any certificate signed by any of the default trust roots provided by my OS, certificate pinning means that I specify which specific certificate I want to trust, and any connection signed by any other certificate should fail to authenticate.

I have tried the following in order to use this:

**1. Got the certificate from our server**

I ran
```
openssl s_client -servername <servername> -host <hostname> -port <port> -showcerts < /dev/null
```

which yielded 2 certificates.

**2. Instantiated the certificates in code**

```
let certificateString1 = """
        -----BEGIN CERTIFICATE-----
        ...
        -----END CERTIFICATE-----
        """

let certificateString2 = """
        -----BEGIN CERTIFICATE-----
        ...
        -----END CERTIFICATE-----
        """

func getCertificate(for certificateString: String) -> NIOSSLCertificate? {
        do {
            let bytes: [Byte] = certificateString.utf8.map { $0 }
            let certificate = try NIOSSLCertificate(bytes: bytes, format: .pem)
            return certificate
        } catch let error {
            print(error)
            return nil
        }
}

let certificate1 = getCertificate(for: certificateString1)!
let certificate2 = getCertificate(for: certificateString2)!
let certificates = [certificate1, certificate2]

let certificateSource1 = NIOSSLCertificateSource.certificate(certificate1)
let certificateSource2 = NIOSSLCertificateSource.certificate(certificate2)
let certificateSources = [certificateSource1, certificateSource2]
```

**3. Created some different TLS configurations**

Configuration A:
- My certificates given as the chain
- Empty trust roots

```
let tlsA = ClientConnection.Configuration.TLS(
        certificateChain: certificateSources,
        trustRoots: .certificates([]))
```

Configuration B:
- Empty certificate chain
- My certificates given as the trust roots

```
let tlsB = ClientConnection.Configuration.TLS(
        certificateChain: [],
        trustRoots: .certificates(self.certificates))
```

Configuration C:
- My certificates given as the chain
- My certificates given as the trust roots

```
let tlsC = ClientConnection.Configuration.TLS(
        certificateChain: certificateSources,
        trustRoots: .certificates(self.certificates))
```

Configuration D:
- Using the default certificate chain
- My certificates given as the trust roots

```
let tlsD = ClientConnection.Configuration.TLS(
        trustRoots: .certificates(self.certificates))
```

Configuration E:
- Using my certificates as the chain
- Using the default trust roots

```
let tlsE = ClientConnection.Configuration.TLS(
        certificateChain: certificateSources)
```

**4. Created and used a client connection with each of these TLS configurations.**

I won't post the code here, but I will note that this part works fine if I do not attempt to set up certificate pinning; when I just pass `tls: .init()` to my client connection configuration, my endpoint calls authenticate correctly and everything. Things only start breaking once I use one of the configurations above.

**5. Results**

With configurations A-D, I get errors like these:

Errors:
```
2020-02-14T09:07:46+0200 error: error=handshakeFailed(NIOSSL.BoringSSLError.sslError([Error: 268435581 error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED])) error caught before TLS was verified
2020-02-14T09:07:46+0200 error: error=handshakeFailed(NIOSSL.BoringSSLError.sslError([Error: 268435581 error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED])) grpc client error
2020-02-14T09:07:46+0200 notice: connection_id=21C06D0C-1C94-4903-BCDF-8BE8B6DD86D2 error=handshakeFailed(NIOSSL.BoringSSLError.sslError([Error: 268435581 error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED])) connection attempt failed
```

When I use configuration E, the call succeeds; however, when I change the certificate (by substituting some of the caracters with `0`s), the call *still* succeeds, which indicates that my certificates are not being used for authentication at all. This makes sense; the certificates are signed by trusted CAs, so using the default trust roots would cause my connection to succeed.

**Overview**

My own inexperience with TLS makes it difficult for to pinpoint whether my problems originate from my own code, the Swift GRPC / NIO SSL libraries, or the server. So it would be extremely helpful if someone could tell me whether Swift GRPC supports certificate pinning, and if so, how to use it.

### LINKED ISSUE #1104: Public key pinning support in grpc-swift

SSL Pinning can be achieved using Certificate pinning or public key pinning. I found there is support for  [certificate pinning](https://medium.com/@ambrose12silveira/ios-swift-grpc-integration-with-tls-client-authentication-f2e2164ed125) in grpc-swift but it has a downside that if the server rotates its certificate on a regular basis, then the application would need to be updated regularly.  

Is there any support for **public key pinning** in grpc-swift?

franck-clement-ug:
Hi,

This PR adds the ability for client connections to override the certificate verification logic implemented in nio-ssl.

The main goal is to allow client apps to perform their own certificate verification logic, such as SSL public key pinning.
Nio-ssl provide this feature as described here: [nio-ssl-issue-25](https://github.com/apple/swift-nio-ssl/issues/25) 
But I wasn't able to use it through grpc-swift because the `NIOSSLCustomVerificationCallback` is not accessible when initializing a secure `ClientConnection`.

I exposed a `NIOSSLCustomVerificationCallback` property in `ClientConnection.Configuration` struct and provided a `ClientConnection.Builder` func as well.

## Usage example for client public key pinning

```swift
let myPublicKey = Array(Data(base64Encoded: """
.... (Your base64 encoded public key)
""", options: .ignoreUnknownCharacters)!)

ClientConnection.secure(group: PlatformSupport.makeEventLoopGroup(loopCount: 1))
            .withTLSCustomVerificationCallback({ cert, eventLoopVerification in
                // Extract the leaf certificate's public key,
                // then compare it to the one you have.
                if let leaf = cert.first,
                   let publicKey = try? leaf.extractPublicKey() {

                    if let certSPKI = try? publicKey.toSPKIBytes(),
                       myPublicKey == certSPKI {
                        eventLoopVerification.succeed(.certificateVerified)
                    } else {
                        eventLoopVerification.fail(NIOSSLError.unableToValidateCertificate)
                    }
                } else {
                    eventLoopVerification.fail(NIOSSLError.noCertificateToValidate)
                }
            })
            .connect(host: host, port: port)
```

This PR + the example above should resolve #723 and resolve #1104

### DIFF
--- a/Sources/GRPC/ClientConnection.swift
+++ b/Sources/GRPC/ClientConnection.swift
@@ -312,6 +312,9 @@ extension ClientConnection {
       return self.tls == nil ? .http : .https
     }
 
+    /// A custom verification callback that allows completely overriding the certificate verification logic for this connection.
+    public var customVerificationCallback: NIOSSLCustomVerificationCallback?
+
     /// A logger for background information (such as connectivity state). A separate logger for
     /// requests may be provided in the `CallOptions`.
     ///
@@ -362,6 +365,7 @@ extension ClientConnection {
         label: "io.grpc",
         factory: { _ in SwiftLogNoOpLogHandler() }
       ),
+      customVerificationCallback: NIOSSLCustomVerificationCallback? = nil,
       debugChannelInitializer: ((Channel) -> EventLoopFuture<Void>)? = nil
     ) {
       self.target = target
@@ -411,7 +415,8 @@ extension Channel {
     connectionIdleTimeout: TimeAmount,
     errorDelegate: ClientErrorDelegate?,
     requiresZeroLengthWriteWorkaround: Bool,
-    logger: Logger
+    logger: Logger,
+    customVerificationCallback: NIOSSLCustomVerificationCallback?
   ) -> EventLoopFuture<Void> {
     // We add at most 8 handlers to the pipeline.
     var handlers: [ChannelHandler] = []
@@ -426,16 +431,31 @@ extension Channel {
     #endif
 
     if let tlsConfiguration = tlsConfiguration {
-      do {
-        let sslClientHandler = try NIOSSLClientHandler(
-          context: try NIOSSLContext(configuration: tlsConfiguration),
-          serverHostname: tlsServerHostname
-        )
-        handlers.append(sslClientHandler)
-        handlers.append(TLSVerificationHandler(logger: logger))
-      } catch {
-        return self.eventLoop.makeFailedFuture(error)
-      }
+        if let customVerificationCallback = customVerificationCallback {
+            do {
+                let sslClientHandler = try NIOSSLClientHandler(
+                    context: try NIOSSLContext(configuration: tlsConfiguration),
+                    serverHostname: tlsServerHostname,
+                    customVerificationCallback: customVerificationCallback
+                )
+                handlers.append(sslClientHandler)
+                handlers.append(TLSVerificationHandler(logger: logger))
+            } catch {
+                return self.eventLoop.makeFailedFuture(error)
+            }
+        } else {
+            do {
+                let sslClientHandler = try NIOSSLClientHandler(
+                    context: try NIOSSLContext(configuration: tlsConfiguration),
+                    serverHostname: tlsServerHostname
+                )
+                handlers.append(sslClientHandler)
+                handlers.append(TLSVerificationHandler(logger: logger))
+            } catch {
+                return self.eventLoop.makeFailedFuture(error)
+            }
+        }
+
     }
 
     // We could use 'configureHTTP2Pipeline' here, but we need to add a few handlers between the
--- a/Sources/GRPC/ClientConnection.swift
+++ b/Sources/GRPC/ClientConnection.swift
@@ -312,9 +312,6 @@ extension ClientConnection {
       return self.tls == nil ? .http : .https
     }
 
-    /// A custom verification callback that allows completely overriding the certificate verification logic for this connection.
-    public var customVerificationCallback: NIOSSLCustomVerificationCallback?
-
     /// A logger for background information (such as connectivity state). A separate logger for
     /// requests may be provided in the `CallOptions`.
     ///
@@ -365,7 +362,6 @@ extension ClientConnection {
         label: "io.grpc",
         factory: { _ in SwiftLogNoOpLogHandler() }
       ),
-      customVerificationCallback: NIOSSLCustomVerificationCallback? = nil,
       debugChannelInitializer: ((Channel) -> EventLoopFuture<Void>)? = nil
     ) {
       self.target = target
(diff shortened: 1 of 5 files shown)


Sources/GRPC/ClientConnection.swift
```diff
@@ -312,6 +312,9 @@ extension ClientConnection {
       return self.tls == nil ? .http : .https
     }
 
+    /// A custom verification callback that allows completely overriding the certificate verification logic for this connection.
+    public var customVerificationCallback: NIOSSLCustomVerificationCallback?
```
glbrntt, line 316: Since this relates to TLS it should be on the `ClientConnection.Configuration.TLS` struct rather than the general configuration one.

Sources/GRPC/ClientConnection.swift
```diff
@@ -426,16 +431,31 @@ extension Channel {
...
-        let sslClientHandler = try NIOSSLClientHandler(
-          context: try NIOSSLContext(configuration: tlsConfiguration),
-          serverHostname: tlsServerHostname
-        )
-        handlers.append(sslClientHandler)
-        handlers.append(TLSVerificationHandler(logger: logger))
-      } catch {
-        return self.eventLoop.makeFailedFuture(error)
-      }
+        if let customVerificationCallback = customVerificationCallback {
```
glbrntt, line 434: There's a lot of duplicated code here: let's move the if-else inside the do-catch so that we don't duplicate the error handling or adding of handlers.

Sources/GRPC/GRPCChannel/GRPCChannelBuilder.swift
```diff
@@ -273,6 +273,16 @@ extension ClientConnection.Builder {
   }
 }
 
+extension ClientConnection.Builder {
```
glbrntt, line 276: This should be an extension on `ClientConnection.Builder.Secure`: it doesn't apply to non-secure connections.

Sources/GRPC/GRPCChannel/GRPCChannelBuilder.swift
```diff
@@ -273,6 +273,16 @@ extension ClientConnection.Builder {
...
+
+  /// A custom verification callback that allows completely overriding the certificate verification logic.
+  @discardableResult
+    public func withCustomVerification(callback: @escaping NIOSSLCustomVerificationCallback) -> Self {
```
glbrntt, line 280: This naming isn't clear at the call site:

```swift
builder.withCustomVerification { 
  // ...
}
```

I suggest: 

```swift
public func withTLSCustomVerificationCallback(_ callback: @escaping NIOSSLCustomVerificationCallback) -> Self {
  ...
}
```

glbrntt (changes requested):
Thanks for the PR, I've left some initial comments inline. You'll need to reformat the code to be consistent with the rest of the project (running `./scripts/format.sh` should do it). We'll also need a test for this; you should be able to use the `ClientTLSFailureTests` as a starting point.

franck-clement-ug:
@glbrntt Thank you for the detailed comments, they all make a lot of sense, and I fixed them. 
I also added a test when the validation logic is forced to fail within the `customVerificationCallback`.

Sources/GRPC/GRPCChannel/GRPCChannelBuilder.swift
```diff
@@ -235,6 +235,16 @@ extension ClientConnection.Builder.Secure {
     self.tls.certificateVerification = certificateVerification
     return self
   }
+
+  /// A custom verification callback that allows completely overriding the certificate verification logic.
+  @discardableResult
+  public func withCustomVerificationCallback(
```
glbrntt, line 241: This should be `withTLSCustomVerificationCallback`: without `TLS` it's unclear what is being verified.

Sources/GRPC/GRPCChannel/GRPCChannelBuilder.swift
```diff
@@ -235,6 +235,16 @@ extension ClientConnection.Builder.Secure {
...
+    _ callback: @escaping NIOSSLCustomVerificationCallback
+  )
+    -> Self {
```
glbrntt, line 244:
```suggestion
  ) -> Self {
```

glbrntt (changes requested):
Thanks @franck-clement-ug this looks good, just one naming nit to fix.

glbrntt (approved):
This looks good to me -- assuming CI passes -- thanks @franck-clement-ug!

PR #1110, opened 2021-01-21
Emit fewer flushes when handling RPCs on the server

glbrntt:
Motivation:

The changes in #1101 included a change in when flushes were emitted.
For unary RPCs this led to an additional flush, which in some
circumstances led to a drop in QPS of approximately 5%.

Modifications:

- Introduce an option as to whether a flush should be emitted when
  writing initial response metadata
- Fix a bug where the request to flush or not for messages was ignored

Result:

A 5% increase in QPS for the unary-single benchmark (single connection,
one rpc at a time).

### DIFF
--- a/Sources/GRPC/HTTP2ToRawGRPCServerCodec.swift
+++ b/Sources/GRPC/HTTP2ToRawGRPCServerCodec.swift
@@ -170,7 +170,9 @@ internal final class HTTP2ToRawGRPCServerCodec: ChannelDuplexHandler, GRPCServer
 
     switch responsePart {
     case let .metadata(headers):
-      self.sendMetadata(headers, promise: promise)
+      // We're in 'write' so we're using the old type of RPC handler which emits its own flushes,
+      // no need to emit an extra one.
+      self.sendMetadata(headers, flush: false, promise: promise)
 
     case let .message(buffer, metadata):
       self.sendMessage(buffer, metadata: metadata, promise: promise)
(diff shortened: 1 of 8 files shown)


Sources/GRPC/HTTP2ToRawGRPCServerCodec.swift
```diff
@@ -335,15 +326,21 @@ internal final class HTTP2ToRawGRPCServerCodec: ChannelDuplexHandler, GRPCServer
...
     }
   }
+
+  /// Mark a flush as pending - to be emitted once the read completes - if we're currently reading,
+  /// or emit a flush now if we are not.
+  private func markFlushPoint() {
+    if self.isReading {
+      self.flushPending = true
+    } else {
+      self.context.flush()
```
Lukasa, line 343: Don't we need to unset `flushPending` here?

glbrntt: I don't think it's strictly necessary; we'll only set it when we're reading and we'll unset it before we flush in channel read complete. That said, that seems a little brittle and there's no harm in unsetting it here too.

Lukasa: For that matter, do we do anything with `flushPending` at all?

Lukasa: Oh never mind, I'm clearly looking in the wrong place.

Lukasa (approved):

PR #1115, opened 2021-01-26
Remove deprecated/dead code

glbrntt:
Motivation:

We made no commitment to carry API which was deprecated before 1.0.0, so
let's remove all shims, deprecated and unreachable code.

Modifications:

- Remove shims
- Remove some call contexts, since they're for use with the old RPC
  handling

Result:

Less code.


Lukasa (approved):
Love deleting code, world's best patch, we can retire happy now.

PR #1119, opened 2021-01-29
Move the pcap example into Source/Examples/

glbrntt:
Motivation:

The PCAP example depends on swift-nio-extras. Since grpc-swift now
depends on swift-nio-extras (and we have target based dependency
resolution) the example can live in-source with other examples.

Modifications:

- Move Examples/PCAPExample to Source/Examples/PCAP.
- Use swift-argument-parser
- Make it a target in the GRPC package, rather than its own package.

Result:

Easier to avoid bit-rot.


Package.swift
```diff
@@ -257,5 +257,18 @@ let package = Package(
       ],
       path: "Sources/Examples/RouteGuide/Server"
     ),
+
+    // Client for the PCAP example
+    .target(
+      name: "PCAP",
```
Lukasa, line 263: Does this target need a clearer name?

Lukasa (approved):
Basically looks good to me, one question about the target name.

Package.swift
glbrntt: Sure does. Changed it to 'PacketCapture'

Package.swift
```diff
@@ -257,5 +257,18 @@ let package = Package(
       ],
       path: "Sources/Examples/RouteGuide/Server"
     ),
+
+    // Client for the PacketCapture example
+    .target(
+      name: "PacketCapture",
```
weissi, line 263: shouldn't all targets be `GRPC` prefixed? Targets need to be unique across the whole binary (so all dependencies' targets must be globally unique)

glbrntt: Isn't that only necessary for targets which are also products?

weissi: @glbrntt no, there can only be _one module_ of a given name in the whole binary. Some people call it the `Utilities` module problem.

Lukasa: Sure, but we have target specific dependencies right? So this module won't _be_ in the product, because it's an executable target.

weissi: And I think in your particular case it will appear to work. But I think there's some weird issue when you do `swift package edit` where then even modules that aren't officially exported at all start to clash.

Lukasa: Put another way, it's a leaf.

weissi: @Lukasa I _think_ it's still an issue with `swift package edit` but let me try to create a repo.

weissi: there we go: https://github.com/apple/swift-nio-ssl/commit/f9ac0c3486f8e9333e791caf135e3fe067e0e66c

weissi: So I think the repro is: Create a package `C` that depends on `A` and `B`. Make `A` and `B` both have a target called `Target`. After that, in `C`, do `swift package edit A && swift package edit B`. And I think now you'll get the error :)

Lukasa: Well that's stupid. Is there a Swift bug report for this?

weissi: @Lukasa I don't think I filed one but this is a long time ago.

glbrntt: @weissi I can't repro that. Plus, I thought the reasoning for the package description API in 5.2 where packages are explicitly named was to avoid this problem?

weissi: @glbrntt that _is_ possible. Maybe it's just a problem in < 5.2? Tbh I don't fully understand what's going on but it used to break the release tools. But we do test with 5.0 there...

glbrntt: Yeah, I can totally believe that it didn't work before 5.2. Any objections to leaving the naming as is @weissi?

weissi: @glbrntt go for it. We can fix that at any time if it causes trouble.

PR #1121, opened 2021-01-29
Fix up some docs, add an faqs doc

glbrntt:
Motivation:

Documentation is -- generally speaking -- useful. More so when it's
correct.

Modifications:

- Add some missing documentation. There's still more to go but it's a
  start.
- Add an 'faqs' doc which answers some questions which aren't necessarily
  obvious from reading source documentation. It's in no way
  comprehensive and is intended to be built upon over time.
- Remove some dead code.

Result:

Better docs, hopefully.


docs/faqs.md
```diff
@@ -0,0 +1,136 @@
...
+  supplied which logs information at the connection level, relating to things
+  such as connectivity state changes, or connection errors.
+- `CallOptions` allows `logger` to be specified. It is used to log information
+  at the RPC level, such as changes to the state of the RPC. Any connection
+  level metadata will be attached to the logger so that RPC logs may be
+  correlated with connection level changes.
+
+For the server:
+
+- `Server.Builder.withLogger` allows a `logger` to be specified. It is used a
```
Lukasa, line 22:
```suggestion
- `Server.Builder.withLogger` allows a `logger` to be specified. It is used as a
```

Lukasa (approved):

Davidde94 (approved):
Looks really solid, especially the FAQ.

PR #1123, opened 2021-02-02
Add a codegen option to allow stub names to not have their first character lowercased

glbrntt:
Motivation:

Protobuf suggests that RPC methods should always start with an uppercase
character. Function names in Swift typically start with a lowercase
character. As such, by default, protoc-gen-grpc-swift will lowercase the
first character.

Having an uppercase RPC method is by no means a requirement and since
routing is not case sensitive, defining two methods 'Foo' and 'foo'
within the same service will lead to compilation errors since both
generated Swift function names will be 'foo'.

Modifications:

- Add an escape hatch to the plugin to allow for method casing to be
  kept, this defaults to false
- Add a test service using these options
- Update docs

Result:

Users can generate code which dos not alter the casing of the RPC name


Lukasa (approved):

glbrntt:
(Merging over the test failure; it is unrelated to this change.)

PR #1128, opened 2021-02-03
Add a 'version' identifier and a more specific user-agent

glbrntt:
Motivation:

User-agent string should, ideally, contain a version number.

Modifications:

- Add a 'Version' struct
- Use the version string in the user-agent
- Only check user-provided headers for a user-agent string rather than
  all headers

Result:

- User-agent string includes version number

### DIFF
--- a/Sources/GRPC/Version.swift
+++ b/Sources/GRPC/Version.swift
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2021, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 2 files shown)


Sources/GRPC/Version.swift
```diff
@@ -0,0 +1,29 @@
...
+
+internal struct Version {
+  /// The major version.
+  internal static let major = 1
+
+  /// The minor version.
+  internal static let minor = 0
+
+  /// The patch version.
+  internal static let patch = 0
```
Lukasa, line 25: How are we going to enforce updating this?

Lukasa, line 25: An extra question: should this also contain any version qualifiers, e.g. `-alpha24`?

Lukasa (approved):
Nothing wrong with the patch as-it-stands, though I have a few questions in the diff.

Sources/GRPC/Version.swift
glbrntt: I was also wondering about this. I'm not really sure at the moment because we have to update this before tagging a release, and anything that automates this check needs to know when a release is happening first.

One idea was to validate in CI against git tags, of course that only works after-the-fact so is of limited use, you can at least update it in a patch release straight after tough if the tag was wrong.

A release checklist has the problem as we have now the same as now: you need to remember to look at it. 

The other idea I had was to build a check into the release notes drafter since I always run that before a release, definitely a hack though.

Any other ideas?

glbrntt: I'm being optimistic about what the next version is!

Lukasa: I think that hack may be what we need to do for now, sadly.

PR #1139, opened 2021-02-16
Better handle DATA frames with end stream set

### LINKED ISSUE #1136: Client doesn't handle DATA frames with end stream set gracefully
### Describe the bug

If the client receives a DATA frame with the end stream flag set then the RPC will fail with a "Transport became inactive" error; while that's true -- the transport _did_ become inactive -- it's not really representative of the actual problem: the remote peer is not conforming to the spec; it should be sending trailers with the end stream flag set.

### Expected behaviour

The RPC should fail with an accurate error message.

glbrntt:
Motivation:

If a peer doesn't follow spec and sends the client a DATA frame with end
stream set, the client will ignore the fact that end stream was set.
When the stream channel closes just after, the call is unaware of why it
was closed (because it hasn't received a final status) which leads to a
pretty non-descript error message.

Modifications:

Add a 'receiveEndOfResponseStream' to the client state machine which
produces a 'GRPCStatus' if the end stream is unexpected and should not
be ignored.

Result:

Calls will terminate with an 'internal error' status and a message about
a protocol violation if the server sends a data frame with end stream
set.

Resolves #1136

### DIFF
--- a/Sources/GRPC/GRPCClientStateMachine.swift
+++ b/Sources/GRPC/GRPCClientStateMachine.swift
@@ -346,6 +346,21 @@ struct GRPCClientStateMachine {
     }
   }
 
+  /// Receive a DATA frame with the end stream flag set. Determines whether it is safe for the
+  /// caller to ignore the end stream flag or whether a synthesised status should be forwarded.
+  ///
+  /// Receiving a DATA frame with the end stream flag set is unexpected: the specification dictates
+  /// that an RPC should be ended by the server sending the client a HEADERS frame with end stream
+  /// set. However, we will tolerate end stream on a DATA frame if we believe the RPC has already
+  /// completed (i.e. we are in the 'clientClosedServerClosed' state). In cases where we don't
+  /// expect end of stream on a DATA frame we will emit a status with a message explaining
+  /// the protocol violation.
+  mutating func receiveEndOfResponseStream() -> GRPCStatus? {
+    return self.withStateAvoidingCoWs { state in
+      state.receiveEndOfResponseStream()
+    }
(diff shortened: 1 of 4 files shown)


Sources/GRPC/GRPCClientStateMachine.swift
```diff
@@ -346,6 +346,21 @@ struct GRPCClientStateMachine {
...
 
+  /// Receive a DATA frame with the end stream flag set. Determines whether it is safe for the
+  /// caller to ignore the end stream flag or whether a synthesised status should be forwarded.
+  ///
+  /// Receiving a DATA frame with the end stream flag set is unexpected: the specification dictates
+  /// that an RPC should be ended by the server sending the client a HEADERS frame with end stream
+  /// set. However, we will tolerate end stream on a DATA frame providing we have received trailers
+  /// containing 'grpc-status' (i.e. we are in the 'clientClosedServerClosed' state). In cases where
+  /// we don't expect end of stream on a DATA frame we will emit a status with a message explaining
+  /// the protocol violation.
```
Lukasa, line 357: I'm a bit confused about this state description. When will we receive a DATA frame _after_ having received a HEADERS frame that contains trailers? NIOHTTP2 enforces that the trailers header block must have END_STREAM, so no DATA frame on that stream may validly follow it. What have I missed?

glbrntt: Ah, right, that's a good point. That was me assuming a state in which it could be valid. Really the case we tolerate it is "the rpc is already complete".

Lukasa (approved):

Issue #1137, opened 2021-02-12
[Flatbuffers] content-type suffix


mustiikhalil:
Good evening, 

 While working on implementing FlatBuffers support for other programming languages I found out that Go requires the suffix `+flatbuffers` to the `content-type` example: `content-type: application/grpc+flatbuffers`. I was wondering if this is something the library can support? I've dig into the code and found that we might need to change the following enum to support it `ContentType` which won't be a huge issue. since according to the [Https2 protocol](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) the `content-type` should allow any custom suffix if needed.
 
I am willing to implement all the details behind this feature, however I wanted to ask before doing so, since the library might have a different direction and isn't willing to support custom content-type suffix

glbrntt:
I'm okay with adding this.

You're right, adding a case to the `ContentType` is part of this. We also need a way for users to set the value and then tests of course. The easiest way to do this would probably be by using any any value provided in the `customMetadata` and then falling back to "application/grpc" if nothing was provided by the user.

mustiikhalil:
Sounds good to me. I can definitely work on this. This would be the plan:
- [ ] Adding it to the `ContentType` enum.
- [ ] Allowing the user to select the content type they would be sending over the network. 
- [ ] Tests
    - GRPCTests
    - GRPCInteroperabilityTests?

What do you think would be better:

1 - Maybe add a field on the `CallOptions` object itself, and appending it to the `HPACKHeaders` after [init](https://github.com/grpc/grpc-swift/blob/c85a7efe87b6fe5415d95e26f36b05cc5e9bf6f8/Sources/GRPC/CallOptions.swift#L71)
 
```swift
/// example
init(
customMetadata: HPACKHeaders = HPACKHeaders(),
contentType: ContentType = .protoc,
) {
  customMetaData.append((GRPCHeaderName.contentType, contentType.canonicalValue)
}
```

2 - Add it into the `HPACKHeaders`, as in allow the user to append it himself into the `HPACKHeaders` object, then within the code we try to see if there is a `content-type` appended if not, we default to `application/grpc`

glbrntt:
> Sounds good to me. I can definitely work on this. This would be the plan:
> * [ ]  Tests
>   
>   * GRPCTests
>   * GRPCInteroperabilityTests?

[...]

PR #1141, opened 2021-02-17
Adding Flatbuffers content-type

mustiikhalil:
The following PR adds the ability to send custom "implemented" content-types `application/grpc+{(proto || flatbuffers)}`. It does that by checking the `customMetadata` object for the type `content-type` on the sender side, If there is an object of type content-type we use that as the `content-type` for the request. However if `customMetadata` doesn't have any header for the content-type we default to `application/grpc`. 

- Tests implemented:
     - Checks for sending custom content-types and if they would be appended in the header
     - Checks if the server would be able to parse those content types.
 
Notes: Tests were implemented on the valid server state only.

### DIFF
--- a/Tests/GRPCTests/GRPCClientStateMachineTests.swift
+++ b/Tests/GRPCTests/GRPCClientStateMachineTests.swift
@@ -78,6 +78,22 @@ extension GRPCClientStateMachineTests {
     }
   }
 
+  func doTestCustomContentType(_ contentType: ContentType) {
+    var stateMachine = self
+      .makeStateMachine(.clientIdleServerIdle(pendingWriteState: .one(), readArity: .one))
+    stateMachine.sendRequestHeaders(requestHead: .init(
+      method: "POST",
+      scheme: "http",
+      path: "/echo/Get",
+      host: "host",
+      deadline: .distantFuture,
+      customMetadata: ["content-type": contentType.canonicalValue],
+      encoding: .disabled
+    )).assertSuccess { headers in
+      XCTAssertEqual(headers.first(name: "content-type"), contentType.canonicalValue)
+    }
+  }
+
   func testSendRequestHeadersFromIdle() {
     var stateMachine = self
       .makeStateMachine(.clientIdleServerIdle(pendingWriteState: .one(), readArity: .one))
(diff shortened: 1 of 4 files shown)


Tests/GRPCTests/GRPCClientStateMachineTests.swift
```diff
@@ -92,6 +92,58 @@ extension GRPCClientStateMachineTests {
     )).assertSuccess()
   }
 
+  func testSendingDefaultContentypeRequestHeader() {
```
glbrntt, line 95: nit: `Contentype` -> `ContentType`

Tests/GRPCTests/GRPCClientStateMachineTests.swift
```diff
@@ -92,6 +92,58 @@ extension GRPCClientStateMachineTests {
...
+      host: "host",
+      deadline: .distantFuture,
+      customMetadata: [:],
+      encoding: .disabled
+    )).assertSuccess { headers in
+      XCTAssertEqual(headers.first(name: "content-type"), "application/grpc")
+    }
+  }
+
+  func testSendingProtoContentypeRequestHeader() {
```
glbrntt, line 111: nit: let's have a helper test which takes a `ContentType`, something like:

```swift
func doTestCustomContentType(_ contentType: ContentType) {
  var stateMachine = ...
  stateMachine.sendRequestHeaders(requestHead: .init(
    ...
    customMetadata: ["content-type": contentType.canonicalValue]
    ...
  )).assertSuccess { 
    ...
  } 
}
```

then we can just call that from our tests with `.flatbuffers` and `.protobuf`

glbrntt (changes requested):
Looks good so far, just a couple of nits in tests.

> Checks if the server would be able to parse those content types.

This isn't actually tested; you're testing receiving response headers from the server. `HTTP2ToRawGRPCStateMachineTests` is what you're after.

mustiikhalil:
Thanks for the review! Yeah I was looking for that test class so I can see how to implement items there!

mustiikhalil:
converted to draft until the outcome of what's explained here https://github.com/grpc/grpc-swift/issues/1137#issuecomment-780951438

mustiikhalil:
Closing this since the go lang added a `ForceServerCodec`. Until the main grpc team fixes the issue with the headers not being sent properly. 

ref:
https://github.com/grpc/grpc/issues/25478

PR #1142, opened 2021-02-23
Bump version of Google's SpeechToText Example Podfile

Raduan77:
After lurking around and playing with tutorials in `Sources/Examples/` and `Examples/` directories, I've noticed that I can't run Google's SpeechToText example. After reading similar problems in the Issues section, I've found out that the problem was in the `Podfile` of this SpeechToText tutorial. 

Here is a fix, which makes the tutorial's project have a correct behavior.

---

Also, the first time I've skipped the part of enabling Google Cloud's SpeechToText API's, which is mentioned in overall Google examples [readme](https://github.com/grpc/grpc-swift/tree/main/Examples/Google/README.md) and jumped straight to the steps of SpeechToText [readme](https://github.com/grpc/grpc-swift/blob/main/Examples/Google/SpeechToText/README.md). It resulted in some confusing errors, so I propose adding a `0.` step of enabling Google Cloud's Speech To Text API, but maybe that's solely my problem, and it's not that necessary :)

I would like to hear your thoughts on that as well.

### DIFF
--- a/Examples/Google/SpeechToText/Podfile
+++ b/Examples/Google/SpeechToText/Podfile
@@ -6,5 +6,5 @@ target 'SpeechToText-gRPC-iOS' do
   use_frameworks!
 
   pod 'SnapKit'
-  pod 'gRPC-Swift', '1.0.0-alpha.20'
+  pod 'gRPC-Swift', '1.0.0'
 end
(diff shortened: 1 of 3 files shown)


glbrntt:
> It resulted in some confusing errors, so I propose adding a `0.` step of enabling Google Cloud's Speech To Text API, but maybe that's solely my problem, and it's not that necessary :)

That sounds like a sensible addition, would you mind adding it?

Raduan77:
Done!

glbrntt (approved):
Thank you!

PR #1147, opened 2021-03-10
Add a 'closeFuture' to the server context

### LINKED ISSUE #1145: Detect When Client Closes Client-Side Streaming Call

I need to detect server-side inside a client-side streaming call handler function when a client disconnects from the streaming call in order to clean up state / free resources and subscriptions.

For server-side streaming calls I can subscribe to `context.statusPromise.futureResult` where `context` is a `StreamingResponseCallContext`. This seems to always be called eventually, even if the client (in our case an iOS app) looses connection / is killed. However, for client-side streaming calls the context is of type `UnaryResponseCallContext` which doesn't seem to have a corresponding future that I can subscribe to.

How do I get a callback when the client-side streaming call is closed / connection is lost?

glbrntt:
Motivation:

Freeing up user-allocated resources for RPCs is difficult to do
correctly: there isn't a single obvious way to do it for all types, and
in some cases it's impossible since the handler doesn't always know when
the RPC has been terminated. That makes it nigh on impossible for users
to correctly manage resources.

Modifications:

- Add a 'closeFuture' to the server context allowing users to register a
  callbacks to be executed when the RPC completes.
- In existing context implementations this is just the stream channels
  'closeFuture'
- Tests for each RPC style: when the RPC completes successfully, when
  the RPC handler future fails, and when the connection is killed
  mid-rpc

Results:

- It's easier for users to tear down resources when the RPC has
  completed
- Resolves #1145

### DIFF
--- a/Sources/GRPC/CallHandlers/BidirectionalStreamingServerHandler.swift
+++ b/Sources/GRPC/CallHandlers/BidirectionalStreamingServerHandler.swift
@@ -167,6 +167,7 @@ public final class BidirectionalStreamingServerHandler<
         logger: self.context.logger,
         userInfoRef: self.userInfoRef,
         compressionIsEnabled: self.context.encoding.isEnabled,
+        closeFuture: self.context.closeFuture,
         sendResponse: self.interceptResponse(_:metadata:promise:)
       )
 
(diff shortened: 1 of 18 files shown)


Lukasa (commented):
Is it a good idea to make the tear down block a single function, rather than a future that is completed? Having it be a single function means it doesn't compose very well with other things.

glbrntt:
> Is it a good idea to make the tear down block a single function, rather than a future that is completed? Having it be a single function means it doesn't compose very well with other things.

Oh just a `closeFuture` on the context? I guess we can avoid the extra allocation by re-using the stream channel's future. We'd lose the error though but I'm not sure how helpful that is anyway.

Lukasa (approved):

PR #1150, opened 2021-03-17
Add allocation counting tests

glbrntt:
Motivation:

Memory allocations are a useful tool when evaluating performance. As
such it's useful to have a framework where tests can run continually and
be checked against an expected limit.

Modifications:

There are three groups of related changes in this PR.

First, shuffling existing performance test code so that it may be used
with NIOs allocation counting test framework. This includes dropping the
dependency on the 'EchoModel' target and symlinking in the generated
protobuf and gRPC code services instead. Some of the benchmark code was
split into separate files as well (we'll symlink some code to the
allocation counting test). The 'Benchmark' protocol for the tests now
returns an 'Int' from the 'run' function (a requirement of the NIO
allocation counting test framework).

Second, the addition of various scripts to run allocation counting
tests. These were lifted from SwiftNIO and slightly modified to fit our
use case. As such a NOTICES.txt was also added. Three tests were also
added which make use of code from the existing performance tests. The
shared code is symlinked in (to avoid adding an additional library
product to the package).

Third, the Travis configuration was updated to run the tests on 5.2 and
5.3 on Linux only. A script was also added to parse the output from
running tests and produce an allocation limit.

Result:

We can track our allocation counts.

### DIFF
--- a/.travis-script.sh
+++ b/.travis-script.sh
@@ -56,6 +56,15 @@ make_test() {
   echo -en 'travis_fold:end:make.test\\r'
 }
 
+run_allocation_tests() {
+  echo -en 'travis_fold:start:allocation_tests\\r'
+
+  info "Running allocation counter tests"
+  ./Performance/allocations/test-allocation-counts.sh
+
+  echo -en 'travis_fold:end:allocation_tests\\r'
+}
+
 make_test_plugin() {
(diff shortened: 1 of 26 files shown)


glbrntt:
cc @weissi

Lukasa (approved):
Nice, LGTM.

weissi (approved):
Awesome!

PR #1151, opened 2021-03-17
Use static trailers if possible

glbrntt:
Motivation:

Happy path RPCs end with the 'ok' status code. If the user doesn't
specify any trailers then the trailers are static and we can save an
allocation.

Modifications:

- If we're sending the '.ok' status and the user provides no trailers,
  use statically allocated trailers.
- Update allocation counts.

Result:

We save an allocation on the happy path.


weissi (approved):
yay!

glbrntt:
Ping @Lukasa

Lukasa (approved):

PR #1153, opened 2021-03-23
Update paths to protoc-gen-swift in basic-tutorial

dengjeffrey:
Based on https://github.com/grpc/grpc-swift/blob/main/Makefile#L11 the plugins are placed in `./build/release` not `./build/debug`. These changes fixes the documentation to point to the correct path of the plugin.

### DIFF
--- a/docs/basic-tutorial.md
+++ b/docs/basic-tutorial.md
@@ -162,10 +162,10 @@ Let's look at how to run the same command manually:
 ```sh
 $ protoc Sources/Examples/RouteGuide/Model/route_guide.proto \
     --proto_path=Sources/Examples/RouteGuide/Model \
-    --plugin=./.build/debug/protoc-gen-swift \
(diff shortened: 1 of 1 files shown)


dengjeffrey:
Feel free to copy these changes in a different PR if you have signed the CLA. I am not going to go through the hoops of signing the CLA for documentation changes.

Lukasa:
Copying the changes doesn’t really address the IP issue here. Signing the CLA should be extremely straightforward, but if you don’t want to that’s totally fine.

dengjeffrey:
@Lukasa Yea unfortunately in my case it wasn't as straightforward. It wouldn't let me change my email to a non primary email associated with this GitHub account. My primary email is linked to an organization, so it would have been odd to use it to sign as an individual. 

I suppose maybe not copy the changes, but take this PR as a suggestion of what could be fixed.

PR #1155, opened 2021-04-01
Allocation counting tests for full RPC flows

glbrntt:
Motivation:

Now that we have an allocation counting test framework we should add
tests to it!

Modifications:

- Add some common boilerplate for making an echo server and client
- Add an allocation test for doing 1000 unary RPCs sequentially
- Add allocation tests for doing 1000 bidi streaming RPCs sequentially,
  one where each RPC sends 1 request and another where each RPC sends 10
  request.

Result:

More test coverage for allocations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,25 +37,33 @@ jobs:
       script: ./.travis-script.sh -t -a # tests with tsan, run allocation tests
       env:
         - SWIFT_VERSION=5.3.3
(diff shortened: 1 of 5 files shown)


Lukasa (approved):

weissi (approved):
🎉

glbrntt:
Hmm `test_bidi_1k_rpcs_10_requests` is pretty unstable, the range is something like 20,000 allocs (usually around 600,000) 👀

PR #1156, opened 2021-04-06
Alllow calls to run on a specified event loop

glbrntt:
Motivation:

It can be helpful for callers to know which event loop their RPCs are
completed on. Currently the caller must hop to the desired event loop if
it differs from the event loop used by the underlying transport. This
can be tedious and error prone since it must be done in a number of
places.

Modifications:

- Separate the event loop an RPC runs on (i.e. the one used for
  interceptors and response futures) and the event loop the RPC is
  transported on (i.e. the one used by the underlying 'Channel').
- Add an 'EventLoopPreference' to 'CallOptions' allowing users to
  specify they are 'indifferent' (the existing and default behaviour,
  use an event loop provided by the framework) or use an 'exact' event
  loop.
- Add tests.

Result:

Callers can specify an event loop to run their RPCs on.


Lukasa (approved):
Nice, this LGTM.

PR #1159, opened 2021-04-08
Add a Homebrew command for installing Protocol Buffers in docs

graciborski:


docs/quick-start.md
```diff
@@ -11,7 +11,10 @@ gRPC requires Swift 5.0 or higher.
 #### Install Protocol Buffers v3
 
 Install the protoc compiler that is used to generate gRPC service code. The
-simplest way to do this is to download pre-compiled binaries for your
+simplest way to do this is to use Homebrew.
```
glbrntt, line 14: This is only the simplest way if you're using macOS (Homebrew exists on Linux too but I'm not sure it's so ubiquitous) and have Homebrew installed so I think we should prefix this by saying something to that effect.

docs/quick-start.md
```diff
@@ -11,7 +11,10 @@ gRPC requires Swift 5.0 or higher.
...
+```brew install swift-protobuf```
```
glbrntt, line 15: This section is about install `protoc` rather than the plugins, so this should be `protobuf` instead of `swift-protobuf`:

```sh
$ brew install protobuf
```

glbrntt (changes requested):
I'm okay adding instructions for install with homebrew but I think we need a few clarifications.

Also, you'll need to agree to and sign the CLA for this PR to be accepted.

PR #1165, opened 2021-04-22
Add a static-source Logger

glbrntt:
Motivation:

swift-log includes a `source` field which -- in most cases -- should be
the name of the module. Currently swift-log parses the `#file` to use
the directory containing the file in the hope that it is the module
name.

In our tests we have a tear down step which validates that all caputred
logs have 'GRPC' as their source. However, this requires manually
providing the source in a bunch of places, that's a bit of a pain and
only validates the logs we emit.

Modifications:

- Add an internal `GRPCLogger` which wraps a `Logger` providing only
  `trace` and `debug` (for now) and always sets the source to `GRPC`.

Result:

It's harder to get the source wrong.

### DIFF
--- a/Sources/GRPC/GRPCLogger.swift
+++ b/Sources/GRPC/GRPCLogger.swift
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2021, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


Lukasa (approved):
Nice one, seems like a very good idea.

weissi:
@glbrntt I fixed that in SwiftLog

PR #1173, opened 2021-05-11
Extract the connection monitor from the connection manager

glbrntt:
Motivation:

The connection monitor allows users to register a delegate for changes
in connectivity state as well as access the current state. Because the
user code may call out it is executed on a `DispatchQueue` (rather than
the connection's `EventLoop`).

However, there are cases within gRPC where state change notifications
on the event loop are very useful (such as in a pool of connection
managers all running on the same event loop).

Moreover, it's not necessary for the connection manager to hold the
connection monitor, it just needs a way to inform it about any changes
so that it can call any user provided delegates.

Modifications:

- Extract the connectivity monitor from the connection manager
- Allow on-event-loop callbacks to be passed to the connection manager
  for connection state changes and for certain HTTP/2 events (stream
  closed and receiving max concurrent streams settings). While not
  strictly necessary for this change they will be helpful down the line.
- Update a bunch of tests.

Result:

The connection manager no longer holds a connection monitor and may
execute state change callbacks on the event loop.

### DIFF
--- a/Sources/GRPC/ClientConnection.swift
+++ b/Sources/GRPC/ClientConnection.swift
@@ -93,9 +93,7 @@ public class ClientConnection {
   internal let authority: String
 
   /// A monitor for the connectivity state.
-  public var connectivity: ConnectivityStateMonitor {
-    return self.connectionManager.monitor
-  }
+  public let connectivity: ConnectivityStateMonitor
 
   /// The `EventLoop` this connection is using.
   public var eventLoop: EventLoop {
@@ -112,8 +110,16 @@ public class ClientConnection {
     self.configuration = configuration
     self.scheme = configuration.tls == nil ? "http" : "https"
     self.authority = configuration.tls?.hostnameOverride ?? configuration.target.host
+
+    let monitor = ConnectivityStateMonitor(
+      delegate: configuration.connectivityStateDelegate,
+      queue: configuration.connectivityStateDelegateQueue
+    )
+
+    self.connectivity = monitor
     self.connectionManager = ConnectionManager(
       configuration: configuration,
+      connectivityDelegate: monitor,
       logger: configuration.backgroundActivityLogger
     )
   }
--- a/Sources/GRPC/ConnectionManager.swift
+++ b/Sources/GRPC/ConnectionManager.swift
@@ -176,29 +176,43 @@ internal final class ConnectionManager {
     }
   }
 
+  /// The last 'external' state we are in, a subset of the internal state.
+  private var externalState: ConnectivityState = .idle
+
+  /// Update the external state, potentially notifying a delegate about the change.
+  private func updateExternalState(to nextState: ConnectivityState) {
+    if self.externalState != nextState {
+      self.externalState = nextState
+      self.connectivityDelegate?.connectionStateDidChange(
+        self, from: self.externalState, to: nextState
+      )
+    }
+  }
+
+  /// Our current state.
   private var state: State {
     didSet {
       switch self.state {
       case .idle:
-        self.monitor.updateState(to: .idle, logger: self.logger)
+        self.updateExternalState(to: .idle)
         self.updateConnectionID()
 
       case .connecting:
-        self.monitor.updateState(to: .connecting, logger: self.logger)
+        self.updateExternalState(to: .connecting)
 
       // This is an internal state.
       case .active:
         ()
 
       case .ready:
-        self.monitor.updateState(to: .ready, logger: self.logger)
+        self.updateExternalState(to: .ready)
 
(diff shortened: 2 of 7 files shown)


Sources/GRPC/ClientConnection.swift
```diff
@@ -112,9 +110,23 @@ public class ClientConnection {
...
       configuration: configuration,
-      logger: configuration.backgroundActivityLogger
+      logger: configuration.backgroundActivityLogger,
+      onConnectivityStateChange: { manager, _, newState in
+        let logger = manager.logger
+        monitor.updateState(to: newState, logger: logger)
+      },
+      onConnectionIsQuiescing: { _ in
+        monitor.beginQuiescing()
+      }
```
fabianfett, line 129: Are we sure we want to have the communication `ConnectionManager` -> `ConnectivityStateMonitor` closure based? Is this the right tradeoff vs. a protocol here? This is not really a change suggestion, more a question how we deal with these type of situations here.

Sources/GRPC/ConnectionManager.swift
```diff
@@ -176,38 +176,64 @@ internal final class ConnectionManager {
...
+  private var _onConnectionIsQuiescing: OnConnectionIsQuiescing?
+  internal typealias OnConnectionIsQuiescing = (ConnectionManager) -> Void
+
+  /// A closure which is notified when an HTTP/2 stream is closed.
+  private var _onHTTP2StreamClosed: OnHTTP2StreamClosed?
+  internal typealias OnHTTP2StreamClosed = (ConnectionManager) -> Void
+
+  /// A closure which is notified when the value of HTTP/2 SETTINGS_MAX_CONCURRENT_STREAMS changes.
+  private var _onHTTP2MaxConcurrentStreamsChanged: OnHTTP2MaxConcurrentStreamsChanged?
+  internal typealias OnHTTP2MaxConcurrentStreamsChanged = (ConnectionManager, Int) -> Void
```
fabianfett, line 236: With a protocol this could be just plain functions. No `typealias` needed. IIRC `typealias` on functions doesn't play nicely with code completion.

Sources/GRPC/ClientConnection.swift
glbrntt: Changed to a protocol (two protocols actually, one for connectivity another for http/2 things).

> Using a protocol existential would be cheaper, if we can do it.

Because we save on allocating closure contexts?

Sources/GRPC/ConnectionManager.swift
glbrntt: I didn't have any issues with the `typealias` with the Xcode I'm using. Either way it's protocol based now.

Sources/GRPC/ClientConnection.swift
Lukasa: Yes: protocol existentials only require heap allocation if you're stuffing a closure larger than 24 bytes into them, otherwise they're stack allocated. You can't optimise through them, but often you can't do that with a closure _anyway_.

Sources/GRPC/ConnectionManager.swift
```diff
@@ -176,38 +176,55 @@ internal final class ConnectionManager {
...
+  /// The last 'external' state we are in, a subset of the internal state.
+  private var externalState: ConnectivityState = .idle
+
+  /// Update the external state, potentially notifying a delegate about the change.
+  private func updateExternalState(to nextState: ConnectivityState) {
+    if self.externalState != nextState {
+      self.connectivityDelegate?.connectionStateDidChange(
+        self, from: self.externalState, to: nextState
+      )
+      self.externalState = nextState
```
Lukasa, line 188: Is this order of operations correct? Generally my advice is to update our internal state _first_ and then make arbitrary outcalls second. Should we be following that advice here?

Lukasa (commented):
One brief note inline.

Sources/GRPC/ConnectionManager.swift
glbrntt: Oof yes, good catch.

Lukasa (approved):
Nice, LGTM.

Tests/GRPCTests/ConnectionManagerTests.swift
```diff
@@ -1025,6 +1012,106 @@ extension ConnectionManagerTests {
       XCTAssertNoThrow(try shutdown.wait())
     }
   }
+
+  func testHTTP2Delegates() throws {
```
fabianfett, line 1016: Nit: do we need the `throws` here?

fabianfett (approved):
Cool. LGTM!

PR #1176, opened 2021-05-13
Add a per event loop connection pool

glbrntt:
Motivation:

One way of building up an client which efficiently pools connections is
to group connections by the event loop in which they run. Each sub-pool
can then synchronize on its event loop and run almost independently of
whatever is managing the sub-pools.

Modifications:

- Add a `ConnectionPool` (the 'sub-pool' mentioned above) which pools
  `ConnectionManager`s all running on the same event loop. Callers can
  ask the pool to create an HTTP/2 stream channel which will be made
  'now' if the pool has a connection with capacity, or queued to be made
  later when capacity becomes available. The pool also has API to
  'return' streams to an outer pool or pool manager.
- Add a `sync` view to `ConnectionManager` to expose some API which
  requires being on the correct event loop.
- `ConnectionManager.shutdown()` now checks if we're already on the
  right event loop before executing (rather than always `flatSubmit`ing
  onto the event loop)

Result:

We have a per event loop connection pool.

### DIFF
--- a/Sources/GRPC/ConnectionManager.swift
+++ b/Sources/GRPC/ConnectionManager.swift
@@ -182,10 +182,9 @@ internal final class ConnectionManager {
   /// Update the external state, potentially notifying a delegate about the change.
   private func updateExternalState(to nextState: ConnectivityState) {
     if self.externalState != nextState {
+      let oldState = self.externalState
       self.externalState = nextState
-      self.connectivityDelegate?.connectionStateDidChange(
-        self, from: self.externalState, to: nextState
-      )
+      self.connectivityDelegate?.connectionStateDidChange(self, from: oldState, to: nextState)
     }
   }
 
@@ -214,6 +213,29 @@ internal final class ConnectionManager {
       case .shutdown:
         self.updateExternalState(to: .shutdown)
       }
+    }
+  }
+
+  /// Returns whether the state is 'idle'.
+  private var isIdle: Bool {
+    self.eventLoop.assertInEventLoop()
+    switch self.state {
+    case .idle:
+      return true
+    case .connecting, .transientFailure, .active, .ready, .shutdown:
+      return false
+    }
+  }
+
+  /// Returns the `HTTP2StreamMultiplexer` from the 'ready' state or `nil` if it is not available.
+  private var multiplexer: HTTP2StreamMultiplexer? {
+    self.eventLoop.assertInEventLoop()
+    switch self.state {
+    case let .ready(state):
+      return state.multiplexer
+
+    case .idle, .connecting, .transientFailure, .active, .shutdown:
+      return nil
     }
   }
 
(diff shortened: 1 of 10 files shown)


Sources/GRPC/ConnectionManager.swift
```diff
@@ -413,77 +435,85 @@ internal final class ConnectionManager {
...
+      // connect the application shouldn't should have access to the channel.
+      state.readyChannelMuxPromise.fail(shutdown.reason)
+
+    // We're already shutdown; nothing to do.
+    case let .shutdown(state):
+      shutdown = state
     }
+
+    return shutdown.closeFuture
   }
```
glbrntt, line 517: This hasn't changed, it's just reindented thanks to the if-in-event-loop-else dance added just above.

Sources/GRPC/ConnectionPool/ConnectionPool.swift
```diff
@@ -0,0 +1,660 @@
...
+  private var connections: [ConnectionManagerID: PerConnectionState]
+
+  /// The threshold which if exceeded when creating a stream determines whether the pool will
+  /// start connecting an idle connection (if one exists).
+  ///
+  /// The 'load' is calculated as the ratio of demand for streams (the sum of the number of waiters
+  /// and the number of reserved streams) and the total number of streams which non-idle connections
+  /// could support (this includes the streams that a connection in the connecting state could
+  /// support).
+  private let reservationLoadThreshold: Double
```
glbrntt, line 51: I hate this name, suggestions gladly welcomed!

Sources/GRPC/ConnectionPool/ConnectionPool+Waiter.swift
```diff
@@ -0,0 +1,101 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import NIO
+import NIOHTTP2
+
+extension ConnectionPool {
+  internal struct Waiter {
```
Lukasa, line 20: I think this object is probably a `class`, not a `struct`. In particular it has identity, as manifested by the fact that we cancel it. Given that we're also closing over it in some cases, redesigning this as a `class` is likely better.

Sources/GRPC/ConnectionPool/ConnectionPool.swift
```diff
@@ -0,0 +1,660 @@
...
+  }
+
+  /// A connection is now in the ready state, it can provide streams to waiters.
+  private func connectionAvailable(_ id: ConnectionManagerID) {
+    self.eventLoop.assertInEventLoop()
+    self.connections[id]?.available(maxConcurrentStreams: self.maxConcurrentStreams)
+
+    // A connection becoming ready is usually followed by an update to max concurrent streams.
+    // Instead of servicing waiters right now we'll do this in a jiffy, i.e. once we receive the
+    // update to max concurrent streams. This allows us to avoid succeeding too many waiters if max
```
Lukasa, line 517: This seems a little brittle: what if we _don't_ receive the update to max concurrent streams in a timely fashion? Would it be better to trigger this based on the receipt of that frame?

Lukasa (changes requested):

Sources/GRPC/ConnectionPool/ConnectionPool.swift
```diff
@@ -0,0 +1,660 @@
...
+  internal func receivedSettingsMaxConcurrentStreams(
+    _ manager: ConnectionManager,
+    maxConcurrentStreams: Int
+  ) {
+    self.eventLoop.assertInEventLoop()
+    self.connections[manager.id]?.updateMaxConcurrentStreams(maxConcurrentStreams)
+
+    if maxConcurrentStreams != self.maxConcurrentStreams {
+      self.maxConcurrentStreams = maxConcurrentStreams
+      let maxAvailable = maxConcurrentStreams * self.connections.count
```
fabianfett, line 556: Is this an abstraction that we are allowed to make? Can we assume that if we are allowed to send more streams on one connection, this will apply to all connections?

Sources/GRPC/ConnectionPool/ConnectionPool.swift
glbrntt: I also didn't like this -- feels like a massive hack. I wasn't sure about receiving the update though. IIRC the settings frame after the connection preface is allowed to be empty, what can we do if this is the case, can we assume a value of max concurrent streams and go about our business until we know better or do we have to wait until we have a value before opening streams?

glbrntt: Maybe this isn't a problem in practice though 🤷‍♂️

Lukasa: If the settings is empty then the default value applies, we are using the heuristic that the value won't change.

Sources/GRPC/ConnectionPool/ConnectionPool.swift
Lukasa: It would be better if we didn't.

Sources/GRPC/ConnectionPool/ConnectionPool.swift
glbrntt: Okay, that makes sense, I'll incorporate this suggestion. I think emitting an update of the default value if we don't receive one in the first settings frame should work fine.

Lukasa (changes requested):

Sources/GRPC/ConnectionManager.swift
```diff
@@ -860,6 +891,62 @@ extension ConnectionManager {
...
 }
 
+extension ConnectionManager {
+  /// Returns a synchronous view of the connection manager; each operation requires the caller to be
+  /// executing on the same `EventLoop` as the connection manager.
+  internal var sync: Sync {
+    return Sync(self)
+  }
+
+  internal struct Sync {
```
Davidde94, line 901: Nit: I know this isn't public, but I'm not in love with the name `Sync`

Davidde94 (commented):
Just a question about delegates

Sources/GRPC/ConnectionManager.swift
```diff
@@ -860,6 +891,62 @@ extension ConnectionManager {
...
+    private let manager: ConnectionManager
+
+    fileprivate init(_ manager: ConnectionManager) {
+      self.manager = manager
+    }
+
+    /// A delegate for connectivity changes.
+    internal var connectivityDelegate: ConnectionManagerConnectivityDelegate? {
```
Davidde94, line 909: Comment didn't post the first time :(

The `Delegate` pattern usually requires the delegate to be `weak` to avoid a reference cycle. I can't see/find the `connectivityDelegate` in `ConnectionManager`, but is it weak?

Davidde94: Only applies if your delegates are classes of course

Lukasa: It only requires a weak reference in cases where that cycle isn't broken manually due to some other lifecycle event. Generally in server-side code we prefer to do that (because weak is sloooooow).

Davidde94: Ah fair enough! Ignore then, I'll resolve the comment :)

Sources/GRPC/ConnectionManager.swift
glbrntt: "sync" is _sort of_ common enough now (synchronous channel options, synchronous pipeline operations) to have meaning (i.e. you must be on the right event loop). That said, if you have a better name I'm all ears!

Lukasa (approved):
This is a _really_ nice patch @glbrntt, well done. Looks like it covers all the bases, I'm happy with it.

PR #1177, opened 2021-05-13
In-source fuzzing support

glbrntt:
Motivation:

gRPC Swift was recently added to OSS Fuzz. Doing so required patching
gRPC to make the server pipeline configurator public and a small driver
program to setup and run a server on an embedded channel. We should make
supporting oss-fuzz a little nicer!

Modifications:

- Add a FuzzTesting sub package with a 'ServerFuzzer' target
- Symlink in the echo model and implementation
- Add an underscore-public API addition to GRPC to configure an
  `EmbeddedChannel` for fuzzing

Result:

Better support for oss-fuzz

### DIFF
--- a/FuzzTesting/.gitignore
+++ b/FuzzTesting/.gitignore
@@ -1,0 +1,7 @@
+.DS_Store
+/.build
+/Packages
+/*.xcodeproj
+xcuserdata/
+DerivedData/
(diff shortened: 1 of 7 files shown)


Davidde94 (approved):

Lukasa (approved):
Ace, LGTM.

PR #1179, opened 2021-05-14
Use `PlatformSupport.makeEventLoopGroup` on BasicTutorial

e-sung:
Even though [Apple Platform](https://github.com/grpc/grpc-swift/blob/master/docs/apple-platforms.md) documentation is suggesting to use `PlatformSupport.makeEventLoopGroup` where possible, I'm afraid it may not be obvious for beginners to checkout that documentation. 

I think it would be better if we use `PlatformSupport.makeEventLoopGroup` API in [Basic-Tutorial](https://github.com/grpc/grpc-swift/blob/master/docs/basic-tutorial.md) document which is much more approachable. 

How do you think about this idea?

### DIFF
--- a/Sources/Examples/RouteGuide/Client/main.swift
+++ b/Sources/Examples/RouteGuide/Client/main.swift
@@ -201,7 +201,7 @@ struct RouteGuide: ParsableCommand {
     // Load the features.
     let features = try loadFeatures()
 
-    let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+    let group = PlatformSupport.makeEventLoopGroup(loopCount: 1)
     defer {
       try? group.syncShutdownGracefully()
     }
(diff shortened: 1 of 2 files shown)


glbrntt:
Hi @e-sung -- I think that's a good idea, would you mind updating the client code as well (https://github.com/grpc/grpc-swift/blob/main/Sources/Examples/RouteGuide/Client/main.swift)? Also you'll need to sign the CLA before we can accept any contribution.

e-sung:
Thank you for great feedback @glbrntt! I updated corresponding example code on [RoutGuide](https://github.com/grpc/grpc-swift/blob/main/Sources/Examples/RouteGuide/Client/main.swift) too!

glbrntt (approved):
Thank you!

PR #1180, opened 2021-05-14
Echo: add an --iterations option

weissi:
Motivation:

For performance analysis, it can be useful to have multiple iterations of RPCs. Especially in gRPC where you may want to ignore the connection setup overhead.

Modification:

Add `--iterations` to `Echo` which does the RPC that many times.

Result:

Pretty graphs (the one below shows all allocations made by one RPC).

Screenshot 2021-05-14 at 3 20 46 pm

### DIFF
--- a/Sources/Examples/Echo/Runtime/main.swift
+++ b/Sources/Examples/Echo/Runtime/main.swift
@@ -79,6 +79,9 @@ struct Echo: ParsableCommand {
     @Option(help: "RPC to call ('get', 'collect', 'expand', 'update').")
     var rpc: RPC = .get
 
+    @Option(help: "How many RPCs to do.")
+    var iterations: Int = 1
+
     @Argument(help: "Message to echo")
     var message: String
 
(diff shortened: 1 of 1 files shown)


Sources/Examples/Echo/Runtime/main.swift
```diff
@@ -98,7 +101,9 @@ struct Echo: ParsableCommand {
         try! client.channel.close().wait()
       }
 
-      callRPC(self.rpc, using: client, message: self.message)
+      for _ in 0..<self.iterations {
```
glbrntt, line 104: I think the formatter is gonna get mad at you, I think you want:

```swift
for _ in 0 ..< self.iterations {
```

Or you can run `./scripts/format.sh`

weissi, line 104:
```suggestion
      for _ in 0 ..< self.iterations {
```

glbrntt (approved):
Ta!

PR #1184, opened 2021-05-17
Make 'NIOSSLContext' ahead of time

glbrntt:
Motivation:

A `NIOSSLContext` may be used for multiple connections. Moreover,
creating `NIOSSLContext`s is _expensive_ and we should avoid creating
them unnecessarily. At the moment we create one context per connection
for the client and the server.

Modifications:

- Make `NIOSSLContext`s when setting up the client (rather than when
  initializing a channel).
- Make `NIOSSLContext`s when preparing a server bootstrap

Result:

- We only create `NIOSSLContext` once per server/client connection


Davidde94 (approved):
Nice!

Lukasa (approved):
Beautiful, this is a great patch.

PR #1187, opened 2021-05-19
Handle multiple requests for gRPC-Web

glbrntt:
Motivation:

Our gRPC Web implementation was _somewhat_ underbaked. It would, for
example, disregard the 'connection' header and was not able to handle
multiple requests. That's pretty bad!

Modifications:

- Rework part of the grpc-web state machine
- Extend the http/2 to grpc state machine and codec to support multiple
  requests per instance
- grpc-web state machine tests

Result:

gRPC Web is better tested and supports multiple requests.


Lukasa (approved):
This seems like a really good clean-up George, nice one!

glbrntt:
CI failures are related to https://github.com/grpc/grpc-swift/pull/1188

Lukasa (approved):
Nice, LGTM.

PR #1189, opened 2021-05-21
Add a connection pool manager

glbrntt:
Motivation:

In #1176 we added a per-event-loop connection pool. In order to make a
client which uses these pools we need a way to manage them. This PR adds
a connection pool manager.

Modifications:

- Add a pool manager state machine which tracks connection pools and
  per-pool state such as the number of streams available and reserved
  for each pool.
- Add a pool manager which wraps the state machine.
- Add a pool manager state machine tests; the pool manager isn't tested
  here but will be tested indirectly in a later PR (when a client is
  added to wrap the pool manager).

Result:

We can manage connection pools.

### DIFF
--- a/Sources/GRPC/ConnectionPool/PoolManager.swift
+++ b/Sources/GRPC/ConnectionPool/PoolManager.swift
@@ -57,6 +57,7 @@ internal final class PoolManager {
   /// The current state of the pool manager, `lock` must be held when accessing or
   /// modifying `state`.
   private var state: PoolManagerStateMachine
+  private var pools: [ConnectionPool]
   private let lock = Lock()
 
   /// The `EventLoopGroup` providing `EventLoop`s for connection pools. Once initialized the manager
@@ -86,6 +87,7 @@ internal final class PoolManager {
 
   private init(group: EventLoopGroup) {
     self.state = PoolManagerStateMachine(.inactive)
+    self.pools = []
     self.group = group
 
     // The pool relies on the identity of each `EventLoop` in the `EventLoopGroup` being unique. In
@@ -124,10 +126,11 @@ internal final class PoolManager {
     var logger = logger
     logger[metadataKey: Metadata.id] = "\(ObjectIdentifier(self))"
 
-    let pools = self.makePools(perPoolConfiguration: configuration, logger: logger)
+    assert(self.pools.isEmpty)
+    self.pools = self.makePools(perPoolConfiguration: configuration, logger: logger)
 
     logger.debug("initializing connection pool manager", metadata: [
-      Metadata.poolCount: "\(pools.count)",
+      Metadata.poolCount: "\(self.pools.count)",
       Metadata.connectionsPerPool: "\(configuration.maxConnections)",
       Metadata.waitersPerPool: "\(configuration.maxWaiters)",
     ])
@@ -135,13 +138,22 @@ internal final class PoolManager {
     // The assumed maximum number of streams concurrently available in each pool.
     let assumedCapacity = configuration.assumedStreamCapacity
 
+    // The state machine stores the per-pool state keyed by the pools `EventLoopID` and tells the
+    // pool manager about which pool to use/operate via the pools index in `self.pools`.
+    let poolKeys = self.pools.enumerated().map { poolIndex, pool in
+      return ConnectionPoolKey(
+        index: ConnectionPoolIndex(poolIndex),
+        eventLoopID: pool.eventLoop.id
+      )
+    }
+
     self.lock.withLockVoid {
       // We'll blow up if we've already been initialized, that's fine, we don't allow callers to
       // call `initialize` directly.
-      self.state.activate(pools: pools, assumingPerPoolCapacity: assumedCapacity)
+      self.state.activatePools(keyedBy: poolKeys, assumingPerPoolCapacity: assumedCapacity)
     }
 
-    for pool in pools {
+    for pool in self.pools {
       pool.initialize(connections: configuration.maxConnections)
     }
   }
@@ -212,12 +224,18 @@ internal final class PoolManager {
     streamInitializer initializer: @escaping (Channel) -> EventLoopFuture<Void>
   ) -> PooledStreamChannel {
     let reservation = self.lock.withLock {
-      self.state.reserveStream(preferringPoolOnEventLoop: preferredEventLoop)
+      self.state.reserveStream(
+        preferringPoolWithEventLoopID: preferredEventLoop.map { EventLoopID($0) }
+      )
     }
 
     switch reservation {
-    case let .success(pool):
-      let channel = pool.makeStream(deadline: deadline, logger: logger, initializer: initializer)
+    case let .success(poolIndex):
+      let channel = self.pools[poolIndex.value].makeStream(
+        deadline: deadline,
+        logger: logger,
+        initializer: initializer
+      )
       return PooledStreamChannel(futureResult: channel)
 
     case let .failure(error):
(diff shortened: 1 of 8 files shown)


glbrntt:
cc @fabianfett

Sources/GRPC/ConnectionPool/PoolManager.swift
```diff
@@ -0,0 +1,270 @@
...
+ */
+import Logging
+import NIO
+import NIOConcurrencyHelpers
+
+internal final class PoolManager {
+  /// Configuration used for each connection pool.
+  internal struct PerPoolConfiguration {
+    /// The maximum number of connections per pool.
+    var connections: Int
```
fabianfett, line 24: Maybe call this `maxConnections`? You need to read the docu to know that this is about the max number...

Sources/GRPC/ConnectionPool/PoolManager.swift
```diff
@@ -0,0 +1,270 @@
...
+    logger: GRPCLogger
+  ) -> [ConnectionPool] {
+    let eventLoops = self.group.makeIterator()
+    return eventLoops.map { eventLoop in
+      // We're creating a retain cycle here as each pool will reference the manager and the per-pool
+      // state will hold the pool which will in turn be held by the pool manager. That's okay: when
+      // the pool is shutdown the per-pool state and in turn each connection pool will be dropped.
+      // and we'll break the cycle.
+      return ConnectionPool(
+        eventLoop: eventLoop, maxWaiters: configuration.maximumWaiters,
```
fabianfett, line 165:
```suggestion
        eventLoop: eventLoop, 
        maxWaiters: configuration.maximumWaiters,
```

Sources/GRPC/ConnectionPool/PoolManager.swift
```diff
@@ -0,0 +1,270 @@
...
+    self.lock.withLockVoid {
+      self.state.returnStreams(count, to: pool)
+    }
+  }
+
+  internal func increaseStreamCapacity(by delta: Int, for pool: ConnectionPool) {
+    self.lock.withLockVoid {
+      self.state.increaseStreamCapacity(by: delta, for: pool)
+    }
+  }
```
fabianfett, line 261: Could a streams capacity also be decreased?

Sources/GRPC/ConnectionPool/PoolManagerStateMachine+PerPoolState.swift
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    internal var description: String {
+      return String(describing: self.id)
+    }
+  }
+
+  internal struct PerPoolState {
+    /// A pool of connections using the same `EventLoop`.
+    internal var pool: ConnectionPool
```
fabianfett, line 33: Do we need to hold a reference to the actual `ConnectionPool` in the state machine?

fabianfett: Could the StateMachine only hold state info, and the `ConnectionPool` itself is hold by the `ConnectionPoolManager`?

fabianfett: Maybe we can replace that with the `EventLoopID`.

Sources/GRPC/ConnectionPool/PoolManager.swift
Lukasa, line 165: Nit: I don't love having `maxWaiters` stashed over here to the side. Can we give it its own line?

Sources/GRPC/ConnectionPool/PoolManager.swift
```diff
@@ -0,0 +1,270 @@
...
+  ///   - initializer: A closure to initialize the `Channel` with.
+  /// - Returns: An `EventLoopFuture<Channel>` on the same `EventLoop` as the connection pool
+  ///     providing the `Channel`. The future will be failed if the pool manager has been shutdown,
+  ///     the deadline has passed before a stream was created or if the selected connection pool
+  ///     is unable to create a stream (if there is too much demand on that pool, for example).
+  internal func makeStream(
+    preferredEventLoop: EventLoop?,
+    deadline: NIODeadline,
+    logger: GRPCLogger,
+    streamInitializer initializer: @escaping (Channel) -> EventLoopFuture<Void>
```
Lukasa, line 198: This API is likely to be somewhat hard to hold: the future can run on an essentially arbitrary event loop, making chaining from it difficult. We should consider either using a `promise` or requiring an event loop for the returned future to execute on.

Sources/GRPC/ConnectionPool/PoolManager.swift
```diff
@@ -0,0 +1,270 @@
...
+
+    case let .failure(error):
+      return (preferredEventLoop ?? self.group.next()).makeFailedFuture(error)
+    }
+  }
+
+  // MARK: Shutdown
+
+  /// Shutdown the pool manager and all connection pools it manages.
+  internal func shutdown() -> EventLoopFuture<Void> {
```
Lukasa, line 216: Again, this API has a tricky time with thread-safety. Do we really want it?

fabianfett (commented):
Some nits and room for one discussion.

Lukasa (changes requested):

Sources/GRPC/ConnectionPool/PoolManager.swift
glbrntt: I agree it's slightly awkward but I'm not sure how to work around it best. The problem is that if the caller doesn't have an event loop preference then the event loop of the selected pool (and returned future) dictate which event loop the RPC should run on.

I suppose one option would be to more explicitly indicate which event loop it's running on:

```swift
enum SelectedStreamEventLoop {
  case preferredEventLoop(EventLoopFuture<Channel>)
  case other(EventLoopFuture<Channel>)
}
```

Although I'm not sure how much this actually adds since that's information already encoded by the future.

Lukasa: If this is load-bearing then I think this needs to call be more explicit, yeah. That information is very _implicit_ in the future. I think we shouldn't just return the future but the loop as well, to make it very clear that we're trying to communicate "this is the loop on which the RPC will run".

Sources/GRPC/ConnectionPool/PoolManager.swift
glbrntt: Yes, a negative `delta` is acceptable here. I'll change this to `changeStreamCapacity`.

Sources/GRPC/ConnectionPool/PoolManager.swift
```diff
@@ -0,0 +1,314 @@
...
+      Metadata.connectionsPerPool: "\(configuration.maxConnections)",
+      Metadata.waitersPerPool: "\(configuration.maxWaiters)",
+    ])
+
+    // The assumed maximum number of streams concurrently available in each pool.
+    let assumedCapacity = configuration.assumedStreamCapacity
+
+    // The state machine stores the per-pool state keyed by the pools `EventLoopID` and tells the
+    // pool manager about which pool to use/operate via the pools index in `self.pools`.
+    let poolKeys = self.pools.enumerated().map { poolIndex, pool in
```
Lukasa, line 143: A minor defensiveness note: `poolIndex` is an `index` _only_ if the stored data type is `Array`. Otherwise, `enumerated` does offsets, not indices. It may be better to rename this, or to provide an alternative implementation that actually does use indices and not offsets.

Sources/GRPC/ConnectionPool/PoolManager.swift
```diff
@@ -0,0 +1,314 @@
...
+  ///     is unable to create a stream (if there is too much demand on that pool, for example).
+  internal func makeStream(
+    preferredEventLoop: EventLoop?,
+    deadline: NIODeadline,
+    logger: GRPCLogger,
+    streamInitializer initializer: @escaping (Channel) -> EventLoopFuture<Void>
+  ) -> PooledStreamChannel {
+    let reservation = self.lock.withLock {
+      self.state.reserveStream(
+        preferringPoolWithEventLoopID: preferredEventLoop.map { EventLoopID($0) }
```
Lukasa, line 228: Can we lift this `preferredEventLoop.map` call out of the lock? It's nice to do as little as possible while holding that lock.

Sources/GRPC/ConnectionPool/PoolManager.swift
```diff
@@ -0,0 +1,314 @@
...
+      )
+    }
+
+    switch reservation {
+    case let .success(poolIndex):
+      let channel = self.pools[poolIndex.value].makeStream(
```
Lukasa, line 234: `self.pools` is `var`. Presumably we need to hold a lock here.

Sources/GRPC/ConnectionPool/PoolManager.swift
```diff
@@ -0,0 +1,314 @@
...
+  internal func shutdown(promise: EventLoopPromise<Void>) {
+    let (action, pools): (PoolManagerStateMachine.ShutdownAction, [ConnectionPool]?) = self.lock
+      .withLock {
+        let action = self.state.shutdown(promise: promise)
+
+        switch action {
+        case .shutdownPools:
+          // Clear out the pools; we need to shut them down.
+          let pools = self.pools
+          self.pools.removeAll()
```
Lukasa, line 259: Suggest using `keepingCapacity: true` to avoid Array thinking we want to free memory.

Lukasa (changes requested):

Sources/GRPC/ConnectionPool/PoolManager.swift
glbrntt: Great point; I appreciate the nitty comments like this.

Lukasa (approved):
Cool, :shipit:

PR #1192, opened 2021-05-26
Fix compiler crash on Swift 5.2 when inlining test helper

Davidde94:
Running the command
```
swift test -c release -Xswiftc -enable-testing --enable-test-discovery
```

On Linux with Swift 5.2 will result in a compiler crash. The bug in the compiler is to do with inlining a test helper. The solution is to make the helper never inline. This fixes the issue, so that the code is now testable on Swift 5.2.


Lukasa (commented):
Can we add a code comment to clarify why we're doing this?

Davidde94:
@Lukasa done :)

Lukasa (approved):

PR #1195, opened 2021-06-02
Add remote and local address metadata to loggers

glbrntt:
Motivation:

In some cases having the local and remote address for a connection in
log messages can make debugging easier.

Modifications:

- Add the local and remote address to logs in the idle handler, error
  handler and client transport
- Adopt `GRPCLogger` a little more widely to avoid having to explicitly
  set the `source`

Result:

More logging metadata

### DIFF
--- a/Sources/GRPC/GRPCLogger.swift
+++ b/Sources/GRPC/GRPCLogger.swift
@@ -14,10 +14,12 @@
  * limitations under the License.
  */
 import Logging
+import NIO
 
 /// Wraps `Logger` to always provide the source as "GRPC".
 ///
 /// See https://github.com/apple/swift-log/issues/145 for rationale.
+@usableFromInline
 internal struct GRPCLogger {
   private var logger: Logger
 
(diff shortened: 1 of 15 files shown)


Sources/GRPC/GRPCLogger.swift
```diff
@@ -71,6 +73,46 @@ internal struct GRPCLogger {
...
+    _ message: @autoclosure () -> Logger.Message,
+    metadata: @autoclosure () -> Logger.Metadata? = nil,
+    file: String = #file,
+    function: String = #function,
+    line: UInt = #line
+  ) {
+    self.logger.notice(
+      message(),
+      metadata: metadata(),
+      source: "GRPC",
```
Davidde94, line 87: Not `gRPC`? ;)

Davidde94 (approved):

Sources/GRPC/GRPCLogger.swift
glbrntt: Logging source is usually the module name.

glbrntt:
Looks like an allocation regression 👀

Lukasa (approved):
Allocation regression could well be the strings you need to construct to represent the socket addresses. Those will surely be longer than 15 bytes and so will need to heap-allocate.

PR #1201, opened 2021-06-08
Update Cocoapods after 1.1.0

glbrntt:


gRPC-Swift.podspec
```diff
@@ -19,7 +19,7 @@ Pod::Spec.new do |s|
 
     s.dependency 'CGRPCZlib', s.version.to_s
     s.dependency 'Logging', '>= 1.4.0', '< 2.0.0'
-    s.dependency 'SwiftNIO', '>= 2.22.0', '< 3.0.0'
+    s.dependency 'SwiftNIO', '>= 2.28.0', '< 3.0.0'
```
Davidde94, line 22: Should we do 2.29 as it's the latest version?

Lukasa (approved):

gRPC-Swift.podspec
glbrntt: No, this mirrors the requirements from `Package.swift` rather than the resolved version

PR #1211, opened 2021-07-05
Bump the minimum version of argument parser

glbrntt:


Lukasa (approved):

Lukasa (changes requested):
Can I suggest we use a range here? `"0.3.0" ..< "0.5.0"` would be best, I suspect.

Lukasa (approved):

PR #1213, opened 2021-07-05
Check Swift reserved names when generating function names

ialimz:
This PR fixes a bug that causes the Generator not to consider Swift reserved names such as the `do` keyword.

For example, assume we have an RPC like below:

```
service JsonRpc {
    rpc Do(Request) returns (Response) {}
}
```

If we don't consider Swift's `do` keyword, we will be going to have a compile error.

With these changes, the generated code should be like below:
```
public protocol JsonRpc_JsonRpcClientProtocol: GRPCClient {
  var serviceName: String { get }
  var interceptors: JsonRpc_JsonRpcClientInterceptorFactoryProtocol? { get }

  func `do`(
    _ request: JsonRpc_Request,
    callOptions: CallOptions?
  ) -> UnaryCall<JsonRpc_Request, JsonRpc_Response>
}
```

> The changes idea is the [same](https://github.com/apple/swift-protobuf/blob/4b11fbc4f2fea757d15fb6d392c7155f1e598bdd/Sources/SwiftProtobufPluginLibrary/NamingUtils.swift) as SwiftProtobuf library

### DIFF
--- a/Sources/protoc-gen-grpc-swift/Generator-Names.swift
+++ b/Sources/protoc-gen-grpc-swift/Generator-Names.swift
@@ -69,8 +69,7 @@ extension Generator {
     return nameForPackageServiceMethod(file, service, method) + "Call"
   }
 
-  internal let quotableFieldNames: Set<String> = {
-    () -> Set<String> in
+  internal let quotableFieldNames: Set<String> = { () -> Set<String> in
     var names: Set<String> = []
 
     names = names.union(swiftKeywordsUsedInDeclarations)
@@ -97,14 +96,13 @@ extension Generator {
     "Self", "throw", "throws", "true", "try"
   ]
 
-  internal let quotableFieldNames: Set<String> = {
-    () -> Set<String> in
+  internal let quotableFieldNames: Set<String> = { () -> Set<String> in
     var names: Set<String> = []
 
     names = names.union(swiftKeywordsUsedInDeclarations)
     names = names.union(swiftKeywordsUsedInStatements)
     names = names.union(swiftKeywordsUsedInExpressionsAndTypes)
-      return names
+    return names
   }()
 
   internal var methodFunctionName: String {
--- a/Sources/protoc-gen-grpc-swift/Generator-Names.swift
+++ b/Sources/protoc-gen-grpc-swift/Generator-Names.swift
@@ -32,6 +32,35 @@ internal func nameForPackageServiceMethod(_ file: FileDescriptor,
   return nameForPackageService(file, service) + method.name
 }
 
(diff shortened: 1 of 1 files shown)


Sources/protoc-gen-grpc-swift/Generator-Names.swift
```diff
@@ -69,13 +69,58 @@ extension Generator {
     return nameForPackageServiceMethod(file, service, method) + "Call"
   }
 
+  internal let quotableFieldNames: Set<String> = {
+    () -> Set<String> in
```
glbrntt, line 73:
```suggestion
  internal let quotableFieldNames: Set<String> = { () -> Set<String> in
```

Sources/protoc-gen-grpc-swift/Generator-Names.swift
```diff
@@ -69,13 +69,58 @@ extension Generator {
...
+  ]
+
+  internal let quotableFieldNames: Set<String> = {
+    () -> Set<String> in
+    var names: Set<String> = []
+
+    names = names.union(swiftKeywordsUsedInDeclarations)
+    names = names.union(swiftKeywordsUsedInStatements)
+    names = names.union(swiftKeywordsUsedInExpressionsAndTypes)
+      return names
```
glbrntt, line 107:
```suggestion
    return names
```

Sources/protoc-gen-grpc-swift/Generator-Names.swift
```diff
@@ -69,13 +69,58 @@ extension Generator {
...
+    "while"
+  ]
+
+  internal let swiftKeywordsUsedInExpressionsAndTypes: Set<String> = [ "as",
+    "Any", "catch", "false", "is", "nil", "rethrows", "super", "self",
+    "Self", "throw", "throws", "true", "try"
+  ]
+
+  internal let quotableFieldNames: Set<String> = {
+    () -> Set<String> in
```
glbrntt, line 101:
```suggestion
  internal let quotableFieldNames: Set<String> = { () -> Set<String> in
```

glbrntt (commented):
Thanks, this looks good although I've left a couple of nits.

ialimz:
Thanks @glbrntt 
I've committed your suggestions

glbrntt:
@ialimz could you run `./scripts/format.sh` to make the formatting checks happy?

glbrntt:
The other failures here are https://github.com/grpc/grpc-swift/issues/1215 -- we can ignore those for now.

ialimz:
@glbrntt SwiftFormat changes have been applied

glbrntt (approved):
Thanks @ialimz!

PR #1214, opened 2021-07-06
Migrate from travis.org to GitHub actions.

MrMage:


glbrntt (approved):
Looks great! I can see why you'd want to move away from Travis.

It looks like `testBidirectionalStreamingOnCloseAfterUserFunctionFails` is getting wedged, but given this PR doesn't change any of the library we can look into that separately.

PR #1221, opened 2021-07-12
Add support for Network.framework TLS options

glbrntt:
Motivation:

gRPC Swift currently half supports Network.framework via
NIOTransportServices, that is, we make use of the network stack where
it's available but we always use NIOSSL for TLS. We should offer users a
way to use Network.framework's TLS stack.

Modifications:

- Add 'GRPCTLSConfiguration' which is a wrapper around configuration
  structures for both NIOSSL and Network.framework. This is for both
  client and server configuration.
- Add static factory methods making it easier for callers to construct
  the right configuration with the right defauls.
- Add static factory methods allowing the caller to construct a
  `GRPCTLSConfiguration` from the 'base' configuration (i.e. a
  `NIOSSL.TLSConfiguration` or `NWProtocolTLS.Options`)
- Add new methods to the client and server builders to make a builder
  backed by a given TLS backend. In addition the client has a helper for
  automatically choosing the right backend from the type of the provided
  event loop group.
- Add backend specific options to each builder.
- Deprecate the generic 'secure' builders for both the client and the
  server.

Result:

Users can configure their client/server to use the TLS implementation
provided by Network.framework where appropriate.


Sources/GRPC/ClientConnection.swift
```diff
@@ -303,7 +303,18 @@ extension ClientConnection {
     public var connectivityStateDelegateQueue: DispatchQueue?
 
     /// TLS configuration for this connection. `nil` if TLS is not desired.
```
Lukasa, line 305: Is it worth updating the doc comment here to note the deprecation?

Sources/GRPC/ConnectionManagerChannelProvider.swift
```diff
@@ -76,32 +74,36 @@ internal struct DefaultChannelProvider: ConnectionManagerChannelProvider {
...
     // configuration. We do it now and surface any error during channel creation (we're limited by
     // our API in when we can throw any error).
-    let sslContext: Result<NIOSSLContext, Error>? = configuration.tls.map { tls in
-      return Result {
-        try NIOSSLContext(configuration: tls.configuration)
-      }
+    //
+    // 'nil' means we're not using TLS, or we're using the Network.framework TLS backend. We'll
+    // check and apply the Network.framework TLS options when we create a bootstrap.
+    let sslContext: Result<NIOSSLContext, Error>?
```
Lukasa, line 80: Hmm, `Optional<Result>` is sufficiently opaque that I'm never going to be quite sure what each thing means. I wonder if this needs a temporary type.

Sources/GRPC/ConnectionManagerChannelProvider.swift
```diff
@@ -117,25 +119,42 @@ internal struct DefaultChannelProvider: ConnectionManagerChannelProvider {
...
     let needsZeroLengthWriteWorkaround = self.requiresZeroLengthWorkaround(eventLoop: eventLoop)
 
-    let bootstrap = PlatformSupport.makeClientBootstrap(group: eventLoop, logger: logger)
+    let bootstrap = PlatformSupport.makeClientBootstrap(
+      group: eventLoop,
+      tlsConfiguration: self.tlsConfiguration,
+      logger: logger
+    )
+
+    _ = bootstrap
```
Lukasa, line 128: While I appreciate that our builder doesn't _actually_ require you follow the builder pattern, it would be _much_ less surprising if you actually chained off it (i.e. make `bootstrap` `var` instead of `let` and then assign it again here.

Sources/GRPC/GRPCTLSConfiguration.swift
```diff
@@ -0,0 +1,603 @@
...
+        config.hostnameOverride = newValue
+        self.backend = .nio(config)
+
+      #if canImport(Network)
+      case var .network(config):
+        if #available(macOS 10.14, iOS 12.0, watchOS 5.0, tvOS 12.0, *) {
+          if let hostnameOverride = newValue {
+            config.updateHostnameOverride(to: hostnameOverride)
+          } else {
+            // We can't unset the value so error instead.
```
Lukasa, line 98: In principle we could allow unsetting by lazily setting the value (i.e. set it at usage time). For now this is probably fine, but it may be worth documenting the caveat.

Sources/GRPC/GRPCTLSConfiguration.swift
```diff
@@ -0,0 +1,603 @@
...
+    return GRPCTLSConfiguration.makeClientConfigurationBackedByNIOSSL(
+      configuration: configuration,
+      hostnameOverride: hostnameOverride,
+      customVerificationCallback: customVerificationCallback
+    )
+  }
+
+  /// Creates a gRPC TLS Configuration using the given `NIOSSL.TLSConfiguration`.
+  ///
+  /// This gRPC TLS Configuration is only compatible with `MultiThreadedEventLoopGroup`.
```
Lukasa, line 238: Is that true?

Sources/GRPC/GRPCTLSConfiguration.swift
```diff
@@ -0,0 +1,603 @@
...
+    configuration.trustRoots = trustRoots
+    configuration.applicationProtocols = GRPCApplicationProtocolIdentifier.server
+
+    return GRPCTLSConfiguration.makeServerConfigurationBackedByNIOSSL(
+      configuration: configuration,
+      requireALPN: requireALPN
+    )
+  }
+
+  public static func makeServerConfigurationBackedByNIOSSL(
```
Lukasa, line 302: This lacks a comment.

Sources/GRPC/GRPCTLSConfiguration.swift
```diff
@@ -0,0 +1,603 @@
...
+    }
+
+    /// Always a NWProtocolTLS.Options.
+    ///
+    /// This somewhat insane type-erasure is necessary because we need to availability-guard the NWProtocolTLS.Options
+    /// (it isn't available in older SDKs), but we cannot have stored properties guarded by availability in this way, only
+    /// computed ones. To that end, we have to erase the type and then un-erase it. This is fairly silly.
+    private var _options: Any
+
+    // This is set privately via `updateHostnameOverride(to:)` because we availability guards to
```
Lukasa, line 414: This comment needs rewording.

Sources/GRPC/ConnectionManagerChannelProvider.swift
glbrntt: Yeah, it is a little unclear. I extracted this into a `TLSMode`.

Sources/GRPC/GRPCTLSConfiguration.swift
glbrntt: Good idea! I added FIXME and documented in the relevant places.

Lukasa (approved):

PR #1226, opened 2021-07-17
Add setting for max number of concurrent streams for HTTP/2 connection

qusc:
... in order to circumvent the limitation of a maximum of 100 concurrent calls as imposed by the default settings of NIOHTTP2Handler.

### DIFF
--- a/Tests/GRPCTests/HTTP2MaxConcurrentStreamsTests.swift
+++ b/Tests/GRPCTests/HTTP2MaxConcurrentStreamsTests.swift
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2021, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import EchoImplementation
+import EchoModel
+@testable import GRPC
+import NIO
+import NIOHTTP2
+import XCTest
+
+class HTTP2MaxConcurrentStreamsTests: GRPCTestCase {
+  struct Constants {
+    static let testTimeout: TimeInterval = 10
+
+    static let defaultMaxNumberOfConcurrentStreams =
+      nioDefaultSettings.first(where: { $0.parameter == .maxConcurrentStreams })!.value
+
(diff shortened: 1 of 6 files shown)


glbrntt (changes requested):
Thanks for the PR, this broadly looks good! I left a few notes inline but I think we'll also want a test for this. It could do something like:

1. start a server setting max concurrent streams to N (where N > 100)
2. connect to the server
3. start N client streaming (or bidirectional streaming) RPCs
4. send a message on each RPC (but do not send end)
5. send end on RPC N
6. wait for the status on RPC N (i.e. if this RPC does not complete then we didn't apply the setting correctly)
7. send on all other RPCs, wait for them to complete

qusc:
@glbrntt awesome, thanks for your comments, let me know what you think about my changes

Tests/GRPCTests/HTTP2MaxConcurrentStreamsTests.swift
```diff
@@ -0,0 +1,85 @@
...
+    static let testTimeout: TimeInterval = 10
+
+    static let defaultMaxNumberOfConcurrentStreams =
+      nioDefaultSettings.first(where: { $0.parameter == .maxConcurrentStreams })!.value
+
+    static let testNumberOfConcurrentStreams: Int = defaultMaxNumberOfConcurrentStreams + 20
+  }
+
+  func testHTTP2MaxConcurrentStreamsSetting() {
+    let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
```
glbrntt, line 34: Could you add a `defer` block here to shutdown the event loop group?

Tests/GRPCTests/HTTP2MaxConcurrentStreamsTests.swift
```diff
@@ -0,0 +1,85 @@
...
+
+    let server = try! Server.insecure(group: eventLoopGroup)
+      .withLogger(self.serverLogger)
+      .withHTTPMaxConcurrentStreams(Constants.testNumberOfConcurrentStreams)
+      .withServiceProviders([EchoProvider()])
+      .bind(host: "localhost", port: 0)
+      .wait()
```
glbrntt, line 41: Same here, can you add a `defer` block for shutting down the server?

Tests/GRPCTests/HTTP2MaxConcurrentStreamsTests.swift
```diff
@@ -0,0 +1,85 @@
...
+
+    let clientConnection = ClientConnection.insecure(group: eventLoopGroup)
+      .withBackgroundActivityLogger(self.clientLogger)
+      .connect(host: "localhost", port: server.channel.localAddress!.port!)
```
glbrntt, line 45: And one more `defer` here to close the connection

glbrntt (changes requested):
This looks great! We're just missing a bit of resource cleanup in the tests then this should be good to go.

glbrntt (approved):
Great, thank you @qusc! 🙏

PR #1228, opened 2021-07-20
Implement `_preconditionSafeToSyncShutdown` extension point

glbrntt:
Motivation:

NIO recently added a `_preconditionSafeToSyncShutdown` extension point
to `EventLoopGroup`. In our tests we have a custom `EventLoopGroup`
called `EmbeddedEventLoopGroup`. When testing against the `main` branch of
swift-nio we hit this precondition (because the default implementation
preconditions that we aren't on any of the event loops in the group, but
we are always on an `EmbeddedEventLoop`).

Modifications:

- Implement `_preconditionSafeToSyncShutdown` for `EmbeddedEventLoopGroup`

Result:

Tests shouldn't crash against NIOs `main` branch.

### DIFF
--- a/Tests/GRPCTests/ConnectionPool/PoolManagerStateMachineTests.swift
+++ b/Tests/GRPCTests/ConnectionPool/PoolManagerStateMachineTests.swift
@@ -347,6 +347,12 @@ private final class EmbeddedEventLoopGroup: EventLoopGroup {
     return EventLoopIterator(self.loops)
   }
 
+  internal func _preconditionSafeToSyncShutdown(file: StaticString, line: UInt) {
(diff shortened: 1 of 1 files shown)


Lukasa:
Yeah, I think we shouldn't merge this: this is an indication that the default implementation in NIO is bad. We should just do nothing there. I'll make a patch to NIO.

glbrntt:
> Yeah, I think we shouldn't merge this: this is an indication that the default implementation in NIO is bad. We should just do nothing there. I'll make a patch to NIO.

Oh I just took `EmbeddedEventLoop` to be a bit of a special case... don't mind either way.

Lukasa:
No, there's no reason not to be able to write an event loop group in this way IMO: the original enhancement is wrong, I think.

glbrntt:
Fair enough. I'll close this.

PR #1229, opened 2021-07-21
Use sync options to get the http2 stream ID

glbrntt:
Motivation:

NIO added API for getting channel options synchronously. NIO HTTP/2
added support for this in 1.17.0. When configuring a server we configure
our new stream channels to get the stream ID to attach to the loggers.
We can now do this synchronously and save some allocations.

Modifications:

- Bump minimum H2 version to 1.17.0
- Use sync options to get the stream ID and sync pipeline operations to
  add the handler.

Result:

Fewer allocations.


.github/workflows/ci.yaml
```diff
@@ -82,4 +82,4 @@ jobs:
     - name: 🧮 Allocation Counting Tests
       run: ./Performance/allocations/test-allocation-counts.sh
       env: ${{ matrix.env }}
-      timeout-minutes: 20
\ No newline at end of file
+      timeout-minutes: 20
```
Davidde94, line 85: Wowsers

Sources/GRPC/GRPCServerPipelineConfigurator.swift
```diff
@@ -103,18 +103,29 @@ final class GRPCServerPipelineConfigurator: ChannelInboundHandler, RemovableChan
...
-        logger[metadataKey: MetadataKey.h2StreamID] = "<unknown>"
-        return logger
-      }.flatMap { logger in
+      let streamID: Int?
+
+      // Sync options were added to the HTTP/2 stream channel in 1.17.0 (we require at least this)
+      // so this shouldn't be `nil`, but not a problem if it is.
+      if let syncOptions = stream.syncOptions {
+        do {
+          streamID = Int(try syncOptions.getOption(HTTP2StreamChannelOptions.streamID))
```
Davidde94, line 112: Nit: As you don't care about the error you could condense the code by using `try?` with a `flatmap`.

Sources/GRPC/GRPCServerPipelineConfigurator.swift
```diff
@@ -103,18 +103,29 @@ final class GRPCServerPipelineConfigurator: ChannelInboundHandler, RemovableChan
...
+        } catch {
+          streamID = nil
+        }
+      } else {
+        streamID = nil
```
Davidde94, line 117: Why not just start the `streamID` as nil?

Davidde94 (approved):
Some nits, otherwise looks good.

Sources/GRPC/GRPCServerPipelineConfigurator.swift
glbrntt: A few (minor) reasons: I prefer being explicit, similar to not having a `default` case in a `switch` you have to make sure you initialize it on each path. The other reason is that I don't need or want it to be mutable.

Sources/GRPC/GRPCServerPipelineConfigurator.swift
glbrntt: Good idea!

Sources/GRPC/GRPCServerPipelineConfigurator.swift
```diff
@@ -103,18 +103,22 @@ final class GRPCServerPipelineConfigurator: ChannelInboundHandler, RemovableChan
...
-        return logger
-      }.recover { _ in
-        logger[metadataKey: MetadataKey.h2StreamID] = "<unknown>"
-        return logger
-      }.flatMap { logger in
+      // Sync options were added to the HTTP/2 stream channel in 1.17.0 (we require at least this)
+      // so this shouldn't be `nil`, but it's not a problem if it is.
+      let http2StreamID = try? stream.syncOptions?.getOption(HTTP2StreamChannelOptions.streamID)
+      let streamID = http2StreamID.map { streamID in
+        return String(Int(streamID))
```
Davidde94, line 110: More nit, the return shouldn't be needed right? Entirely personal opinion, this would look really neat on one line. (I already approved, you can ignore).

glbrntt: I tend to prefer explicit returns so I'l leave as is.

fabianfett (approved):
LGTM

PR #1231, opened 2021-07-27
Proposal: Async-await client and server APIs

simonjbeaumont:
I have spent the last couple of weeks with a few folks at Apple thinking about how we might offer async/await APIs for this project. Specifically how users might be able to implement and call gRPC services using APIs that abstract away NIO and expose functionality in terms of the new native Swift concurrency features.

I have written up our thinking so far in the form of a proposal which is contained in this PR with the hope of gathering feedback from a wider audience. Anyone who has feedback should feel welcome to comment directly on the contents of this PR.

I also have a branch that implements the proposal for those who wish to take a closer look at how it looks and feels: https://github.com/grpc/grpc-swift/compare/main...simonjbeaumont:async-await-poc-squashed.

Thanks to @glbrntt for the help so far!

/cc @glbrntt @Lukasa @fabianfett @PeterAdams-A

### DIFF
--- a/docs/async-await-proposal.md
+++ b/docs/async-await-proposal.md
@@ -1,0 +1,442 @@
+# Proposal: Async/await support
+
+## Introduction
+
+With the introduction of [async/await][SE-0296] in Swift 5.5, it is
+now possible to write asynchronous code without the need for callbacks.
+Language support for [`AsyncSequence`][SE-0298] also allows for writing
+functions that return values over time.
+
+We would like to explore how we could offer APIs that make use of these new
+language features to allow users to implement and call gRPC services using
+these new idioms.
+
+This proposal describes what these APIs could look like and explores some of
+the potential usability concerns.
+
+## Motivation
+
+Consider the familiar example Echo service which exposes all four types of
+call: unary, client-streaming, server-streaming, and bidirectional-streaming.
+It is defined as follows:
+
+### Example Echo service
+
+```proto
+service Echo {
+  // Immediately returns an echo of a request.
+  rpc Get(EchoRequest) returns (EchoResponse) {}
+
+  // Splits a request into words and returns each word in a stream of messages.
+  rpc Expand(EchoRequest) returns (stream EchoResponse) {}
+
+  // Collects a stream of messages and returns them concatenated when the caller closes.
+  rpc Collect(stream EchoRequest) returns (EchoResponse) {}
+
+  // Streams back messages as they are received in an input stream.
+  rpc Update(stream EchoRequest) returns (stream EchoResponse) {}
+}
+
+message EchoRequest {
+  // The text of a message to be echoed.
+  string text = 1;
+}
+
+message EchoResponse {
+  // The text of an echo response.
+  string text = 1;
+}
+```
+
+### Existing server API
+
+To implement the Echo server, the user must implement a type that conforms to
+the following generated protocol:
+
+```swift
+/// To build a server, implement a class that conforms to this protocol.
+public protocol Echo_EchoProvider: CallHandlerProvider {
+  var interceptors: Echo_EchoServerInterceptorFactoryProtocol? { get }
+
+  /// Immediately returns an echo of a request.
+  func get(request: Echo_EchoRequest, context: StatusOnlyCallContext) -> EventLoopFuture<Echo_EchoResponse>
+
+  /// Splits a request into words and returns each word in a stream of messages.
+  func expand(request: Echo_EchoRequest, context: StreamingResponseCallContext<Echo_EchoResponse>) -> EventLoopFuture<GRPCStatus>
+
+  /// Collects a stream of messages and returns them concatenated when the caller closes.
+  func collect(context: UnaryResponseCallContext<Echo_EchoResponse>) -> EventLoopFuture<(StreamEvent<Echo_EchoRequest>) -> Void>
+
+  /// Streams back messages as they are received in an input stream.
+  func update(context: StreamingResponseCallContext<Echo_EchoResponse>) -> EventLoopFuture<(StreamEvent<Echo_EchoRequest>) -> Void>
+}
+```
+
+### Existing example server implementation
+
+Here is an example implementation of the bidirectional streaming handler for `update`:
+
+```swift
+public func update(
+  context: StreamingResponseCallContext<Echo_EchoResponse>
+) -> EventLoopFuture<(StreamEvent<Echo_EchoRequest>) -> Void> {
+  var count = 0
+  return context.eventLoop.makeSucceededFuture({ event in
+    switch event {
+    case let .message(message):
+      let response = Echo_EchoResponse.with {
+        $0.text = "Swift echo update (\(count)): \(message.text)"
+      }
+      count += 1
+      context.sendResponse(response, promise: nil)
+
+    case .end:
+      context.statusPromise.succeed(.ok)
+    }
+  })
+}
+```
+
+This API exposes a number incidental types and patterns that the user need
+concern themselves with that are not specific to their application:
+
+1. The fact that gRPC is implemented on top of NIO is not hidden from the user
+   and they need to implement an API in terms of an `EventLoopFuture` and access
+   an `EventLoop` from the call context.
+2. There is a different context type passed to the user function for each
+   different type of call and this context is generic over the response type.
+3. In the server- and bidirectional-streaming call handlers, an added layer of
+   asynchrony is exposed. That is, the user must return a _future_ for
+   a closure that will handle incoming events.
+4. The user _must_ fulfil the `statusPromise` when it receives `.end`, but there
+is nothing that enforces this.
+
+### Existing client API
+
+Turning our attention to the client API, in order to make calls to the Echo server, the user must instantiate the generated `Echo_EchoClient` which provides the following API:
+
+```swift
+public protocol Echo_EchoClientProtocol: GRPCClient {
+  var serviceName: String { get }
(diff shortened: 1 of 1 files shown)


Lukasa:
ping @weissi and @ktoso, both of whom likely have thoughts about this as well.

weissi:
Thanks you, this looks really great to me!

docs/async-await-proposal.md
```diff
@@ -0,0 +1,442 @@
...
+    requests: GRPCAsyncStream<Echo_EchoRequest>,
+    context: AsyncServerCallContext
+  ) async throws -> Echo_EchoResponse
+
+  /// Streams back messages as they are received in an input stream.
+  func update(
+    requests: GRPCAsyncStream<Echo_EchoRequest>,
+    responseStreamWriter: AsyncResponseStreamWriter<Echo_EchoResponse>,
+    context: AsyncServerCallContext
+  ) async throws
```
ktoso, line 215: This boils down to opinions™ but worth considering that this still allows the error of forgetting to complete the response stream etc.

We went through the same design dance way back when with akka-grpc and went with accepting and emitting a stream (`Source` in Akka's wording); https://doc.akka.io/docs/akka-grpc/current/server/walkthrough.html#implementing-the-service

It may be that Swift is still too rigid to allow that... we may first need the ability to say `some AsyncSequence where Element == Echo_EchoResponse` to be honest... 🤔 I believe this feature is coming "soon". 

But just throwing it out there for consideration.

docs/async-await-proposal.md
```diff
@@ -0,0 +1,442 @@
...
+) async throws {
+  var count = 0
+  for try await request in requests {
+    let response = Echo_EchoResponse.with {
+      $0.text = "Swift echo update (\(count)): \(request.text)"
+    }
+    count += 1
+    try await responseStreamWriter.sendResponse(response)
+  }
+}
```
ktoso, line 236: Hm now that I think about it (after the comment above) I guess the writer is easier here... but you may want to figure out how you want to implement the backpressure here -- would you want to suspend here until more demand is abailable to push more responses? Internally I guess this would be implemented the ill named but final `AsyncStream` API, so you'll have all the tools to notice backpressure on that boundary, so that's good news 👍

docs/async-await-proposal.md
ktoso: (I amend this opinion a bit below)

docs/async-await-proposal.md
```diff
@@ -0,0 +1,442 @@
...
+>   to access an `EventLoop` from the call context.
+
+There is no longer a need for the adopter to `import NIO` nor implement anything
+in terms of NIO types. Instead they now implement an `async` function.
+
+> 2. There is a different context type passed to the user function for each
+>   different type of call and this context is generic over the response type.
+
+The same non-generic `AsyncServerCallContext` is passed to the user function
+regardless of the type of RPC.
```
ktoso, line 252: 👍

docs/async-await-proposal.md
```diff
@@ -0,0 +1,442 @@
...
+>   a closure that will handle incoming events.
+
+The user function consumes requests from an `AsyncSequence`, using the new
+language idioms.
+
+> 4. The user _must_ fulfil the `statusPromise` when it receives `.end` but there
+>   is nothing that enforces this.
+
+The user need simply return from the function or throw an error. The closing of
+the call is handled by the library.
```
ktoso, line 265: 👍

docs/async-await-proposal.md
```diff
@@ -0,0 +1,442 @@
...
+
+If the user function throws a `GRPCStatus` (which already conforms to `Error`)
+or a value of a type that conforms to `GRPCStatusTransformable` then the library
+will take care of setting the RPC status appropriately. If the user throws
+anything else then the library will still take care of setting the status
+appropriately, but in this case it will use `internalError` for the RPC status.
```
ktoso, line 271: nitpick, perhaps `GRPCStatusError` if it is intended to be thrown...? 🤔

docs/async-await-proposal.md
```diff
@@ -0,0 +1,442 @@
...
+
+// Consume responses as an AsyncSequence.
+for try await response in update.responseStream {
+  print("update received: \(response.text)")
+}
+
+// Wait for the call to terminate, but using `await` rather than a future.
+let status = await update.status
+print("update completed with status: \(status.code)")
+```
```
ktoso, line 326: This all looks good 👍 

Not in love with `sendEnd()` I guess (one can miss sending it), but it matches the general style of the stream writer so I guess it's fine 👍

docs/async-await-proposal.md
```diff
@@ -0,0 +1,442 @@
...
+It is for this reason that the above proposed client APIs have slightly obscured
+names (e.g. `makeUpdateCall` instead of `update`) and we propose also generating
+additional, less expressive, but safer APIs. These APIs will not expose the RPC
+metadata (e.g. the status, initial metadata, trailing metadata), but will
+instead either simply return the response(s) or throw an error.
+
+In addition to avoiding the pitfall of the expressive counterparts, the client-
+and bidirectional-streaming calls provide the ability to pass an `AsyncSequence`
+of requests. In this way, the underlying library takes care of ensuring that no
+part of the RPC goes unterminated (both the request and response streams). It
```
ktoso, line 367: ok good, this addresses my annoyance with `sendEnd` mentioned above, good 👍

docs/async-await-proposal.md
```diff
@@ -0,0 +1,442 @@
...
+  public func expand(
+    _ request: Echo_EchoRequest,
+    callOptions: CallOptions? = nil
+  ) -> GRPCAsyncStream<Echo_EchoResponse> { ... }
+
+  public func update<RequestStream>(
+    requests: RequestStream,
+    callOptions: CallOptions? = nil
+  ) -> GRPCAsyncStream<Echo_EchoResponse>
+  where RequestStream: AsyncSequence, RequestStream.Element == Echo_EchoRequest { ... }
```
ktoso, line 401: Signature looks good (that's the "akka grpc way" I linked/mentioned above, very happy to see it be offered as well here) 👍

ktoso (approved):
This looks very good, thanks for the writeup! I think it'll fit the concurrency story very well 👍 

Tracing should also be able to integrate here; I'm not sure about the "return response stream"-one yet but we'll figure that out (it's a general streaming + tracing question, not specific to grpc)

docs/async-await-proposal.md
Lukasa: Stream writer probably handles backpressure by refusing to return from the write call when the window is full. This doesn't prevent multithreaded writing to the stream writer so we need to tolerate that, but this still works pretty well.

Also, the stream writer is a base abstraction: we can absolutely use it to implement an API that accepts an AsyncSequence.

ktoso: > Stream writer probably handles backpressure by refusing to return from the write call when the window is full.

Ok good 👍 

> Also, the stream writer is a base abstraction: we can absolutely use it to implement an API that accepts an AsyncSequence.

Yeah, good -- I saw that bit later on in the proposal, happy to see it 👍

docs/async-await-proposal.md
simonjbeaumont: I agree that we would like `some AsyncSequence where Element == ...`. George and I were mulling over the same as this would allow us to use opaque return type without having to provide the `GRPCAsyncStream` wrapper in this proposal.

docs/async-await-proposal.md
simonjbeaumont: I could add a sentence or two about backpressure to this doc. As you noted, there is no attention given to it. This was a deliberate omission from the prototype because the underlying NIO gRPC library doesn't handle backpressure today. But, as discussed in this thread, we should be able to accommodate backpressure with this API.

docs/async-await-proposal.md
simonjbeaumont: That's a fair callout. `GRPCStatus` already exists in the API and already conforms to `Error` so I was working with that. It isn't only used on the failure path which is why I guess it does not have `Error` in the name.

I suppose we could wrap the `GRPCStatus` in a struct with a more explicit name. @glbrntt do you have any thoughts on this?

docs/async-await-proposal.md
simonjbeaumont: Yep, this is why we provide the "simpler" API variants below as you already noticed in https://github.com/grpc/grpc-swift/pull/1231#discussion_r678802185.

docs/async-await-proposal.md
glbrntt: I agree `GRPCStatusError` is a better name (and shouldn't allow an 'ok' status) but personally I don't think we buy ourselves much by wrapping a `GRPCStatus`. 

The existing name would still be present in various places throughout the API (we could deprecate it but it's quite pervasive so I'd rather not) and having two very similar types would make the overall picture less clear, I think.

We should revisit when we do a major breaking change though.

docs/async-await-proposal.md
simonjbeaumont: I added a note about the desire for an opaque `AsyncSequence` return type in https://github.com/grpc/grpc-swift/pull/1231/commits/da20f7d14a74af1ca521a75ffab171600596944b.

docs/async-await-proposal.md
simonjbeaumont: I added a note about (the lack of) backpressure support in https://github.com/grpc/grpc-swift/pull/1231/commits/71d3728062afe2afd850f92df825650e5cdb813a.

glbrntt (approved):

Issue #1236, opened 2021-08-12
Get path property of call


ialimz:
### What are you trying to achieve?

How can I access `path` property of `call`?
I'd like to setup performance monitoring and need to merge `path` property into `host` in order to have a valid identifier for monitoring

### What have you tried so far?

I see that `Call` property of `UnaryCall` struct has `path` property, but it's private and I can't access it.

glbrntt:
Interceptors might be the way to do this (they make the path available to you but not the host, you'd have to pass that in).

Alternatively I'm not opposed to making the `path` available to the `{Unary,ClientStreaming,...}Call`s.

ialimz:
Thanks @glbrntt.
Since I have not access to interceptors in the performance monitor layer, I prefer to have `path` available in `call` object.
I'll create a PR in next few hours and make these changes.

PR #1237, opened 2021-08-12
Make path available in ClientCalls

ialimz:
This PR's purpose is making `path` variable available in `ClientCall`s.

Related Issue: #1236


glbrntt (approved):
Great, thanks @ialimz!

PR #1238, opened 2021-08-18
Prepare CI for async/await.

MrMage:
Motivation: To test async/await, we actually need a Swift version in CI that supports it.


glbrntt (approved):
Looks great, thank you!

glbrntt (approved):

PR #1239, opened 2021-08-19
Use more specific NIO imports

glbrntt:
Motivation:

The `NIO` module is now a 'shell' module which re-exports `NIOCore`,
`NIOPosix` and `NIOEmbeeded`.

Modifications:

For the GRPC and GRPCTests modules:
- Replace "import NIO" with "import NIOCore"
- Add imports for NIOEmbeeded and NIOPosix as necessary

Result:

Imports are closer to what we actually use.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -26,13 +26,13 @@ let package = Package(
   dependencies: [
     // GRPC dependencies:
     // Main SwiftNIO package
-    .package(url: "https://github.com/apple/swift-nio.git", from: "2.28.0"),
+    .package(url: "https://github.com/apple/swift-nio.git", from: "2.32.0"),
     // HTTP2 via SwiftNIO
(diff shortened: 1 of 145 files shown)


Package.swift
```diff
@@ -56,6 +56,9 @@ let package = Package(
       name: "GRPC",
       dependencies: [
         .product(name: "NIO", package: "swift-nio"),
+        .product(name: "NIOCore", package: "swift-nio"),
+        .product(name: "NIOPosix", package: "swift-nio"),
+        .product(name: "NIOEmbedded", package: "swift-nio"),
```
fabianfett, line 61: I know that this is to setup a client for a throughput test, but linking `NIOEmbedded` looks wrong :) maybe we can change this with a follow up PR.

fabianfett (approved):
We should really try to not link `NIOEmbedded` in `GRPC` in the future

PR #1241, opened 2021-08-19
Use more specific NIO imports (#1239)

glbrntt:
Motivation:

The `NIO` module is now a 'shell' module which re-exports `NIOCore`,
`NIOPosix` and `NIOEmbeeded`.

Modifications:

For the GRPC and GRPCTests modules:
- Replace "import NIO" with "import NIOCore"
- Add imports for NIOEmbeeded and NIOPosix as necessary

Result:

Imports are closer to what we actually use.

(cherry picked from commit c88f038cdee150543ffcbb44a987db58abaf4d9c)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -26,13 +26,13 @@ let package = Package(
   dependencies: [
     // GRPC dependencies:
     // Main SwiftNIO package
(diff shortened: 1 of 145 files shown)


fabianfett (approved):
Have fun!

PR #1245, opened 2021-08-20
Add an internal pausable writer

glbrntt:
Motivation:

Async RPCs which stream outbound messages (i.e. requests on the client,
responses on the server) will be provided with a means to asynchronously
send messages. This object should suspend if the underlying channel is
not currently writable.

Modifications:

Add an 'AsyncWriter' as the underlying type for these objects which
suspends if the writer is 'paused' and resumes when the writer is
resumed.

Result:

We have a writer capable of suspending writes.

### DIFF
--- a/Sources/GRPC/AsyncAwaitSupport/AsyncWriter.swift
+++ b/Sources/GRPC/AsyncAwaitSupport/AsyncWriter.swift
@@ -13,9 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 #if compiler(>=5.5)
-import NIOConcurrencyHelpers
 import NIOCore
 
 /// An asynchronous writer which forwards messages to a delegate.
--- a/Sources/GRPC/AsyncAwaitSupport/AsyncWriter.swift
+++ b/Sources/GRPC/AsyncAwaitSupport/AsyncWriter.swift
@@ -1,0 +1,314 @@
+/*
+ * Copyright 2021, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#if compiler(>=5.5)
+import NIOConcurrencyHelpers
+import NIOCore
+
+/// An asynchronous writer which forwards messages to a delegate.
+///
+/// Forwarding of messages to the delegate may be paused and resumed by controlling the writability
+/// of the writer. This may be controlled by calls to ``toggleWritability()``. When the writer is
+/// paused (by becoming unwritable) calls to ``write(_:)`` may suspend. When the writer is resumed
+/// (by becoming writable) any calls which are suspended may be resumed.
+///
+/// The writer must also be "finished" with a final value: as for writing, calls to ``finish(_:)``
+/// may suspend if the writer has been paused.
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+@usableFromInline
+internal struct AsyncWriter<Delegate: AsyncWriterDelegate> {
+  @usableFromInline
+  internal typealias Element = Delegate.Element
+
+  @usableFromInline
+  internal typealias End = Delegate.End
+
+  @usableFromInline
+  internal let _storage: AsyncWriterStorage<Delegate>
+
+  @inlinable
+  internal init(maxPendingElements: Int = 16, maxWritesBeforeYield: Int = 5, delegate: Delegate) {
+    self._storage = AsyncWriterStorage(
+      maxPendingElements: maxPendingElements,
+      maxWritesBeforeYield: maxWritesBeforeYield,
+      delegate: delegate
+    )
+  }
+
+  /// Toggles whether the writer is writable or not. The writer is initially writable.
+  ///
+  /// If the writer becomes writable then it may resume writes to the delegate. If it becomes
+  /// unwritable then calls to `write` may suspend until the writability changes again.
+  ///
+  /// This API does not offer explicit control over the writability state so the caller must ensure
+  /// calls to this function correspond with changes in writability. The reason for this is that the
+  /// underlying type is an `actor` and updating its state is therefore asynchronous. However,
+  /// this functions is not called from an asynchronous context so it is not possible to `await`
+  /// state updates to complete. Instead, changing the state on the underlying type is via
+  /// a `nonisolated` function on the `actor` which spawns a new task. If this or a similar API
+  /// allowed the writability to be explicitly set then calls to that API are not guaranteed to be
+  /// ordered which may lead to deadlock.
+  @usableFromInline
+  internal func toggleWritability() {
+    self._storage.toggleWritability()
+  }
+
+  /// As ``toggleWritability`` but `async` -- intended for testing only.
+  internal func testingOnly_toggleWritability() async {
+    await self._storage._toggleWritability()
+  }
+
+  /// Write an `element`.
+  ///
+  /// The call may be suspend if the writer is paused.
+  ///
+  /// Throws: ``AyncWriterError`` if the writer has already been finished or too many write tasks
+  ///   have been suspended.
+  @inlinable
+  internal func write(_ element: Element) async throws {
+    try await self._storage.write(element)
+  }
+
+  /// Write the final element
+  @inlinable
+  internal func finish(_ end: End) async throws {
+    try await self._storage.finish(end)
+  }
+}
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+@usableFromInline
+internal final actor AsyncWriterStorage<Delegate: AsyncWriterDelegate> {
+  @usableFromInline
+  internal typealias Element = Delegate.Element
+
+  @usableFromInline
+  internal typealias End = Delegate.End
+
+  /// A value pending a write.
+  @usableFromInline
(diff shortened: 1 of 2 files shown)


glbrntt:
cc @simonjbeaumont

Sources/GRPC/Concurrency/AsyncWriter.swift
```diff
@@ -0,0 +1,257 @@
...
+/// Any calls which are suspended may be continued at later (after ``resume()`` has been called).
+///
+/// The writer must also be "finished" with a final value: as for writing, calls to ``finish(_:)``
+/// may suspend if the writer has been paused.
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+internal struct AsyncWriter<
+  Element,
+  End,
+  Delegate: AsyncWriterDelegate
+> where Delegate.Element == Element, Delegate.End == End {
```
fabianfett, line 34:
```suggestion
internal struct AsyncWriter<Delegate: AsyncWriterDelegate> {
```

And then just use `Delegate.Element` and `Delegate.End` in the function signatures.

fabianfett: This makes stuff IMHO easier to read. Since it shows right away where this `End` and `Element` comes from

Sources/GRPC/Concurrency/AsyncWriter.swift
```diff
@@ -0,0 +1,257 @@
...
+  }
+}
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+fileprivate final class AsyncWriterStorage<
+  Element,
+  End,
+  Delegate: AsyncWriterDelegate
+> where Delegate.Element == Element, Delegate.End == End {
+  /// A value pending a write.
```
fabianfett, line 76: Not a hill I will die on, but if we put the `Storage` into the `AsyncWriter`, we don't need to redefine the types here, since they would be inherited from their enclosing struct.

Sources/GRPC/Concurrency/AsyncWriter.swift
```diff
@@ -0,0 +1,257 @@
...
+  private var delegate: Delegate
+
+  fileprivate init(maxPendingElements: Int, delegate: Delegate) {
+    self.lock = Lock()
+    self.maxPendingElements = maxPendingElements
+    self.pendingElements = CircularBuffer(initialCapacity: maxPendingElements)
+    self.completionState = .incomplete
+    self.delegate = delegate
+  }
+
```
fabianfett, line 142: I think we need a `deinit` that validates final state. If the Storage is released and the user has not called `finish` is there some message we want to emit?

glbrntt: That's a good point -- we _probably_ want an assertion/precondition.

Sources/GRPC/Concurrency/AsyncWriter.swift
glbrntt: It'd rather have this than the extra level of indentation tbh!

fabianfett: Okay, In this case we should still remove the `Element` and `End` from the generic parameters. They are implicit in the `Delegate`.

Sources/GRPC/Concurrency/AsyncWriter.swift
```diff
@@ -0,0 +1,266 @@
...
+  /// Write the final element
+  internal func finish(_ end: End) async throws {
+    try await self.storage.finish(end)
+  }
+}
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+fileprivate final class AsyncWriterStorage<
```
fabianfett, line 70: Did you consider making this an actor?

Sources/GRPC/Concurrency/AsyncWriter.swift
```diff
@@ -0,0 +1,266 @@
...
+  private let storage: AsyncWriterStorage<Element, End, Delegate>
+
+  internal init(maxPendingElements: Int = 16, delegate: Delegate) {
+    self.storage = AsyncWriterStorage(maxPendingElements: maxPendingElements, delegate: delegate)
+  }
+
+  /// Pauses the writer to stop value from being written to the delegate.
+  ///
+  /// This will cause subsequent calls to ``write(_:)`` to suspend until the writer is resumed.
+  internal func pause() {
```
fabianfett, line 43: Maybe we should mention that this method is normally called, by the Delegate to signal backpressure.

Sources/GRPC/Concurrency/AsyncWriter.swift
glbrntt: No -- `pause` and `resume` won't be called from asynchronous contexts.

Sources/GRPC/Concurrency/AsyncWriter.swift
glbrntt: Agh, yes, missed this one.

glbrntt: FWIW I think this change wouldn't be great for public APIs since Xcode suggests the "full" type name rather than any type aliases, e.g.: `"AsyncWriter<DelegateImplType<String, Int>>.Element"` rather than `"Element"`.

Sources/GRPC/Concurrency/AsyncWriter.swift
```diff
@@ -0,0 +1,266 @@
...
+      case .incomplete:
+        return false
+      case .pending, .completed:
+        return true
+      }
+    }
+  }
+
+  /// A lock which must be help when accessing or modifying mutable state held by this storage.
+  private var lock: Lock
```
fabianfett, line 115: Why is this a var?

Sources/GRPC/Concurrency/AsyncWriter.swift
```diff
@@ -0,0 +1,266 @@
...
+  fileprivate func resume() {
+    self.lock.withLockVoid {
+      self.isPaused = false
+    }
+
+    while let nextWrite = self.nextWrite() {
+      switch nextWrite {
+      case let .element(pendingElement):
+        self.delegate.write(pendingElement.value)
+        pendingElement.continuation.resume()
```
fabianfett, line 167: I'm not sure about this. Let's consider we have a buffer full with 16 writes waiting. After we have written the first element, do we want to allow more writes from the users side? There are still 15 writes pending... 

Further this is in a non detached loop. This means we will def. write all 16 items, and we will signal to everyone waiting, you can write again...

If this was an actor, we could consider using `await Task.yield` to check after a number of writes if the target is still writable and only in those cases to start succeed the continuations.

Sources/GRPC/Concurrency/Result+AsyncAwait.swift
```diff
@@ -0,0 +1,30 @@
...
+    do {
+      let success = try await operation()
+      self = .success(success)
+    } catch {
+      self = .failure(error)
+    }
+  }
+}
+
+#endif // compiler(>=5.5)
```
fabianfett, line 30: Is this used anywhere? CMD+F doesn't show me a use.

Tests/GRPCTests/Concurrency/AsyncWriterTests.swift
```diff
@@ -0,0 +1,168 @@
...
+
+      writer.pause()
+      async let written1: Void = writer.write("wunch")
+      async let written2: Void = writer.write("meat")
+      XCTAssert(delegate.elements.isEmpty)
+
+      writer.resume()
+      try await written1
+      try await written2
+      XCTAssertEqual(delegate.elements.sorted(), ["wunch", "meat"].sorted())
```
fabianfett, line 52: Why do we have `sorted()` here? I would expect that we check that the elements arrive in order here.

Tests/GRPCTests/Concurrency/AsyncWriterTests.swift
```diff
@@ -0,0 +1,168 @@
...
+
+  func testQueueUpEverything() throws {
+    XCTAsyncTest {
+      let delegate = CollectingDelegate<String, Int>()
+      let writer = AsyncWriter(delegate: delegate)
+
+      writer.pause()
+
+      async let w1: Void = writer.write("hitchcock")
+      await Task.yield()
```
fabianfett, line 72: why do we have the `Task.yield()` here?

Tests/GRPCTests/Concurrency/AsyncWriterTests.swift
```diff
@@ -0,0 +1,168 @@
...
+      writer.pause()
+
+      await XCTAssertThrowsError {
+        try await writer.write("pontiac")
+      } verify: { error in
+        XCTAssertEqual(error as? AsyncWriterError, .tooManyPendingWrites)
+      }
+
+      writer.resume()
+      try await writer.finish(0)
```
fabianfett, line 105: Maybe check `delegate.element.isEmpty`

Tests/GRPCTests/Concurrency/AsyncWriterTests.swift
```diff
@@ -0,0 +1,168 @@
...
+      let writer = AsyncWriter(delegate: delegate)
+
+      try await writer.finish(0)
+      XCTAssertEqual(delegate.end, 0)
+
+      await XCTAssertThrowsError {
+        try await writer.write("cheddar")
+      } verify: { error in
+        XCTAssertEqual(error as? AsyncWriterError, .alreadyFinished)
+      }
```
fabianfett, line 121: Maybe check that write was not forwarded to delegate

Tests/GRPCTests/Concurrency/XCTAwait.swift
```diff
@@ -0,0 +1,59 @@
...
+      }
+      expectation.fulfill()
+    }
+    self.wait(for: [expectation], timeout: timeout)
+  }
+}
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+internal func XCTAssertThrowsError<T>(
+  _ expression: @escaping () async throws -> T,
```
fabianfett, line 46: just out of interest for AHC, can we make this an `@autoclosure`?

fabianfett (approved):
Great! Really like how it turned out! Thanks!

PR #1247, opened 2021-08-25
Improve error message when length exceeds receive limit

glbrntt:
Motivation:

The error message emitted when a received message exceeds the configured
receive limit is non-obvious which make diagnosis more difficult than it
needs to be.

Modifications:

- Add a new error rather than relying on a deserialization error
- Use the 'resource exhausted' status code rather than 'internal error'
  when converting to a grpc status
- Update tests.

Result:

More helpful error when received messages exceed permitted limit


Sources/GRPC/GRPCError.swift
```diff
@@ -92,6 +92,19 @@ public enum GRPCError {
     }
   }
 
+  /// The length of the received payload was longer than is permitted.
```
fabianfett, line 95: If we want to be really nice to the user, we could point them to the config struct that they need to tune.

Sources/GRPC/GRPCClientChannelHandler.swift
```diff
@@ -463,6 +463,8 @@ extension GRPCClientChannelHandler: ChannelInboundHandler {
       case let .decompressionLimitExceeded(compressedSize):
         return GRPCError.DecompressionLimitExceeded(compressedSize: compressedSize)
           .captureContext()
+      case let .lengthExceedsLimit(underlyingError):
+        return underlyingError.captureContext()
```
fabianfett, line 467: When we create the error in `LengthPrefixedMessageReader.swift` we already capture the context. Is it right, that we capture the context here again, even though we pass the original error onwards?

fabianfett (approved):
Two small notes... Looks good though.

PR #1248, opened 2021-08-25
Raise the default HTTP/2 target window size

glbrntt:
Motivation:

The default flow control window size is 64kb which is quite small. As a
result connections may spend longer than necessary waiting for
window updates.

Modifications:

Increase the target window size to 8MB.

Result:

HTTP target window size is 8MB by default.


fabianfett (approved):
LGTM. 8MB is exactly the value that Cory proposed for AHC.

fabianfett (approved):
💨

PR #1249, opened 2021-08-25
[async-await] Base types for server implementation

simonjbeaumont:
This PR implements some of the types required by the proposal for async/await support, added in #1231.

To aid reviewing, only the types required for the server are included. They have been pulled in from the proof-of-concept implementation linked from the proposal PR. It is a complimentary PR to #1243 ("Async-await: Base types for client implementation").

It provides a unified `AsyncServerHandler` for all of the RPC types which avoids a substantial amount of code duplication that is found in the existing handlers. Wrappers are provided for the four RPC types. Otherwise it is analogous to the existing `BidirectionalStreamingServerHandler`.

It's worth calling out that this PR makes use of some placeholder types which are not intended to be final. Specifically:

* `AsyncResponseStreamWriter` is expected to be superseded by the `AsyncWriter` from #1245.
* `AsyncServerCallContext` conformance has been added to the existing `ServerCallContextBase`. It is expected that we will provide a new implementation of `AsyncServerCallContext` that is independent from the existing call context types.


Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,519 @@
...
+ */
+import _NIOConcurrency
+import Logging
+import NIOCore
+import NIOHPACK
+
+#if compiler(>=5.5)
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+public final class GRPCAsyncServerHandler<
```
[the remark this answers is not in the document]
simonjbeaumont: > Who creates an `GRPCAsyncServerHandler`?

It's created in in generated code (an example of which is not present in this PR). The generated code will make use of the initializers defined in the extension near the bottom of this file: https://github.com/grpc/grpc-swift/pull/1249/files#diff-fcf632a779a41e9ebc055fb40ac26d96379a72ca8506cc6e43d022834b9b287fR402-R517.

> For which request/response type is this class used? For all? Only unary?

There is one such initializer per RPC type, so this is a common handler for all of them, which was something @glbrntt was keen we try and do with this API (the existing one has different handler API types per RPC type which we wanted to avoid).

Taking the Echo service, the codegen would generate an `Echo_AsyncEchoProvider` protocol that they would need to conform to (provide just the user functions). The generated code would also contain the following extension which glues it all together:

```swift
@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
extension Echo_AsyncEchoProvider {
  public var serviceName: Substring { return "echo.Echo" }

  /// Determines, calls and returns the appropriate request handler, depending on the request's method.
  /// Returns nil for methods not handled by this service.
  public func handle(
    method name: Substring,
    context: CallHandlerContext
  ) -> GRPCServerHandlerProtocol? {
    switch name {
    case "Get":
      return AsyncServerHandler(
        context: context,
        requestDeserializer: ProtobufDeserializer<Echo_EchoRequest>(),
        responseSerializer: ProtobufSerializer<Echo_EchoResponse>(),
        interceptors: self.interceptors?.makeGetInterceptors() ?? [],
        wrapping: self.get(request:context:)
      )

    case "Expand":
      return AsyncServerHandler(
        context: context,
        requestDeserializer: ProtobufDeserializer<Echo_EchoRequest>(),
        responseSerializer: ProtobufSerializer<Echo_EchoResponse>(),
        interceptors: self.interceptors?.makeExpandInterceptors() ?? [],
        wrapping: self.expand(request:responseStreamWriter:context:)
      )

    case "Collect":
      return AsyncServerHandler(
        context: context,
        requestDeserializer: ProtobufDeserializer<Echo_EchoRequest>(),
        responseSerializer: ProtobufSerializer<Echo_EchoResponse>(),
        interceptors: self.interceptors?.makeCollectInterceptors() ?? [],
        wrapping: self.collect(requests:context:)
      )

    case "Update":
      return AsyncServerHandler(
        context: context,
        requestDeserializer: ProtobufDeserializer<Echo_EchoRequest>(),
        responseSerializer: ProtobufSerializer<Echo_EchoResponse>(),
        interceptors: self.interceptors?.makeUpdateInterceptors() ?? [],
        wrapping: self.update(requests:responseStreamWriter:context:)
      )

    default:
      return nil
    }
  }
}
```

This is structurally similar to what we have today except that we implement everything in terms of a single handler and use the various `init(...:wrapping:)` initializers in the codegen.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,519 @@
...
+  @usableFromInline
+  internal let userInfoRef: Ref<UserInfo>
+
+  /// The user provided function to execute.
+  @usableFromInline
+  internal let observer: (
+    GRPCAsyncStream<Request>,
+    GRPCAsyncResponseStreamWriter<Response>,
+    GRPCAsyncServerCallContext
+  ) async throws -> Void
```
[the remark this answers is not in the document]
simonjbeaumont: This is one of many places where I kept the spellings from the existing handlers. In this case, `observer` is what is used in the `BidirectionalStreamingServerHandler`.

I'm not attached to this spelling and happy to change it to something else that you and/or @glbrntt think is clearer.

One thing to consider here though is that there is an outer "handler" which _actually_ handles the stream of incoming events. This is the closure in the associated data in the `State.observing` enum value:

```swift
  internal enum State {
...
    case observing((StreamEvent<Request>) -> Void, _StreamingResponseCallContext<Request, Response>)
...
```

This "handling" of stream events (message or end) is abstracted away from the user function/observer/handler/thing. The user "thing" has already been invoked with a request stream from which it is consuming requests. When the stream is done, the user "thing" returns. And, if there are any errors along the way, the user "thing" throws.

So... @fabianfett @glbrntt what should we call this user "thing"? `userFunction`, `userHandler`, something else?

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,519 @@
...
+  public init(
+    context: CallHandlerContext,
+    requestDeserializer: Deserializer,
+    responseSerializer: Serializer,
+    interceptors: [ServerInterceptor<Request, Response>],
+    observer: @escaping @Sendable(
+      GRPCAsyncStream<Request>,
+      GRPCAsyncResponseStreamWriter<Response>,
+      GRPCAsyncServerCallContext
+    ) async throws -> Void
```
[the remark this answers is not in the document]
simonjbeaumont: Oh, it definitely does. This _is_ the user function.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,519 @@
...
+      // Create a request stream to pass to the user function and capture the
+      // handler in the updated state to allow us to produce more results.
+      let requestStream = GRPCAsyncStream<Request>(AsyncThrowingStream { continuation in
+        self.state = .observing({ streamEvent in
+          switch streamEvent {
+          case let .message(request): continuation.yield(request)
+          case .end: continuation.finish()
+          }
+        }, context)
+      })
```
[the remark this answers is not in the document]
simonjbeaumont: You're right that this type is a bit obscure because it is the glue layer from NIO world to async-await world.

This handler uses an EL internally because it interacts with the rest of the internal GRPC library. Its internal implementation is similar in structure to the extant `BidirectionalStreamingServerHandler`.

In order to call the user function, we construct an `AsyncSequence` of requests and set the `state` to `.observing` with a closure that handles incoming stream events (message or end).

The closure passed into the `.obeserving` enum value will always be called on the right EL.

If I've misunderstood this then I guess we'll need to do something different here. 

Worth noting that when we have #1245 we will not be doing things this way and the code here is a stopgap so may not be worth agonising over.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,528 @@
...
+      )
+
+      // Send response headers back via the interceptors.
+      self.interceptors.send(.metadata([:]), promise: nil)
+
+      // Register callbacks on the status future.
+      statusPromise.futureResult.whenComplete(self.userFunctionStatusResolved(_:))
+
+      // Spin up a task to call the async user handler.
+      self.task = statusPromise.completeWithTask {
```
glbrntt, line 223: We'll need to revisit this, I think. We may not be writable when we complete the user function (which means there may also be buffered response messages to send before the status). Completing the user function should instead enqueue a status to send rather immediately send one back. I suspect we'll need a more detailed `State` enum to represent state as well (I don't have a clear picture what this looks like yet but completing a promise is probably not the right way to go).

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,528 @@
...
+        // Check for cancellation before calling the user function.
+        // This could be the case if the RPC has been cancelled or had an error before this task
+        // has been scheduled.
+        guard !Task.isCancelled else {
+          throw CancellationError()
+        }
+
+        // Call the user function.
+        try await self.observer(requestStream, responseStreamWriter, context)
```
glbrntt, line 232: We should make sure that we correctly handle (not necessarily here) the user function throwing `GRPCStatus.ok`

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,528 @@
...
+    self.init(
+      context: context,
+      requestDeserializer: requestDeserializer,
+      responseSerializer: responseSerializer,
+      interceptors: interceptors,
+      observer: { requestStream, responseStreamWriter, context in
+        for try await request in requestStream.prefix(1) {
+          let response = try await unary(request, context)
+          try await responseStreamWriter.sendResponse(response)
+        }
```
glbrntt, line 453: I think with the current implementation if there are no request messages we'll still finish the RPC with `.ok`. I think we need to throw an appropriate status error here if that's the case.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,528 @@
...
+      GRPCAsyncServerCallContext
+    ) async throws -> Void
+  ) {
+    self.init(
+      context: context,
+      requestDeserializer: requestDeserializer,
+      responseSerializer: responseSerializer,
+      interceptors: interceptors,
+      observer: { requestStream, responseStreamWriter, context in
+        for try await request in requestStream.prefix(1) {
```
glbrntt, line 499: Same here re: empty request stream.

glbrntt (commented):
Apologies: this was spread over a couple of days so some comments are probably out-of-date / answered already.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
```diff
@@ -0,0 +1,102 @@
...
+ * limitations under the License.
+ */
+import Logging
+import NIOConcurrencyHelpers
+import NIOHPACK
+
+#if compiler(>=5.5)
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+public final class GRPCAsyncServerCallContext {
```
simonjbeaumont, line 23: @glbrntt FYI. This is probably new since you last looked :)

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
```diff
@@ -0,0 +1,48 @@
...
+import Logging
+import NIOHPACK
+
+#if compiler(>=5.5)
+
+/// A context provided to RPC handlers.
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+public protocol GRPCAsyncServerCallContext {
+  /// Request headers for this request.
+  var headers: HPACKHeaders { get }
```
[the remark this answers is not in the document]
simonjbeaumont: > We'll need a way to set response headers too (we don't currently support this but we should!)

Do you mean it isn't even supported in the existing, non-async-await, handlers? I took a look and I don't see it; I just see (request) `headers` and (response) `trailers`.

I could play around with what you suggest.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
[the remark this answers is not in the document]
simonjbeaumont: OK, I must have misremembered or conflated this with another discussion because I did this (excluding the protocol) deliberately based on my (potentially flawed) memory of what was asked.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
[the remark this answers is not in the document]
simonjbeaumont: Unless I'm mistaken, the bracketing and the indentation attach this `async throws` to the `observer` parameter of the `init`.

If the `init` were marked with effects, they would appear on line 90.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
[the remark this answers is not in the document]
simonjbeaumont: OK, unless @fabianfett objects, I'll go with `userHandler` and change the state from `.observing` to `.active`?

I couldn't find a term for an in-progress RPC in the official lifecycle docs https://grpc.io/docs/what-is-grpc/core-concepts/ but if anything springs to mind, shout now 😁

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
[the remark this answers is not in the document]
glbrntt: I'm not sure if we actually discussed this to be honest 🤷‍♂️

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
simonjbeaumont: Um, this is an interesting one.

I'm looking at the semantics of the existing handlers where the user function is expected to complete a promise. If they fail the promise with `.ok` then that is equivalent to the user function throwing `GRPCStatus.ok` in the async-await handler. 

It's my claim that the current implementation in this PR is semantically equivalent to the existing one. That is to say that it will end up calling `handleError(error, thrownFromHandler: true)` in both cases and it ends up in the common layer of `ServerErrorProcessor.processObserverError`.

So first off, do you agree that the PR currently implements the same semantics as the existing handler? If not, then I'll need to go away and look again.

And, if you do, are you saying you would like for us to have different semantics in the async-await–based API?

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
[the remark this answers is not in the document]
glbrntt: > Do you mean it isn't even supported in the existing, non-async-await, handlers?

That's right (although there have been requests to add it). We don't have to add this now but we should leave ourselves enough room in the API that we can add it without the naming being ambiguous.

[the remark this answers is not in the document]
glbrntt: > I could play around with what you suggest.

Up to you; play around and see what feels most natural. 

Side note: I wonder also if we should use "metadata" rather than "headers" and  "{initial,trailing}metadata" rather than response headers and response trailers to align more with the client API.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
```diff
@@ -0,0 +1,48 @@
...
+
+  /// A 'UserInfo' dictionary which is shared with the interceptor contexts for this RPC.
+  var userInfo: UserInfo { get set }
+
+  /// The logger used for this call.
+  var logger: Logger { get }
```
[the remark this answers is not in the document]
glbrntt: We use a new logger per-request so that shouldn't be the case.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
simonjbeaumont: I can at least add a test for this.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncResponseStreamWriter.swift
```diff
@@ -0,0 +1,65 @@
...
+  @inlinable
+  internal func shouldCompress(_ compression: Compression) -> Bool {
+    guard self._compressionEnabledOnServer else {
+      return false
+    }
+    return compression.isEnabled(callDefault: self._context.compressionEnabled)
+  }
+
+  @inlinable
+  public func sendResponse(
```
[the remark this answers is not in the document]
simonjbeaumont: Fixed in a68386d3. But yes, agree this will need rework when we have the new writer.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
```diff
@@ -0,0 +1,48 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import Logging
+import NIOHPACK
+
+#if compiler(>=5.5)
```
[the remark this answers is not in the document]
simonjbeaumont: Moved compiler guards above imports in 31db0234 and removed all unused imports in c11a9e1a.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,519 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import _NIOConcurrency
+import Logging
+import NIOCore
+import NIOHPACK
+
+#if compiler(>=5.5)
```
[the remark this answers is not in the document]
simonjbeaumont: Moved compiler guards above imports in 31db0234 and removed all unused imports in c11a9e1a.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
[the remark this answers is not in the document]
simonjbeaumont: > Side note: I wonder also if we should use "metadata" rather than "headers" and "{initial,trailing}metadata" rather than response headers and response trailers to align more with the client API.

That would be my thinking if we decide to add this.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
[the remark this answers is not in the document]
simonjbeaumont: Either way, I'll rework this to conform to a protocol 👍.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
simonjbeaumont: Added `testHandlerThrowsGRPCStatusOK` in 2611a99d.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
simonjbeaumont: I've added some code to guard against this in 71d85d57 but I don't want to mark this as resolved because I want to confirm that this is what you were expecting to see?

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncStream.swift
```diff
@@ -0,0 +1,53 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#if compiler(>=5.5)
+
+/// This is currently a wrapper around AsyncThrowingStream because we want to be
+/// able to swap out the implementation for something else in the future.
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+public struct GRPCAsyncStream<Element>: AsyncSequence {
```
simonjbeaumont, line 22: @glbrntt in the client PR you asked for us to rename this stream to be client specific, i.e. `GRPCAsyncResponseStream`.

Would you like this PR to do the same, i.e. this would be a `GRPCAsyncRequestStream` that is provided to the user handler?

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncStream.swift
```diff
@@ -0,0 +1,53 @@
...
+
+  @usableFromInline
+  internal let stream: WrappedStream
+
+  @inlinable
+  internal init(_ stream: WrappedStream) {
+    self.stream = stream
+  }
+
+  __consuming public func makeAsyncIterator() -> Iterator {
```
simonjbeaumont, line 34: Note to self: remove `__consuming`.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncStream.swift
glbrntt: Yes please, `GRPCAsyncRequestStream` sounds good. 

(I'm wondering whether we should drop the "Async" from a bunch of places, `async`/`await` are implementation details which aren't necessarily relevant. Either way I think we should defer that discussion for later.)

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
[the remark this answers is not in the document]
simonjbeaumont: Added the following fixup commit. Happy to rename again if we come up with something even better before merging.

```
a758ccba fixup: Rename State.observing to active and observer to userHandler
```

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncStream.swift
simonjbeaumont: I speculatively made this change anyway: 05951c05.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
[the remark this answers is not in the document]
simonjbeaumont: I've had a crack at this in b73edc3e and c80b35f8. LMK if this was what you meant.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
simonjbeaumont: @glbrntt does this type also need the "make it an internal class and provide a public wrapper struct" treatment?

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
simonjbeaumont: Is this something we revisit when we have the new writer or something we do in this PR?

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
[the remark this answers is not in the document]
glbrntt: Looks good!

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
glbrntt: I don't think so; this class is just storage so I don't see any value in it.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
glbrntt: > So first off, do you agree that the PR currently implements the same semantics as the existing handler? If not, then I'll need to go away and look again.

Sort of. This handler now does the job of four handlers and reviewing those revealed they have somewhat inconsistent semantics:
- unary: return a failed future of `GRPCStatus.ok` would result in no response message and an 'ok' status (seems incorrect to me)
- client streaming: 
  - caller fails stream-event handler callback with `.ok`; no response message and an 'ok' status (seems wrong)
  - caller fails response promise with `.ok`; no response message and an 'ok' status (seems wrong)
- server streaming: API requires user to return a `GRPCStatus` future; 'ok' is always fine.
- bidirectional streaming:
  - caller fails stream-event handler callback with '.ok': no messages but '.ok' sent back (seems wrong but perhaps much less so than in the client streaming case because empty response streams are allowed) 
  - user must complete a `GRPCStatus` promise; 'ok' is fine.
  
To sum up: when we expect a single response (unary and client streaming) a thrown 'ok' status is _weird_ because the client will throw when awaiting response but the status will be okay. For bidirectional and server streaming throwing an 'ok' is fine (albeit slightly weird) because we allow empty streams. Ideally `.ok` wouldn't be an error but that ship has sadly already sailed.

> And, if you do, are you saying you would like for us to have different semantics in the async-await–based API?

Yes, I think we should catch thrown `.ok` statuses and handle them appropriately. For unary and client streaming we should turn them into a different status, for server streaming and bidirectional streaming we should treat them as an early return from the user function.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncRequestStream.swift
```diff
@@ -0,0 +1,53 @@
...
+
+/// This is currently a wrapper around AsyncThrowingStream because we want to be
+/// able to swap out the implementation for something else in the future.
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+public struct GRPCAsyncRequestStream<Element>: AsyncSequence {
+  @usableFromInline
+  internal typealias WrappedStream = AsyncThrowingStream<Element, Error>
+
+  @usableFromInline
+  internal let stream: WrappedStream
```
glbrntt, line 27: If something would be `private` but can't be because it requires a `@usableFromInline`/`@inlinable` then by convention we prefix the name with an underscore.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncRequestStream.swift
```diff
@@ -0,0 +1,53 @@
...
+
+  @inlinable
+  internal init(_ stream: WrappedStream) {
+    self.stream = stream
+  }
+
+  public func makeAsyncIterator() -> Iterator {
+    Self.AsyncIterator(self.stream)
```
glbrntt, line 35: I might be missing something obvious here, but how does this work? The iterator type here defined by this `struct` is `Iterator` and there's no typealias for `AsyncIterator` 🤔

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
```diff
@@ -0,0 +1,102 @@
...
+    headers: HPACKHeaders,
+    logger: Logger,
+    userInfo: UserInfo = UserInfo()
+  ) {
+    self.init(
+      headers: headers,
+      logger: logger,
+      userInfoRef: .init(userInfo)
+    )
+  }
```
glbrntt, line 88: Let's get rid of this `init` for now -- we only have the equivalent in existing contexts for backwards compatibility reasons.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
```diff
@@ -0,0 +1,102 @@
...
+ * limitations under the License.
+ */
+#if compiler(>=5.5)
+
+import Logging
+import NIOConcurrencyHelpers
+import NIOHPACK
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+public final class GRPCAsyncServerCallContext {
```
glbrntt, line 23: FWIW: I'm considering this "fine for now" rather than complete and signed off.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,564 @@
...
+    wrapping unary: @escaping @Sendable(Request, GRPCAsyncServerCallContext) async throws
+      -> Response
+  ) {
+    self._handler = .init(
+      context: context,
+      requestDeserializer: requestDeserializer,
+      responseSerializer: responseSerializer,
+      interceptors: interceptors,
+      userHandler: { requestStream, responseStreamWriter, context in
+        guard let request = try await requestStream.prefix(1).first(where: { _ in true }) else {
```
glbrntt, line 71: I thought I asked elsewhere but I couldn't find the comment: do we need both `prefix` and `first`?

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,564 @@
...
+  public typealias Request = Deserializer.Output
+  public typealias Response = Serializer.Input
+
+  @inlinable
+  public init(
+    context: CallHandlerContext,
+    requestDeserializer: Deserializer,
+    responseSerializer: Serializer,
+    interceptors: [ServerInterceptor<Request, Response>],
+    wrapping unary: @escaping @Sendable(Request, GRPCAsyncServerCallContext) async throws
```
glbrntt, line 62:
```suggestion
    wrapping unary: @escaping @Sendable (Request, GRPCAsyncServerCallContext) async throws
```

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,564 @@
...
+    )
+  }
+
+  @inlinable
+  public init(
+    context: CallHandlerContext,
+    requestDeserializer: Deserializer,
+    responseSerializer: Serializer,
+    interceptors: [ServerInterceptor<Request, Response>],
+    wrapping clientStreaming: @escaping @Sendable(
```
glbrntt, line 86:
```suggestion
    wrapping clientStreaming: @escaping @Sendable (
```

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,564 @@
...
+    )
+  }
+
+  @inlinable
+  public init(
+    context: CallHandlerContext,
+    requestDeserializer: Deserializer,
+    responseSerializer: Serializer,
+    interceptors: [ServerInterceptor<Request, Response>],
+    wrapping serverStreaming: @escaping @Sendable(
```
glbrntt, line 109:
```suggestion
    wrapping serverStreaming: @escaping @Sendable (
```

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,564 @@
...
+    )
+  }
+
+  @inlinable
+  public init(
+    context: CallHandlerContext,
+    requestDeserializer: Deserializer,
+    responseSerializer: Serializer,
+    interceptors: [ServerInterceptor<Request, Response>],
+    wrapping bidirectional: @escaping @Sendable(
```
glbrntt, line 135:
```suggestion
    wrapping bidirectional: @escaping @Sendable (
```

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,564 @@
...
+      responseSerializer: responseSerializer,
+      interceptors: interceptors,
+      userHandler: bidirectional
+    )
+  }
+}
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+@usableFromInline
+internal final class _GRPCAsyncServerHandler<
```
glbrntt, line 153: We can drop the `_GRPC` now this is internal.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,564 @@
...
+    case completed
+  }
+
+  @inlinable
+  public init(
+    context: CallHandlerContext,
+    requestDeserializer: Deserializer,
+    responseSerializer: Serializer,
+    interceptors: [ServerInterceptor<Request, Response>],
+    userHandler: @escaping @Sendable(
```
glbrntt, line 228:
```suggestion
    userHandler: @escaping @Sendable (
```

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,564 @@
...
+      // Make a context to invoke the user handler with.
+      let context = GRPCAsyncServerCallContext(
+        headers: headers,
+        logger: self.context.logger,
+        userInfoRef: self.userInfoRef
+      )
+
+      // Create a request stream to pass to the user function and capture the
+      // handler in the updated state to allow us to produce more results.
+      let requestStream = GRPCAsyncRequestStream<Request>(AsyncThrowingStream { continuation in
```
glbrntt, line 327: FYI https://github.com/grpc/grpc-swift/pull/1252 has been merged now. It has two parts a `PassthroughMessageSource` and a `PassthroughMessageSequence` (an adapter for the source to turn it into an `AsyncSequence`)

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
glbrntt: It'll be clearer when we have the new writer.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,564 @@
...
+
+        // Call the user function.
+        try await self.userHandler(requestStream, responseStreamWriter, context)
+
+        // Check for cancellation after the user function has returned so we don't return OK in the
+        // event the RPC was cancelled. This is probably overkill because the `handleError(_:)`
+        // will also fail the `statusPromise`.
+        guard !Task.isCancelled else {
+          throw CancellationError()
+        }
```
glbrntt, line 370: I'm wondering if we actually want to check this: if the user handler has completed without throwing then -- presumably -- most of the hard work is done and we should just continue?

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
simonjbeaumont: > Yes, I think we should catch thrown .ok statuses and handle them appropriately. For unary and client streaming we should turn them into a different status, ...

Should we use the protocol violation error (in this case, it would be the server that violated the protocol)?

> ... for server streaming and bidirectional streaming we should treat them as an early return from the user function.

So, turn this into a status `.ok` with no responses?

> Ideally .ok wouldn't be an error but that ship has sadly already sailed.

Could we explicitly catch `.ok` from the user handler and fail in all cases with a protocol violation, or somehow deprecate the use of throwing a `GRPCStatus` in favour of throwing something that doesn't include `.ok`, given that this is a new API?

glbrntt: > > Yes, I think we should catch thrown .ok statuses and handle them appropriately. For unary and client streaming we should turn them into a different status, ...
> 
> Should we use the protocol violation error (in this case, it would be the server that violated the protocol)?

Not sure if this is protocol violation, it's really a user error. `.unknown` might be the right status code here (since it's really not clear what the caller meant if they threw `.ok`). 

> > ... for server streaming and bidirectional streaming we should treat them as an early return from the user function.
> 
> So, turn this into a status `.ok` with no responses?

Yes, I think that's probably fine. WDYT? (I keep changing my mind on this one.)

> > Ideally .ok wouldn't be an error but that ship has sadly already sailed.
> 
> Could we explicitly catch `.ok` from the user handler and fail in all cases with a protocol violation, or somehow deprecate the use of throwing a `GRPCStatus` in favour of throwing something that doesn't include `.ok`, given that this is a new API?

I'm not sure it's worth it. It's such a pervasive type that deprecating and replacing the whole type to handle this one case arguably isn't worth the toil. Moreover users can still use deprecated types.

There is something to be said for always catching `.ok` and treating them as invalid for all four types of RPC though. Perhaps that's better than handling it when we stream responses.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncRequestStream.swift
glbrntt: That's much better type inference than I expected!

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
glbrntt: Ah, what a pain. No, let's leave it as is and sort it out later.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
glbrntt: I'm don't mind. It shouldn't be a particularly large change though.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
glbrntt: Yes, I think we should. The only harm in doing this is that we have to wait for the end of the request stream before we can invoke the function which is potentially slower (as the end might be delivered in a separate http/2 frame to the request).

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
glbrntt: Okay, let's leave it then.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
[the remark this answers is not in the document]
fabianfett: I like `userHandler`!

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
simonjbeaumont: > There is something to be said for always catching .ok and treating them as invalid for all four types of RPC though. Perhaps that's better than handling it when we stream responses.

I prefer the consistency of this if it's acceptable to you? We could always document it clearly.

glbrntt: Yes that's fine by me!

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncRequestStream.swift
simonjbeaumont: Does this go for type aliases and functions too?

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncResponseStreamWriter.swift
```diff
@@ -0,0 +1,65 @@
...
+    compressionIsEnabled: Bool,
+    send: @escaping (Response, MessageMetadata) async throws -> Void
+  ) {
+    self._context = context
+    self._compressionEnabledOnServer = compressionIsEnabled
+    self._send = send
+  }
+
+  @inlinable
+  internal func shouldCompress(_ compression: Compression) -> Bool {
```
simonjbeaumont, line 48: @glbrntt Should this have an `_` prepended?

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
```diff
@@ -0,0 +1,102 @@
...
+      self.userInfoRef.value
+    } }
+    set { self.lock.withLock {
+      self.userInfoRef.value = newValue
+    } }
+  }
+
+  /// A reference to an underlying `UserInfo`. We share this with the interceptors.
+  @usableFromInline
+  internal let userInfoRef: Ref<UserInfo>
```
simonjbeaumont, line 63: Should this also have `_` prepended?

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
simonjbeaumont: Done in afdb71ac.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncRequestStream.swift
simonjbeaumont: I did `stream` and `WrappedStream` in 4fa1c3a2.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
simonjbeaumont: Great. I've updated this logic in d8f645e2.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
simonjbeaumont: Done in 1cb21b98.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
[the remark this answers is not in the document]
simonjbeaumont: OK, well I made it mutable in b46a4c19.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
glbrntt: It's fine without.

fabianfett (commented):
Couple of small questions...

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,566 @@
...
+ * limitations under the License.
+ */
+#if compiler(>=5.5)
+
+import _NIOConcurrency
+import NIOCore
+import NIOHPACK
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+public struct GRPCAsyncServerHandler<
```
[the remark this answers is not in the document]
glbrntt: The underlying type _is_ a reference type. We only wrap this up as a `struct` because we don't want to expose the underlying type to public API (generated code will return instances of this) so that we can change the implementation later.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
simonjbeaumont: @fabianfett I asked whether this was needed in the above resolved thread and apparently no. 
https://github.com/grpc/grpc-swift/pull/1249/files#r700021649

To avoid ping-ponging a fixup, I'll wait for you and @glbrntt to agree on this one.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncRequestStream.swift
```diff
@@ -0,0 +1,53 @@
...
+  internal let _stream: _WrappedStream
+
+  @inlinable
+  internal init(_ stream: _WrappedStream) {
+    self._stream = stream
+  }
+
+  public func makeAsyncIterator() -> Iterator {
+    Self.AsyncIterator(self._stream)
+  }
```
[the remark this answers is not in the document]
simonjbeaumont: Because `Self.Iterator.init(_:)` is `fileprivate`. If we wanted `makeAsyncIterator()` to be `@inlinable` I guess I would need to make `Self.Iterator.init(_:)` `@usableFromInline internal`?

simonjbeaumont:
Urgh. @glbrntt in order to incorporate the new stream source and writers I merged from the feature branch but this now makes the files-changed view look like these are part of this PR. Apologies for that. I'm happy to rebase or do something else if it would clean things up for review?

I only made use of the new stream source in the server code in this PR. Intend to follow up with another PR for client once we land this.

I have added the additional state which should take care of ensuring that the response stream writer has been emptied before we send `.end` on the stream in the happy case, and that the responses in the writer are dropped when there is an error.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,689 @@
...
+  }
+
+  @inlinable
+  internal func receiveInterceptedMessage(_ request: Request) {
+    switch self.state {
+    case .idle:
+      self.handleError(GRPCError.ProtocolViolation("Message received before headers"))
+    case let .active(requestStreamSource, _, _, _):
+      // TODO: Handle YieldResult from `yield()`
+      requestStreamSource.yield(request)
```
simonjbeaumont, line 478: @glbrntt what should we be doing with the `YieldResult` here? Should we be checking that it is `.accepted`? And if it's `.dropped` do we consider this to be a fatal error that requires tearing down the RPC (because messages may have been lost)?

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,689 @@
...
+  }
+
+  @inlinable
+  internal func receiveInterceptedEnd() {
+    switch self.state {
+    case .idle:
+      self.handleError(GRPCError.ProtocolViolation("End of stream received before headers"))
+    case let .active(requestStreamSource, _, _, _):
+      // TODO: Handle YieldResult from `finish()`
+      requestStreamSource.finish()
```
simonjbeaumont, line 495: Similar question re: `YieldResult` here.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,689 @@
...
+          /// error, cancel whatever you're doing." If we want to preserve the API of the
+          /// non-async-await handlers in this error flow we could add conformance to
+          /// `GRPCStatusTransformable` to `CancellationError`, but we still cannot control _how_
+          /// the user handler will handle the `CancellationError` which could even be swallowed.
+          ///
+          /// - NOTE: Currently we _have_ added `GRPCStatusTransformable` conformance to
+          /// `CancellationError` to convert it into `GRPCStatus.unavailable` and expect to
+          /// document that user handlers should always rethrow `CacellationError` if handled, after
+          /// optional cleanup.
+          requestStreamSource.finish(throwing: CancellationError())
```
simonjbeaumont, line 454: @glbrntt do you think we should mark `finish(throwing:)` as `@discardableResult`. In this instance I don't think there's anything we can meaningfully do with the `YieldResult`?

simonjbeaumont:
@glbrntt as requested I rebased on `async-await` to reduce the PR diff. Some comment history may be affected.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,689 @@
...
+        // The user handler is done but there may be pending writes in the response stream writer.
+
+        // Create a promise to hang a callback off when the response stream is drained.
+        let responseStreamDrainedPromise: EventLoopPromise<GRPCStatus> = self.context.eventLoop
+          .makePromise()
+
+        // Register callback for the response stream being drained.
+        responseStreamDrainedPromise.futureResult.whenComplete(self.responseStreamDrained(_:))
+
+        self.responseStreamDrainTask = responseStreamDrainedPromise.completeWithTask {
```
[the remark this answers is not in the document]
simonjbeaumont: I actually had it this way and decided it was a race condition. Once the `userHandler` has returned we await the finish, we still have things coming in the handler. We want to have moved the state on. I tried moving the state on in this block but that meant a data race because the state was being moved in two different contexts.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncResponseStreamWriter.swift
simonjbeaumont: Done

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
simonjbeaumont: Any convergence between reviewers on this one?

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,689 @@
...
+    GRPCAsyncRequestStream<Request>,
+    GRPCAsyncResponseStreamWriter<Response>,
+    GRPCAsyncServerCallContext
+  ) async throws -> Void
+
+  /// The state of the handler.
+  @usableFromInline
+  internal var state: State = .idle
+
+  /// The task used to run the async user function.
```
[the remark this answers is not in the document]
simonjbeaumont: Done

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,689 @@
...
+        )
+
+      // Set the state to active and bundle in all the associated data.
+      self.state = .active(requestStreamSource, context, responseStreamWriter, userHandlerPromise)
+
+      // Register callback for the completion of the user handler.
+      userHandlerPromise.futureResult.whenComplete(self.userHandlerCompleted(_:))
+
+      // Send response headers back via the interceptors.
+      self.interceptors.send(.metadata([:]), promise: nil)
```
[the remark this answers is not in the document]
simonjbeaumont: I've dropped in a `TODO`.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,689 @@
...
+            responseStreamWriter._asyncWriter.cancelAsynchronously()
+            // Throwing GRPCStatus.ok is considered to be invalid.
+            if (error as? GRPCStatus)?.isOk ?? false {
+              throw GRPCStatus(
+                code: .unknown,
+                message: "Handler threw GRPCStatus error with code .ok"
+              )
+            }
+            throw error
+          }
```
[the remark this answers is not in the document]
simonjbeaumont: Done

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
simonjbeaumont: Done.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,689 @@
...
+          /// non-async-await handlers in this error flow we could add conformance to
+          /// `GRPCStatusTransformable` to `CancellationError`, but we still cannot control _how_
+          /// the user handler will handle the `CancellationError` which could even be swallowed.
+          ///
+          /// - NOTE: Currently we _have_ added `GRPCStatusTransformable` conformance to
+          /// `CancellationError` to convert it into `GRPCStatus.unavailable` and expect to
+          /// document that user handlers should always rethrow `CacellationError` if handled, after
+          /// optional cleanup.
+          requestStreamSource.finish(throwing: CancellationError())
+          responseStreamWriter._asyncWriter.cancelAsynchronously()
```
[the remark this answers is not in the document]
simonjbeaumont: Done.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
simonjbeaumont: OK, I've implemented these semantics and captured the reasoning in some commentary. I also dropped a `TODO` for the future intention to action the queue depth.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,689 @@
...
+  @inlinable
+  internal func receiveInterceptedMessage(_ request: Request) {
+    switch self.state {
+    case .idle:
+      self.handleError(GRPCError.ProtocolViolation("Message received before headers"))
+    case let .active(requestStreamSource, _, _, _):
+      // TODO: Handle YieldResult from `yield()`
+      requestStreamSource.yield(request)
+    case .finishingSuccessfully:
+      self.handleError(GRPCError.ProtocolViolation("Message received after end of stream"))
```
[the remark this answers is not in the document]
simonjbeaumont: Nah, this isn't right at all. This whole state is gone now.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
simonjbeaumont: Actioned.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,689 @@
...
+  @inlinable
+  internal func receiveInterceptedEnd() {
+    switch self.state {
+    case .idle:
+      self.handleError(GRPCError.ProtocolViolation("End of stream received before headers"))
+    case let .active(requestStreamSource, _, _, _):
+      // TODO: Handle YieldResult from `finish()`
+      requestStreamSource.finish()
+    case .finishingSuccessfully:
+      self.handleError(GRPCError.ProtocolViolation("Multiple end of stream blocks received on RPC"))
```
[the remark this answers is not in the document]
simonjbeaumont: State removed.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,689 @@
...
+      // The user handler cannot send responses before it has been invoked.
+      preconditionFailure()
+
+    case .active:
+      self.interceptors.send(.message(response, metadata), promise: promise)
+
+    case .finishingSuccessfully:
+      // TODO: Add test for this...
+      // The user handler cannot send responses once it has finished.
+      // ... or can it? (is this called from the delegate?)
```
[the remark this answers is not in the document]
simonjbeaumont: State removed and the delegate end call back is now used to complete the message stream back to the client now.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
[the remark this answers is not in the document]
simonjbeaumont: I've updated this now. Should be more what you were expecting to see.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,689 @@
...
+
+        // Register callback for the response stream being drained.
+        responseStreamDrainedPromise.futureResult.whenComplete(self.responseStreamDrained(_:))
+
+        self.responseStreamDrainTask = responseStreamDrainedPromise.completeWithTask {
+          try await withTaskCancellationHandler {
+            // Await the writer finish.
+            try await responseStreamWriter._asyncWriter.finish(())
+            // Return the original status from the user handler.
+            return status
```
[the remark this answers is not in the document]
simonjbeaumont: Done. This also resulted in us using just a void promise for the task callback so we are just handling the status in the stream writer.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,632 @@
...
+            try await self.userHandler(requestStream, responseStreamWriter, context)
+          } catch let status as GRPCStatus where status.isOk {
+            // The user handler throwing `GRPCStatus.ok` is considered to be invalid.
+            responseStreamWriter.asyncWriter.cancelAsynchronously()
+            throw GRPCStatus(
+              code: .unknown,
+              message: "Handler threw GRPCStatus error with code .ok"
+            )
+          } catch {
+            // TODO: Should we be instead doing `finish(throwing: error)`?
```
glbrntt, line 393: Hmm, we don't currently offer a way of finishing with an error and given we only finish with a cancellation error via `cancel()` I'm not sure this is worth adding now.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,632 @@
...
+    }
+  }
+
+  // MARK: - User Function To Interceptors
+
+  @inlinable
+  internal func interceptResponse(
+    _ response: Response,
+    metadata: MessageMetadata,
+    promise: EventLoopPromise<Void>?
```
glbrntt, line 497: I think `promise` is always `nil` here -- let's just remove it

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,632 @@
...
+      }
+
+    case .completed:
+      ()
+    }
+  }
+
+  @inlinable
+  internal func responseStreamDrained(_ status: GRPCStatus) {
+    switch self.state {
```
glbrntt, line 540: Let's assert we're on the correct event loop in here

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
glbrntt: Ah, apologies, I missed this one.

---

@fabianfett It is fundamentally a value type but not one we want copy-on-write semantics with. The instance that the async handler holds must not diverge from the instance the implementor of the RPC holds. They hold these instances on different threads (EventLoop vs Task).

We could wrap this `class` up as a `struct` and pass it `inout` to the RPC. This would make the fact that the type stores mutable state more obvious. However without copy-on-write semantics this would be weird and surprising I think.

We also considered an `actor` but that felt clunky ("why do I have to `await` retrieving a logger or updating trailers?") and requires a `promise` each time the async server handler requires reading a property.

fabianfett: Okay, can we get this as a `//` comment? I think this context is rather important when we look at it the next time.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,632 @@
...
+      let requestStream = GRPCAsyncRequestStream(.init(consuming: requestStreamSource))
+
+      // TODO: In future use `AsyncWriter.init(maxPendingElements:maxWritesBeforeYield:delegate:)`?
+      let responseStreamWriter =
+        GRPCAsyncResponseStreamWriter(
+          wrapping: AsyncWriter(delegate: AsyncResponseStreamWriterDelegate(
+            context: context,
+            compressionIsEnabled: self.context.encoding.isEnabled,
+            send: { response, metadata in
+              self.interceptResponse(response, metadata: metadata, promise: nil)
```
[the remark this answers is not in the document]
simonjbeaumont: Oh, this is slightly embarrassing. I had the event loop dance in my make-the-handler-a-delegate branch. It dropped off. 😅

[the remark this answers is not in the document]
simonjbeaumont: Done in 0c819a55

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,632 @@
...
+            },
+            finish: self.responseStreamDrained(_:)
```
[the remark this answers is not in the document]
simonjbeaumont: Done in 0c819a55

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -0,0 +1,632 @@
...
+    /// consumed by the user handler.
+    ///
+    /// - The `GRPCAsyncServerContext` is a reference to the context that was passed to the user
+    /// handler.
+    ///
+    /// - The `GRPCAsyncResponseStreamWriter` is the response stream writer that is being written to
+    /// by the user handler. Because this is pausable, it may contain responses after the user
+    /// handler has completed that have yet to be written. We need to keep hold of this so we can
+    /// propagate it to the next state which waits for these pending responses to be written before
+    /// sending `.end` on the interceptors.
```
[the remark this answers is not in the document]
simonjbeaumont: See 33b65823

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
simonjbeaumont: Immortalised this reasoning in a comment in 732f8648 😉

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
simonjbeaumont: Done when adding the EL dance in 0c819a55

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
simonjbeaumont: I think we resolved this now.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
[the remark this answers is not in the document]
glbrntt: If you use an `EmbeddedEventLoop` then `execute` requires you to `run()` the loop which can be a bit of a pain in tests (and unclear if it's at few layers down). As `inEventLoop` is defined to be `true` for the embedded event loop we can side step the problem.

glbrntt (approved):
I think this is good to be merged now! 🎉

glbrntt:
> I think this is good to be merged now! 🎉

Modulo any feedback from @fabianfett

PR #1252, opened 2021-08-27
Add an async throwing source

glbrntt:
Motivation:

`AsyncThrowingStream` provides an implementation of `AsyncSequence`
which allows the holder to provide new values to that sequence from
within a closure provided to the initializer. This API doesn't fit our
needs: we must be able to provide the values 'from the outside' rather
than during initialization.

Modifications:

- Add an `AsyncThrowingSource`, an implementation of `AsyncSequence`
  which may have values provided to it via `yield(_:)` and terminated
  with `finish()` or `finish(throwing:)`.
- Add tests and a few `AsyncSequence` helpers.

Result:

We have an `AsyncSequence` implementation which can have values provided
to it.

### DIFF
--- a/Tests/GRPCTests/GRPCAsyncClientCallTests.swift
+++ b/Tests/GRPCTests/GRPCAsyncClientCallTests.swift
@@ -178,10 +178,7 @@ class GRPCAsyncClientCallTests: GRPCTestCase {
 
     await assertThat(try await update.initialMetadata, .is(.equalTo(Self.OKInitialMetadata)))
 
-    actor TestResults {
-      static var numResponses = 0
-      static var numRequests = 0
-    }
+    let counter = RequestResponseCounter()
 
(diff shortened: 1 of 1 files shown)


Sources/GRPC/AsyncAwaitSupport/PassthroughMessageSource.swift
```diff
@@ -0,0 +1,152 @@
...
+    }
+  }
+
+  // MARK: - Next
+
+  internal func consumeNextElement() async throws -> Element? {
+    return try await withCheckedThrowingContinuation {
+      self.consumeNextElement(continuation: $0)
+    }
+  }
```
fabianfett, line 132: I assume this should be `@inlinable`?

fabianfett (approved):

PR #1253, opened 2021-09-07
New configuration parameter for HTTP2 maxFrameSize.

gcjenkinson:
Motivation:

In gRPC calls that process large payloads, the default settings of the
HTTP2 maxFrameSize (16384) within swift-nio-http2 results in significant
performance impact (for large payloads this performance impact has been
demonstrated to result in timeouts where RPCs fail to complete within
their assigned deadline).

Allowing programmatic configuration of the HTTP2 maxFrameSize when the
server is built allows the parameter to be optimised for the expected
payload sizes.

Modifications:

Added new gRPC configuration parameter for the HTTP2 parameter
maxFrameSize (which can be configured in the Server.Builder using the
new method `withHTTPMaxFrameSize()`).

Result:

The Server.Builder can be used to configure the HTTP2 maxFrameSize
parameter as:

Server.insecure(group: serverGroup).withHTTPMaxFrameSize(1024 * 1024)...

### DIFF
--- a/Sources/GRPC/Server.swift
+++ b/Sources/GRPC/Server.swift
@@ -332,6 +332,14 @@ extension Server {
       }
     }
 
+    /// The HTTP/2 max frame size. Defaults to 16384. Value is clamped between 2^14 and 2^24-1
+    /// octets inclusive (the minimum and maximum allowable values - HTTP/2 RFC 7540 4.2).
+    public var httpMaxFrameSize: Int = 16384 {
+      didSet(httpMaxFrameSize) {
+        self.httpMaxFrameSize = httpMaxFrameSize.clamped(to: 16384 ... 16_777_215)
+      }
+    }
+
     /// The root server logger. Accepted connections will branch from this logger and RPCs on
     /// each connection will use a logger branched from the connections logger. This logger is made
     /// available to service providers via `context`. Defaults to a no-op logger.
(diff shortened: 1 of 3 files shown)


Sources/GRPC/Server.swift
```diff
@@ -447,3 +455,7 @@ private extension ServerBootstrapProtocol {
     }
   }
 }
+
+private func clamped<T>(_ value: T, lower: T, upper: T) -> T where T : Comparable {
+    return min(max(value, lower), upper)
+}
```
glbrntt, line 461: This can be expressed a little more Swift-ly as:

```swift
extension Comparable {
  fileprivate func clamped(to range: ClosedRange<Self>) -> Self {
    return min(max(self, range.lowerBound), range.upperBound)
  }
}
```

Sources/GRPC/Server.swift
```diff
@@ -332,6 +332,14 @@ extension Server {
       }
     }
 
+    /// The HTTP/2 max frame size. Defaults to 16384. Must be value between 2^14 and 2^24-1 octets
+    /// inclusive.
+    public var httpMaxFrameSize: Int = 16384 {
+      didSet(httpMaxFrameSize) {
+        self.httpMaxFrameSize = clamped(httpMaxFrameSize, lower: 16_384, upper: 16_777_216)
```
glbrntt, line 339: The upper-bound is off-by-one here; the upper limit should be 2^24 - 1.

Sources/GRPC/Server.swift
```diff
@@ -332,6 +332,14 @@ extension Server {
       }
     }
 
+    /// The HTTP/2 max frame size. Defaults to 16384. Must be value between 2^14 and 2^24-1 octets
+    /// inclusive.
```
glbrntt, line 336: Could you add a note saying that we clamp the value if it's outside the range?

glbrntt (changes requested):

glbrntt (approved):
Great, thanks @gcjenkinson!

PR #1255, opened 2021-09-08
Expose more http/2 configuration knobs

glbrntt:
Motivation:

Having more control over the configuration of the http/2 layer is useful
for certain performance sensitive situations. We recently exposed the
max frame size setting on the server (#1253); we should do the same on
the client.

Modifications:

- Add configuration for the max frame size to the client
- Fix the clamping of the max frame size on the server (this was missed
  when reviewing #1253)
- Add tests to validate that the configured values are correctly clamped
- Send `SETTINGS_INITIAL_WINDOW_SIZE` in the initial SETTINGS frame set to the
  value of the target window size.

Result:

Users have more control over HTTP/2 settings.

### DIFF
--- a/Sources/GRPC/ClientConnection.swift
+++ b/Sources/GRPC/ClientConnection.swift
@@ -16,6 +16,7 @@
 import Foundation
 import Logging
 import NIOCore
+import NIOHPACK
 import NIOHTTP2
 import NIOSSL
 import NIOTLS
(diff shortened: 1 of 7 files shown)


Sources/GRPC/ClientConnection.swift
```diff
@@ -340,8 +341,21 @@ extension ClientConnection {
     /// Defaults to `waitsForConnectivity`.
     public var callStartBehavior: CallStartBehavior = .waitsForConnectivity
 
-    /// The HTTP/2 flow control target window size. Defaults to 65535.
-    public var httpTargetWindowSize = 65535
+    /// The HTTP/2 flow control target window size. Defaults to 65535. Values are clamped between
+    /// 0 and 2^31-1 inclusive.
```
Lukasa, line 345: Probably we should forbid `0`?

Sources/GRPC/ClientConnection.swift
```diff
@@ -500,12 +514,22 @@ extension ChannelPipeline.SynchronousOperations {
     connectionKeepalive: ClientConnectionKeepalive,
     connectionIdleTimeout: TimeAmount,
     httpTargetWindowSize: Int,
+    httpMaxFrameSize: Int,
     errorDelegate: ClientErrorDelegate?,
     logger: Logger
   ) throws {
+    let initialSettings = [
+      // As per the default settings for swift-nio-http2:
```
Lukasa, line 522: NIOHTTP2 makes its default settings available in a global-scope variable (yeah I know) so you can just extend that array if you'd prefer.

Lukasa (approved):
LGTM, two minor notes.

Sources/GRPC/ClientConnection.swift
glbrntt: Good ol' `nioDefaultSettings`! I'd rather be explicit here, I think.

Just a thought: should we set `maxConcurrentStreams` to `0`? The client doesn't tolerate or expect the server to open streams so explicitly setting it to 100 feels odd.

Lukasa: Sure, why not.

PR #1256, opened 2021-09-09
Bump version number to 1.4.0

glbrntt:
Motivation:

We plan on tagging a release soon.

Modifications:

- Bump the version to 1.4.0

Result:

The version in the default user-agent string will match the released
version.


fabianfett (approved):
Change of the year!

PR #1259, opened 2021-09-10
Code generation for async-await

glbrntt:
Motivation:

Manually constructing clients and servers is an error prone nightmare.
We should generate them instead!

Modifications:

- Add async-await code-generation for server and client.
- The client code generation is missing "simple-safe" wrappers for now,
  this can be added later.
- Naming represents the current state of the branch rather than anything
  final
- Add options for "ExperimentalAsyncClient" and
  "ExperimentalAsyncServer" -- these may be used in conjunction with the
  'regular' "Client" and "Server" options.

Result:

We can generate async-await style grpc clients and servers.


fabianfett (approved):
LGTM. Works on my machine.

PR #1260, opened 2021-09-10
Generate async echo client and server, add async server implementation and basic tests

glbrntt:
Motivation:

In #1259 code generation was added for async client and servers. We
should generate the echo service and have tests exercising the generated
async code.

Modifications:

- Switch on async code generation for echo
- Implement an async echo service and add some basic tests

Results:

We run tests with an async client and server

### DIFF
--- a/Sources/Examples/Echo/Model/echo.grpc.swift
+++ b/Sources/Examples/Echo/Model/echo.grpc.swift
@@ -172,6 +172,101 @@ public final class Echo_EchoClient: Echo_EchoClientProtocol {
   }
 }
 
+#if compiler(>=5.5)
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+public protocol Echo_EchoAsyncClientProtocol: GRPCClient {
+  var serviceName: String { get }
+  var interceptors: Echo_EchoClientInterceptorFactoryProtocol? { get }
+
+  func makeGetCall(
+    _ request: Echo_EchoRequest,
+    callOptions: CallOptions?
+  ) -> GRPCAsyncUnaryCall<Echo_EchoRequest, Echo_EchoResponse>
+
+  func makeExpandCall(
+    _ request: Echo_EchoRequest,
+    callOptions: CallOptions?
+  ) -> GRPCAsyncServerStreamingCall<Echo_EchoRequest, Echo_EchoResponse>
+
+  func makeCollectCall(
+    callOptions: CallOptions?
+  ) -> GRPCAsyncClientStreamingCall<Echo_EchoRequest, Echo_EchoResponse>
+
+  func makeUpdateCall(
+    callOptions: CallOptions?
+  ) -> GRPCAsyncBidirectionalStreamingCall<Echo_EchoRequest, Echo_EchoResponse>
+}
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+extension Echo_EchoAsyncClientProtocol {
+  public var serviceName: String {
+    return "echo.Echo"
+  }
+
+  public var interceptors: Echo_EchoClientInterceptorFactoryProtocol? {
+    return nil
+  }
+
+  public func makeGetCall(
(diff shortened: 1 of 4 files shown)


Sources/Examples/Echo/Model/echo.grpc.swift
```diff
@@ -204,7 +299,7 @@ public final class Echo_EchoTestClient: Echo_EchoClientProtocol {
   public func enqueueGetResponse(
     _ response: Echo_EchoResponse,
     _ requestHandler: @escaping (FakeRequestPart<Echo_EchoRequest>) -> () = { _ in }
-  )  {
+  ) {
```
fabianfett, line 302: nice fix

Tests/GRPCTests/AsyncAwaitSupport/AsyncIntegrationTests.swift
```diff
@@ -0,0 +1,153 @@
...
+      for (i, name) in ["boyle", "jeffers", "holt"].enumerated() {
+        try await update.sendMessage(.with { $0.text = name })
+        let response = try await responseIterator.next()
+        XCTAssertEqual(response, "Swift echo update (\(i)): \(name)")
+      }
+
+      try await update.sendEnd()
+
+      let initialMetadata = try await update.initialMetadata
+      initialMetadata.assertFirst("200", forName: ":status")
```
fabianfett, line 136: Could we check this before the loop already?

Tests/GRPCTests/AsyncAwaitSupport/AsyncIntegrationTests.swift
```diff
@@ -0,0 +1,153 @@
...
+
+      let trailingMetadata = try await update.trailingMetadata
+      trailingMetadata.assertFirst("0", forName: "grpc-status")
+
+      let status = await update.status
+      XCTAssertTrue(status.isOk)
```
fabianfett, line 142: We have to `try` the trailingMetadata, but not the status? Is this intended? When do we get the status?

fabianfett (commented):
Some questions just to be sure.

Tests/GRPCTests/AsyncAwaitSupport/AsyncIntegrationTests.swift
glbrntt: At the moment, yes, we could. However, there's no guarantee that initial metadata is sent back on receipt of request metadata. In particular some gRPC implementations offer RPC implementors a way to set initial metadata which is delayed until the server sends a response message. We don't currently do this but it's been requested and I think we will add it in the future. When that's the case then we can't wait for this before sending the first request.

Tests/GRPCTests/AsyncAwaitSupport/AsyncIntegrationTests.swift
glbrntt: It is intended but it's something I'm not sure of API-wise going forward. In the ELF-flavoured API the status promise will always be succeeded. We receive the status at the end of every RPC -- even it doesn't complete (we synthesise one locally if that's the case).

Tests/GRPCTests/AsyncAwaitSupport/AsyncIntegrationTests.swift
fabianfett: Maybe we want to add this as a comment?

fabianfett (approved):

Tests/GRPCTests/AsyncAwaitSupport/AsyncIntegrationTests.swift
simonjbeaumont: See #1262 which adds support for our adopters to do just this.

glbrntt: Added a comment in 5be83ddbd4ba5c772b3e6317b84c429b87ce859e

PR #1261, opened 2021-09-11
[async-await] Code generation for "simple, but safe" wrapper client calls.

simonjbeaumont:
This PR adds codegen support for the "simple, but safe wrappers" that were part of the proposal for async/await support, added in #1231.

Codegen has also been re-run for the example Echo service.

### DIFF
--- a/Sources/Examples/Echo/Model/echo.grpc.swift
+++ b/Sources/Examples/Echo/Model/echo.grpc.swift
@@ -282,6 +282,7 @@ extension Echo_EchoAsyncClientProtocol {
       callOptions: callOptions ?? self.defaultCallOptions
     )
   }
+
   public func collect<RequestStream>(
     _ requests: RequestStream,
     callOptions: CallOptions? = nil
@@ -303,6 +304,7 @@ extension Echo_EchoAsyncClientProtocol {
       callOptions: callOptions ?? self.defaultCallOptions
     )
   }
+
   public func update<RequestStream>(
     _ requests: RequestStream,
     callOptions: CallOptions? = nil
--- a/Sources/Examples/Echo/Model/echo.grpc.swift
+++ b/Sources/Examples/Echo/Model/echo.grpc.swift
@@ -254,69 +254,44 @@ extension Echo_EchoAsyncClientProtocol {
     _ request: Echo_EchoRequest,
     callOptions: CallOptions? = nil
   ) async throws -> Echo_EchoResponse {
-    return try await self.makeGetCall(
-      request,
-      callOptions: callOptions
-    ).response
+    return try await self.performAsyncUnaryCall(
+      path: "/echo.Echo/Get",
+      request: request,
+      callOptions: callOptions ?? self.defaultCallOptions
+    )
   }
 
   public func expand(
     _ request: Echo_EchoRequest,
     callOptions: CallOptions? = nil
   ) -> GRPCAsyncResponseStream<Echo_EchoResponse> {
-    return self.makeExpandCall(
-      request,
-      callOptions: callOptions
-    ).responses
+    return self.performAsyncServerStreamingCall(
+      path: "/echo.Echo/Expand",
+      request: request,
+      callOptions: callOptions ?? self.defaultCallOptions
+    )
   }
 
   public func collect<RequestStream>(
     requests: RequestStream,
     callOptions: CallOptions? = nil
   ) async throws -> Echo_EchoResponse where RequestStream: AsyncSequence, RequestStream.Element == Echo_EchoRequest {
-    let call = self.makeCollectCall(
-      callOptions: callOptions
+    return try await self.performAsyncClientStreamingCall(
+      path: "/echo.Echo/Collect",
+      requests: requests,
+      callOptions: callOptions ?? self.defaultCallOptions
     )
-    return try await withTaskCancellationHandler {
-      try Task.checkCancellation()
-      for try await request in requests {
-        try Task.checkCancellation()
-        try await call.sendMessage(request)
-      }
-      try Task.checkCancellation()
-      try await call.sendEnd()
-      try Task.checkCancellation()
-      return try await call.response
-    } onCancel: {
-      Task.detached {
-        try await call.cancel()
-      }
-    }
   }
 
   public func update<RequestStream>(
     requests: RequestStream,
     callOptions: CallOptions? = nil
   ) -> GRPCAsyncResponseStream<Echo_EchoResponse> where RequestStream: AsyncSequence, RequestStream.Element == Echo_EchoRequest {
-    let call = self.makeUpdateCall(
-      callOptions: callOptions
+    return self.performAsyncBidirectionalStreamingCall(
+      path: "/echo.Echo/Update",
+      requests: requests,
+      callOptions: callOptions ?? self.defaultCallOptions
     )
-    Task {
-      try await withTaskCancellationHandler {
-        try Task.checkCancellation()
-        for try await request in requests {
-          try Task.checkCancellation()
-          try await call.sendMessage(request)
-        }
-        try Task.checkCancellation()
-        try await call.sendEnd()
-      } onCancel: {
-        Task.detached {
-          try await call.cancel()
-        }
-      }
-    }
-    return call.responses
   }
 }
 
--- a/Sources/Examples/Echo/Model/echo.grpc.swift
+++ b/Sources/Examples/Echo/Model/echo.grpc.swift
@@ -275,6 +275,16 @@ extension Echo_EchoAsyncClientProtocol {
   public func collect<RequestStream>(
     requests: RequestStream,
     callOptions: CallOptions? = nil
+  ) async throws -> Echo_EchoResponse where RequestStream: Sequence, RequestStream.Element == Echo_EchoRequest {
+    return try await self.performAsyncClientStreamingCall(
+      path: "/echo.Echo/Collect",
+      requests: requests,
+      callOptions: callOptions ?? self.defaultCallOptions
+    )
+  }
+  public func collect<RequestStream>(
+    requests: RequestStream,
+    callOptions: CallOptions? = nil
   ) async throws -> Echo_EchoResponse where RequestStream: AsyncSequence, RequestStream.Element == Echo_EchoRequest {
     return try await self.performAsyncClientStreamingCall(
       path: "/echo.Echo/Collect",
@@ -283,6 +293,16 @@ extension Echo_EchoAsyncClientProtocol {
     )
   }
 
+  public func update<RequestStream>(
+    requests: RequestStream,
+    callOptions: CallOptions? = nil
+  ) -> GRPCAsyncResponseStream<Echo_EchoResponse> where RequestStream: Sequence, RequestStream.Element == Echo_EchoRequest {
+    return self.performAsyncBidirectionalStreamingCall(
+      path: "/echo.Echo/Update",
+      requests: requests,
+      callOptions: callOptions ?? self.defaultCallOptions
+    )
+  }
   public func update<RequestStream>(
     requests: RequestStream,
     callOptions: CallOptions? = nil
--- a/Sources/Examples/Echo/Model/echo.grpc.swift
+++ b/Sources/Examples/Echo/Model/echo.grpc.swift
@@ -249,6 +249,78 @@ extension Echo_EchoAsyncClientProtocol {
 }
 
 @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+extension Echo_EchoAsyncClientProtocol {
+  public func get(
+    _ request: Echo_EchoRequest,
+    callOptions: CallOptions? = nil
+  ) async throws -> Echo_EchoResponse {
+    return try await self.makeGetCall(
+      request,
+      callOptions: callOptions
+    ).response
+  }
+
+  public func expand(
+    _ request: Echo_EchoRequest,
+    callOptions: CallOptions? = nil
+  ) -> GRPCAsyncResponseStream<Echo_EchoResponse> {
+    return self.makeExpandCall(
+      request,
+      callOptions: callOptions
+    ).responses
+  }
+
+  public func collect<RequestStream>(
+    requests: RequestStream,
+    callOptions: CallOptions? = nil
+  ) async throws -> Echo_EchoResponse where RequestStream: AsyncSequence, RequestStream.Element == Echo_EchoRequest {
+    let call = self.makeCollectCall(
+      callOptions: callOptions
+    )
+    return try await withTaskCancellationHandler {
+      try Task.checkCancellation()
+      for try await request in requests {
+        try Task.checkCancellation()
+        try await call.sendMessage(request)
+      }
+      try Task.checkCancellation()
+      try await call.sendEnd()
+      try Task.checkCancellation()
+      return try await call.response
+    } onCancel: {
+      Task.detached {
+        try await call.cancel()
+      }
+    }
+  }
+
+  public func update<RequestStream>(
+    requests: RequestStream,
+    callOptions: CallOptions? = nil
+  ) -> GRPCAsyncResponseStream<Echo_EchoResponse> where RequestStream: AsyncSequence, RequestStream.Element == Echo_EchoRequest {
+    let call = self.makeUpdateCall(
+      callOptions: callOptions
+    )
+    Task {
+      try await withTaskCancellationHandler {
+        try Task.checkCancellation()
+        for try await request in requests {
+          try Task.checkCancellation()
+          try await call.sendMessage(request)
+        }
+        try Task.checkCancellation()
+        try await call.sendEnd()
+      } onCancel: {
+        Task.detached {
+          try await call.cancel()
+        }
+      }
+    }
+    return call.responses
+  }
+}
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
 public struct Echo_EchoAsyncClient: Echo_EchoAsyncClientProtocol {
   public var channel: GRPCChannel
   public var defaultCallOptions: CallOptions
--- a/Sources/Examples/Echo/Model/echo.grpc.swift
+++ b/Sources/Examples/Echo/Model/echo.grpc.swift
@@ -273,7 +273,7 @@ extension Echo_EchoAsyncClientProtocol {
   }
 
   public func collect<RequestStream>(
-    requests: RequestStream,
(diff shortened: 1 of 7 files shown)


Sources/Examples/Echo/Model/echo.grpc.swift
```diff
@@ -172,6 +172,173 @@ public final class Echo_EchoClient: Echo_EchoClientProtocol {
...
+
+  public func collect<RequestStream>(
+    requests: RequestStream,
+    callOptions: CallOptions? = nil
+  ) async throws -> Echo_EchoResponse where RequestStream: AsyncSequence, RequestStream.Element == Echo_EchoRequest {
+    let call = self.makeCollectCall(
+      callOptions: callOptions
+    )
+    return try await withTaskCancellationHandler {
+      try Task.checkCancellation()
```
glbrntt, line 281: What's the right amount of checking for cancellation? This _feels_ like a lot.

Sources/Examples/Echo/Model/echo.grpc.swift
```diff
@@ -172,6 +172,173 @@ public final class Echo_EchoClient: Echo_EchoClientProtocol {
...
+      for try await request in requests {
+        try Task.checkCancellation()
+        try await call.sendMessage(request)
+      }
+      try Task.checkCancellation()
+      try await call.sendEnd()
```
glbrntt, line 287: It's possible for the server to respond before the client has sent all of their requests. In this case we'd be blocking until we tried to write them all.

Sources/Examples/Echo/Model/echo.grpc.swift
```diff
@@ -172,6 +172,173 @@ public final class Echo_EchoClient: Echo_EchoClientProtocol {
...
+  ) -> GRPCAsyncBidirectionalStreamingCall<Echo_EchoRequest, Echo_EchoResponse> {
+    return self.makeAsyncBidirectionalStreamingCall(
+      path: "/echo.Echo/Update",
+      callOptions: callOptions ?? self.defaultCallOptions
+    )
+  }
+}
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+extension Echo_EchoAsyncClientProtocol {
```
glbrntt, line 252: I'm a bit hesitant to put the implementation of the simple-safe wrappers in the generated code for a couple of reasons:

1. It stops users from using the simple-safe wrappers who aren't using generated code (not using generated code is not all that rare)
2. Update are driven by the user regenerating their generated code

I think we add extensions to `GRPCClient` and call that in the generated code.

glbrntt (changes requested):

Sources/Examples/Echo/Model/echo.grpc.swift
simonjbeaumont: Hm, OK. How would you like these spelled?

Sources/Examples/Echo/Model/echo.grpc.swift
simonjbeaumont: Well I guess these are all legitimate early exit points. We could omit as many of these as we like, each omission resulting in some different semantics in the presence of cancellation. 

We might want to enumerate the things we are OK with happening after a cancellation?

Sources/Examples/Echo/Model/echo.grpc.swift
simonjbeaumont: That’s a good point I hadn’t considered for the PoC. So we might need a task sending the requests which we can cancel if we get a response. I’ll have a play with that.

Sources/Examples/Echo/Model/echo.grpc.swift
simonjbeaumont: I have spelled these `perform<type>Call` and made use of them in codgen in the update to this PR.

Sources/GRPC/AsyncAwaitSupport/GRPCClient+AsyncAwaitSupport.swift
```diff
@@ -148,4 +148,194 @@ extension GRPCClient {
...
+  public func performAsyncBidirectionalStreamingCall<
+    Request: GRPCPayload,
+    Response: GRPCPayload,
+    RequestStream: AsyncSequence
+  >(
+    path: String,
+    requests: RequestStream,
+    callOptions: CallOptions? = nil,
+    interceptors: [ClientInterceptor<Request, Response>] = []
+  ) -> GRPCAsyncResponseStream<Response>
```
simonjbeaumont, line 278: @glbrntt The existing `make<mumble>Call` extensions take additional, defaulted type params. Could you confirm why and if they are needed here or not?

Sources/Examples/Echo/Model/echo.grpc.swift
simonjbeaumont: Also... I added the extensions to `GRPCClient` but I also see we have a bunch of extensions on `GRPCChannel`. Do we want these wrappers to be on the channel API too?

Sources/Examples/Echo/Model/echo.grpc.swift
simonjbeaumont: I've started looking at this but I'm unsure what to do in the event we receive a response before exhausting the request stream.

* Do we consider that to be a success? I think yes.
* Should the client send end?
* Should the client cancel the call?
* Should the client do neither?

simonjbeaumont: How does f6b1293b look?

simonjbeaumont:
@glbrntt I rebased now that you merged the other PR.

Sources/Examples/Echo/Model/echo.grpc.swift
glbrntt: Apologies -- I had a comment written but didn't publish it:

---

Good question; we're not returning `Async{...}Call`s any more so `makeAsync{...}Call` certainly isn't right (and in some cases we'd be overloading on the return type which is a very bad idea).

Nothing jumps out at me. A couple of ideas though:
- `invoke{Unary....}RPC(...)`
- `invoke{Unary,...}Call(...)`
- `invoke{Unary....}(...)`

The could instead be called from wrapper struct with "friendlier" API. This is nice in that it separates the safe and unsafe calls from 

```swift
extension GRPCChannel {
  // Names are placeholders...
  public var asyncWrappers: GRPCAsyncWrappers { 
    return GRPCAsyncWrappers(self)
  }
}

public struct GRPCAsyncWrappers {
  public func unary<Request, Response>(...) async throws -> Response { ... }
  public func clientStreaming<Request, Response>(...) async throws -> Response { ... }
  // ... etc
}
```

glbrntt: > Also... I added the extensions to `GRPCClient` but I also see we have a bunch of extensions on `GRPCChannel`. Do we want these wrappers to be on the channel API too?

Good spot; we probably don't actually need or want the extensions on `GRPCChannel`.

glbrntt: > I have spelled these `perform<type>Call` and made use of them in codgen in the update to this PR.

I don't love it (or any of the other names I proposed) but it's fine for now.

Sources/GRPC/AsyncAwaitSupport/GRPCClient+AsyncAwaitSupport.swift
glbrntt: Let's keep them: they're useful when part of the type signature can't be inferred as they allow you to not spell out the complete types.

Sources/Examples/Echo/Model/echo.grpc.swift
```diff
@@ -248,6 +248,53 @@ extension Echo_EchoAsyncClientProtocol {
...
+      path: "/echo.Echo/Expand",
+      request: request,
+      callOptions: callOptions ?? self.defaultCallOptions
+    )
+  }
+
+  public func collect<RequestStream>(
+    requests: RequestStream,
+    callOptions: CallOptions? = nil
+  ) async throws -> Echo_EchoResponse where RequestStream: AsyncSequence, RequestStream.Element == Echo_EchoRequest {
```
glbrntt, line 278: It'd be useful to add a `Sequence` variant for this and bidi streaming too.

Sources/GRPC/AsyncAwaitSupport/GRPCClient+AsyncAwaitSupport.swift
```diff
@@ -148,4 +148,204 @@ extension GRPCClient {
...
+    return self.channel.makeAsyncServerStreamingCall(
+      path: path,
+      request: request,
+      callOptions: callOptions ?? self.defaultCallOptions,
+      interceptors: interceptors
+    ).responses
+  }
+
+  public func performAsyncClientStreamingCall<
+    Request: SwiftProtobuf.Message,
```
glbrntt, line 213: ~Does this need to be generic on `Request`? I'd think we'd be able to pull the request type out from `RequestStream`?~

Oh, probably not as it's used for `ClientInterceptor`

Sources/GRPC/AsyncAwaitSupport/GRPCClient+AsyncAwaitSupport.swift
```diff
@@ -148,4 +148,204 @@ extension GRPCClient {
...
+    return self.perform(call, requests)
+  }
+}
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+extension GRPCClient {
+  @inlinable
+  internal func perform<Request, Response, RequestStream>(
+    _ call: GRPCAsyncClientStreamingCall<Request, Response>,
+    _ requests: RequestStream
```
glbrntt, line 294: nit: can we have `with requests: RequestStream`?

Sources/GRPC/AsyncAwaitSupport/GRPCClient+AsyncAwaitSupport.swift
```diff
@@ -148,4 +148,204 @@ extension GRPCClient {
...
+  )
+  async throws -> Response
```
glbrntt, line 296: This looks weird on its own line -- was this you or SwiftFormat?

glbrntt (changes requested):
This broadly looks good. I left a couple of comments inline. Could we also have some very simple tests to exercise these? (You can bootstrap these: https://github.com/grpc/grpc-swift/blob/async-await/Tests/GRPCTests/AsyncAwaitSupport/AsyncIntegrationTests.swift)

Sources/Examples/Echo/Model/echo.grpc.swift
simonjbeaumont: Done in 98026b68.

Sources/GRPC/AsyncAwaitSupport/GRPCClient+AsyncAwaitSupport.swift
simonjbeaumont: Done in 01e0f4f1.

Sources/GRPC/AsyncAwaitSupport/GRPCClient+AsyncAwaitSupport.swift
simonjbeaumont: Done in 71086554.

Sources/GRPC/AsyncAwaitSupport/GRPCClient+AsyncAwaitSupport.swift
simonjbeaumont: 🤷 

Put it "right" in ded6df74.

Sources/GRPC/AsyncAwaitSupport/GRPCClient+AsyncAwaitSupport.swift
```diff
@@ -148,4 +148,314 @@ extension GRPCClient {
...
+        }
+      }
+    }
+    return call.responses
+  }
+}
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+extension AsyncStream {
+  fileprivate init<T>(wrapping sequence: T) where T: Sequence, T.Element == Element {
```
glbrntt, line 450: Mind adding a note here that we could special case the request stream being a `Sequence` to avoid going via `AsyncStream`?

Sources/Examples/Echo/Model/echo.grpc.swift
glbrntt: We're missing the codegen for this now.

Sources/protoc-gen-grpc-swift/Generator-Client+AsyncAwait.swift
```diff
@@ -130,6 +130,115 @@ extension Generator {
...
+      // 'Safe' calls.
+      for (i, method) in self.service.methods.enumerated() {
+        if i > 0 {
+          self.println()
+        }
+        self.method = method
+
+        let rpcType = streamingType(self.method)
+        let callTypeWithoutPrefix = Types.call(for: rpcType, withGRPCPrefix: false)
+        let responseStreamType = "GRPCAsyncResponseStream"
```
glbrntt, line 148: nit: use `Types.responseStream(of: self.methodOutputName)`

Sources/protoc-gen-grpc-swift/Generator-Client+AsyncAwait.swift
```diff
@@ -130,6 +130,115 @@ extension Generator {
...
+              "return try await self.perform\(callTypeWithoutPrefix)",
+              braces: .round
+            ) {
+              self.println("path: \(self.methodPath),")
+              self.println("request: request,")
+              self.println("callOptions: callOptions ?? self.defaultCallOptions")
+            }
+          }
+
+        case .serverStreaming:
```
glbrntt, line 173: This is the same as `.unary`, I think?

Sources/protoc-gen-grpc-swift/Generator-Client+AsyncAwait.swift
```diff
@@ -130,6 +130,115 @@ extension Generator {
...
+              "return try await self.perform\(callTypeWithoutPrefix)",
+              braces: .round
+            ) {
+              self.println("path: \(self.methodPath),")
+              self.println("requests: requests,")
+              self.println("callOptions: callOptions ?? self.defaultCallOptions")
+            }
+          }
+
+        case .bidirectionalStreaming:
```
glbrntt, line 216: This is the same as `.clientStreaming` I think. We also need the `Sequence` variant codegen.

Sources/Examples/Echo/Model/echo.grpc.swift
simonjbeaumont: 🤦

Sources/protoc-gen-grpc-swift/Generator-Client+AsyncAwait.swift
simonjbeaumont: You are partially right! This wasn't the case before we factored the code into the `GRPCClient` extension which was probably why I missed it, but there's more similarity now. However there are still some notable differences:

* whether the input is generic
* input name param: request vs requests
* return type: response vs. response stream
* effects: `async throws` vs none
* return statement requiring `try await` vs not

What I have done is created commit a62c2e96 which refactors this. Do you prefer it or does it add too much indirection to read the codegen clearly?

Sources/protoc-gen-grpc-swift/Generator-Client+AsyncAwait.swift
simonjbeaumont: See comment on other thread https://github.com/grpc/grpc-swift/pull/1261/files#r709267054.

simonjbeaumont:
@glbrntt I see a rebase will be necessary since the merge of #1264. Would you like time to review _before_ that?

Sources/protoc-gen-grpc-swift/Generator-Client+AsyncAwait.swift
```diff
@@ -130,6 +130,65 @@ extension Generator {
...
+
+        let rpcType = streamingType(self.method)
+        let callTypeWithoutPrefix = Types.call(for: rpcType, withGRPCPrefix: false)
+
+        let streamsResponses = [.serverStreaming, .bidirectionalStreaming].contains(rpcType)
+        let streamsRequests = [.clientStreaming, .bidirectionalStreaming].contains(rpcType)
+
+        let sequenceProtocols = streamsRequests ? ["Sequence", "AsyncSequence"] : [nil]
+
+        for sequenceProtocol in sequenceProtocols {
```
glbrntt, line 153: nit: can we print an empty line between each?

Sources/protoc-gen-grpc-swift/Generator-Client+AsyncAwait.swift
glbrntt: Ah, I missed that -- I'm happy with how it is now.

Sources/protoc-gen-grpc-swift/Generator-Client+AsyncAwait.swift
glbrntt, line 153: nit: can we get a blank line between each?

glbrntt (commented):
Cool -- this looks great modulo one nit. There's also an outstanding comment from a previous review:

> Could we also have some very simple tests to exercise these? (You can bootstrap these: https://github.com/grpc/grpc-swift/blob/async-await/Tests/GRPCTests/AsyncAwaitSupport/AsyncIntegrationTests.swift)

simonjbeaumont:
@glbrntt wrote:

> Cool -- this looks great modulo one nit. There's also an outstanding comment from a previous review:
> 
> > Could we also have some very simple tests to exercise these? (You can bootstrap these: https://github.com/grpc/grpc-swift/blob/async-await/Tests/GRPCTests/AsyncAwaitSupport/AsyncIntegrationTests.swift)

Oops, missed this one. Done in 57998ff1.

Tests/GRPCTests/AsyncAwaitSupport/AsyncIntegrationTests.swift
```diff
@@ -73,6 +73,13 @@ final class AsyncIntegrationTests: GRPCTestCase {
     }
   }
 
+  func testUnaryWrapper() {
+    XCTAsyncTest {
+      let response = try await self.echo.get(.with { $0.text = "hello" })
```
glbrntt, line 78: The 'wrapped' API is _so_ much nicer!

glbrntt (approved):
Awesome! Thank you Si!

PR #1263, opened 2021-09-13
Update swift-argument-parser version

glbrntt:
Motivation:

Stable APIs are better than unstable APIs. swift-argument-parser has
released a 1.0.0 and now has a stable API.

Modifications:

Change swift-argument-parser version requirements to "from: 1.0.0"

Result:

Dependencies are more stable.


Lukasa (approved):
:tada:

PR #1265, opened 2021-09-15
Extend lifetime of client interceptor pipeline

glbrntt:
Motivation:

A client call (i.e. the object the user holds) may live longer than the
transport associated with it (roughly speaking, the http/2 stream channel). An
example of this is when interceptors are use to retry and RPC and
redirect responses back to the original call.

However, the interceptor pipeline is held by the transport and is
currently set to nil when the transport is removed from the channel.
This means events invoked from the call object (such as cancellation)
which go via the transport (holding the interceptor pipeline) are
incorrectly failed.

Modifications:

- Have the client interceptor pipeline break the ref cycle between the
  transport and itself when the interceptor pipeline closes rather than
  when the transport is closed
- Emit a cancellation status rater than error on cancellation
- Update the ordering of when close is called in the interceptor
  pipeline.
- Add and update tests

Result:

"sub"-RPCs may be cancelled.


Sources/GRPC/Interceptor/ClientInterceptorPipeline.swift
```diff
@@ -296,6 +298,9 @@ internal final class ClientInterceptorPipeline<Request, Response> {
 
     // Emit the unwrapped error.
     self._onError(unwrappedError)
+
+    // Close the pipeline.
+    self.close()
```
Davidde94, line 303: Should `self.close()` not set `self._isOpen`.

Tests/GRPCTests/EchoHelpers/Interceptors/File.swift
```diff
@@ -0,0 +1,87 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import EchoModel
+import GRPC
+
+// MARK: - Client
+
+internal final class EchoClientInterceptors: Echo_EchoClientInterceptorFactoryProtocol {
```
Davidde94, line 21: Should this file have a different name?

Sources/GRPC/Interceptor/ClientInterceptorPipeline.swift
glbrntt: It does -- but we also need to ensure we do it here because we call out on `_onError`

Tests/GRPCTests/EchoHelpers/Interceptors/File.swift
glbrntt: Ugh, Xcode 😤 -- thanks for spotting!

Lukasa (approved):
LGTM, nice change.

Sources/GRPC/Interceptor/ClientInterceptorPipeline.swift
```diff
@@ -79,16 +79,16 @@ internal final class ClientInterceptorPipeline<Request, Response> {
...
-  internal let _onCancel: (EventLoopPromise<Void>?) -> Void
+  internal var _onCancel: ((EventLoopPromise<Void>?) -> Void)?
 
   @usableFromInline
-  internal let _onRequestPart: (GRPCClientRequestPart<Request>, EventLoopPromise<Void>?) -> Void
+  internal var _onRequestPart: ((GRPCClientRequestPart<Request>, EventLoopPromise<Void>?) -> Void)?
 
   @usableFromInline
-  internal let _onResponsePart: (GRPCClientResponsePart<Response>) -> Void
+  internal var _onResponsePart: ((GRPCClientResponsePart<Response>) -> Void)?
```
fabianfett, line 92: If those are now optional, shouldn't we change the initializer?

fabianfett: Since, we nil them ourselves later, maybe we can make them:
```
internal private(set) var
```

This would communicate the intent a little better.

glbrntt: > If those are now optional, shouldn't we change the initializer?

We only want to `nil` them out when we close.

> Since, we nil them ourselves later, maybe we can make them:
> 
> ```
> internal private(set) var
> ```

Sadly not as they are `@usableFromInline` (the leading `_` indicates they'd be `private` otherwise)

glbrntt: I take that back; to my surprise `internal private(set)` works with `@usableFromInline`!

Sources/GRPC/Interceptor/ClientInterceptorPipeline.swift
```diff
@@ -437,7 +444,14 @@ extension ClientInterceptorPipeline {
     self._scheduledClose = nil
 
     // Cancel the transport.
-    self._onCancel(nil)
+    self._onCancel?(nil)
```
fabianfett, line 447: I think we should update the method here. 

```swift
  /// Closes the pipeline. This must be called once, by the tail interceptor, to indicate that
  /// the RPC has completed. If this is not called, we will leak.
```

fabianfett (approved):

PR #1269, opened 2021-09-16
Add note to README about async/await support.

glbrntt:
Motivation:

The `async-await` branch has enough code in place that we can tag an
experimental async/await version. We should, however, make a few things
clear:

- that it is experimental API and subject to change
- which versions contain async/await code
- how to generate async/await code

Modifications:

- Add an async/await section to README.md
- Update docs/plugin.md to include the async/await codegen options

Result:

We're clearer about async/await support


Lukasa (approved):
Nice one!

PR #1272, opened 2021-09-19
[async-await] Added generation of interceptors

Nekitosss:
Currently interceptors does not propagated to calls. Seems like it was forgotten to be implemented in this alpha implementation.

### DIFF
--- a/Tests/GRPCTests/InterceptorsAsyncTests.swift
+++ b/Tests/GRPCTests/InterceptorsAsyncTests.swift
@@ -1,0 +1,174 @@
+/*
+ * Copyright 2021, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import EchoImplementation
+import EchoModel
+import GRPC
+import HelloWorldModel
+import NIOCore
+import NIOHPACK
+import NIOPosix
+import SwiftProtobuf
+import XCTest
+
+
+#if compiler(>=5.5)
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+class InterceptorsAsyncTests: GRPCTestCase {
+  private var group: EventLoopGroup!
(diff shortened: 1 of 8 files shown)


Lukasa:
Thanks for this! Can you regenerate the various grpc examples? `make generate-echo generate-hello world generate-route-guide generate-normalization` should do it.

Nekitosss:
> Thanks for this! Can you regenerate the various grpc examples? `make generate-echo generate-helloworld generate-route-guide generate-normalization` should do it.

Done. Async flag set only for echo generation, but I still launch whole generation. It added additional newline to the bottom of files. Not sure is it ok or not :)

Lukasa:
The extra newlines are fine :smile:

Lukasa:
Do you feel up to writing some tests? I'd love to see a couple of new tests, probably in `InterceptorsTests`, that validate that the interceptors continue to work correctly in async/await land.

Nekitosss:
I decided to create separate test class for async interceptors in analogy with other async tests like `GRPCAsyncClientCallTests`. I called file `InterceptorsAsyncTests` so that its be close to default counterpart `InterceptorsTests`.
Also, writing tests directly in current `InterceptorsTests` file will cause it suffer from lots of `ifdef`s and `@available`s

Waiting for further improvements suggestion 👌

Nekitosss:
Hello, @Lukasa! What are next steps to make further progress in this PR? Should I wait more for your review when you will have time or should I improve it somehow?

Tests/GRPCTests/InterceptorsAsyncTests.swift
```diff
@@ -0,0 +1,173 @@
+/*
```
glbrntt, line 1: Could you move this file to the `Tests/GRPCTests/AsyncAwaitSupport` directory?

Tests/GRPCTests/InterceptorsAsyncTests.swift
```diff
@@ -0,0 +1,173 @@
...
+import GRPC
+import HelloWorldModel
+import NIOCore
+import NIOHPACK
+import NIOPosix
+import SwiftProtobuf
+import XCTest
+
+
+#if compiler(>=5.5)
```
glbrntt, line 27:
```suggestion
#if compiler(>=5.5)

import EchoImplementation
import EchoModel
import GRPC
import HelloWorldModel
import NIOCore
import NIOHPACK
import NIOPosix
import SwiftProtobuf
import XCTest

```

glbrntt (commented):
Oops..! Thanks for catching and fixing this @Nekitosss. Would you mind running `./scripts/format.sh` to make the CI happy?

Tests/GRPCTests/InterceptorsAsyncTests.swift
Nekitosss: Done

glbrntt (approved):
Great, thanks @Nekitosss 🙏

PR #1278, opened 2021-09-28
Add Echo server and client example using async-await API

simonjbeaumont:
This creates a new executable target, `AsyncAwaitEcho`, which is equivalent to `Echo` except that is makes use of the new async-await–based client and provider APIs.

Signed-off-by: Si Beaumont <beaumont@apple.com>

### DIFF
--- a/Sources/Examples/Echo/AsyncAwaitRuntime/ArgumentParser+AsyncAwait.swift
+++ b/Sources/Examples/Echo/AsyncAwaitRuntime/ArgumentParser+AsyncAwait.swift
@@ -17,7 +17,7 @@
 /// NOTE: This file should be removed when the `async` branch of `swift-argument-parser` has been
 ///       released: https://github.com/apple/swift-argument-parser/tree/async
 
-#if compiler(>=5.5)
+#if compiler(>=5.5) && canImport(_Concurrency)
 
 import ArgumentParser
 
--- a/Sources/Examples/Echo/AsyncAwaitRuntime/ArgumentParser+AsyncAwait.swift
+++ b/Sources/Examples/Echo/AsyncAwaitRuntime/ArgumentParser+AsyncAwait.swift
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2021, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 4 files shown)


Sources/Examples/Echo/AsyncAwaitRuntime/ArgumentParser+AsyncAwait.swift
```diff
@@ -0,0 +1,52 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/// NOTE: This file should be removed when the `async` branch of `swift-argument-parser` has been
+///       released: https://github.com/apple/swift-argument-parser/tree/async
+
+#if compiler(>=5.5)
```
glbrntt, line 20: nit: `&& canImport(_Concurrency)` and elsewhere

Sources/Examples/Echo/AsyncAwaitRuntime/main.swift
```diff
@@ -0,0 +1,260 @@
...
+      try await echoUpdate(client: client, message: message)
+    }
+  } catch {
+    print("\(rpc) RPC failed: \(error)")
+  }
+}
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+func echoGet(client: Echo_EchoAsyncClient, message: String) async throws {
+  async let response = client.get(.with { $0.text = message })
```
glbrntt, line 207: Why `async let` here (and in collect)? Given we `await` on the next line it would be more idiomatic to await here instead.

Sources/Examples/Echo/AsyncAwaitRuntime/main.swift
```diff
@@ -0,0 +1,260 @@
...
+import Dispatch
+let dg = DispatchGroup()
+dg.enter()
+if #available(macOS 12, iOS 15, tvOS 15, watchOS 8, *) {
+  Task {
+    await Echo.main()
+    dg.leave()
+  }
+} else {
+  dg.leave()
```
glbrntt, line 251: nit: we should print an error here too

glbrntt (commented):
Looks good aside from a few nits. 

Could you also add a note to the README in `Sources/Examples/Echo` to mention that there are two versions, the original one and this new async/await one?

glbrntt (commented):
Could you rebase onto the `1.4.1-async-await` branch as well?

simonjbeaumont (commented):
Addressed the inline comments.

> Could you also add a note to the README in Sources/Examples/Echo to mention that there are two versions, the original one and this new async/await one?

Added in 800b72ec.

simonjbeaumont:
@glbrntt rebased and retargeted PR for the `1.4.1` branch.

Sources/Examples/Echo/AsyncAwaitRuntime/main.swift
glbrntt: This one hasn't been done for 'get'. A victim of the rebase?

simonjbeaumont: Oops. Fixed in 3f5ee89e.

glbrntt (approved):
Great, thanks Si!

PR #1284, opened 2021-09-30
Add a pooled channel

glbrntt:
Motivation:

We added the various internals of the connection pool but are yet to
surface it as public API. This PR does just that.

Modifications:

- Add a `PooledChannel`, a `GRPCChannel` which uses the connection pool
- Add `GRPCPooledChannel` and `GRPCPooledChannel.Configuration` which
  produces `PooledChannel`s as `GRPCChannel`s.
- Add a handful of tests

Result:

We have a usable connection pool in the public API.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -51,16 +51,6 @@ jobs:
         include:
           - image: swift:5.4-focal
             env:
-              MAX_ALLOCS_ALLOWED_bidi_1k_rpcs_10_requests: 503000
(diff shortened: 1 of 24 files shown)


glbrntt:
I have a follow up PR with some allocation counting tests.

Lukasa (approved):
I have one minor note regarding inlinability but otherwise I think this patch is fantastic, and I'm really excited to have this functionality available in grpc-swift.

glbrntt:
Unfortunately we regress 1 alloc per RPC but I think it's worth it since switching to the pooled channel allocates quite a bit less and should be preferred.

PR #1289, opened 2021-10-05
Wire up connection backoff of channel pool

glbrntt:
Motivation:

`GRPCChannelPool.Configuration` has an entry controlling connection
backoff configuration. However this was never wired up to the underlying
connection pool and connections which just use the defaults.

Modifications:

- Propagate connection backoff through the API
- Add a test

Result:

Connection backoff is wired up as expected for `GRPCChannelPool`.


Lukasa (approved):
Nice easy patch, thanks!

PR #1290, opened 2021-10-05
Add a 'cause' to 'GRPCStatus'

glbrntt:
Motivation:

In some causes it's useful to know the underlying error which caused an
RPC to complete with a particular status. The current way of achieving
this is by embeddeding it in the status message. While the information
may be relevant to some callers it is noise to others. Moreover, to
those that want the information, they have to know how to search for it
(some strings may embed errors in different ways and so on). This also
precludes them from being able to switch on a particular error type, for
example.

Modifications:

- Add an optional 'cause' field to 'GRPCStatus', an error associated
  with the status.
- Adding cause directly into 'GRPCStatus' made it a little too wide so a
  backing storage was added for the storage and message. This is a bit
  of a shame since we now unconditionally allocate for messages but the
  happy path ('ok') should be without message or cause and so should not
  allocate.

Result:

- Causal errors can be attached to 'GRPCStatus'

### DIFF
--- a/Sources/GRPC/GRPCStatus.swift
+++ b/Sources/GRPC/GRPCStatus.swift
@@ -20,11 +20,68 @@ import NIOHTTP2
 
 /// Encapsulates the result of a gRPC call.
 public struct GRPCStatus: Error {
-  /// The status message of the RPC.
-  public var message: String?
+  /// Storage for message/cause. In the happy case ('ok') there will not be a message or cause
+  /// and this will reference a static storage containing nil values. Making it optional makes the
+  /// setters for message and cause a little messy.
+  private var storage: Storage = .makeStorage(message: nil, cause: nil)
 
   /// The status code of the RPC.
   public var code: Code
+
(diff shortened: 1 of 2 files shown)


Sources/GRPC/GRPCStatus.swift
```diff
@@ -20,20 +20,81 @@ import NIOHTTP2
 
 /// Encapsulates the result of a gRPC call.
 public struct GRPCStatus: Error {
-  /// The status message of the RPC.
-  public var message: String?
+  /// Storage for message/cause. In the happy case ('ok') there will not be a message or cause
+  /// and this will reference a static storage containing nil values. Making it optional makes the
+  /// setters for message and cause a little messy.
+  private var storage: Storage = .makeStorage(message: nil, cause: nil)
```
Lukasa, line 26: Can we remove the default value here? It risks being unconditionally called if we do anything interesting with inlinability here, and we don't use it on any code path anyway.

PR #1292, opened 2021-10-06
Provide more context to failing RPCs from the connection pool

glbrntt:
Motivation:

When an RPC is started using the connection pool, it waits for a
connection to come up and a stream to be created. It may also time out
waiting for the connection to come up, this may be because the pool is
genuinely too busy to handle the RPC or there is an underlying problem
with the connection.

In the latter case it's not possible for the RPC to see the underlying
reason that it timed out. This makes issue diagnosis more difficult than
necessary when there root cause is at the connection level (e.g. due to TLS
configuration) as important but infrequent connection-level errors are
lost in the noise of RPC-level timeouts.

Modifications:

- Give connection pools a notion of a 'last known error', the most
  recent connection-level error to occur on a connection managed by the
  pool. The error is cleared as soon as any connection within the pool
  becomes available. This error is added to existing waiter errors to
  provide more context.
- Add a '_ConnectivityState' mirroring the public 'ConnectivityState'
  but which holds 'Error' as associated data in some cases.
- Wire this through the `ConnectionManager` to the `ConnectionPool`

Result:

Connection level errors are more obvious.

### DIFF
--- a/Tests/GRPCTests/ConnectionPool/GRPCChannelPoolTests.swift
+++ b/Tests/GRPCTests/ConnectionPool/GRPCChannelPoolTests.swift
@@ -19,6 +19,7 @@ import GRPC
 import GRPCSampleData
 import NIO
 import NIOConcurrencyHelpers
+import NIOSSL
 import XCTest
 
 final class GRPCChannelPoolTests: GRPCTestCase {
@@ -107,7 +108,7 @@ final class GRPCChannelPoolTests: GRPCTestCase {
     self.startChannel(withTLS: tls) {
       // We'll allow any number of waiters since we immediately fire off a bunch of RPCs and don't
       // want to bounce off the limit as we wait for a connection to come up.
-      $0.connectionPool.maxWaitersPerEventLoop = .max
+      $0.connectionPool.maxWaitersPerEventLoop = 1000
     }
   }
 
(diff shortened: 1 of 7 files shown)


Tests/GRPCTests/ConnectionPool/GRPCChannelPoolTests.swift
```diff
@@ -107,7 +108,7 @@ final class GRPCChannelPoolTests: GRPCTestCase {
     self.startChannel(withTLS: tls) {
       // We'll allow any number of waiters since we immediately fire off a bunch of RPCs and don't
       // want to bounce off the limit as we wait for a connection to come up.
-      $0.connectionPool.maxWaitersPerEventLoop = .max
+      $0.connectionPool.maxWaitersPerEventLoop = 1000
```
Lukasa, line 111: The comment above seems to be out of step with this change.

glbrntt, line 111:
```suggestion
      $0.connectionPool.maxWaitersPerEventLoop = .max
```

glbrntt: Thanks, didn't actually need this change.

Issue #1301, opened 2021-10-12
Crash if receiving messages on streaming call that has finished (1.4.1-async-await.2)


qusc:
### Describe the bug

GRPC server crashes with `GRPC/GRPCAsyncServerHandler.swift:529: Fatal error` / `preconditionFailure()`.

### To reproduce

1. Immediately close client-side streaming call from the server
2. Have the client send more messages

### Expected behaviour

Any client behaviour should never cause the server to crash

### Additional information

I'm actually not entirely sure in which setup this error occurs. It also doesn't seem to appear to occur immediately (race condition?). I can however reliably reproduce it after a few seconds with my client / server setup. Not closing the streaming call seems to prohibit this error. Also may be noteworthy that the client restarts the streaming call frequently if it finishes since it's supposed to be open all the time.

Full stack trace:

```
#0	0x00007ff81dbf72d0 in _swift_runtime_on_report ()
#1	0x00007ff81dc762e9 in _swift_stdlib_reportFatalErrorInFile ()
#2	0x00007ff81d936f85 in closure #1 in closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) ()
#3	0x00007ff81d936ca5 in closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) ()
#4	0x00007ff81d935cdc in _assertionFailure(_:_:file:line:flags:) ()
#5	0x00000001000aa3bb in AsyncServerHandler._interceptResponse(_:metadata:) at GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift:529
#6	0x00000001000aa460 in closure #1 in AsyncServerHandler.interceptResponse(_:metadata:) at GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift:539
#7	0x000000010038097c in thunk for @escaping @callee_guaranteed () -> () ()
#8	0x00000001003842b1 in partial apply for thunk for @escaping @callee_guaranteed () -> () ()
#9	0x000000010038099c in thunk for @escaping @callee_guaranteed () -> (@out ()) ()
#10	0x00000001003809e2 in closure #4 in SelectableEventLoop.run() at NIOPosix/SelectableEventLoop.swift:509
#11	0x000000010030713f in thunk for @callee_guaranteed () -> (@error @owned Error) ()
#12	0x0000000100383454 in thunk for @callee_guaranteed () -> (@error @owned Error)partial apply ()
[...]

Issue #1299, opened 2021-10-11
Crash on connection close (1.4.1-async-await.2)


qusc:
I'm experiencing the following bug on `1.4.1-async-await.2`:

### Describe the bug

When I'm closing my connected iOS app via the app switcher my GRPC Swift server (on 1.4.1-async-await.2) crashes with:
`GRPC/HTTP2ToRawGRPCServerCodec.swift:334: Fatal error: Unexpectedly found nil while implicitly unwrapping an Optional value`

### To reproduce

1. Setup server with async service provider
2. Connect client, open streaming call
3. Close (force-)close client app

### Expected behaviour

No crash when I'm closing a connected client app.

### Additional information

Full stack trace:

```
#0	0x00007ff81dbf72d0 in _swift_runtime_on_report ()
#1	0x00007ff81dc762e9 in _swift_stdlib_reportFatalErrorInFile ()
#2	0x00007ff81d936913 in closure #1 in closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) ()
#3	0x00007ff81d93667c in closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) ()
#4	0x00007ff81d936355 in closure #1 in _assertionFailure(_:_:file:line:flags:) ()
#5	0x00007ff81d935faa in _assertionFailure(_:_:file:line:flags:) ()
#6	0x00000001001ac1d7 in HTTP2ToRawGRPCServerCodec.sendTrailers(_:promise:) at /SourcePackages/checkouts/grpc-swift/Sources/GRPC/HTTP2ToRawGRPCServerCodec.swift:334
#7	0x00000001001abe28 in HTTP2ToRawGRPCServerCodec.sendEnd(status:trailers:promise:) at /SourcePackages/checkouts/grpc-swift/Sources/GRPC/HTTP2ToRawGRPCServerCodec.swift:317
#8	0x00000001001ac445 in protocol witness for GRPCServerResponseWriter.sendEnd(status:trailers:promise:) in conformance HTTP2ToRawGRPCServerCodec ()
#9	0x00000001000a4e1f in AsyncServerHandler.sendInterceptedPart(_:promise:) at /SourcePackages/checkouts/grpc-swift/Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift:669
[...]

PR #1302, opened 2021-10-12
async-await: Move to completed state before cancelling task during finish()

simonjbeaumont:
As reported in #1299 and #1301, there are some scenarios where the closing the channel or using the response stream writer after the channel has been closed will lead to a crash.

Specifically, when `finish()` is called the state was not progressed to `.completed` before cancelling the task. This was to maintain parity with the ELG-based API where the status and the trailers were still sent after `finish()` is called. We now believe this to be misguided and we shouldn't expect to be able to send anything on the channel at this point because we are tearing the handler and the channel down.

This changes `finish()` to move to the `.completed` state before cancelling the `userHandlerTask`. As a result, when the completion handler for the user function fires, it will call `handleError(_:)` with `CancellationError` (as before) but now the error handler will not attempt to send the status or trailers back via the interceptors because the state will be in `.completed`.

Tests for receiving an error after headers and after a message have been added.

### DIFF
--- a/Tests/GRPCTests/AsyncAwaitSupport/AsyncIntegrationTests.swift
+++ b/Tests/GRPCTests/AsyncAwaitSupport/AsyncIntegrationTests.swift
@@ -202,7 +202,7 @@ final class AsyncIntegrationTests: GRPCTestCase {
       try await update.requestStream.send(.with { $0.text = "hello" })
       _ = try await update.responses.first(where: { _ in true })
       XCTAssertNoThrow(try self.server.close().wait())
-      self.server = nil  // So that tearDown() does not call close() again.
+      self.server = nil // So that tearDown() does not call close() again.
       try await update.requestStream.finish()
     }
   }


simonjbeaumont:
@glbrntt what do we want to do about the [`deinit` check here](https://github.com/grpc/grpc-swift/blob/1.4.1-async-await/Sources/GRPC/AsyncAwaitSupport/AsyncWriter.swift#L137):

```swift
// AsyncWriter.swift

  deinit {
    switch self._completionState {
    case .completed:
      ()
    case .incomplete, .pending:
      assertionFailure("writer has not completed is pending completion")
    }
  }
```

This will currently cause an `assertionFailure` if I add the following test to the integration tests, which shuts down the server before completing the RPC.

```diff
diff --git a/Tests/GRPCTests/AsyncAwaitSupport/AsyncIntegrationTests.swift b/Tests/GRPCTests/AsyncAwaitSupport/AsyncIntegrationTests.swift
index bc5f7eb4..298055cb 100644
--- a/Tests/GRPCTests/AsyncAwaitSupport/AsyncIntegrationTests.swift
+++ b/Tests/GRPCTests/AsyncAwaitSupport/AsyncIntegrationTests.swift
@@ -197,2 +197,11 @@ final class AsyncIntegrationTests: GRPCTestCase {
   }
+
+  func testServerCloseAfterMessage() {
+    XCTAsyncTest {
+      let update = self.echo.makeUpdateCall()
+      try await update.requestStream.send(.with { $0.text = "hello" })
+      _ = try await update.responses.first(where: { _ in true })
+      // server is closed in tearDown.
+    }
+  }
 }
```

PR #1306, opened 2021-11-08
Add 'cause' to more transformed statuses

glbrntt:
Motivation:

In #1290 a 'cause' was added to `GRPCStatus` so use have access the error
which resulted in the status. We use the `GRPCStatusTransformable`
protocol to transform errors to statuses, but we missed adding passing
along the cause in a bunch of places.

Modifications:

- Add a cause to the status for a handful of additional error types

Result:

More errors provide a cause when converted to a status


fabianfett (approved):
Love it!

PR #1307, opened 2021-11-09
test: Add tests to verify mTLS behaviour

simonjbeaumont:
This adds some tests to check that configuring for mTLS behaves as expected.

The tests aim to show the following:
* Mutually trusted certs succeed
* Untrusted certs from either side results in failure
* Plaintext peer on either side results in failure

Other changes included in this PR:
* Not all certs in the sample data are actually generated in the `makecert` script (e.g. the `localhost` server cert).
* The `makecert` script and sample data were extended to include some certs signed by a different CA for mTLS failure mode tests.


Tests/GRPCTests/MutualTLSTests.swift
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  func performTestWith(
+    _ serverTLSConfiguration: GRPCTLSConfiguration?,
+    _ clientTLSConfiguration: GRPCTLSConfiguration?,
+    expect expectedOutcome: ExpectedOutcome
+  ) throws {
+    // Setup the server.
+    let serverEventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+    var serverConfiguration = Server.Configuration.default(
```
glbrntt, line 41: Can you add a logger in here too? `GRPCTestCase` provides `self.serverLogger` which will emit all trace logs on test failures.

Tests/GRPCTests/MutualTLSTests.swift
```diff
@@ -0,0 +1,139 @@
...
+      eventLoopGroup: clientEventLoopGroup
+    )
+    clientConfiguration.tlsConfiguration = clientTLSConfiguration
+    clientConfiguration.connectionBackoff = nil
+    clientConfiguration.backgroundActivityLogger = self.clientLogger
+    let clientErrorExpectation = self.expectation(description: "client error")
+    clientErrorExpectation.isInverted = expectedOutcome == .success
+    clientErrorExpectation.assertForOverFulfill = false
+    let clientErrorDelegate = ErrorRecordingDelegate(expectation: clientErrorExpectation)
+    clientConfiguration.errorDelegate = clientErrorDelegate
```
glbrntt, line 69: Rather than using an error delegate, if you switch the the channel pool you will be able to retrieve locally produced error from the status (more new API!):

```swift
let call = client.get(...)
let status = try call.status.wait()
if let cause = status.cause {
  // ...
} else {
  // ...
}
```

Tests/GRPCTests/MutualTLSTests.swift
```diff
@@ -0,0 +1,139 @@
...
+    case clientError
+  }
+
+  func performTestWith(
+    _ serverTLSConfiguration: GRPCTLSConfiguration?,
+    _ clientTLSConfiguration: GRPCTLSConfiguration?,
+    expect expectedOutcome: ExpectedOutcome
+  ) throws {
+    // Setup the server.
+    let serverEventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
```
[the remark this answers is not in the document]
glbrntt: Also you can use just one ELG if you like.

Tests/GRPCTests/MutualTLSTests.swift
```diff
@@ -0,0 +1,139 @@
...
+    serverErrorExpectation.assertForOverFulfill = false
+    let serverErrorDelegate = ServerErrorRecordingDelegate(expectation: serverErrorExpectation)
+    serverConfiguration.errorDelegate = serverErrorDelegate
+
+    let server = try! Server.start(configuration: serverConfiguration).wait()
+    let port = server.channel.localAddress!.port!
+
+    // Setup the client.
+    let clientEventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+    var clientConfiguration = ClientConnection.Configuration.default(
```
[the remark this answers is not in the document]
simonjbeaumont: I'm playing with it now. There was a nice symmetry between the client and the server setups in this test which made it easy to follow using the pool and stopping using the error delegate removes this. I'll play some more with it but it makes the test less clear IMO.

Tests/GRPCTests/MutualTLSTests.swift
simonjbeaumont: So I have some _interesting behaviour_ when I try and use the new pool API including tests blocking and also the error delegate just not working. I'm sure I'm probably holding it wrong.

In the meantime I think the PR is in a good shape o/w (pending what CI has to say, ofc).

glbrntt: When an RPC requests a stream from the pool it will by default wait at most 30 seconds for a stream to become available, I suspect you're running into that rather than blocking indefinitely. FWIW I'm okay with sticking to client connection.

Tests/GRPCTests/MutualTLSTests.swift
```diff
@@ -0,0 +1,156 @@
...
+    )
+    serverConfiguration.tlsConfiguration = serverTLSConfiguration
+    serverConfiguration.logger = self.serverLogger
+    let serverErrorExpectation = self.expectation(description: "server error")
+    serverErrorExpectation.isInverted = expectedOutcome != .serverError
+    serverErrorExpectation.assertForOverFulfill = false
+    let serverErrorDelegate = ServerErrorRecordingDelegate(expectation: serverErrorExpectation)
+    serverConfiguration.errorDelegate = serverErrorDelegate
+
+    let server = try! Server.start(configuration: serverConfiguration).wait()
```
glbrntt, line 71: We still need to shutdown the server

Tests/GRPCTests/MutualTLSTests.swift
```diff
@@ -0,0 +1,156 @@
...
+    let clientTLSConfiguration = GRPCTLSConfiguration.makeClientConfigurationBackedByNIOSSL(
+      certificateChain: [.certificate(SampleCertificate.client.certificate)],
+      privateKey: .privateKey(SamplePrivateKey.client),
+      trustRoots: .certificates([SampleCertificate.ca.certificate]),
+      certificateVerification: .fullVerification
+    )
+    try self.performTestWith(serverTLSConfiguration, clientTLSConfiguration, expect: .success)
+  }
+
+  func test_noServerCert_ClientError() throws {
```
glbrntt, line 137: I'm not sure how useful this test is as an _mTLS test_: it's validates that a client configured with TLS fails to connect to a plaintext server.

Tests/GRPCTests/MutualTLSTests.swift
```diff
@@ -0,0 +1,156 @@
...
+    let clientTLSConfiguration = GRPCTLSConfiguration.makeClientConfigurationBackedByNIOSSL(
+      certificateChain: [.certificate(SampleCertificate.client.certificate)],
+      privateKey: .privateKey(SamplePrivateKey.client),
+      trustRoots: .certificates([SampleCertificate.ca.certificate]),
+      certificateVerification: .fullVerification
+    )
+    try self.performTestWith(nil, clientTLSConfiguration, expect: .clientError)
+  }
+
+  func test_noClientCert_ServerError() throws {
```
glbrntt, line 147: Similar to above: this validates that a plaintext client cannot connect to a server configured with TLS. The more interesting test would be to configure the client to not present any certificates and for the server to expect mTLS.

simonjbeaumont: Oh, my logic here for testing mutual auth was to check that when both sides provide valid certs things work, then when either side doesn't that things don't work.

Specifically this test is I guess testing client-verification by the server or am I missing something.

Is the thing I'm missing that we should be using a TLS client but with no cert (cf. no TLS)?

glbrntt: > Specifically this test is I guess testing client-verification by the server or am I missing something.

Yes but more specifically I think we want to test that the server verifies the certificate presented by the client.

At the moment because the client isn't configured to use TLS, the first bytes it sends will be the http/2 preface which the server will attempt to interpret as the SSL client hello, so the server will fail the handshake at the step rather than because the client doesn't present a valid certificate.

> Is the thing I'm missing that we should be using a TLS client but with no cert (cf. no TLS)?

Yes

simonjbeaumont: OK, I've updated the PR description, and the `makecert` script which had some limitations, and updated the tests to hopefully make more sense.

Tests/GRPCTests/BasicEchoTestCase.swift
```diff
@@ -87,12 +87,21 @@ class EchoTestCaseBase: GRPCTestCase {
...
       ).withTLS(trustRoots: .certificates([SampleCertificate.ca.certificate]))
+
+    case .mutualAuthentication:
+      return Server.usingTLSBackedByNIOSSL(
+        on: self.serverEventLoopGroup,
+        certificateChain: [SampleCertificate.server.certificate],
+        privateKey: SamplePrivateKey.server
+      )
+      .withTLS(trustRoots: .certificates([SampleCertificate.ca.certificate]))
+      .withTLS(certificateVerification: .noHostnameVerification)
```
simonjbeaumont, line 104: Note: This looked to have been latently wrong since AFAIK the default client-verification for server config is `.none`.

Tests/GRPCTests/MutualTLSTests.swift
```diff
@@ -0,0 +1,269 @@
...
+    clientConfiguration.errorDelegate = clientErrorDelegate
+
+    self.channel = ClientConnection(configuration: clientConfiguration)
+    let client = Echo_EchoClient(channel: channel!)
+
+    // Make the call.
+    let call = client.get(.with { $0.text = "mumble" })
+
+    // Wait for side effects.
+    self.wait(for: [clientErrorExpectation, serverErrorExpectation], timeout: 1)
```
glbrntt, line 103: CI can be pretty slow at times, so best to have a fairly large timeout:

```suggestion
    self.wait(for: [clientErrorExpectation, serverErrorExpectation], timeout: 10)
```

glbrntt (approved):
This is great, thanks Si! Just one tiny nit to avoid CI flakiness.

PR #1308, opened 2021-11-10
Allow clients to shutdown gracefully

glbrntt:
Motivation:

If users wish to shutdown their client after all running RPCs have
completed then it is currently up to them to monitor when all of their
running RPCs have completed. This is somewhat awkward and we should
provide a better way to do this.

Modifications:

- Two new requirements on the `GRPCChannel` protocol which have default
  implementations to avoid breaking API:
  - `closeGracefully(deadline:promise:)`, and
  - `close(promise:)` (the future-based version already exists)
- The internal `ConnectionManager` has been updated to shutdown with a
  'mode', either forceful or graceful and accept a promise.
  - The semantics of shutdown in the connecting and active states has
    changed slightly to complete when the `closeFuture` is resolved.
    This required ignoring SSL unclean shutdown errors (which can safely
    be ignored for gRPC).
- The idle handler state machine was also altered to only send GOAWAY
  frames when the client closes and there are no open streams. If there
  are open streams the layers above (i.e. `ConnectionManager` and
  `ChannelPool`) will not request new streams.

Results:

A `GRPCChannel` may be shutdown gracefully, allowing existing RPCs to
run to completion.

### DIFF
--- a/Sources/GRPC/ClientConnection.swift
+++ b/Sources/GRPC/ClientConnection.swift
@@ -125,9 +125,32 @@ public class ClientConnection {
     )
   }
 
-  /// Closes the connection to the server.
+  /// Close the channel, and any connections associated with it. Any ongoing RPCs may fail.
+  ///
+  /// - Returns: Returns a future which will be resolved when shutdown has completed.
   public func close() -> EventLoopFuture<Void> {
-    return self.connectionManager.shutdown()
+    let promise = self.eventLoop.makePromise(of: Void.self)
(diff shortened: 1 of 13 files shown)


Lukasa (approved):
This is a really nice patch @glbrntt: the API surface makes sense and the implementation is straightforward and clear, along with good test coverage. 10/10 good work.

fabianfett (approved):
Thanks! This is awesome!

PR #1312, opened 2021-11-12
Extend interoperability tests to use async provider

simonjbeaumont:
The interoperability tests exercise a well-defined sequence of requests
for all implementations of gRPC. We already had interop tests for the
ELG-based client and provider. This pull request extends the tests to
exercise an async-await–based provider.

### DIFF
--- a/Sources/GRPCInteroperabilityTestsImplementation/TestServiceAsyncProvider.swift
+++ b/Sources/GRPCInteroperabilityTestsImplementation/TestServiceAsyncProvider.swift
@@ -17,7 +17,7 @@
 import Foundation
 import GRPC
 import GRPCInteroperabilityTestModels
-import NIO
+import NIOCore
 
 /// An async service provider for the gRPC interoperability test suite.
 ///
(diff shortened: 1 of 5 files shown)


Sources/GRPCInteroperabilityTestsImplementation/TestServiceAsyncProvider.swift
```diff
@@ -0,0 +1,220 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#if compiler(>=5.5)
+import Foundation
+import GRPC
+import GRPCInteroperabilityTestModels
+import NIO
```
glbrntt, line 20: nit: use `NIOCore` (`NIO` is an umbrella import for `NIOCore`, `NIOPosix` and `NIOEmbedded`).

glbrntt (approved):
One nit but looks great otherwise, thanks Si!

glbrntt (approved):

Issue #1183, opened 2021-05-17
Get the serviceName in Service discovery


mengxiangyue:
### What are you trying to achieve?
When I am creating the client, I want to use service discovery to get the service name. But I need to create `Client` then I can get the service name, if I want to create `Client` I need to use host/port to create a channel, at here I need to use service name to get host. it become a cycle. could you give a suggestion?

```
// Configure the channel, we're not using TLS so the connection is `insecure`.
let channel = ClientConnection.insecure(group: group)
  .connect(host: "localhost", port: self.port) // I need to use service name to get the host/port

// Provide the connection to the generated client.
let greeter = Helloworld_GreeterClient(channel: channel)
```

glbrntt:
I suspect the correct way of resolving this is to generate some `static` metadata for each service and integrate with [swift-service-discovery](https://github.com/apple/swift-service-discovery). However, that'll take a little bit of time.

---

As a workaround for the meantime you can create a no-op client to get the service name by creating something which conforms to `GRPCChannel`: 

```swift
struct NoOpChannel: GRPCChannel {
  func makeCall<Request: SwiftProtobuf.Message, Response: SwiftProtobuf.Message>(
    path: String,
    type: GRPCCallType,
    callOptions: CallOptions,
    interceptors: [ClientInterceptor<Request, Response>]
  ) -> Call<Request, Response> { 
    fatalError()
  }

  func makeCall<Request: GRPCPayload, Response: GRPCPayload>(
    path: String,
    type: GRPCCallType,
    callOptions: CallOptions,
    interceptors: [ClientInterceptor<Request, Response>]
  ) -> Call<Request, Response> {
    fatalError()
  }

  func close() -> EventLoopFuture<Void> { 
    fatalError()
  }
}
```

Then you can extract the service name for service discovery:

```swift
let noOpGreeter = Helloworld_GreeterClient(channel: NoOpChannel())
let serviceName = noOpGreeter.serviceName
```

I appreciate that it's not the nicest solution but it should at least unblock you!

mengxiangyue:
@glbrntt thanks your response so fast, what do you think if we change the `serviceName` into static property?

glbrntt:
> @glbrntt thanks your response so fast, what do you think if we change the `serviceName` into static property?

Unfortunately we can't just make it `static` since we'd effectively be breaking the API. 

We'd most likely want to generate some static service metadata alongside the protocol and then add a static property with a default implementation to the generated protocol, something like:

[...]

PR #1322, opened 2021-12-08
Generate static service metadata

glbrntt:
Motivation:

Sometimes it's useful to know information about a service, including its
name, methods it offers and so on. An example where this is useful is
service discovery (#1183). However, we currently only provide the
service name and this isn't available statically. For service discovery
this is problematic as it requires users to create a client in order to
get the service name so that a server can be dialled.

For the async/await code, since it is not yet final, we can add
requirements to generated protocols to provide the service metadata
statically.

Modifications:

- Add service and method descriptors to `GRPC`
- Generate service descriptors for client and server separately (it's feasible
  that a user may generate client code into one module and server code
  into separate modules)
- Update the generated code for async/await and NIO based APIs to use
  the descriptors directly rather than generating literals in places
  where they are required.
- Add test for the generated echo service metadata
- Regenerate other services

Result:

Adopters can get static information about services.

### DIFF
--- a/Sources/Examples/Echo/Model/echo.grpc.swift
+++ b/Sources/Examples/Echo/Model/echo.grpc.swift
@@ -67,7 +67,7 @@ extension Echo_EchoClientProtocol {
     callOptions: CallOptions? = nil
   ) -> UnaryCall<Echo_EchoRequest, Echo_EchoResponse> {
     return self.makeUnaryCall(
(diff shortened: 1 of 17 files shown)


glbrntt:
Actual changes are in d47446d (along with regenerated Echo service to illustrate changes)

fabianfett (approved):
Your PRs are a pleasure.

PR #1334, opened 2021-12-17
Add Sendable conformance to various types

glbrntt:
Motivation:

Sendable conformance is used to indicate that a type is thread-safe.
Sendable checks are limited in 5.5 but will tighten in future Swift
versions.

Modifications:

- Add `Sendable` conformance to a number of the new types added just for
  async/await
- Add a `GRPCSendable` typealias to allow `Sendable` to be applied when
  it is available. This is used on the relevant types which existed
  before async/await support was added.

Result:

Wider Sendable conformance.


Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -153,13 +158,10 @@ extension GRPCAsyncServerHandler {
...
-  Deserializer: MessageDeserializer
->: GRPCServerHandlerProtocol {
-  @usableFromInline
-  internal typealias Request = Deserializer.Output
-  @usableFromInline
-  internal typealias Response = Serializer.Input
-
+  Deserializer: MessageDeserializer,
+  Request: Sendable,
+  Response: Sendable
```
fabianfett, line 163: Why do we have an explicit Request and Response here? Cant we say 
```swift
where Deserializer.Output == Sendable, Serializer.Input == Sendable
```

fabianfett (commented):
Except for one question this looks good to my eyes. However it might be good to get a second review on this as well.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
glbrntt: Good idea, we can do that.

glbrntt:
> Except for one question this looks good to my eyes. However it might be good to get a second review on this as well.

Okay, we can wait for @Lukasa to have a look -- no rush with this one tbh.

glbrntt:
Sadly the compiler is tripping over an assertion too: https://bugs.swift.org/browse/SR-15628

Lukasa (approved):
LGTM.

PR #1336, opened 2021-12-20
Use new XCTest async/await support

glbrntt:
Motivation:

Swift 5.5.2 includes XCTest support for async/await. Currently we are
working around the lack of support with XCTest expectations.

Modifications:

- Remove the helper
- Make async tests `async throws`

Result:

Fewer workarounds.


fabianfett (approved):
Love it! ❤️

PR #1349, opened 2022-01-27
Tolerate channel error when connecting

glbrntt:
Motivation:

In some cases it's possible for the connection manager to pick an error
from the channel when in the connecting state.

For some reason we decided this was not a valid transition despite
another function explicitly allowing an error to be handled while
in this state. The difference is that the other path is triggered by the
'connect()' future rather than coming from the channel directly.

Modifications:

- Call the function for handling failed connection attempts when we
  receive an error from the channel in the connecting state.

Result:

Better tolerate connection errors.

### DIFF
--- a/Sources/GRPC/ConnectionManager.swift
+++ b/Sources/GRPC/ConnectionManager.swift
@@ -591,7 +591,7 @@ internal final class ConnectionManager {
       ])
 
     case .connecting:
-      self.invalidState()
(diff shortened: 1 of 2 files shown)


fabianfett (approved):
Looks good. However, are you sure, we don't want to add a test for this?

fabianfett (approved):

PR #1356, opened 2022-02-09
Add cause to GRPCStatus.processingError for uncaught errors

simonjbeaumont:
#1290 and #1306 added a `cause` field to `GRPCStatus` to help understand the underlying error.

This patch adds the `cause` to `GRPCStatus.processingError` which is the status returned when there is an uncaught error when handling requests.


glbrntt (approved):
thanks Si!

PR #1357, opened 2022-02-09
Add cause to GRPCStatus.descrpition if it is non-nil

simonjbeaumont:
Following on from #1290 and #1356, this adds the `cause` to the `GRPCStatus.description` so that it renders usefully in log lines by default.


simonjbeaumont:
//cc @pahfat

glbrntt (approved):

PR #1361, opened 2022-02-12
protoc-gen-grpc-swift: FileHandle for stdin, stdout

lrewega:
This fixes a case where the output response is silently truncated.
Truncation occurs because the `write` syscall is only invoked a single
time, but `write` may return a value indicating a partial write for
various reasons, even in a blocking call. For example, if stdout is a
nonblocking pipe, output is truncated when the pipe is full.

Instead of rewriting Stdout.write to handle this case, let's migrate to
FileHandle for I/O, available as of Swift 3.2. swift-protobuf has
already migrated to FileHandle, see also
github.com/apple/swift-protobuf/commit/2ccad3484e06d180af3bd7be53c96d41e8f03a62


glbrntt (approved):
Nice fix; thank you!

PR #1363, opened 2022-02-15
Release stream callback, once the stream has finished

fabianfett:
### Motivation

Currently we don't release the stream callback once we a stream has ended. This makes it really easy for adopters to run into retain cycles, they don't expect.

### Changes

- Release callback once the grpc stream has ended.

### DIFF
--- a/Sources/GRPC/ClientCalls/ResponseContainers.swift
+++ b/Sources/GRPC/ClientCalls/ResponseContainers.swift
@@ -98,7 +98,7 @@ internal class StreamingResponseParts<Response> {
   private let eventLoop: EventLoop
 
   /// A callback for response messages.
-  private let responseCallback: (Response) -> Void
+  private var responseCallback: ((Response) -> Void)?
 
   /// Lazy promises for the status, initial-, and trailing-metadata.
   private var initialMetadataPromise: LazyEventLoopPromise<HPACKHeaders>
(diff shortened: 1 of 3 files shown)


Sources/GRPC/ClientCalls/ResponseContainers.swift
```diff
@@ -118,17 +118,19 @@ internal class StreamingResponseParts<Response> {
...
+    return self.statusPromise.futureResult
   }
 
   internal init(on eventLoop: EventLoop, _ responseCallback: @escaping (Response) -> Void) {
     self.eventLoop = eventLoop
     self.responseCallback = responseCallback
     self.initialMetadataPromise = eventLoop.makeLazyPromise()
     self.trailingMetadataPromise = eventLoop.makeLazyPromise()
-    self.statusPromise = eventLoop.makeLazyPromise()
+    self.statusPromise = eventLoop.makePromise()
```
glbrntt, line 129: Can we keep this as a lazy promise and just replace the callback in `handle` and `handleError`?

fabianfett: much better! will change!

Sources/GRPC/ClientCalls/ResponseContainers.swift
```diff
@@ -98,7 +98,7 @@ internal class StreamingResponseParts<Response> {
   private let eventLoop: EventLoop
 
   /// A callback for response messages.
-  private let responseCallback: (Response) -> Void
+  private var responseCallback: ((Response) -> Void)?
```
glbrntt, line 101:
```suggestion
  private var responseCallback: Optional<((Response) -> Void)>
```

glbrntt (approved):

PR #1374, opened 2022-03-24
Ignore state machine inputs in unexpected states

glbrntt:
Motivation:

When we ratchet down the last stream ID in a GOAWAY frame, we should
only do it if we're quiescing. If we aren't quiescing then we should
just ignore the request to do it.

Modifications:

- Ignore requests to ratchet down the last stream ID in a GOAWAY if
  we're not quiescing.

Result:

Fewer traps.


Lukasa (approved):
Nice catch

PR #1380, opened 2022-03-30
ConnectionPool Waiter should store its timeout task

glbrntt:
Motivation:

Connection pool waiters have a scheduled tasks for to timeout waiting.
This is cancelled and set to `nil` when the waiter is succeeded or
failed. Hoever, when scheduling a task the task was never stored! This
would lead to the potentially firing when the deadline expired although
the effect of this would be a no-op.

Modifications:

- Capture the waiters scheduled task
- Add a test

Result:

Cancelled tasks are actually cancelled.


FranzBusch (approved):
LGTM!

PR #1381, opened 2022-03-30
Update formatter

glbrntt:
Motivation:

We need to use a newer version of SwiftFormat on the async/await code.
First we should update main so that merging changes into the async
branch is cleaner.

Modifications:

- Update SwiftFormat
- Reformat

Result:

Our version of SwiftFormat is new enough to correctly handle async/await
code.


FranzBusch (approved):
Nice!

PR #1384, opened 2022-04-01
Always generate async code

glbrntt:
Motivation:
    
Async code generation currently requires opting in. We should make it
the default alongside the NIO-based code so that it's easier to find.
    
Modifications:
    
- Remove the async options from the generator.
- Update the generator to always emit async code behind compiler guards.
- Update plugin docs.
- Regenerate code.
    
Result:
    
Async code is easier to find.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ ${XCODEPROJ}:
 	protoc $< \
 		--proto_path=$(dir $<) \
 		--plugin=${PROTOC_GEN_GRPC_SWIFT} \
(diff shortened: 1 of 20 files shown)


glbrntt:
Note:
- [eac9861](https://github.com/grpc/grpc-swift/pull/1384/commits/eac9861a3b6e57d8b6f446c536b8d32e7daa5aa2) contains the changes to be reviewed
- [879a554](https://github.com/grpc/grpc-swift/pull/1384/commits/879a5549b29d24dd76e33e5d999a2a07181fb6f7) regenerates the code

Lukasa (approved):

Issue #1353, opened 2022-01-28
Is there a way to initialise a client using a file descriptor or an XPC connection?


jvimal-eg:
### What are you trying to achieve?

There are 3 ways to start a grpc client today: host and port, socket address, and unix domain socket path. Is there a way to start one using a file descriptor?

My use case is that I have a server and client running in two different sandboxes.  For some reason, I am unable to share a unix domain socket across the two, and I've resorted to using a TCP socket, which is insecure as any process can connect to it.  

Ideally, it would be great if grpc client can work over an XPC connection, which offers stronger authentication between client/server using codesign certificates.  Alternatively, I could have the server pass a file descriptor to an authenticated client over XPC and use that file descriptor to start the client.

### What have you tried so far?

I tried looking at the source code to see how else I can create connections.

Lukasa:
grpc-over-xpc is unlikely to be a good design pattern: XPC is not really built for bulk data transfer in this way. Passing an FD over XPC and using that would be better, and grpc could absolutely support that as a design pattern, but I'll defer to @glbrntt as to whether that exists today.

jvimal-eg:
@Lukasa that's great, thank you. I remember reading in Apple dev forums that XPC is pretty fast, even faster than unix domain sockets: https://developer.apple.com/forums/thread/74498. I have't benchmarked it myself, so I'll defer it to the experts.

@glbrntt is there a way to use the grpc-swift library to achieve it today without an additional implementation?

glbrntt:
There's no API in grpc to do this at the moment, but NIO provides a [`withConnectedSocket(_:)`](https://github.com/apple/swift-nio/blob/4f2c6a3e0bf5fd2a855d5821b6b397508e46da86/Sources/NIOPosix/Bootstrap.swift#L662-L694) which we should be able to just wire through the grpc API.

We'd need:
- A new entry on [`ConnectionTarget`](https://github.com/grpc/grpc-swift/blob/f79ef102add7a6a920e559a93a9fb309006f794c/Sources/GRPC/ClientConnection.swift#L264)
- New API on the `ClientConnection` builder [here](https://github.com/grpc/grpc-swift/blob/f79ef102add7a6a920e559a93a9fb309006f794c/Sources/GRPC/GRPCChannel/GRPCChannelBuilder.swift#L67) (the channel pool uses `ConnectionTarget` directly so does not need new API)
[...]

PR #1385, opened 2022-04-03
allow client/server to be initialised with a connected socket

jvimal-eg:
Motivation: https://github.com/grpc/grpc-swift/issues/1353

Modifications: As outlined by @glbrntt:
- A new entry on ConnectionTarget - New API on the ClientConnection builder
here (the channel pool uses ConnectionTarget directly so does not need new API)
- Some validation that we're not trying to use this API with a
NIOTSEventLoopGroup or NIOTSEventLoop (as far as I can tell, using a file
descriptor directly is not possible with Network.framework)
- Tests

Result:
This allows greater flexibility in spawning the client/server; in particular, it allows unix domain sockets for sandboxed apps on macOS.

### DIFF
--- a/Sources/GRPC/PlatformSupport.swift
+++ b/Sources/GRPC/PlatformSupport.swift
@@ -124,6 +124,12 @@ public protocol ClientBootstrapProtocol {
   func channelInitializer(_ handler: @escaping (Channel) -> EventLoopFuture<Void>) -> Self
 }
 
+extension ClientBootstrapProtocol {
+  public func withConnectedSocket(_ socket: NIOBSDSocket.Handle) -> EventLoopFuture<Channel> {
+    preconditionFailure("withConnectedSocket(_:) is not implemented")
+  }
+}
+
 extension ClientBootstrap: ClientBootstrapProtocol {}
 
 #if canImport(Network)
@@ -148,6 +154,12 @@ public protocol ServerBootstrapProtocol {
 
   func childChannelInitializer(_ initializer: @escaping (Channel) -> EventLoopFuture<Void>) -> Self
   func childChannelOption<T>(_ option: T, value: T.Value) -> Self where T: ChannelOption
+}
+
+extension ServerBootstrapProtocol {
+  public func withBoundSocket(_ connectedSocket: NIOBSDSocket.Handle) -> EventLoopFuture<Channel> {
+    preconditionFailure("withBoundSocket(_:) is not implemented")
+  }
 }
 
 extension ServerBootstrap: ServerBootstrapProtocol {}
--- a/Sources/GRPC/PlatformSupport.swift
+++ b/Sources/GRPC/PlatformSupport.swift
@@ -117,6 +117,7 @@ public protocol ClientBootstrapProtocol {
   func connect(to: SocketAddress) -> EventLoopFuture<Channel>
   func connect(host: String, port: Int) -> EventLoopFuture<Channel>
   func connect(unixDomainSocketPath: String) -> EventLoopFuture<Channel>
+  func withConnectedSocket(_ socket: NIOBSDSocket.Handle) -> EventLoopFuture<Channel>
 
   func connectTimeout(_ timeout: TimeAmount) -> Self
   func channelOption<T>(_ option: T, value: T.Value) -> Self where T: ChannelOption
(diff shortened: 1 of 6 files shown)


glbrntt:
re: https://github.com/grpc/grpc-swift/issues/1353#issuecomment-1086899166

I would just create a new test class, something like https://github.com/grpc/grpc-swift/blob/main/Tests/GRPCTests/GRPCKeepaliveTests.swift should be a useful starting point.

Sources/GRPC/PlatformSupport.swift
```diff
@@ -127,7 +128,11 @@ extension ClientBootstrap: ClientBootstrapProtocol {}
 
 #if canImport(Network)
 @available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
-extension NIOTSConnectionBootstrap: ClientBootstrapProtocol {}
+extension NIOTSConnectionBootstrap: ClientBootstrapProtocol {
```
glbrntt, line 131: We need to provide a default implementation on `ClientBootstrapProtocol` and `ServerBootstrapProtocol` as well (otherwise this would be a breaking change). 

We could also just remove the implementation on the `NIOTS*Bootstrap`s and defer to the default ones.

glbrntt:
@jvimal-eg you'll need to run `./scripts/format.sh` to fixup the formatting.

jvimal-eg:
@glbrntt Sounds good.  I am new to Swift / NIO, and I am stumped at how to create a `NIOBSDSocket.Handle` that's already "connected" to use with a client's withConnectedSocket().  Any pointers on how to do it?  `NIOBSDSocket.Handle` on mac/Linux appears to be just a file descriptor but I am a bit hesitant to use low-level APIs just to create a UDS.

Lukasa:
It's a file descriptor on all currently supported platforms, so you can safely use the regular libc APIs to create the Unix domain socket.

jvimal-eg:
@Lukasa thanks, done!

glbrntt (changes requested):

jvimal-eg:
@glbrntt thanks for your review.  Apologies -- I blindly copypasta the old test without really paying attention to what it was doing. I fixed the test now.

Tests/GRPCTests/WithConnectedSocketTests.swift
```diff
@@ -0,0 +1,64 @@
...
+    let server = try Server.insecure(group: group)
+      .withServiceProviders([EchoProvider()])
+      .withLogger(self.serverLogger)
+      .bind(unixDomainSocketPath: path)
+      .wait()
+    defer {
+      XCTAssertNoThrow(try server.close().wait())
+    }
+
+    let clientSocket = socket(AF_UNIX, SOCK_STREAM, 0)
```
glbrntt, line 41: I think you need something like:

```swift
#if os(Linux)
let sockStream = CInt(SOCK_STREAM.rawValue)
#else
let sockStream = SOCK_STREAM
#endif
let clientSocket = socket(AF_UNIX, sockStream, 0)
```

Tests/GRPCTests/WithConnectedSocketTests.swift
```diff
@@ -0,0 +1,64 @@
...
+      let ret = connect(clientSocket, addr, UInt32(size))
+      XCTAssert(ret != -1)
+    }
+    let flags = fcntl(clientSocket, F_GETFL, 0)
+    XCTAssert(flags != -1)
+    XCTAssert(fcntl(clientSocket, F_SETFL, flags | O_NONBLOCK) == 0);
+
+    let connection = ClientConnection.insecure(group: group)
+      .withBackgroundActivityLogger(self.clientLogger)
+      .withKeepalive(.init(interval: .seconds(1), timeout: .milliseconds(100)))
```
glbrntt, line 54: Don't need this one either
```suggestion
```

glbrntt:
> @glbrntt thanks for your review. Apologies -- I blindly copypasta the old test without really paying attention to what it was doing. I fixed the test now.

No worries!

You need to rebase your branch on main and reformat, see: https://github.com/grpc/grpc-swift/pull/1385#issuecomment-1087298846

jvimal-eg:
@glbrntt Thanks.  Rebased, reformatted, and addressed your comments.

glbrntt (approved):
Great, thank you!

PR #1386, opened 2022-04-05
Make clients sendable

glbrntt:
Motivation:

gRPC clients should be `Sendable`.

Modifications:

This commit contains a number of changes to enable clients to be
'Sendable'. The are:

- Require `GRPCChannel` to be 'Sendable' since the underlying API should
  be thread-safe.
- Require `GRPCClient` to be 'Sendable' as this just provides wrappers
  around a `GRPCChannel` which is also thread-safe. This propagates to
  generated client protocols which refine `GRPCClient`.
- Generated clients using the NIO API are classes (!) holding mutable
  call options and interceptor factories. Codegen was updated so make
  these `@unchecked Sendable` by protecting mutable state with a lock.
  Because this generates additional allocations and should never have
  been a class it is now marked deprecated in favor of a struct based
  client.
- `AnyServiceClient` is also a class and has been deprecated in favor of
  the struct based `GRPCAnyServiceClient`.
- A separate commit fixes these deprecation warnings.
- The test clients have also been marked as deprecated: they rely on an
  `EmbeddedChannel` which is not and will never be `Sendable` (without
  also being deprecated). The test clients are also marked
  as `@unchecked Sendable` to satisfy the requirement on `GRPCClient`
  despite being deprecated. The same is true for the `FakeChannel` (a
  `GRPCChannel`). We reccomend using a client and server on localhost as
  a replacement.
- Various supporting value types are marked as `Sendable`.

Result:

- Clients are now `Sendable`.
- Clients are now always `struct`s.
- Generated test clients are deprecated.


glbrntt:
Note to reviewers:

- The bulk of the changes are in: [e0ff686](https://github.com/grpc/grpc-swift/pull/1386/commits/e0ff68615998baf0df157746300c8a1682a0be91)
- Generated clients are updated in: [fb767ad](https://github.com/grpc/grpc-swift/pull/1386/commits/fb767ad5b72eb3e8b6197774e1fdac58d4053d38)
- Deprecation warnings are fixed in: [c6599ea](https://github.com/grpc/grpc-swift/pull/1386/commits/c6599ea586a86e2f431fa66309d1cebdd3f8e9d5)

Sources/GRPC/GRPCChannel/EmbeddedGRPCChannel.swift
```diff
@@ -1,113 +0,0 @@
-/*
```
glbrntt, line 1: This file was only used in tests, its new home is in `Tests/GRPCTests/ClientTimeoutTests.swift`.

Sources/GRPC/CallOptions.swift
```diff
@@ -160,14 +174,22 @@ extension CallOptions {
     }
 
     /// Provide a factory to generate request IDs.
+    #if swift(>=5.6)
+    public static func generated(
+      _ requestIDFactory: @escaping @Sendable () -> String
```
Lukasa, line 179: Should we promote the `StringFactory` typealias to public here to avoid the extra `#if`?

Sources/GRPC/Interceptor/ClientInterceptors.swift
```diff
@@ -97,3 +97,10 @@ open class ClientInterceptor<Request, Response> {
     context.cancel(promise: promise)
   }
 }
+
+#if compiler(>=5.6)
+// Interceptors sendability is unchecked: they have their own documented rules for thread safety
+// which we can't check. We require them to be sendable in order for interceptor factories to be
+// sendable and in turn, clients.
+extension ClientInterceptor: @unchecked Sendable {}
```
Lukasa, line 105: Does this need to be `preconcurrency`, given that the class already existed and is `open` for subclassing?

glbrntt: Ah, yes it sure does!

glbrntt:
Also not sure if `...NIOClient` is the right new spelling for the future/callback based API (the client with the `async` based API is `...AsyncClient`). Happy to take suggestions.

Sources/GRPC/Interceptor/ClientInterceptors.swift
Lukasa: I think it'd have to be both `@preconcurrency` _and_ `@unchecked`. But I think it's API breaking if we don't add the `@preconcurrency` annotation, because any adopters who are not concurrency aware but build in 5.6 mode will break.

Lukasa:
I think `NIOClient` is totally fine there.

Sources/GRPC/Interceptor/ClientInterceptors.swift
glbrntt: I think they inherit the `@unchecked` (none of the subclasses within `GRPCTests` complained about not being sendable, some of which have mutable state)?

Lukasa: I don't think `@unchecked` should be inherited, so far as I know.

Lukasa: However, I can't find any immediate citation for that belief.

Lukasa: Hmmm, [this](https://forums.swift.org/t/improving-sendable/52377/21) seems relevant.

glbrntt: As does the final post https://forums.swift.org/t/improving-sendable/52377/26:

> There are a number of unresolved issues with Sendable that could be tackled separately. Here's the list I know of:
> ...
> - Sendable with class types needs to be revisited. We likely want to require subclasses to restate Sendable conformance, decide if there are any cases where that conformance need not be @unchecked (e.g., all state is immutable and Sendable) and whether it can ever be inferred (same conditions as for removing @unchecked?).

I'm not sure exactly where that leaves us but I think `@preconcurrency` and `@unchecked` is probably the safest bet here.

Lukasa (approved):

Issue #1337, opened 2021-12-20
Example code to enable compression and decompression on client side ?


uchiha-itachi95:
### What are you trying to achieve?

I want to enable compression and decompression on the client side call. But I am not able to find any documentation regarding this.

### What have you tried so far?

I have looked into issue #611 and #707 but could not find any sample code to achieve the same. How do I do that?

glbrntt:
You can configure compression via the [`messageEncoding`](https://github.com/grpc/grpc-swift/blob/5cc08ebf77e68598bfe4c793927d0f2838598607/Sources/GRPC/CallOptions.swift#L43) property on `CallOptions`.

As an example:

```swift
// Configure encoding
let encodingConfiguration = ClientMessageEncoding.Configuration(
  forRequests: .gzip, // use gzip for requests
  acceptableForResponses: .all, // accept all supported algorithms for responses
  decompressionLimit: .ratio(20) // reject messages and fail the RPC if a response decompresses to over 20x its compressed size
)

// Enable compression with configuration
let options = CallOptions(messageEncoding: .enabled(encodingConfiguration))

// Use the options to make a request
let rpc = echo.get(request, callOptions: options)
// ...
```

PR #1388, opened 2022-04-12
Add very basic documentation on how to enable compression

Lutzifer:
As I did not find any documentation on how to enable compression, I think it is a good idea to make the answer given by @glbrntt in https://github.com/grpc/grpc-swift/issues/1337#issuecomment-997917294 part of the official documentation.


glbrntt (commented):
Thanks @Lutzifer -- would you mind adding this to a section in https://github.com/grpc/grpc-swift/blob/main/docs/faqs.md instead of its own doc?

glbrntt (approved):
Great, thank you!

PR #1394, opened 2022-04-28
Server handler interceptor state machine

glbrntt:
Motivation:

The existing async server handler has a state machine which deals with
the interceptor pipeline and the lifecycle of user handler. These are
really two separate entities each with its own state machine.

This PR is the first in a series which will eventually update the async
server handler to make use of the two state machines.

Modifications:

- Add a 'ServerInterceptorStateMachine' and tests. This acts as a filter
  for inbound and outbound request and response messages, both as they
  enter and exit the interceptor pipeline. Message parts delievered in the
  wrong order lead to the RPC being cancelled, parts delievered after
  the RPC has finished lead to them being dropped. Cancelling the RPC
  informs the caller to nil out the associated interceptor pipeline.
- Tests
- This is not used anywhere (yet).

Result:

Interceptor state machine is in place.

### DIFF
--- a/Sources/GRPC/AsyncAwaitSupport/AsyncServerHandler/ServerInterceptorStateMachine/ServerInterceptorStateMachine+Actions.swift
+++ b/Sources/GRPC/AsyncAwaitSupport/AsyncServerHandler/ServerInterceptorStateMachine/ServerInterceptorStateMachine+Actions.swift
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2022, gRPC Authors All rights reserved.
+ *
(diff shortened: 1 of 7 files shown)


Lukasa (approved):
Nice one, LGTM

FranzBusch (commented):
Nice work! A more general question, does the separation of the state machines bring us any closer to having Interceptors on a Channel level or maybe does it make it harder?

glbrntt:
> Nice work! A more general question, does the separation of the state machines bring us any closer to having Interceptors on a Channel level or maybe does it make it harder?

No, it has no bearing on that change.

PR #1395, opened 2022-04-29
Update Github Actions version to v3

Gumichocopengin8:
Update `actions/checkout@v2` to `actions/checkout@v3` because of Node12 will be out of life after Aril 30, 2022 [[Reference](https://nodejs.org/en/about/releases/)].  
`actions/xxxx@v3` use Node16 whose support lasts until April 30, 2024.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,7 +11,7 @@ jobs:
     container:
       image: swift
     steps:
-    - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


Gumichocopengin8:
@glbrntt Could you please review this PR? Thanks.

glbrntt (approved):
Awesome, thanks for updating this @Gumichocopengin8!

PR #1396, opened 2022-05-03
Server handler state machine

glbrntt:
Motivation:

See #1394.

Modifications:

- Add a 'ServerHandlerStateMachine' and tests.
- This is not used anywhere (yet).

Result:

Handler state machine is in place.

### DIFF
--- a/Sources/GRPC/AsyncAwaitSupport/AsyncServerHandler/ServerHandlerStateMachine/ServerHandlerStateMachine+Actions.swift
+++ b/Sources/GRPC/AsyncAwaitSupport/AsyncServerHandler/ServerHandlerStateMachine/ServerHandlerStateMachine+Actions.swift
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2022, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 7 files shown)


Sources/GRPC/AsyncAwaitSupport/AsyncServerHandler/ServerHandlerStateMachine/ServerHandlerStateMachine+Actions.swift
```diff
@@ -0,0 +1,60 @@
...
+
+  enum SendMessageAction: Equatable {
+    /// Intercept the message, but first intercept the headers if they are non-nil. Must go via
+    /// the interceptor state machine first.
+    case intercept(headers: HPACKHeaders?)
+    /// Drop the message.
+    case drop
+  }
+
+  enum SendStatusAction: Equatable {
```
Lukasa, line 46: Are these not `Hashable` for a reason?

Lukasa (approved):
Basically good, one question.

Sources/GRPC/AsyncAwaitSupport/AsyncServerHandler/ServerHandlerStateMachine/ServerHandlerStateMachine+Actions.swift
glbrntt: Only because `HPACKHeaders` is not `Hashable`.

We don't _really_ need `Equatable` or `Hashable` here, but it's useful for testing.

PR #1397, opened 2022-05-04
Raise minimum supported Swift version to 5.4

glbrntt:
Motivation:

SwiftNIO raised its minimum supported Swift version to 5.4 in the 2.40.0
release. To be able to use the latest updates, we must do the same.

Modifications:

- Bump swift-tools-version to 5.4 in the various package manifests
  (including QPS and FuzzTesting)
- Add excludes for non-source files (READMEs and protos)
- Use 'executableTarget' instead of 'target' for executables
- Update the README to reflect supported versions
- Remove 5.2 and 5.3 from CI
- Remove LinuxMain.swift as test discovery is enabled by default

Result:

Minimum supported Swift version is 5.4


Lukasa (approved):
Good stuff.

PR #1399, opened 2022-05-04
Make all tests discoverable on Linux

### LINKED ISSUE #1392: Several test cases do not run on Linux
grpc-swift uses test discovery on Linux to find the CI tests. This mostly works well, but unlike on Darwin (where tests are discovered by querying the Objective-C runtime), on Linux the tests are discovered only by source code analysis. This analysis does not work with subclasses. As a result, any testing strategy that relies on subclasses discovering tests declared in parent classes fails to run on Linux. At a bare minimum, the following do not run on Linux:

- `FunctionalTestsInsecureTransport` subclasses `FunctionalTestsAnonymousClient` and `FunctionalTestsMutualAuthentication` do not run.
- `GRPCInsecureInteroperabilityTests` subclass `GRPCSecureInteroperabilityTests` does not run.
- The `ServerThrowingTests` subclasses `ServerDelayedThrowingTests`, `ClientThrowingWhenServerReturningErrorTests`, `ServerErrorTransformingTests` do not run.

All of these need to be rewritten to expose an entry point for the test in all subclasses.

glbrntt:
Motivation:

Test discovery on Linux relies on source code analysis. Tests in
subclasses which do not explicitly redeclare each test are missed. As a
result, a number of tests are not run on Linux.

Modifications:

- Explicitly redeclare tests in subclasses.

Result:

- Resolves #1392

### DIFF
--- a/Tests/GRPCTests/FunctionalTests.swift
+++ b/Tests/GRPCTests/FunctionalTests.swift
@@ -259,11 +259,107 @@ class FunctionalTestsAnonymousClient: FunctionalTestsInsecureTransport {
   override var transportSecurity: TransportSecurity {
     return .anonymousClient
   }
+
+  override func testUnary() throws {
+    try super.testUnary()
+  }
+
+  override func testUnaryLotsOfRequests() throws {
+    try super.testUnaryLotsOfRequests()
+  }
+
+  override func testUnaryWithLargeData() throws {
+    try super.testUnaryWithLargeData()
+  }
+
+  override func testUnaryEmptyRequest() throws {
+    try super.testUnaryEmptyRequest()
+  }
+
+  override func testClientStreaming() {
+    super.testClientStreaming()
+  }
+
+  override func testClientStreamingLotsOfMessages() throws {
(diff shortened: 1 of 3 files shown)


glbrntt:
I diffed the output of `swift test --list-tests` on macOS and linux to determine which tests were missing.

simonjbeaumont:
OOI, what's going to stop this from diverging when a new test is added to the base class? Do you think it's worth adding a CI check for the diff you performed manually for this PR?

Lukasa:
I'm +1 with @simonjbeaumont: I'd love to see a tool written for this that we can reuse on our other repos.

glbrntt:
Oh I totally agree, a tool would be great and that's what I looked at doing first:

> I was hoping we might be able to use `swift package dump-package` to analyse functions in test modules and diff them with the list of tests (so we could check if any tests are missing as part of CI) but `dump-package` does not include test targets.

If you have any other ideas I'm all ears!

Lukasa (approved):

Issue #1393, opened 2022-04-28
Google_Cloud_Speech_V1_SpeechAdaptation is not found


csdf-ssm:
eg:SpeechToText-gRPC-iOS ,when i get make protos, make generate,and run, it tips  Cannot find type 'Google_Cloud_Speech_V1_SpeechAdaptation' in scope

glbrntt:
Did each of the `make` steps succeed?

GoluKumar1:
Still facing this issue when try to run the speech example also Generated file are not available. Only the references are shown.

PR #1402, opened 2022-05-10
Migrate speech to text example app to swift package manager & fix compile errors

Jake-Prickett:
Motivation:

After reading that [SwiftNIO was dropping support for CocoaPods](https://forums.swift.org/t/swiftnio-is-dropping-support-for-cocoapods/56840) I thought it'd be fun to revisit the example App and refresh it. 

I also believe this PR resolves https://github.com/grpc/grpc-swift/issues/1393

Modifications:

- Removed Cocoapods from the Speech to Text Example App
- Added missing `resource.proto` to generation step (Must have been added recently)
- Added the required Swift Packages to the project


Examples/Google/SpeechToText/Sources/SpeechService.swift
```diff
@@ -16,6 +16,7 @@
 
 import GRPC
 import Logging
+import Foundation
```
glbrntt, line 19: The formatter doesn't like the order of these imports, I don't think we need `Foundation` here so let's just remove it.

```suggestion
```

glbrntt (commented):
Thanks @Jake-Prickett, just one suggestion left inline.

Examples/Google/SpeechToText/Sources/SpeechService.swift
Jake-Prickett: When I remove the import, the compiler has trouble finding `Data` for the `stream(_ data: Data, ..)` method below. Not 100% certain, my hunch would be that when removing Cocoapods a project setting changed that allowed it to not be required. 

I've corrected the order of the imports 👍

glbrntt: Ah, great, thanks!

glbrntt (approved):

PR #1407, opened 2022-05-16
Refactor the async server call context

glbrntt:
Motivation:

The async call context provided to async server methods provides an API
to access request headers and the logger as well as modify the response
headers, trailers and user info.

This is currently implemented as a class with its mutable state
protected by a lock. This works fine but is racy: the response headers
can be updated after writing the first message and still be sent before
that message (as writing the message requires executing onto the event
loop).

Modifications:

- Turn the context into a `struct` and break it into request and
  response components.
- Request components are immutable.
- Response components are mutable via `async` calls which execute onto
  the underlying event loop.
- This introduces allocations as the work is submitted onto the
  event loop, however setting headers is not a frequent operation and is
  worth the trade off for a less surprising API.
- Trap when headers/trailers are set after they have been sent
  (this is an assertion failre).

Result:

- Async server context has a less surprising API
- Setting headers/trailers is not racy

### DIFF
--- a/Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
+++ b/Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
@@ -15,114 +15,104 @@
  */
 #if compiler(>=5.6)
 
-import Logging
+@preconcurrency import Logging
 import NIOConcurrencyHelpers
-import NIOHPACK
+@preconcurrency import NIOHPACK
 
-// We use a `class` here because we do not want copy-on-write semantics. The instance that the async
-// handler holds must not diverge from the instance the implementor of the RPC holds. They hold these
-// instances on different threads (EventLoop vs Task).
-//
-// We considered wrapping this in a `struct` and pass it `inout` to the RPC. This would communicate
-// explicitly that it stores mutable state. However, without copy-on-write semantics, this could
-// make for a surprising API.
-//
-// We also considered an `actor` but that felt clunky at the point of use since adopters would need
-// to `await` the retrieval of a logger or the updating of the trailers and each would require a
-// promise to glue the NIO and async-await paradigms in the handler.
-//
-// Note: this is `@unchecked Sendable`; all mutable state is protected by a lock.
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
-public final class GRPCAsyncServerCallContext: @unchecked Sendable {
-  private let lock = Lock()
-
-  /// Metadata for this request.
-  public let requestMetadata: HPACKHeaders
-
-  /// The logger used for this call.
-  public var logger: Logger {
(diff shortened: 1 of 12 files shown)


Tests/GRPCTests/GRPCAsyncServerHandlerTests.swift
```diff
@@ -230,34 +230,6 @@ class AsyncServerHandlerTests: GRPCTestCase {
     try await responseStream.next().assertNil()
   }
 
-  func testResponseHeadersDroppedIfSetAfterFirstResponse() async throws {
-    throw XCTSkip("Setting metadata is racy. This test will not reliably pass until that is fixed.")
```
glbrntt, line 234: This was removed because it no longer applies: we now trap (on debug builds) if headers are set after writing the first response

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
```diff
@@ -15,114 +15,104 @@
...
+    public func setHeaders(_ headers: HPACKHeaders) async throws {
+      try await self.contextProvider.setResponseHeaders(headers)
+    }
+
+    /// Set the metadata to return at the end of the RPC.
+    ///
+    /// If this is required it must be updated before returning from the handler.
+    public func setTrailers(_ trailers: HPACKHeaders) async throws {
+      try await self.contextProvider.setResponseTrailers(trailers)
+    }
```
FranzBusch, line 103: The new API only allows to set headers/trailers. Should we expose APIs that allow to mutate the currently set headers?

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
```diff
@@ -15,114 +15,104 @@
...
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension GRPCAsyncServerCallContext {
+  public struct Request: Sendable {
+    /// The request headers received from the client at the start of the RPC.
+    public var headers: HPACKHeaders
+
+    /// A logger.
+    public var logger: Logger
+
```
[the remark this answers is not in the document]
glbrntt: There's no harm in having them: the caller would have to take a copy of `Request` to mutate these anyway.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
glbrntt: Good question. I thought about this a bit and I don't think we should provide it.

They are empty by default and the caller _knows_ what the value is: it can only have been set by them so I'm not sure it's actually useful to have (any update could be achieved by just setting the value a bit later).

FranzBusch (approved):
LGTM. Nice work!

PR #1410, opened 2022-05-24
Close client request stream on error and end

glbrntt:
Motivation:

It's possible to write messages on the client request stream after the
RPC has closed. It shouldn't be.

Modifications:

- Close the request stream on end and on error for client streaming and
  bidirectional streaming rpcs

Result:

Callers can't write to a request stream after an RPC has finished.


fabianfett (approved):
LGTM

PR #1413, opened 2022-05-27
Make client RPC cancellation non-async

glbrntt:
Motivation:

Cancelling an RPC as a client is done via a cancellation function on the
call which `throws` and is also `async`. That it `throws` is odd as an
API because by virtue of cancelling we no longer care about the result
so it doesn't matter if cancellation was successful or not (i.e. if the
RPC was already cancelled). That the cancellation is `async` does not
fit with task cancellation handlers which are also not `async`.

Modifications:

- Make all async call `cancel()` sync and non throwing.
- Add tests for the 'wrapped' RPCs (and fix some bugs along the way)

Result:

Cancelling RPCs from the client is async and more robust.

### DIFF
--- a/Sources/GRPC/AsyncAwaitSupport/GRPCAsyncBidirectionalStreamingCall.swift
+++ b/Sources/GRPC/AsyncAwaitSupport/GRPCAsyncBidirectionalStreamingCall.swift
@@ -36,8 +36,8 @@ public struct GRPCAsyncBidirectionalStreamingCall<Request: Sendable, Response: S
   }
 
   /// Cancel this RPC if it hasn't already completed.
-  public func cancel() async throws {
-    try await self.call.cancel().get()
+  public func cancel() {
+    self.call.cancel(promise: nil)
   }
 
   // MARK: - Response Parts
--- a/Sources/GRPC/AsyncAwaitSupport/GRPCAsyncResponseStream.swift
+++ b/Sources/GRPC/AsyncAwaitSupport/GRPCAsyncResponseStream.swift
@@ -44,7 +44,8 @@ public struct GRPCAsyncResponseStream<Element>: AsyncSequence {
 
     @inlinable
     public mutating func next() async throws -> Element? {
-      try await self.iterator.next()
+      if Task.isCancelled { throw GRPCStatus(code: .cancelled) }
+      return try await self.iterator.next()
     }
   }
 }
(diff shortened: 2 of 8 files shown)


Sources/GRPC/AsyncAwaitSupport/GRPCClient+AsyncAwaitSupport.swift
```diff
@@ -401,30 +413,27 @@ extension GRPCClient {
...
-      // If we have a response, we can stop sending requests.
-      requestsTask.cancel()
-      // Return the response.
-      return response
-    } onCancel: {
-      requestsTask.cancel()
-      // If this outer task is cancelled then we should also cancel the RPC.
-      Task.detached {
-        try await call.cancel()
+    return try await withThrowingTaskGroup(of: Void.self, returning: Response.self) { group in
```
fabianfett, line 416: Why do we need TaskGroup here? Only one task is attached.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncResponseStream.swift
```diff
@@ -44,7 +44,8 @@ public struct GRPCAsyncResponseStream<Element>: AsyncSequence {
 
     @inlinable
     public mutating func next() async throws -> Element? {
-      try await self.iterator.next()
+      if Task.isCancelled { throw GRPCStatus(code: .cancelled) }
```
FranzBusch, line 47: Hm, I am bit unsure about that shouldn't we throw the `try checkCancellation()`  here? The `Task` got cancelled and we did not receive `GRPCStatus` from the remote and it could be confusing to the user why he got a `GRPCStatus` here.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncBidirectionalStreamingCall.swift
```diff
@@ -36,8 +36,8 @@ public struct GRPCAsyncBidirectionalStreamingCall<Request: Sendable, Response: S
   }
 
   /// Cancel this RPC if it hasn't already completed.
-  public func cancel() async throws {
-    try await self.call.cancel().get()
+  public func cancel() {
+    self.call.cancel(promise: nil)
```
fabianfett, line 40: I wonder if, we should make the underlying promise succeed always in a follow up pr?

fabianfett (commented):
Just remove the TaskGroup and we should be fine!

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncResponseStream.swift
glbrntt: We do it for consistency with other throwing gRPC apis. We always produce a `GRPCStatus` _even_ if the remote did not send one (this is expected behaviour).

When cancelling client calls with `cancel()` the resulting status of the call will be `.cancelled` which is _also_ used to fail any remaining promises/continuations (i.e. any `try await call.response` will throw a `GRPCStatus(code: .cancelled)`).

FranzBusch (approved):

PR #1414, opened 2022-05-27
Add co-operative cancellation to async writer and passthrough source

glbrntt:
Motivation:

Whenever we create continuations we should be careful to add support for
co-operative cancellation via a cancellation handler.

Modifications:

- Add co-operative cancellation to the async write and passthrough
  source
- Tests

Result:

Better cancellation support

### DIFF
--- a/Sources/GRPC/AsyncAwaitSupport/AsyncWriter.swift
+++ b/Sources/GRPC/AsyncAwaitSupport/AsyncWriter.swift
@@ -245,50 +245,45 @@ internal final actor AsyncWriter<Delegate: AsyncWriterDelegate>: Sendable {
   ///   have been suspended.
   @inlinable
   internal func write(_ element: Element) async throws {
-    try await withCheckedThrowingContinuation { continuation in
-      self._write(element, continuation: continuation)
-    }
-  }
-
-  @inlinable
-  internal func _write(_ element: Element, continuation: CheckedContinuation<Void, Error>) {
     // There are three outcomes of writing:
     // - write the element directly (if the writer isn't paused and no writes are pending)
     // - queue the element (the writer is paused or there are writes already pending)
     // - error (the writer is complete or the queue is full).
-
-    if self._completionState.isPendingOrCompleted {
-      continuation.resume(throwing: GRPCAsyncWriterError.alreadyFinished)
-    } else if !self._isPaused, self._pendingElements.isEmpty {
-      self._delegate.write(element)
-      continuation.resume()
-    } else if self._pendingElements.count < self._maxPendingElements {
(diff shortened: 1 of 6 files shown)


Sources/GRPC/AsyncAwaitSupport/AsyncWriter.swift
```diff
@@ -245,50 +245,45 @@ internal final actor AsyncWriter<Delegate: AsyncWriterDelegate>: Sendable {
...
+        self._completionState = .completed
+        self._delegate.writeEnd(end)
+      } else {
+        try await withCheckedThrowingContinuation { continuation in
+          // Either we're paused or there are pending writes which must be consumed first.
+          self._completionState = .pending(PendingEnd(end, continuation: continuation))
+        }
+      }
+    } onCancel: {
+      self.cancelAsynchronously()
```
fabianfett, line 286: I think, this should just be `cancel`. No need to create another task.
```suggestion
      self.cancel()
```

Sources/GRPC/AsyncAwaitSupport/AsyncWriter.swift
```diff
@@ -245,50 +245,45 @@ internal final actor AsyncWriter<Delegate: AsyncWriterDelegate>: Sendable {
...
+      } else if self._pendingElements.count < self._maxPendingElements {
+        // The continuation will be resumed later.
+        try await withCheckedThrowingContinuation { continuation in
+          self._pendingElements.append(PendingElement(element, continuation: continuation))
+        }
+      } else {
+        throw GRPCAsyncWriterError.tooManyPendingWrites
+      }
+    } onCancel: {
+      self.cancelAsynchronously()
```
fabianfett, line 266: No need for redispatch into the `AsyncWriter`
```suggestion
      self.cancel()
```

fabianfett (commented):
Two small issues.

Sources/GRPC/AsyncAwaitSupport/AsyncWriter.swift
glbrntt: This doesn't work unfortunately: `Actor-isolated instance method 'cancel()' can not be referenced from a Sendable closure`

IIUC `onCancel` is not executed on the actor but a different thread (which makes sense as it runs immediately when cancelled), so I think this makes sense.

fabianfett (approved):

Sources/GRPC/AsyncAwaitSupport/AsyncWriter.swift
fabianfett: urgh. makes sense.

PR #1415, opened 2022-05-27
Don't fail on cancellation

glbrntt:
Motivation:

Cancellation should not throw if the RPC has already been cancelled.

Modifications:

- Don't fail a cancellation promise if the call has not yet been started
  or has already been cancelled.

Result:

Cancellation does not fail.


fabianfett (approved):
Love it! Makes cancellation much easier to use!

PR #1419, opened 2022-05-30
Fix a handful of Sendable warnings

glbrntt:
Motivation:

We require all request/response types to be `Sendable` as they always cross a
thread boundary. Our various async sequences and async writers are conditionally
Sendable if their `Element` is `Sendable`, however, in practice they always will
be. We require them to be `Sendable` as they are arguments to `async` functions
(see SE-0338). For the same reason, the async server context (and by extension
the async server handler) must also be `Sendable`.

Modifications:

- Require passthrough message/source sequence to have `Sendable` elements so
  that they are unconditionally `Sendable`
- Make the async server context and handler `Sendable`
- Make the async writer error Sendable
- Remove a global variable from protoc-gen-grpc-swift
- Add appropriate conditions to extensions on `Call` which define functions
  requiring `Call` to be `Sendable`

Result:

Better `Sendable` support.

### DIFF
--- a/Sources/GRPC/AsyncAwaitSupport/AsyncWriter.swift
+++ b/Sources/GRPC/AsyncAwaitSupport/AsyncWriter.swift
@@ -296,11 +296,11 @@ extension AsyncWriter where End == Void {
   }
 }
 
-public struct GRPCAsyncWriterError: Error, Hashable {
+public struct GRPCAsyncWriterError: Error, Sendable, Hashable {
   private let wrapped: Wrapped
 
   @usableFromInline
-  internal enum Wrapped {
+  internal enum Wrapped: Sendable {
     case tooManyPendingWrites
     case alreadyFinished
   }
--- a/Sources/GRPC/AsyncAwaitSupport/Call+AsyncRequestStreamWriter.swift
+++ b/Sources/GRPC/AsyncAwaitSupport/Call+AsyncRequestStreamWriter.swift
@@ -16,7 +16,7 @@
 #if compiler(>=5.6)
 
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
-extension Call {
+extension Call where Request: Sendable, Response: Sendable {
   internal func makeRequestStreamWriter() -> GRPCAsyncRequestStreamWriter<Request> {
     let delegate = GRPCAsyncRequestStreamWriter<Request>.Delegate(
       compressionEnabled: self.options.messageEncoding.enabledForRequests
--- a/Sources/GRPC/AsyncAwaitSupport/GRPCAsyncResponseStream.swift
+++ b/Sources/GRPC/AsyncAwaitSupport/GRPCAsyncResponseStream.swift
@@ -18,7 +18,7 @@
 /// This is currently a wrapper around AsyncThrowingStream because we want to be
 /// able to swap out the implementation for something else in the future.
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
-public struct GRPCAsyncResponseStream<Element>: AsyncSequence {
+public struct GRPCAsyncResponseStream<Element: Sendable>: AsyncSequence {
   @usableFromInline
   internal typealias WrappedStream = PassthroughMessageSequence<Element, Error>
 
@@ -51,8 +51,8 @@ public struct GRPCAsyncResponseStream<Element>: AsyncSequence {
 }
 
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
-extension GRPCAsyncResponseStream: Sendable where Element: Sendable {}
+extension GRPCAsyncResponseStream: Sendable {}
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
-extension GRPCAsyncResponseStream.Iterator: Sendable where Element: Sendable {}
+extension GRPCAsyncResponseStream.Iterator: Sendable {}
 
 #endif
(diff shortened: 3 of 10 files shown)


FranzBusch (approved):

Sources/GRPC/AsyncAwaitSupport/AsyncWriter.swift
```diff
@@ -296,11 +296,11 @@ extension AsyncWriter where End == Void {
   }
 }
 
-public struct GRPCAsyncWriterError: Error, Hashable {
+public struct GRPCAsyncWriterError: Error, Sendable, Hashable {
```
fabianfett, line 299: Sendability is already required from `Error` conformance.

Davidde94: https://developer.apple.com/documentation/swift/error

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -798,7 +803,7 @@ protocol AsyncServerCallContextProvider {
 
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
 @usableFromInline
-internal struct ServerHandlerComponents<Request, Delegate: AsyncWriterDelegate> {
+internal struct ServerHandlerComponents<Request: Sendable, Delegate: AsyncWriterDelegate> {
```
fabianfett, line 806: I don't think, we should enforce this.

Sources/GRPC/AsyncAwaitSupport/PassthroughMessageSequence.swift
```diff
@@ -18,7 +18,7 @@
 /// An ``AsyncSequence`` adapter for a ``PassthroughMessageSource``.`
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
 @usableFromInline
-internal struct PassthroughMessageSequence<Element, Failure: Error>: AsyncSequence {
+internal struct PassthroughMessageSequence<Element: Sendable, Failure: Error>: AsyncSequence {
```
fabianfett, line 21: I don't think, we should enforce this.

Sources/GRPC/AsyncAwaitSupport/PassthroughMessageSource.swift
```diff
@@ -29,7 +29,7 @@ import NIOCore
 /// indicate that the sequence should end with an error.
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
 @usableFromInline
-internal final class PassthroughMessageSource<Element, Failure: Error> {
+internal final class PassthroughMessageSource<Element: Sendable, Failure: Error> {
```
fabianfett, line 32: I don't think, we should enforce this.

fabianfett (changes requested):
I still think we don't need to require Request and Response to be Sendable.

Sources/GRPC/AsyncAwaitSupport/Call+AsyncRequestStreamWriter.swift
```diff
@@ -16,7 +16,7 @@
 #if compiler(>=5.6)
 
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
-extension Call {
+extension Call where Request: Sendable, Response: Sendable {
```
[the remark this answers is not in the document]
Davidde94: Conditional extensions aren't enforcements - if Response and a Request don't conform then neither will this

Sources/GRPC/AsyncAwaitSupport/PassthroughMessageSource.swift
Lukasa: I'm curious as to why not. A non-`Sendable` `AsyncSequence` is next to useless: you can only send to it and receive from it on the same `Task`. When is anyone ever going to do that?

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncResponseStream.swift
```diff
@@ -18,7 +18,7 @@
 /// This is currently a wrapper around AsyncThrowingStream because we want to be
 /// able to swap out the implementation for something else in the future.
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
-public struct GRPCAsyncResponseStream<Element>: AsyncSequence {
+public struct GRPCAsyncResponseStream<Element: Sendable>: AsyncSequence {
```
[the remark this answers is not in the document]
Lukasa: I think we have to. The underlying elements are produced from a NIO thread, so they are necessarily `Sendable`. The type is useless without that.

Sources/GRPC/AsyncAwaitSupport/Call+AsyncRequestStreamWriter.swift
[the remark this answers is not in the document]
glbrntt: As `GRPCAsyncRequestStreamWriter` expects `Request` to be `Sendable` we require it be here. We also require `Response` to be `Sendable` to that `Call` is `Sendable` which we require to capture it in the `Sendable` closures for the delegate.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
glbrntt: We need this one to enforce it for the `PassthroughMessageSource`

glbrntt (commented):
We talked a little offline about whether or not the various `AsyncSequence`s should be required to be `Sendable` here and whether `Request` and `Response` are required to be `Sendable`. 

The summary is that `Request` and `Response` must always be `Sendable`: they are always transferred across a thread boundary (event loop to concurrency context). For async sequences in some cases it would be useful to be able to _move_ the ownership of the `AsyncSequence` to another concurrency context (in particular moving it from the event-loop context to a concurrency context). However we can't currently represent that and so we must make them `Sendable`.

PR #1423, opened 2022-06-01
Remove async main workarounds for Echo

glbrntt:
Motivation:

The async echo example was written before ArgumentParser
supported async/await. As such it has a few hacks to make it
work with `@main`.

Modifications:

- Remove hacks and use `AsyncParsableCommand`

Result:

- Fewer hacks.


FranzBusch (approved):
LGTM!

PR #1439, opened 2022-06-14
Remove Cocoapods script and references

glbrntt:
Motivation:

We no longer support Cocoapods, so we should remove any
references to them from the repo.

Modifications:

- Remove pod related tooling and documentation.

Result:

More accurate documentation. Cleaner repo.


Lukasa (approved):

drcapulet:
Any reason for the dropped Cocoapods support? We still use it for a number of reasons, but wrt this library it's so we can dynamically compile protos to Swift using the plugins pod.

PR #1444, opened 2022-06-23
Raise minimum NIO version to 2.36.0

### LINKED ISSUE #1443: [Version mismatch] 'get()' is only available in iOS 15 or newer
### Describe the bug

It seem like there is a version mismatch issue in the latest v1.8.0 version, build as the project was fine.
But when I import it to my own project, it throws other errors, I can see those 2 places iOS version is not consistent.

Here is the version info:
gRPC-swift version `1.8.0`
Xcode version `14-beta`

### To reproduce

Steps to reproduce the bug you've found:
1. Import the lib from SPM
2. Build the project

### Screenshots
Screen Shot 2022-06-23 at 4 44 38 pm
Screen Shot 2022-06-23 at 4 44 59 pm

Lukasa:
With the merge of the async/await support to main we now require NIO
2.36.0, which is the release that contains the new availability
requirements for EventLoopFuture.get

Resolves #1443

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -40,7 +40,7 @@ let argumentParserMinimumVersion: Version = "1.0.0"
 let packageDependencies: [Package.Dependency] = [
   .package(
     url: "https://github.com/apple/swift-nio.git",
-    from: "2.32.0"
+    from: "2.36.0"
   ),
   .package(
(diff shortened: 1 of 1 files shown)


glbrntt (approved):
ta!

PR #1453, opened 2022-07-11
Use swift-atomics

glbrntt:
Motivation:

NIO deprecated its atomics in apple/swift-nio#2204.

Modifications:

- Use swift-atomics in the `QPSBenchmark` sub-package
- Use a lock in the one test which used `NIOAtomic`

Result:

Not using deprecated code.


FranzBusch (approved):
LGTM!

Issue #1452, opened 2022-07-11
Want to change some sources of plugin: --grpc-swift_out: protoc-gen-grpc-swift: Plugin killed by signal 9.


bimawa:
When I trying change some generator logic and recreate grpc plugin and satart it I got the error:
`--grpc-swift_out: protoc-gen-grpc-swift: Plugin killed by signal 9.`

How I launch the protoc: 
```
$: make plugins; protoc streamlayerTest/load.proto \
                                           --proto_path=streamlayerTest/ \
                                           --plugin=./protoc-gen-swift \
                                           --swift_opt=Visibility=Public \
                                           --swift_out=streamlayerOut/ \
                                           --plugin=./protoc-gen-grpc-swift \
                                           --swift_opt=Visibility=Public \
                                           --grpc-swift_opt=Server=false,GRPCModuleName=GRPC 
```
output: 
```
swift build --build-path=./.build --configuration=release --enable-test-discovery --product protoc-gen-grpc-swift
warning: '--enable-test-discovery' option is deprecated; tests are automatically discovered on all platforms
Building for production...
[2/2] Linking protoc-gen-grpc-swift
Build complete! (3.03s)
cp .build/release/protoc-gen-swift .build/release/protoc-gen-grpc-swift .
--grpc-swift_out: protoc-gen-grpc-swift: Plugin killed by signal 9.
```
If I change Generator.swift back. Its workin. if I change some line, its Exit with error 9
How can I cusomise the grpc generator plugin?

glbrntt:
What did you change in `Generator.swift` that causes it to crash?

glbrntt:
Odd that it works as expected in debug mode though.

bimawa:
![CleanShot 2022-07-11 at 20 11 07@2x](https://user-images.githubusercontent.com/1393809/178297026-e1ed5bf8-1253-42ca-a8be-5d57f51314e1.png)
under verbose mode in release config i got this warnings

bimawa:
hmm ok its worked as expected now.
But. when I execute: `make plugins`
Its make me broken binary of `protoc-gen-grpc-swift`
When I build manualy by: 
```
swift build --build-path=./.build --configuration=release --product protoc-gen-grpc-swift
cp .build/debug/protoc-gen-grpc-swift .
```
Its work as expected.

bimawa:
But with make plugins still not working (
I'll investigate it

bimawa:
[...]

PR #1454, opened 2022-07-11
Update Makefile

bimawa:
After update, source code of `protoc-gen-grpc-swift` binary not work correctly.
Issue described here: #1452 

Maybe it's related with Apple Silicon, but `cp` command not work correctly. Only If I remove old binaries, it's work as expected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,14 +2,14 @@
 SWIFT:=swift
 # Where products will be built; this is the SPM default.
 SWIFT_BUILD_PATH:=./.build
-SWIFT_BUILD_CONFIGURATION=debug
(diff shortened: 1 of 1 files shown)


glbrntt (commented):
Thanks for opening this but I'm not willing to accept it as is because I don't think it's actually solving a problem here.

The `rm` step in the `plugins` target is unnecessary: `cp` will overwrite any files which already exist, so removing them first isn't necessary.

Most of the other build configuration changes are also not necessary as the plugin targets are already built using release configuration (see line 35 and line 38).

bimawa:
Ok thanks.

PR #1456, opened 2022-07-12
Provide an error when cancelling async writer

glbrntt:
Motivation:

The `AsyncWriter` fails any pending writes with a `CancellationError` if
the writer is cancelled.

This can be misleading if the writer is cancelled because a connection
could not be established, for example.

Modifications:

- Pass an error through `AsyncWriter.cancel`

Result:

Pending writes which are failed because the writer has been cancelled
have more relevant errors.

### DIFF
--- a/Tests/GRPCTests/AsyncAwaitSupport/AsyncClientTests.swift
+++ b/Tests/GRPCTests/AsyncAwaitSupport/AsyncClientTests.swift
@@ -405,19 +405,13 @@ final class AsyncClientCancellationTests: GRPCTestCase {
       $0.connectionPool.maxWaitTime = .milliseconds(10)
     }
 
-    do {
-      let update = echo.makeUpdateCall()
-      try await update.requestStream.send(.init())
-      XCTFail("Expected an error to be thrown")
-    } catch {
+    let update = echo.makeUpdateCall()
+    await XCTAssertThrowsError(try await update.requestStream.send(.init())) { error in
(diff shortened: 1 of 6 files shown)


Tests/GRPCTests/AsyncAwaitSupport/AsyncClientTests.swift
```diff
@@ -394,6 +398,29 @@ final class AsyncClientCancellationTests: GRPCTestCase {
...
+      XCTAssertFalse(error is CancellationError)
+    }
+
+    do {
+      let collect = echo.makeCollectCall()
+      try await collect.requestStream.send(.init())
+      XCTFail("Expected an error to be thrown")
+    } catch {
+      XCTAssertFalse(error is CancellationError)
+    }
```
FranzBusch, line 422: Nit: Should we add the [overload](https://github.com/swift-server/async-http-client/blob/1af18d2f95a24358b9e0001d9597a4646d2555ca/Tests/AsyncHTTPClientTests/XCTest%2BAsyncAwait.swift#L64-L77) from AHC to gRPC as well?

FranzBusch (approved):
LGTM! Just a small suggestion

Tests/GRPCTests/AsyncAwaitSupport/AsyncClientTests.swift
glbrntt: No, we should just use the one that AHC copied from gRPC 🙈

PR #1462, opened 2022-07-23
Make path available in async calls

pouyayarandi:
This PR adds a `path` property to async client calls. There was a similar issue with non async calls before. (#1236)


pouyayarandi:
@glbrntt I would appreciate it if you review it

pouyayarandi:
@glbrntt Do you have any idea what causes checks to fail?

glbrntt:
> @glbrntt Do you have any idea what causes checks to fail?

Unclear; I didn't have a chance to take a proper look yet, will try today. (Thanks for the PR by the way!)

glbrntt (approved):

PR #1467, opened 2022-08-02
Prepare for DocC

FranzBusch:
# Motivation
We wanna use DocC for publishing docs.

# Modification
Setup our `Package.swift`  manifests to enable proper doc generation.

# Result
We can now use DocC to generate docs.


Lukasa (approved):

Lukasa:
Can you clean up the license headers?

PR #1470, opened 2022-08-09
Add async version of QPS Benchmark Service

gjcairo:
Motivation:
    
We want to provide an async/await version of the QPS Benchmark service, and be able to choose whether we want to use that or the ELF version when running the tests.
    
Modifications:

Added support for an async/await implementation of the Benchmark Service. This is currently not in use, but a follow-up CR will include changes to the Worker Service and other classes, and will enable choosing between the ELF and async/await implementations when running.
    
Result:
    
An async/await version of the Benchmark Service will be ready for use.

### DIFF
--- a/Performance/QPSBenchmark/Package.swift
+++ b/Performance/QPSBenchmark/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.4
+// swift-tools-version:5.6
 /*
  * Copyright 2020, gRPC Authors All rights reserved.
  *
@@ -19,6 +19,7 @@ import PackageDescription
 
 let package = Package(
   name: "QPSBenchmark",
+  platforms: [.macOS(.v12)],
   products: [
     .executable(name: "QPSBenchmark", targets: ["QPSBenchmark"]),
   ],
--- a/Performance/QPSBenchmark/Sources/QPSBenchmark/Model/benchmark_service.grpc.swift
+++ b/Performance/QPSBenchmark/Sources/QPSBenchmark/Model/benchmark_service.grpc.swift
@@ -350,7 +350,7 @@ extension Grpc_Testing_BenchmarkServiceAsyncClientProtocol {
   public func streamingCall<RequestStream>(
     _ requests: RequestStream,
     callOptions: CallOptions? = nil
-  ) -> GRPCAsyncResponseStream<Grpc_Testing_SimpleResponse> where RequestStream: AsyncSequence, RequestStream.Element == Grpc_Testing_SimpleRequest {
+  ) -> GRPCAsyncResponseStream<Grpc_Testing_SimpleResponse> where RequestStream: AsyncSequence & Sendable, RequestStream.Element == Grpc_Testing_SimpleRequest {
     return self.performAsyncBidirectionalStreamingCall(
       path: Grpc_Testing_BenchmarkServiceClientMetadata.Methods.streamingCall.path,
       requests: requests,
@@ -374,7 +374,7 @@ extension Grpc_Testing_BenchmarkServiceAsyncClientProtocol {
   public func streamingFromClient<RequestStream>(
     _ requests: RequestStream,
     callOptions: CallOptions? = nil
-  ) async throws -> Grpc_Testing_SimpleResponse where RequestStream: AsyncSequence, RequestStream.Element == Grpc_Testing_SimpleRequest {
+  ) async throws -> Grpc_Testing_SimpleResponse where RequestStream: AsyncSequence & Sendable, RequestStream.Element == Grpc_Testing_SimpleRequest {
     return try await self.performAsyncClientStreamingCall(
       path: Grpc_Testing_BenchmarkServiceClientMetadata.Methods.streamingFromClient.path,
       requests: requests,
@@ -410,7 +410,7 @@ extension Grpc_Testing_BenchmarkServiceAsyncClientProtocol {
   public func streamingBothWays<RequestStream>(
     _ requests: RequestStream,
     callOptions: CallOptions? = nil
-  ) -> GRPCAsyncResponseStream<Grpc_Testing_SimpleResponse> where RequestStream: AsyncSequence, RequestStream.Element == Grpc_Testing_SimpleRequest {
+  ) -> GRPCAsyncResponseStream<Grpc_Testing_SimpleResponse> where RequestStream: AsyncSequence & Sendable, RequestStream.Element == Grpc_Testing_SimpleRequest {
     return self.performAsyncBidirectionalStreamingCall(
       path: Grpc_Testing_BenchmarkServiceClientMetadata.Methods.streamingBothWays.path,
       requests: requests,
--- a/Performance/QPSBenchmark/Sources/QPSBenchmark/Model/worker_service.grpc.swift
+++ b/Performance/QPSBenchmark/Sources/QPSBenchmark/Model/worker_service.grpc.swift
@@ -302,7 +302,7 @@ extension Grpc_Testing_WorkerServiceAsyncClientProtocol {
   public func runServer<RequestStream>(
     _ requests: RequestStream,
     callOptions: CallOptions? = nil
-  ) -> GRPCAsyncResponseStream<Grpc_Testing_ServerStatus> where RequestStream: AsyncSequence, RequestStream.Element == Grpc_Testing_ServerArgs {
+  ) -> GRPCAsyncResponseStream<Grpc_Testing_ServerStatus> where RequestStream: AsyncSequence & Sendable, RequestStream.Element == Grpc_Testing_ServerArgs {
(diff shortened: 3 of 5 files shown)


Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/Async/AsyncPingPongRequestMaker.swift
```diff
@@ -0,0 +1,86 @@
...
+    var messagesSent = 0
+    var streamingCall: GRPCAsyncBidirectionalStreamingCall<
+      Grpc_Testing_SimpleRequest,
+      Grpc_Testing_SimpleResponse
+    >
+
+    streamingCall = self.client.makeStreamingCallCall()
+    while !self.stopRequested, self.messagesPerStream == 0 || messagesSent < self.messagesPerStream {
+      try await streamingCall.requestStream.send(self.requestMessage)
+      var responseStream = streamingCall.responseStream.makeAsyncIterator()
```
glbrntt, line 71: Can you pull this out of the while loop? Not all `AsyncSequence`s support having multiple iterators created.

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/Async/AsyncQPSClientImpl.swift
```diff
@@ -0,0 +1,293 @@
...
+      )
+      precondition(exchangedRunning.exchanged, "launchRequests should only be called once")
+
+      // We only decrement the outstanding count when running has been changed back to false.
+      let exchangedOutstanding = self.outstanding.compareExchange(
+        expected: 0,
+        desired: self.maxPermittedOutstandingRequests,
+        ordering: .relaxed
+      )
+      precondition(exchangedOutstanding.exchanged, "launchRequests should only be called once")
```
glbrntt, line 228: I don't think `outstanding` is used anymore?

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/Async/AsyncQPSClientImpl.swift
```diff
@@ -0,0 +1,293 @@
...
+      self.cpuStatsPeriodStart = currentResourceUsage
+    }
+  }
+
+  /// Shutdown the service.
+  /// - parameters:
+  ///     - callbackLoop: Which eventloop should be called back on completion.
+  /// - returns: A future on the `callbackLoop` which will succeed on completion of shutdown.
+  func shutdown() async throws {
+    for repeater in self.channelRepeaters {
```
glbrntt, line 121: We can use a task group here to shutdown the repeaters concurrently. At the moment we're waiting for each to stop before moving onto the next.

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/Async/AsyncQPSClientImpl.swift
```diff
@@ -0,0 +1,293 @@
...
+  /// Class to manage a channel.  Repeatedly makes requests on that channel and records what happens.
+  private class ChannelRepeater {
+    private let channel: GRPCChannel
+    private let eventLoop: EventLoop
+    private let maxPermittedOutstandingRequests: Int
+
+    private let stats: StatsWithLock
+
+    /// Succeeds after a stop has been requested and all outstanding requests have completed.
+    private var stopComplete: EventLoopPromise<Void>
```
glbrntt, line 170: this should be a constant

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/Async/AsyncQPSClientImpl.swift
```diff
@@ -0,0 +1,293 @@
...
+
+    private let running = ManagedAtomic<Bool>(false)
+    private let outstanding = ManagedAtomic<Int>(0)
+
+    private var requestMaker: RequestMakerType
```
glbrntt, line 175: this should also be a constant

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/Async/AsyncPingPongRequestMaker.swift
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+import Foundation
+import GRPC
+import Logging
+import NIOCore
+
+/// Makes streaming requests and listens to responses ping-pong style.
+/// Iterations can be limited by config.
+final class AsyncPingPongRequestMaker: AsyncRequestMaker {
```
glbrntt, line 24: I think we have some thread safety issues. We need to adopt `Sendable` on the request makers and channel repeaters (at a minimum). No doubt other things will need to become `Sendable` as a result of the changes.

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/Async/AsyncQPSServerImpl.swift
```diff
@@ -0,0 +1,99 @@
...
+      .withServiceProviders([workerService])
+      .withLogger(self.logger)
+      .bind(host: "localhost", port: Int(config.port))
+      .get()
+  }
+  
+  func getServerInfo() -> ServerInfo {
+    let port = self.server.channel.localAddress?.port ?? 0
+    return ServerInfo(threadCount: self.threadCount, port: port)
+  }
```
glbrntt, line 63: This can just be a computed property ("getter" functions aren't very Swifty):

```suggestion
  var serverInfo: ServerInfo {
    let port = self.server.channel.localAddress?.port ?? 0
    return ServerInfo(threadCount: self.threadCount, port: port)
  }
```

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/Async/AsyncQPSServerImpl.swift
```diff
@@ -0,0 +1,99 @@
...
+      if reset {
+        self.statsPeriodStart = currentTime
+        self.cpuStatsPeriodStart = currentResourceUsage
+      }
+  }
+
+  /// Shut down the service.
+  func shutdown() async throws {
+    do {
+      try await self.server.close().get()
```
glbrntt, line 91: There should be an `initiateGracefulShutdown()` API (or something along those lines), we should use that instead as it allows active RPCs to finish etc.

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/Async/AsyncServerProtocol.swift
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+import GRPC
+import NIOCore
+
+/// Interface server types must implement when using async APIs.
+protocol AsyncQPSServer {
+  /// Returns the server information for the server.
+  /// - returns: an instance of `ServerInfo`
+  func getServerInfo() -> ServerInfo
```
glbrntt, line 24:
```suggestion
  var serverInfo: ServerInfo { get }
```

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/Async/AsyncWorkerServiceImpl.swift
```diff
@@ -0,0 +1,330 @@
...
+
+// Implementation of the control service for communication with the driver process.
+class AsyncWorkerServiceImpl: Grpc_Testing_WorkerServiceAsyncProvider {
+  let interceptors: Grpc_Testing_WorkerServiceServerInterceptorFactoryProtocol? = nil
+
+  private let finishedPromise: EventLoopPromise<Void>
+  private let serverPortOverride: Int?
+
+  private var runningServer: AsyncQPSServer?
+  private var runningClient: AsyncQPSClient?
```
glbrntt, line 27: This is not thread safe unfortunately. In this instance we can just turn the `class` into an `actor`.

glbrntt (changes requested):
This is broadly in the right shape but there are some thread safety issues so we need to take a dive into adding support for `Sendable`.

I think the benchmark service implementation is totally fine as it is so it may make sense to split this PR up a bit i.e. let's land the benchmark service and the updated generated code etc. and then do another PR with the worker service and all of those bits.

gjcairo:
I've updated the PR to only include the new async/await version of the Benchmark Service, recompiled autogenerated code, and Package.swift changes. 
I'll create a follow-up CR with the changes for the Worker Service, the client code, and all related classes (including changes requested by @glbrntt in this PR). Having to modify some of these classes to conform to Sendable may end up requiring substantially more work so it's better to break them up.

glbrntt (approved):

PR #1471, opened 2022-08-10
Add async implementation of QPSBenchmark

gjcairo:
Motivation:

We want to provide an async/await version of the QPSBenchmark implementation, and be able to choose whether we want to use that or the ELF version when running the performance tests via the gRPC JSON driver.

Modifications:

- Added full async/await-based implementations of the worker, server, and client.
- Created a new command line flag (`--use-async`) to choose between the EventLoopFuture and async/await versions when running the tests.

Result:

QPSBenchmark can optionally be executed using an async/await implementation.

### DIFF
--- a/Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/Async/AsyncPingPongRequestMaker.swift
+++ b/Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/Async/AsyncPingPongRequestMaker.swift
@@ -22,6 +22,8 @@ import NIOCore
 
 /// Makes streaming requests and listens to responses ping-pong style.
 /// Iterations can be limited by config.
+/// Class is marked as `@unchecked Sendable` because `ManagedAtomic<Bool>` doesn't conform
+/// to `Sendable`, but we know it's safe.
 final class AsyncPingPongRequestMaker: AsyncRequestMaker, @unchecked Sendable {
   private let client: Grpc_Testing_BenchmarkServiceAsyncClient
   private let requestMessage: Grpc_Testing_SimpleRequest
--- a/Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/Async/AsyncPingPongRequestMaker.swift
+++ b/Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/Async/AsyncPingPongRequestMaker.swift
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020, gRPC Authors All rights reserved.
+ * Copyright 2022, gRPC Authors All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import Atomics
 import Foundation
 import GRPC
 import Logging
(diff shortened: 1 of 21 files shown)


Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/Async/AsyncPingPongRequestMaker.swift
```diff
@@ -0,0 +1,83 @@
...
+
+import Atomics
+import Foundation
+import GRPC
+import Logging
+import NIOCore
+
+/// Makes streaming requests and listens to responses ping-pong style.
+/// Iterations can be limited by config.
+final class AsyncPingPongRequestMaker: AsyncRequestMaker, @unchecked Sendable {
```
glbrntt, line 25: Can you add a code comment why `@unchecked` is okay here? IIRC it's just because of `ManagedAtomic` but it's useful to write this stuff down so we can fix up when appropriate.

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/Async/AsyncQPSClientImpl.swift
```diff
@@ -0,0 +1,293 @@
...
+            )
+          }
+        }
+      }
+    }
+  }
+
+  /// Make a request which can be sent to the server.
+  private static func makeClientRequest(payloadConfig: Grpc_Testing_PayloadConfig) throws
+    -> Grpc_Testing_SimpleRequest {
```
glbrntt, line 140:
```suggestion
  private static func makeClientRequest(
    payloadConfig: Grpc_Testing_PayloadConfig
  ) throws -> Grpc_Testing_SimpleRequest {
```

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/Async/AsyncQPSClientImpl.swift
```diff
@@ -0,0 +1,293 @@
...
+      var result = Grpc_Testing_SimpleRequest()
+      result.responseType = .compressable
+      result.responseSize = 0
+      result.payload.type = .compressable
+      return result
+    }
+  }
+
+  /// Class to manage a channel.  Repeatedly makes requests on that channel and records what happens.
+  private final class ChannelRepeater: @unchecked Sendable {
```
glbrntt, line 171: nit: can you add a comment as to why it's `@unchecked`?

Performance/QPSBenchmark/Sources/QPSBenchmark/Runtime/Async/AsyncQPSClientImpl.swift
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    /// Launch as many requests as allowed on the channel. Must only be called once.
+    private func launchRequests() async throws {
+      // The plan here is:
+      // - store the max number of outstanding requests in an atomic
+      // - start that many requests asynchronously
+      // - when a request finishes it will either start a new request or decrement the
+      //   the atomic counter (if we've been told to stop)
+      // - if the counter drops to zero we're finished.
```
glbrntt, line 222: I think we can just remove this comment: it's no longer accurate and the code is more straightforward than it was previously so not sure it needs it.

glbrntt (commented):
This is pretty much there, just a couple of tiny things to fixup.

glbrntt (approved):

PR #1473, opened 2022-08-24
Add index docc page and fix comment links

FranzBusch:
# Motivation
We want to be able to generate proper docc pages.

# Modification
This PR adds a new index page for the `GRPC` module which is adapting the current `README`. Furthermore, it fixes a bunch of links in the doc comments so that they resolve properly.

# Result
We can now build better docc pages.

It would be great if we can convert some of the other documentations and examples over into the docc format


Sources/GRPC/CallHandlers/ServerHandlerProtocol.swift
```diff
@@ -20,7 +20,7 @@ import NIOHPACK
 /// On receiving a new RPC, gRPC will ask all available service providers for an instance of this
 /// protocol in order to handle the RPC.
 ///
-/// See also: `CallHandlerProvider.handle(method:context:)`
+/// See also:  ``CallHandlerProvider/handle(method:context:)``.
```
glbrntt, line 23:
```suggestion
/// See also: ``CallHandlerProvider/handle(method:context:)``.
```

Sources/GRPC/CallOptions.swift
```diff
@@ -52,20 +54,20 @@ public struct CallOptions: GRPCSendable {
   /// Whether the call is cacheable.
   public var cacheable: Bool
 
-  /// How IDs should be provided for requests. Defaults to `.autogenerated`.
+  /// How IDs should be provided for requests. Defaults to  ``RequestIDProvider-swift.struct/autogenerated``.
```
glbrntt, line 57:
```suggestion
  /// How IDs should be provided for requests. Defaults to ``RequestIDProvider-swift.struct/autogenerated``.
```

Sources/GRPC/CallOptions.swift
```diff
@@ -52,20 +54,20 @@ public struct CallOptions: GRPCSendable {
...
   ///
   /// The request ID is used for logging and will be added to the headers of a call if
   /// `requestIDHeader` is specified.
   ///
-  /// - Important: When setting `CallOptions` at the client level, `.userDefined` should __not__ be
+  /// - Important: When setting  ``CallOptions`` at the client level,  ``RequestIDProvider-swift.struct/userDefined(_:)`` should __not__ be
```
glbrntt, line 62:
```suggestion
  /// - Important: When setting ``CallOptions`` at the client level, ``RequestIDProvider-swift.struct/userDefined(_:)`` should __not__ be
```

Sources/GRPC/Server.swift
```diff
@@ -309,13 +309,13 @@ extension Server {
     /// The compression configuration for requests and responses.
     ///
     /// If compression is enabled for the server it may be disabled for responses on any RPC by
-    /// setting `compressionEnabled` to `false` on the context of the call.
+    /// setting  `compressionEnabled` to `false` on the context of the call.
```
glbrntt, line 312:
```suggestion
    /// setting `compressionEnabled` to `false` on the context of the call.
```

glbrntt (approved):
Fab, thanks Franz.

PR #1474, opened 2022-08-30
Add SPM plugin for GRPC code generation

gjcairo:
## Motivation

After adding an SPM plugin for protobuf generation, we want to offer the same feature for GRPC code generation.

## Modifications

* Added a GRPC codegen SPM plugin

## Result

GRPC codegen SPM plugin is now available.


Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md
```diff
@@ -0,0 +1,113 @@
...
+
+> Note: paths to your `.proto` files will have to include the full path from the target source to the `.proto` file location. 
+
+### Defining the path to the protoc binary
+
+The plugin needs to be able to invoke the `protoc` binary to generate the Swift types. 
+There are two ways how this can be achieved. First, by default, the package manager looks into
+the `$PATH` to find binaries named `protoc`. This works immediately if you use `swift build` to build
+your package and `protoc` is installed in the `$PATH` (`brew` is adding it to your `$PATH` automatically).
+However, this doesn't work if you want to compile from Xcode since Xcode is not passed the `$PATH`.
```
t089, line 99: This should also work (not tested): alternatively you can start Xcode from a terminal, which will make Xcode inherit the environment from the shell.

```zsh
$ xed Package.swift
```

glbrntt:
@gjcairo can you split this into two PRs? One to add an SPM plugin and another to adopt it in the QPS benchmark?

Package.swift
```diff
@@ -48,7 +48,7 @@ let packageDependencies: [Package.Dependency] = [
   ),
   .package(
     url: "https://github.com/apple/swift-protobuf.git",
-    from: "1.19.0"
+    from: "1.20.0"
```
glbrntt, line 51: Can we use 1.20.1? (1.20.0 has warnings and there's no harm in setting to the newest)

Package.swift
```diff
@@ -89,6 +88,7 @@ extension Target.Dependency {
   static let interopTestModels: Self = .target(name: "GRPCInteroperabilityTestModels")
   static let interopTestImplementation: Self =
     .target(name: "GRPCInteroperabilityTestsImplementation")
+  static let protocGenGRPCSwift: Self = .target(name: "protoc-gen-grpc-swift")
```
glbrntt, line 91: Can this move back to where it was? It's still available externally.

Package.swift
```diff
@@ -165,6 +165,14 @@ extension Target {
     ]
   )
 
+  static let swiftGRPCPlugin: Target = .plugin(
+    name: "SwiftGRPCPlugin",
```
glbrntt, line 169: This should be "GRPCSwiftPlugin"

Package.swift
```diff
@@ -423,6 +431,11 @@ extension Product {
     name: "protoc-gen-grpc-swift",
     targets: ["protoc-gen-grpc-swift"]
   )
+
+  static let swiftGRPCPlugin: Product = .plugin(
```
glbrntt, line 435: `grpcSwiftPlugin`

Performance/QPSBenchmark/Package.swift
```diff
@@ -53,14 +53,9 @@ let package = Package(
...
         .target(name: "BenchmarkUtils"),
       ],
-      exclude: [
-        "Model/benchmark_service.proto",
-        "Model/control.proto",
-        "Model/core_stats.proto",
-        "Model/messages.proto",
-        "Model/payloads.proto",
-        "Model/stats.proto",
-        "Model/worker_service.proto",
```
glbrntt: Previously the excludes were required to stop SPM warning about unused source files; does it no longer warn? (I have a feeling it only did it on clean builds but I could totally be wrong)

Plugins/SwiftGRPCPlugin/plugin.swift
```diff
@@ -0,0 +1,189 @@
...
+      /// An array of paths to `.proto` files for this invocation.
+      var protoFiles: [String]
+      /// The visibility of the generated files.
+      var visibility: Visibility?
+      /// Whether server code is generated.
+      var server: Bool?
+      /// Whether client code is generated.
+      var client: Bool?
+      /// Whether test client code is generated.
+      var testClient: Bool?
```
glbrntt, line 51: This is deprecated so we can leave out the option (until the time that someone complains...)

Plugins/SwiftGRPCPlugin/plugin.swift
```diff
@@ -0,0 +1,189 @@
...
+      /// Determines whether the casing of generated function names is kept.
+      var keepMethodCasing: Bool?
+    }
+
+    /// The path to the `protoc` binary.
+    ///
+    /// If this is not set, SPM will try to find the tool itself.
+    var protocPath: String?
+
+    /// A list of invocations of `protoc` with the `SwiftGRPCPlugin`.
```
glbrntt, line 61: `GRPCSwiftPlugin`

Plugins/SwiftGRPCPlugin/plugin.swift
```diff
@@ -0,0 +1,189 @@
...
+    var invocations: [Invocation]
+  }
+
+  static let configurationFileName = "swift-grpc-config.json"
```
glbrntt, line 65:
```suggestion
  static let configurationFileName = "grpc-swift-config.json"
```

Plugins/SwiftGRPCPlugin/plugin.swift
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    // We need to find the configuration file at the root of the target
+    let configurationFilePath = target.directory.appending(subpath: Self.configurationFileName)
+    let data = try Data(contentsOf: URL(fileURLWithPath: "\(configurationFilePath)"))
+    let configuration = try JSONDecoder().decode(Configuration.self, from: data)
+
+    try self.validateConfiguration(configuration)
+
+    // We need to find the path of protoc and protoc-gen-swift
```
glbrntt, line 80:
```suggestion
    // We need to find the path of protoc and protoc-gen-grpc-swift
```

Plugins/SwiftGRPCPlugin/plugin.swift
```diff
@@ -0,0 +1,189 @@
...
+
+      // Add the outputPath as an output file
+      outputFiles.append(protobufOutputPath)
+    }
+
+    // Construct the command. Specifying the input and output paths lets the build
+    // system know when to invoke the command. The output paths are passed on to
+    // the rule engine in the build system.
+    return Command.buildCommand(
+      displayName: "Generating GRPC swift files from proto files",
```
glbrntt, line 171:
```suggestion
      displayName: "Generating gRPC Swift files from proto files",
```

Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md
```diff
@@ -0,0 +1,113 @@
+# Using the Swift Package Manager plugin
+
+The Swift Package Manager introduced new plugin capabilities in Swift 5.6, enabling the extension of
+the build process with custom build tools. Learn how to use the SwiftGRPCPlugin plugin for the
```
glbrntt, line 4:
```suggestion
the build process with custom build tools. Learn how to use the `GRPCSwiftPlugin` plugin for the
```

Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md
```diff
@@ -0,0 +1,113 @@
...
+First, you need to add a dependency on `grpc-swift`. Afterwards, you can declare the usage of the plugin
+for your target. Here is an example snippet of a `Package.swift` manifest:
+
+```swift
+let package = Package(
+  name: "YourPackage",
+  products: [...],
+  dependencies: [
+    ...
+    .package(url: "https://github.com/grpc/grpc-swift", from: "1.9.1"),
```
glbrntt, line 45: This would be added in a minor version so should be 1.10.0:

```suggestion
    .package(url: "https://github.com/grpc/grpc-swift", from: "1.10.0"),
```

Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md
```diff
@@ -0,0 +1,113 @@
...
+The plugin works by running the system installed `protoc` compiler with the `protoc-gen-grpc-swift` plugin
+for specified `.proto` files in your targets source folder. Furthermore, the plugin allows defining a
+configuration file which will be used to customize the invocation of `protoc`.
+
+### Installing the protoc compiler
+
+First, you must ensure that you have the `protoc` compiler installed.
+There are multiple ways to do this. Some of the easiest are:
+
+1. If you are on MacOS, installing it via `brew install protoc`
```
glbrntt, line 22:
```suggestion
1. If you are on macOS, installing it via `brew install protoc`
```

Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md
```diff
@@ -0,0 +1,113 @@
...
+    ...
+    .package(url: "https://github.com/grpc/grpc-swift", from: "1.9.1"),
+    ...
+  ],
+  targets: [
+    ...
+    .executableTarget(
+        name: "YourTarget",
+        plugins: [
+            .plugin(name: "SwiftGRPCPlugin", package: "grpc-swift")
```
glbrntt, line 53:
```suggestion
            .plugin(name: "GRPCSwiftPlugin", package: "grpc-swift")
```

Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md
```diff
@@ -0,0 +1,113 @@
...
+    ),
+    ...
+)
+
+```
+
+### Configuring the plugin
+
+Lastly, after you have added the `.proto` files and modified your `Package.swift` manifest, you can now
+configure the plugin to invoke the `protoc` compiler. This is done by adding a `swift-grpc-config.json`
```
glbrntt, line 64:
```suggestion
configure the plugin to invoke the `protoc` compiler. This is done by adding a `grpc-swift-config.json`
```

Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md
```diff
@@ -0,0 +1,113 @@
...
+to the root of your target's source folder. An example configuration file looks like this:
+
+```json
+{
+    "invocations": [
+        {
+            "protoFiles": [
+                "Path/To/Foo.proto",
+            ],
+            "visibility": "internal"
```
glbrntt, line 74: We should include server/client options too.

Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md
```diff
@@ -0,0 +1,113 @@
...
+    ]
+}
+
+```
+
+In the above configuration, you declared two invocations to the `protoc` compiler. The first invocation
+is generating Swift types for the `Foo.proto` file with `internal` visibility. The second invocation
+is generating Swift types for the `Bar.proto` file with the `public` visibility.
+
+> Note: paths to your `.proto` files will have to include the full path from the target source to the `.proto` file location.
```
glbrntt, line 91: To me "full path" means from `/`. I think we want the "relative path from the target source .proto.

Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md
```diff
@@ -0,0 +1,113 @@
+# Using the Swift Package Manager plugin
```
glbrntt, line 1: A lot of this is copied from the protobuf documentation, can we just link to their docs where possible?

glbrntt (changes requested):

Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md
FranzBusch: Oh you are right. This is great that means besides adding the `PROTOC_PATH` env variable we can also use the `PATH`. This is something we should document in both gRPC-swift and swift-protobuf. For the people that have protoc already in their `$PATH` it is way easier.

Package.swift
FranzBusch: I agree but it is kinda sad that this isn't aligned with the `SwiftProotobufPlugin` then

Plugins/SwiftGRPCPlugin/plugin.swift
```diff
@@ -0,0 +1,189 @@
...
+    // We need to find the configuration file at the root of the target
+    let configurationFilePath = target.directory.appending(subpath: Self.configurationFileName)
+    let data = try Data(contentsOf: URL(fileURLWithPath: "\(configurationFilePath)"))
+    let configuration = try JSONDecoder().decode(Configuration.self, from: data)
+
+    try self.validateConfiguration(configuration)
+
+    // We need to find the path of protoc and protoc-gen-swift
+    let protocPath: Path
+    if let configuredProtocPath = configuration.protocPath {
```
FranzBusch, line 82: This has changed on the latest swift-protobuf commit. If the env variable `PROTOC_PATH` is set we also try to respect that. Can you please take a look at how the latest implementation looks inside the protobuf repo and adopt it here.

Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md
FranzBusch: Yes we should include the most extensive config here

Package.swift
gjcairo: I don't mind changing it to `GRPCSwiftPlugin` (it makes more sense since the repo's called grpc-swift), but as Franz said, the reason why I went for SwiftGRPC was to keep it consistent with the `SwiftProtobufPlugin`. If we don't care about that I can change it.

Performance/QPSBenchmark/Package.swift
gjcairo: I tried cleaning and building several times and I got no warnings, that's why I decided to get rid of them.

FranzBusch: These files are not unused anymore since they are input files to the plugin invocation.

Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md
gjcairo: Thanks Tobias, I'll modify the docs slightly to include this option. We should also modify the Protobuf plugin docs.

Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md
gjcairo: Don't we prefer to have "self-contained" docs for each package/plugin? In theory this plugin and the protobuf plugin can evolve separately.
I don't mind making this change, but just curious. I've also ended up making a few changes in this PR that should be copied over (or moved) to the protobuf repo, so I should open a PR to `swift-protobuf`.

Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md
```diff
@@ -0,0 +1,118 @@
...
+There are two ways how this can be achieved. First, by default, the package manager looks into
+the `$PATH` to find binaries named `protoc`. This works immediately if you use `swift build` to build
+your package and `protoc` is installed in the `$PATH` (`brew` is adding it to your `$PATH` automatically).
+However, this may not work if you want to compile from Xcode since Xcode is not passed the `$PATH` by default.
+To still make this work from Xcode you can point the plugin to the concrete location of the `protoc`
+compiler by changing the configuration file like this:
+
+```json
+{
+    "protoCPath": "/path/to/protoc",
```
glbrntt, line 108: I think this should be `protocPath` as per `Configuration`?

Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md
glbrntt: I don't think we have a strong stance. I'm okay with leaving this as is. There's some value in not punting folks over to a different resource. The cost is diverging docs for common things (installing protoc) but we can bear that.

glbrntt (approved):
One nit, looks good otherwise. I'll let @FranzBusch also review as he has more expertise here.

Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md
gjcairo: That is correct, I'll change it.

glbrntt (approved):

Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md
```diff
@@ -0,0 +1,118 @@
...
+            "keepMethodCasing": false
+        }
+    ]
+}
+
+```
+
+In the above configuration, you declared two invocations to the `protoc` compiler. The first invocation
+is generating Swift types for the `Foo.proto` file with `internal` visibility. The second invocation
+is generating Swift types for the `Bar.proto` file with the `public` visibility.
```
FranzBusch, line 92: I think it would be good to add a sentence or two about the `client`, `server` and `keepMethodCasing` options that are set here.

Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md
```diff
@@ -0,0 +1,118 @@
...
+
+> Note: paths to your `.proto` files will have to include the relative path from the target source to the `.proto` file location. 
+
+### Defining the path to the protoc binary
```
FranzBusch, line 96: Can you please pull in the latest doc from `swift-protobuf` for this part of the documentation. It changed a bit since you started this work

FranzBusch (approved):
LGTM!

hudsontrius80A (approved):

hudsontrius80A (approved):

PR #1475, opened 2022-08-31
Add GRPC and protobuf SPM plugins to QPSBenchmark

gjcairo:
## Motivation

We want to make use of both the protobuf and GRPC codegen SPM plugins in QPSBenchmark

## Modifications

* Added support for both GRPC and protobuf plugin to QPSBenchmark

## Result

QPSBenchmark now autogenerates both proto and GRPC code when building.


Performance/QPSBenchmark/Sources/QPSBenchmark/grpc-swift-config.json
```diff
@@ -0,0 +1,12 @@
...
+  "invocations": [
+    {
+      "protoFiles": [
+        "Model/benchmark_service.proto",
+        "Model/worker_service.proto",
+      ],
+      "visibility": "public"
+    }
+  ],
+  "protocPath": "/usr/local/bin/protoc"
```
FranzBusch, line 11: I don't think we should define the path here. Did you do this for local or CI builds?

Performance/QPSBenchmark/Sources/QPSBenchmark/swift-protobuf-config.json
```diff
@@ -0,0 +1,15 @@
...
+        "Model/messages.proto",
+        "Model/control.proto",
+        "Model/core_stats.proto",
+        "Model/payloads.proto",
+        "Model/stats.proto"
+      ],
+      "visibility": "public"
+    }
+  ],
+  "protocPath": "/usr/local/bin/protoc"
```
FranzBusch, line 14: Same here

Performance/QPSBenchmark/Sources/QPSBenchmark/grpc-swift-config.json
gjcairo: Ah, didn't realise about this. I did it so it built locally, because otherwise it fails. CI has passed, I wonder if it would pass without the `protocPath` defined here?

gjcairo: It works fine locally launching with `xed`. I've removed the path - let's see if the CI fails. If it does, I still don't have write access to the repo (have asked @glbrntt) so I won't be able to edit the actions.

FranzBusch: I think the CI is going to work since it doesn't test the QPS package build. Furthermore, you can edit the actions by editing the yml in the `.github` folder at the root of the repo.

gjcairo: Indeed, it has passed :)

glbrntt (commented):
Awesome! Can you also remove the `Makefile` and update the `README.md`?

glbrntt (approved):

PR #1477, opened 2022-09-05
Adopt NIOAsyncSequenceProducer in grpc-swift

gjcairo:
## Motivation

We want to make use of `NIOAsyncSequenceProducer` (or its throwing counterpart) instead of using `PassthroughMessageSequence` and `PassthroughMessageSource`.

## Modifications

* Replaced usages of `PassthroughMessageSequence` and `PassthroughMessageSource` with `NIOThrowingAsyncSequenceProducer`

## Result

grpc-swift now uses `NIOAsyncSequenceProducer`


Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -422,7 +430,10 @@ internal final class AsyncServerHandler<
         contextProvider: self
       )
 
-      let requestSource = PassthroughMessageSource<Request, Error>()
+      let sequenceProducer = AsyncSequenceProducer.makeSequence(
+        backPressureStrategy: .init(lowWatermark: 0, highWatermark: 0),
```
glbrntt, line 434: We should set some values here.. the 10 and 50 as used in other places should be fine.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncRequestStream.swift
```diff
@@ -16,12 +16,19 @@
...
 
+import NIOCore
+
 /// This is currently a wrapper around AsyncThrowingStream because we want to be
 /// able to swap out the implementation for something else in the future.
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
 public struct GRPCAsyncRequestStream<Element: Sendable>: AsyncSequence {
   @usableFromInline
-  internal typealias _WrappedStream = PassthroughMessageSequence<Element, Error>
+  internal typealias _WrappedStream = NIOThrowingAsyncSequenceProducer<
```
FranzBusch, line 26: I think we should use the non-throwing variant here. The only thing we finish the source with is the `CancellationError` which I think is wrong to use here. `CancellationError`s should be used if the consuming `Task` gets cancelled; however, in our case we are finishing the source. @glbrntt WDYT?

gjcairo: [Discussed offline] We not only throw on source finish, but also expose any other errors (see `makeAndInvoke(call:)` in `GRPCAsyncBidirectionalStreamingCall` and `GRPCAsyncServerStreamingCall`); for this, we'll keep the throwing version of the sequence producer but stop finishing the source with `CancellationError` from the server side.

gjcairo:
Performance tests on 5.6 and 5.7 are failing but they're related to the change of `swift-nio` to `main`, and not related to other changes made as part of this PR. I'll spend some time looking into what could be the cause, and will either fix it in a follow-up PR or create a Radar to track for later.

FranzBusch (approved):
LGTM! Nice work! We just have to wait until a new release of NIO is cut to merge this.

glbrntt (approved):

PR #1482, opened 2022-09-08
Use forward looking SwiftProtobuf API

glbrntt:
Motivation:

SwiftProtobuf would like to remove some of the plugin library 
code in its next major release. We would like to allow gRPC to
rely on SwiftProtobuf in both major versions.

We use a few of the APIs marked for removal but can switch
away from them easily as replacement APIs have been backported
to the 1.x branch.

Modifications:

- Raise the minimum protobuf version
- Use the new APIs

Result:

- We can depend on protobuf 1.x and 2.x when it becomes available.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -48,7 +48,7 @@ let packageDependencies: [Package.Dependency] = [
   ),
   .package(
     url: "https://github.com/apple/swift-protobuf.git",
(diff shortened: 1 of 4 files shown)


FranzBusch (approved):
LGTM!

PR #1483, opened 2022-09-13
Raise minimum Swift version to 5.5

glbrntt:
Motivation:

Now that 5.7 has been released NIO is due to drop support for 5.4. gRPC Swift must do the same.

Modifications:

- Remove 5.4 from the CI (the nightly 5.7 CI has not been changed to 5.7 release as the Docker images are not yet available)
- Update Package.swift for the FuzzTesting subpackage to 5.7
- Update the README.md and Docc index page.

Result:

Minimum supported Swift version is 5.4


FuzzTesting/Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.4
+// swift-tools-version:5.7
```
FranzBusch, line 1: Do we need this change?

FranzBusch (approved):
LGTM, just the FuzzingPackage change that we might want to not do.

FuzzTesting/Package.swift
glbrntt: This change is okay: oss-fuzz recently updated their Swift image to be 5.7.

PR #1485, opened 2022-09-16
Expose public creation methods for `GRPCAsyncRequestStream` and `GRPCAsyncResponseStreamWriter`

FranzBusch:
# Motivation
It is highly desirable to be able to write tests against the generated method of a service. Currently, this is close to impossible since both the `GRPCAsyncRequestStream` and the `GRPCAsyncResponseStreamWriter` don't expose a public init so users cannot drive and observe the functions.

# Modification
Adds new public methods to drive and observe the request stream and the response writer.

# Result
We can now properly test gRPC generated server function implementations.

### DIFF
--- a/Sources/GRPC/AsyncAwaitSupport/GRPCAsyncRequestStream.swift
+++ b/Sources/GRPC/AsyncAwaitSupport/GRPCAsyncRequestStream.swift
@@ -21,40 +21,64 @@
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
 public struct GRPCAsyncRequestStream<Element: Sendable>: AsyncSequence {
   @usableFromInline
-  internal typealias _WrappedStream = PassthroughMessageSequence<Element, Error>
+  enum Backing: Sendable {
+    case passthroughMessageSequence(PassthroughMessageSequence<Element, Error>)
+    case asyncStream(AsyncThrowingStream<Element, Error>)
+  }
 
   @usableFromInline
-  internal let _stream: _WrappedStream
+  internal let backing: Backing
 
   @inlinable
-  internal init(_ stream: _WrappedStream) {
-    self._stream = stream
+  internal init(_ sequence: PassthroughMessageSequence<Element, Error>) {
+    self.backing = .passthroughMessageSequence(sequence)
+  }
+
+  @inlinable
+  public init(_ stream: AsyncThrowingStream<Element, Error>) {
+    self.backing = .asyncStream(stream)
   }
 
   @inlinable
   public func makeAsyncIterator() -> Iterator {
-    Self.AsyncIterator(self._stream)
+    switch self.backing {
+    case let .passthroughMessageSequence(sequence):
+      return Self.AsyncIterator(.passthroughMessageSequence(sequence.makeAsyncIterator()))
+    case let .asyncStream(stream):
+      return Self.AsyncIterator(.asyncStream(stream.makeAsyncIterator()))
+    }
   }
 
   public struct Iterator: AsyncIteratorProtocol {
     @usableFromInline
-    internal var iterator: _WrappedStream.AsyncIterator
+    enum BackingIterator {
+      case passthroughMessageSequence(PassthroughMessageSequence<Element, Error>.Iterator)
+      case asyncStream(AsyncThrowingStream<Element, Error>.Iterator)
+    }
 
     @usableFromInline
-    internal init(_ stream: _WrappedStream) {
-      self.iterator = stream.makeAsyncIterator()
+    internal var iterator: BackingIterator
+
+    @usableFromInline
+    internal init(_ iterator: BackingIterator) {
+      self.iterator = iterator
     }
 
     @inlinable
     public mutating func next() async throws -> Element? {
-      try await self.iterator.next()
+      switch self.iterator {
+      case let .passthroughMessageSequence(iterator):
+        return try await iterator.next()
+      case var .asyncStream(iterator):
+        let element = try await iterator.next()
+        self.iterator = .asyncStream(iterator)
+        return element
+      }
     }
   }
 }
 
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
 extension GRPCAsyncRequestStream: Sendable where Element: Sendable {}
-@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
-extension GRPCAsyncRequestStream.Iterator: Sendable where Element: Sendable {}
 
 #endif
--- a/Sources/GRPC/AsyncAwaitSupport/GRPCAsyncRequestStream.swift
+++ b/Sources/GRPC/AsyncAwaitSupport/GRPCAsyncRequestStream.swift
@@ -15,11 +15,66 @@
  */
 
 #if compiler(>=5.6)
-
-/// This is currently a wrapper around AsyncThrowingStream because we want to be
+/// A type for the stream of request messages send to a gRPC server method.
+///
+/// To enable testability this type provides a static ``GRPCAsyncRequestStream/makeTestingRequestStream()``
+/// method which allows you to create a stream that you can drive.
+///
+/// - Note: This is currently a wrapper around AsyncThrowingStream because we want to be
 /// able to swap out the implementation for something else in the future.
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
 public struct GRPCAsyncRequestStream<Element: Sendable>: AsyncSequence {
+  /// A source used for driving a ``GRPCAsyncRequestStream`` during tests.
+  public struct Source {
+    @usableFromInline
+    internal let continuation: AsyncThrowingStream<Element, Error>.Continuation
+
+    @inlinable
+    init(continuation: AsyncThrowingStream<Element, Error>.Continuation) {
+      self.continuation = continuation
+    }
+
+    /// Yields the element to the request stream.
+    ///
+    /// - Parameter element: The element to yield to the request stream.
+    @inlinable
+    public func yield(_ element: Element) {
+      self.continuation.yield(element)
+    }
+
+    /// Finished the request stream.
+    @inlinable
+    public func finish() {
+      self.continuation.finish()
+    }
+
+    /// Finished the request stream.
+    ///
+    /// - Parameter error: An optional `Error` to finish the request stream with.
+    @inlinable
+    public func finish(throwing error: Error?) {
+      self.continuation.finish(throwing: error)
+    }
+  }
+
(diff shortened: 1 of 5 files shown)


Sources/GRPC/AsyncAwaitSupport/GRPCAsyncResponseStreamWriter.swift
```diff
@@ -19,26 +19,90 @@
...
+
+      init(iterator: AsyncStream<(Response, Compression)>.AsyncIterator) {
+        self.iterator = iterator
+      }
+
+      public mutating func next() async -> Element? {
+        await self.iterator.next()
+      }
+    }
+  }
```
Lukasa, line 46: Can we get a bunch of `@inlinable`s here?

FranzBusch:
> Maybe I'm a bit lost, but what's the intended use of `RecordingWriter`?

Sure, maybe I should have put more context in the description. The problem that we have at the moment is that if you implement a method that writing a stream of responses and subsequently you want to write a test for your implementation then you really can't do this since you can't instantiate `GRPCAsyncResponseStreamWriter`. Furthermore, even if you can you want to actually assert which responses you wrote.

The goal of this PR is to make it possible to write unit tests for implementation of generated protocol methods. More specifically for the methods that have either request/response streams.

Lukasa:
Ok, that makes sense. Can you provide this as a detailed doc comment on the type? It'll make it much easier for users to discover the intended use of this object.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncResponseStreamWriter.swift
FranzBusch: I don't think this stuff needs @inlinable it should only be used in tests where performance is moot.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncRequestStream.swift
```diff
@@ -21,40 +21,64 @@
...
 
   @inlinable
-  internal init(_ stream: _WrappedStream) {
-    self._stream = stream
+  internal init(_ sequence: PassthroughMessageSequence<Element, Error>) {
+    self.backing = .passthroughMessageSequence(sequence)
+  }
+
+  @inlinable
+  public init(_ stream: AsyncThrowingStream<Element, Error>) {
```
glbrntt, line 38: Escaping the continuation from the stream is probably the most common way this will be used which is quite an obtuse API. 

I think we'd be better served by hiding `AsyncThrowingStream` from the API and having a `static` `func` returning a _wrapped_ `AsyncThrowingStream.Continuation`. Alternatively it could just be a wrapped `NIOAsyncSequenceProducer.Source` with a no-op delegate and no-op back pressure strategy (we could always swap them out later).

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncResponseStreamWriter.swift
```diff
@@ -19,26 +19,90 @@
 /// Writer for server-streaming RPC handlers to provide responses.
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
 public struct GRPCAsyncResponseStreamWriter<Response: Sendable>: Sendable {
+  public struct RecordingAsyncSequence: AsyncSequence {
```
glbrntt, line 22: I don't think 'recording' is right here: it doesn't keep all the requests it's seen. We should change this name (and elsewhere).

Tests/GRPCTests/AsyncAwaitSupport/GRPCAsyncRequestStreamTests.swift
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright 2021, gRPC Authors All rights reserved.
```
glbrntt, line 2:
```suggestion
 * Copyright 2022, gRPC Authors All rights reserved.
```

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncResponseStreamWriter.swift
```diff
@@ -19,26 +19,90 @@
...
+  }
+
+  /// Creates a new `GRPCAsyncResponseStreamWriter` backed by an ``RecordingAsyncSequence``.
+  /// This is mostly useful for testing purposes where one wants to observe the output of a method invocation.
+  public static func makeRecordingWriter() -> RecordingWriter {
+    var continuation: AsyncStream<(Response, Compression)>.Continuation!
+    let asyncStream = AsyncStream<(Response, Compression)> { cont in
+      continuation = cont
+    }
+    let writer = Self.init { [continuation] in
```
glbrntt, line 101: There's an awkward wrinkle in this code: we never finish the async stream. gRPC does this internally on the async writer (IIRC) after the user method exits. We need to rework this to do it here too.

glbrntt (changes requested):

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncResponseStreamWriter.swift
FranzBusch: I don't know if I fully agree on this. The only way to really finish the sequence of writes would be to to __move__ the writer into the user function and implement a deinit on the writer to finish the sequence. However, I don't think we can spell this right now.
Furthermore, I don't think we necessarily need to finish the sequence. In your tests, you should be able to use `prefix()` and `collect()` to assert how many writes have been issued.
Open to other ideas how you think we should finish this sequence!

glbrntt: My concern is that it's non-obvious that without using `prefix()` or collecting a fixed number of messages that iterating over the sequence will never terminate. Not being able to safely iterate an async sequence is a serious API flaw IMHO.

---

With the end goal of making services testable I think we may need to look at a bigger picture, bearing in mind we also need to provide a context to the caller.

I wonder if we can do something like providing a service wrapper/context/thing which can provide the right "bits" to the caller and also deal with finishing the stream. I'd want to play around with ideas some more to flesh this out but it could be something like:

```swift
struct TestingContext {
    var logger: Logger

    init(logger: Logger) {
        self.logger = logger
    }

    func withRPCContext<Request: Sendable, Response: Sendable>(
        execute function: (Request, GRPCAsyncServerCallContext) async throws -> Response
    ) async throws -> Response {
        fatalError()
    }

    func withRPCContext<Request: Sendable, Response: Sendable>(
        requestType: Request.Type,
        responseType: Response.Type,
        execute function: (GRPCAsyncRequestStream<Request>, GRPCAsyncServerCallContext) async throws -> Response,
        writingRequests: (WrappedAsyncStreamContinuation<Request>) -> Void
    ) async throws -> Response {
        fatalError()
    }

    func withRPCContext<Request: Sendable, Response: Sendable>(
        request: Request,
        responseType: Response.Type,
        execute function: (Request, GRPCAsyncResponseStreamWriter<Response>.AsyncSequence, GRPCAsyncServerCallContext) async throws -> Void
    ) async throws -> GRPCAsyncResponseStreamWriter<Response> {
        fatalError()
    }

    func withRPCContext<Request: Sendable, Response: Sendable>(
        requestType: Request.Type,
        responseType: Response.Type,
        execute function: (GRPCAsyncRequestStream<Request>, GRPCAsyncResponseStreamWriter<Response>..AsyncSequence, GRPCAsyncServerCallContext) async throws -> Void,
        readingAndWriting: (WrappedAsyncStreamContinuation<Request>, GRPCAsyncResponseStreamWriter<Response>) async throws -> Void
    ) async throws {
        fatalError()
    }
}
```

The idea being to have it execute the function and provide the tester with the appropriate pieces to safely test the function. This should also include things like setting request headers and so on (although I haven't included that here). Rough example:

```swift
let context = TestingContext(logger: Logger(label: "foo"))
let echo = EchoAsyncProvider()

let echoResponse = try await context.withRPCContext(
    request: .with { $0.text = "hello" },
    execute: echo.get(request:context:)
)

try await context.withRPCContext(
    requestType: Echo_EchoRequest.self,
    responseType: Echo_EchoResponse.self,
    execute: echo.update(requestStream:responseStream:context:)
) { requests, responses in
        group.addTask {
            requests.write(...)
        }

        group.addTask {
            for try await response in responses {
                print(response)
            }
        }

        try await group.waitForAll()
    }
}
```

FranzBusch: While I fully agree with what you are laying out here, I think I am working on a slightly different but related corner of the testability. I have misphrased my motivation above a bit. My goal is not to test a method implementation per-se which I agree we need a more full blown solution like you pointed out since we need to solve the context and more.

However, my goal is to make both the `GRPCAsyncRequestStream` and the `GRPCAsyncResponseWriter` testable. You can imagine a scenario where you take the writer and pass it further along in your system and you wanna make these methods testable. However, due to the fact that the writer is not testable as of this moment it becomes impossible.

I see two ways out here, either we make the writer testable i.e. by surfacing a never ending `AsyncSequence` from it. Or we say this type is not testable and we require users to re-wrap the writer if they wanna pass it along and make it testable. The latter seems somewhat wrong to me since we would force a lot of adopters to write exactly the same code for testability sake.

On another note, I think that the writer test `AsyncSequence` is not ending is totally fine since the API that the writer exposes is just `write(_:)` so it _must_ produce an infinite sequence. Infinite sequences are quite normal e.g. `CurrentValueSubject<Int, Never>` from Combine. Let me know what you think!

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncResponseStreamWriter.swift
```diff
@@ -17,28 +17,116 @@
...
+      }
+    }
+  }
+
+  /// Simple struct for the return type of ``GRPCAsyncResponseStreamWriter/makeResponseStreamWriter()``.
+  ///
+  /// This struct contains two properties:
+  /// 1. The ``writer`` which is the actual ``GRPCAsyncResponseStreamWriter`` and should be passed to the method under testing.
+  /// 2. The ``stream`` which can be used to observe the written responses.
+  public struct ResponseStreamWriter {
```
glbrntt, line 63: Can we make the naming a little clearer that it's for testing? The names on the request side are 👌

FranzBusch: Yeah slightly awkward naming these types. I ended up with `TestingStreamWriter` and `makeTestingResponseStreamWriter()` which aligns with the request side

glbrntt (approved):

PR #1488, opened 2022-09-20
Clarify relative paths for SPM plugin docs

### LINKED ISSUE #1487: Clarify meaning of relative paths SPM plugin 'protoFiles'
In the [SPM plugin documentation](https://github.com/grpc/grpc-swift/blob/main/Sources/protoc-gen-grpc-swift/Docs.docc/spm-plugin.md), the 'protoFiles' are listed as relative to the target source directory.  This is only partially true (as least as I think of relative paths).

Consider

"protoFiles" : [
    "Local.proto",
    "Model/Child.proto",
    "../Parent.proto",
    "../Common/Sibling.proto"
]

The first two work perfectly.  The later two do not appear to be supported.

Would it be possible to add a note to the documentation to prevent any confusion by others?

glbrntt:
Resolves #1487


FranzBusch (approved):

Jerry-Carter:
Thanks.

PR #1489, opened 2022-09-21
updated NaturalLanguage example to Xcode 14 and latest package versions

filipw:
The current code would not run after "fresh clone".

The PR seems huge but the actual manual changes are only in `main.swift`.


glbrntt (approved):
Ah great, thank you!

glbrntt:
@filipw could you run `./scripts/format.sh` and commit the changes?

filipw:
oops - apologies! done

PR #1490, opened 2022-09-21
Update swift-nio version, allocation test max values and alloc-limits.sh

gjcairo:
## Motivation
This change updates the allocation limits for `grpc-swift` allocation tests. This is necessary because we’ve increased the number of allocations in the new version of `swift-nio` for some of the code paths the tests hit. This change also fixes some warnings as a result of the NIO update.

## Modifications
* Updated max allocation values for allocation tests
* Updated the `alloc-limits.sh` script to work with the output of the GitHub CI logs, which include timestamps. Output from running the script locally doesn’t contain timestamps (and still works).
* Modified the allocation tests to use `Echo_EchoNIOClient` instead of the now deprecated `Echo_EchoClient`
* Fixed warnings related to NIO's `Lock` being deprecated.

## Result
`swift-nio` has been updated to 2.42.0, allocation tests have updated allocation counts, and warnings related to the update fixed.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -56,24 +56,24 @@ jobs:
         include:
           - image: swiftlang/swift:nightly-focal
             env:
-              MAX_ALLOCS_ALLOWED_bidi_1k_rpcs_10_requests: 429000
-              MAX_ALLOCS_ALLOWED_bidi_1k_rpcs_1_request: 177000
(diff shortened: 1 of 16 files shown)


glbrntt (commented):
Looks great but we obviously can't approve and merge this until the next NIO release.

glbrntt (approved):

PR #1496, opened 2022-10-01
Remove invalid command line args from the example

lkstc112233:
These two args has been removed since 1.8 and is on by default. Passing in these two args will result in the grpc stub not being generated at all.


Lukasa (approved):
Thanks, this looks good to me! Can you please sign the CLA?

lkstc112233:
Hey Lukasa, thank you for your review! I have signed the SLA.

PR #1499, opened 2022-10-07
Allow to write a sequence of requests/responses

FranzBusch:
# Motivation
We recently adopted the `NIOAsyncWriter` to back the `GRPCAsyncRequestStreamWriter` and the `GRPCAsyncResponseStreamWriter`; however, we did not expose the functionality to write a sequence of the elements that the `NIOAsyncWriter` offers. This can be useful in cases where you want to write a batch of requests/responses since it reduces the amount of locks and thread hops.

# Modification
Expose new methods on the `GRPCAsyncRequestStreamWriter` and the `GRPCAsyncResponseStreamWriter` to enable to write a sequence. I also fixed up the comments for the `GRPCAsyncRequestStreamWriter` since they were outdated.

# Result
Users can write a batch of requests/responses.

### DIFF
--- a/Sources/GRPC/AsyncAwaitSupport/GRPCAsyncRequestStreamWriter.swift
+++ b/Sources/GRPC/AsyncAwaitSupport/GRPCAsyncRequestStreamWriter.swift
@@ -49,12 +49,7 @@ public struct GRPCAsyncRequestStreamWriter<Request: Sendable>: Sendable {
 
   /// Send a single request.
   ///
-  /// To ensure requests are delivered in order callers should `await` the result of this call
-  /// before sending another request. Callers who do not need this guarantee do not have to `await`
-  /// the completion of this call and may send messages concurrently from multiple ``Task``s.
-  /// However, it is important to note that no more than 16 writes may be pending at any one time
-  /// and attempting to exceed this will result in an ``GRPCAsyncWriterError/tooManyPendingWrites``
-  /// error being thrown.
+  /// It is safe to send multiple requests concurrently by sharing the ``GRPCAsyncRequestStreamWriter`` across tasks.
   ///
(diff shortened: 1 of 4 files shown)


Sources/GRPC/AsyncAwaitSupport/GRPCAsyncRequestStreamWriter.swift
```diff
@@ -72,6 +66,22 @@ public struct GRPCAsyncRequestStreamWriter<Request: Sendable>: Sendable {
...
+  ///
+  /// - Parameters:
+  ///   - contentsOf: The requests and their compression to send.
+  /// - Throws: If the request stream has already been finished.
+  @inlinable
+  public func send<S: Sequence>(
+    contentsOf requests: S
+  ) async throws where S.Element == (Request, Compression) {
+    try await self.asyncWriter.yield(contentsOf: requests)
+  }
```
FranzBusch, line 83: @Lukasa @fabianfett What do you think about using the tuple here? Do you prefer creating a new type like  `MessageWithCompression`?

fabianfett: Fine either way. But generally structs are the better choice for evolving types and APIs. Name should probably be `MessageAndCompression`? With sounds like the message was already compressed.

FranzBusch: Actually I changed the API a bit for now where you can write a sequence of requests/responses and set the compression for all of them. If users ever want to write a sequence with different compression per request/response we can still add that API. WDYT?

Lukasa (approved):

Lukasa (changes requested):
Can we get a unit test to prove this works?

PR #1508, opened 2022-10-19
Log TLS version

dnadoba:
### Motivation
For debugging purposes it is useful to know which TLS version is actually used by gRPC. 
Therefore we should log the TLS version after the TLS handshake completed.

### Modification
- add a method to get the TLS either from Network.framework or NIOSSL
- get the TLS version after the TLS handshake completed and debug log it

### Result
TLS version is logged

Note that this PR depends on changes in NIOTS which are not yet released


glbrntt:
Continued in #1509

PR #1510, opened 2022-10-21
Auto extract IDs from headers on the server

glbrntt:
Motivation:

It's common for service authors to extract a tracing ID from a request header and attach that value to the logger for an RPC. This only requires a few lines of code but must be done in each RPC. gRPC should provide configuration to do this automatically.

Modifications:

- Add server configuration which extracts a value from headers and attaches it to a logger at the start of each RPC.
- Wire up the config. Add tests.

Result:

Trace IDs can be automatically extracted from headers on the server and attached to loggers.


Sources/GRPC/Server.swift
```diff
@@ -451,6 +454,22 @@ extension Server {
   }
 }
 
+extension Server.Configuration {
+  /// Configuration for extracting IDs from metadata and inserting them into a logger.
+  public struct TraceIDExtractor {
```
Lukasa, line 459: Is this sufficiently flexible? Do we need to allow the user to define a type that can do arbitrary searches?

glbrntt: Good question. I suspect not, I can imagine a situation where we could fallback to a different header field for example.

Sources/GRPC/Server.swift
```diff
@@ -451,6 +454,38 @@ extension Server {
...
+
+    /// The key for the logging metadata to set with the extracted trace ID.
+    public var loggerKey: String
+
+    /// Create a new trace ID extractor.
+    ///
+    /// - Parameters:
+    ///   - loggerKey: The key in the logger to set the trace ID on.
+    ///   - extractor: A function to extract a trace ID from
+    public init(loggerKey: String, extractor: @escaping (HPACKHeaders) -> String?) {
```
Lukasa, line 471: Also `@Sendable`, right?

FranzBusch (commented):
In general I love this! I have been doing the exact same thing but needed to extract in each individual method which was kinda annoying.

I am just wondering if we should extend this a bit further to allow more things to get extracted to metadata. Something like this `(HPACKHeaders) -> [String]` or even `(HPACKHeaders, Logger) -> Logger`?

My thinking behind this is that there might be more that needs to get logged than just a trace id. Sometimes you want to log the parent id, span id, trace flags etc. https://www.w3.org/TR/trace-context/#traceparent-header

Maybe @ktoso could chime in here as well.

Sources/GRPC/Server.swift
ktoso: Whoa wait a moment guys -- this is reinventing tracing infrastructure and types. Such extract/inject things are exactly what a Tracer (or other Instrument (a Tracer is-a Instrument)) would be doing. So rather, grpc should instrument itself using the tracing instruments, and make use of those to populate baggage, which then is fed into logging.

Here's the basic API: https://github.com/apple/swift-distributed-tracing/blob/main/Sources/Instrumentation/Instrument.swift

We can set aside a moment to talk this through this week if you want @glbrntt? It'd be really valuable to use the tracing types here; The only holdoff from 1.0 them all is projects like this having a serious look at it, as well as the swift-log discussion we're having right now about propagation, so it'd be really a waste to invent your own "extractor" types right now.

ktoso:
Chimed in in-line - this is exactly what swift-distributed-tracing types are for, so I think we need to catch up a bit, maybe tomorrow?

ktoso:
I don't know why but I couldn't PR "to" this PR branch :-)
Either way, here's how little work it takes to achieve the same with the distributed tracing APIs: https://github.com/ktoso/grpc-swift/commit/cba3cf85ab179df783a635b400dcfce482b41aaf

Let's figure out how to quickly stabilize them so we don't have to do the ad-hoc implementation in here 👍 

We also have the 2 year old PoC over here which even did spans: https://github.com/grpc/grpc-swift/pull/941 so between those two branches I'm sure we can find a way to land the proper solution nicely – and get the tracing 1.0 out there ASAP so this lib can depend on it 👍

ktoso:
CLA signed and all other approvals done. Let's get collaboratin~ 🥳

Package.swift
```diff
@@ -51,19 +51,25 @@ let packageDependencies: [Package.Dependency] = [
     from: "1.20.2"
   ),
   .package(
-    url: "https://github.com/apple/swift-log.git",
-    from: "1.4.4"
+    // url: "https://github.com/apple/swift-log.git",
+    // from: "1.4.4"
+    url: "https://github.com/slashmo/swift-log.git",
+    branch: "feature/baggage"
```
ktoso, line 57: This is now merged and available in 1.5.x

PR #1515, opened 2022-10-26
Add a connection pool delegate

### LINKED ISSUE #1400: Provide connection pool metrics
As a developer using grpc-swift, I would love to instrument my outgoing grpc connections with [swift-metrics](https://github.com/apple/swift-metrics). Since connections are mostly handled by grpc-swift, grpc-swift should expose those metrics.

Metrics I would like to export:

1. How many open connections in a given pool
2. How many streams on a given connection
3. Connect failures and types for a pool
4. Connect latency for a pool

### Describe the solution you'd like

grpc-swift should accept a `MetricsDelegate` on grpc connection pool creation. If the pool state changes the delegate should be informed. The protocol could look something like this. First raw draft.

```swift

protocol GRPCChannelPoolMetricsDelegate {
  func connectionCreated(connectionID: Int, latency: Int)

  func connectionUtilizationChanged(connectionID: Int, available: Int, used: Int)

  func connectionClosed(connectionID: Int, error: Error?)

  func connectFailure(error: Error)
}
```

### Describe alternatives you've considered

GRPC could export metrics directly to `swift-metrics`. However this would add a new dependency, that some adopters might not want. Further adopters may want to report the metrics differently and we want to give them options. Offering the connection metrics through a protocol should give adopters the most flexibility.

glbrntt:
Motivation:

It can be useful to observe what the underlying connection pool is doing over time. This could, for example, be instrumenting it to understand how many streams are active at a given time or to understand when a connection is in the processing of being brought up.

Modifications:

Add new API:
- `GRPCConnectionPoolDelegate` which users can implement and configure on the `GRPCChannelPool` to receive notifications about the connection pool state.
- `GRPCConnectionID`, an opaque ID to distinguish between different connections in the pool.

Modify the connection pool:
- Per-EventLoop pools now hold on to quiescing connections and track additional state, including whether a connection is quiescing and how many streams are currently open on a connection. By contrast the pool manager (which manages a number of these per-EventLoop pools) tracks reserved streams (which are not necessarily open). The delegate tracks _opened_ streams rather than _reserved_ streams (as reserved streams are not allocated to a connection but to an any connection running on an appropriate event-loop).
- Wire through the new delegate and call out to it in appropriate places.
- Add a stream opened function to the internal H2 delegate
- Expose various pieces of state from the connection manager

Result:

- Users can instrument the underlying connection pool.
- Resolves #1400

### DIFF
--- a/Sources/GRPC/ConnectionManager.swift
+++ b/Sources/GRPC/ConnectionManager.swift
@@ -242,12 +242,34 @@ internal final class ConnectionManager {
     }
   }
 
+  /// Returns whether the state is 'shutdown'.
+  private var isShutdown: Bool {
+    self.eventLoop.assertInEventLoop()
+    switch self.state {
+    case .shutdown:
+      return true
+    case .idle, .connecting, .transientFailure, .active, .ready:
+      return false
+    }
+  }
+
   /// Returns the `HTTP2StreamMultiplexer` from the 'ready' state or `nil` if it is not available.
   private var multiplexer: HTTP2StreamMultiplexer? {
     self.eventLoop.assertInEventLoop()
     switch self.state {
     case let .ready(state):
       return state.multiplexer
+
+    case .idle, .connecting, .transientFailure, .active, .shutdown:
+      return nil
+    }
+  }
+
+  private var channel: Channel? {
+    self.eventLoop.assertInEventLoop()
+    switch self.state {
+    case let .ready(state):
+      return state.channel
 
     case .idle, .connecting, .transientFailure, .active, .shutdown:
       return nil
@@ -807,6 +829,11 @@ internal final class ConnectionManager {
     }
   }
 
+  internal func streamOpened() {
+    self.eventLoop.assertInEventLoop()
+    self.http2Delegate?.streamOpened(self)
+  }
+
   internal func streamClosed() {
     self.eventLoop.assertInEventLoop()
     self.http2Delegate?.streamClosed(self)
@@ -1001,10 +1028,21 @@ extension ConnectionManager {
       return self.manager.isIdle
     }
 
+    /// Returne `true` if the connection is in the shutdown state.
+    internal var isShutdown: Bool {
+      return self.manager.isShutdown
+    }
+
     /// Returns the `multiplexer` from a connection in the `ready` state or `nil` if it is any
     /// other state.
     internal var multiplexer: HTTP2StreamMultiplexer? {
       return self.manager.multiplexer
+    }
+
+    /// Returns the `channel` from a connection in the `ready` state or `nil` if it is any
+    /// other state.
+    internal var channel: Channel? {
+      return self.manager.channel
     }
 
     // Start establishing a connection. Must only be called when `isIdle` is `true`.
--- a/Sources/GRPC/ConnectionManager.swift
+++ b/Sources/GRPC/ConnectionManager.swift
@@ -265,17 +265,6 @@ internal final class ConnectionManager {
     }
   }
 
-  private var channel: Channel? {
-    self.eventLoop.assertInEventLoop()
-    switch self.state {
-    case let .ready(state):
-      return state.channel
-
-    case .idle, .connecting, .transientFailure, .active, .shutdown:
-      return nil
-    }
-  }
-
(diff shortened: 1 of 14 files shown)


Sources/GRPC/ConnectionPool/ConnectionPool.swift
```diff
@@ -502,15 +509,30 @@ internal final class ConnectionPool {
...
-        $0.manager.shutdown(mode: mode)
+      let allShutdown: [EventLoopFuture<Void>] = connections.values.map {
+        let id = $0.manager.id
+        let manager = $0.manager
+
+        return manager.eventLoop.flatSubmit {
+          // If the connection was idle/shutdown before calling shutdown then we shouldn't tell
+          // the delegate the connection closed (because it either never connected or was already
+          // informed about this).
+          let connectionIsInactive = manager.sync.isIdle || manager.sync.isShutdown
```
Lukasa, line 528: Is this check too early? Is it possible for us to quiesce between calling `shutdown` and the future completing?

Sources/GRPC/ConnectionPool/ConnectionPool.swift
```diff
@@ -564,27 +586,73 @@ extension ConnectionPool: ConnectionManagerConnectivityDelegate {
...
+    guard let index = self._connections.index(forKey: manager.id) else { return }
+    self._connections.values[index].isQuiescing = true
+    self._connections.values[index].manager.sync.connectivityDelegate = nil
+
+    self.delegate?.connectionQuiescing(id: .init(manager.id))
+
+    // If there's a close future then the underlying channel is open. It will close eventually when
+    // open streams are closed, so drop the H2 delegate and update the pool delegate when that
+    // happens.
+    if let closeFuture = self._connections.values[index].manager.sync.channel?.closeFuture {
```
Lukasa, line 629: Holy Law of Demeter batman! We absolutely need a clearer spelling for this idea.

Sources/GRPC/ConnectionPool/ConnectionPool.swift
glbrntt: Good question, are you getting at the possibility of registering callbacks to the delegate multiple times?

- Call `shutdown()`
- Hit this line, register callbacks to the delegate
- Server tells us to quiesce before future completes
- Code handling quiescing registers callbacks to the delegate

If so that shouldn't happen: we drop all connection state a few lines above (L510/L518) and the quiescing code exits early if the connection state isn't present.

Sources/GRPC/ConnectionPool/ConnectionPool.swift
glbrntt: Agreed. It's certainly more fiddly than I'd like and as it turns out also a little over protective: we will be notified about quiescing only if there are open streams (if there aren't then the connection will just be idled). So the `else` branch here is redundant. I think API for registering an on-close callback on the current connection would probably be sufficient here.

Sources/GRPC/ConnectionManager.swift
```diff
@@ -582,6 +593,31 @@ internal final class ConnectionManager {
     }
   }
 
+  /// Registers a callback which fires when the current active connection is closed.
+  ///
+  /// If there is a connection, the callback will be invoked with `true` when the connection is
+  /// closed. Otherwise the callback is invoked with `false`.
+  internal func onCurrentConnectionClose(_ onClose: @escaping (Bool) -> Void) {
```
Lukasa, line 600: This is a somewhat tricky API to use. Is there a reason it doesn't return whether the callback got registered?

glbrntt: We need to hop onto the right EL first which makes this slightly awkward. A few options:
 - let the caller provide a promise / return a future
 - expose this via the `sync` view and return whether it was registered (could be in addition to above)
 - use a lock in favour of the EL for synchronising here
 - pass whether the callback was registered into the callback (current api)

I don't think any of these stand out as significantly better choices for an internal API. WDYT?

Lukasa: Yeah, that's fair. I dunno man, it's all kinda gross. This works I suppose.

Lukasa (approved):

PR #1518, opened 2022-11-03
Use #fileID/#filePath instead of #file

carolinacass:
Motivation:

#fileID introduced in Swift 5.3, so no longer need to use #file anywhere                                  

Modifications:

Changed #file to #filePath or #fileID depending on the situation

### DIFF
--- a/Sources/Examples/RouteGuide/Client/RouteGuideClient.swift
+++ b/Sources/Examples/RouteGuide/Client/RouteGuideClient.swift
@@ -23,7 +23,7 @@ import RouteGuideModel
 
 /// Loads the features from `route_guide_db.json`, assumed to be in the directory above this file.
 func loadFeatures() throws -> [Routeguide_Feature] {
-  let url = URL(fileURLWithPath: #fileID)
+  let url = URL(fileURLWithPath: #filePath)
     .deletingLastPathComponent() // main.swift
     .deletingLastPathComponent() // Client/
     .appendingPathComponent("route_guide_db.json")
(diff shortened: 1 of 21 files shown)


Sources/Examples/RouteGuide/Client/RouteGuideClient.swift
```diff
@@ -23,7 +23,7 @@ import RouteGuideModel
 
 /// Loads the features from `route_guide_db.json`, assumed to be in the directory above this file.
 func loadFeatures() throws -> [Routeguide_Feature] {
-  let url = URL(fileURLWithPath: #file)
+  let url = URL(fileURLWithPath: #fileID)
```
glbrntt, line 26: This one should be `#filePath` as we're finding a file relative to this source file.

```suggestion
  let url = URL(fileURLWithPath: #filePath)
```

Sources/Examples/RouteGuide/Server/RouteGuideServer.swift
```diff
@@ -24,7 +24,7 @@ import RouteGuideModel
 
 /// Loads the features from `route_guide_db.json`, assumed to be in the directory above this file.
 func loadFeatures() throws -> [Routeguide_Feature] {
-  let url = URL(fileURLWithPath: #file)
+  let url = URL(fileURLWithPath: #fileID)
```
glbrntt, line 27: Same here: 
```suggestion
  let url = URL(fileURLWithPath: #filePath)
```

glbrntt (changes requested):

glbrntt:
Thanks @carolinacass!

Sources/Examples/RouteGuide/Server/RouteGuideServer.swift
Tharolzakariya: Withdraw

PR #1524, opened 2022-11-15
Deploy plugins to `buf` registry

julienfouilhe:
Still a draft, Github Actions job has been tested locally only yet.

- Adds a Dockerfile to package protoc plugins into a docker image so that they can be pushed to the buf registry.
- Adds CI steps to automatically push them to the BSR on new releases.

See buf documentation on [Remote generation](https://docs.buf.build/bsr/remote-generation/overview)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,3 +5,4 @@ FuzzTesting
 Performance
 .github
 .build
+protoc-gen-*
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+docs
+dev
+Examples
+FuzzTesting
+Performance
+.github
+# .build
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,4 +4,4 @@ Examples
 FuzzTesting
 Performance
 .github
-# .build
+.build
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,3 +25,46 @@ jobs:
         uses: softprops/action-gh-release@v1
         with:
           files: protoc-grpc-swift-plugins-linux-x86_64-${{ env.RELEASE_VERSION }}.zip
+
+  buf:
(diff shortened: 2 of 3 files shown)


FranzBusch:
Hey @julienfouilhe,

Thanks for opening this PR. In general, I love what the folks over at Buf are doing in the proto environment. I would love to get a bit more of the bigger picture here though. To what organization are you intending to publish these plugins here? Is this going to end up in the official supported languages of Buf?

FranzBusch:
> I needed these plugins for my personal use, so I first thought of publishing them to my personal organization.

Great that's what I assumed from the PR!

> Is this something that can be imagined?
I think what you are proposing here isn't working since organisations are nothing public on Buf as far as I am aware (maybe that changed since I last looked at it)

> Sorry I can't answer this question, I guess only the Buf team can tell if that's possible?
I think that's the key piece for this. In its current state, I don't think we can use the PR here; however, the Buf team might be able to use your work to setup an official language support. I would recommend reaching out to them in their Slack and see what they think about it. If they need any help from us let them know that we are happy to help!

julienfouilhe:
@FranzBusch Organizations are public, I use a lot of unofficial buf plugins: [prost](https://buf.build/prost/plugins) for instance! 🎉

FranzBusch:
@julienfouilhe Oh that's interesting. We would need somebody from the large gRPC org to setup that organisation probably. In the meantime, I would recommend that you just host the plugins in one of your public orgs.

W.r.t. the PR, I think we have to hold off with this a bit.

julienfouilhe:
I asked the Buf team and they told me they are already [curating a list of plugins](https://github.com/bufbuild/plugins) that they automatically release to the registry. It hasn't been publicly released yet so I didn't know.

`grpc-swift` is part of those plugins, so this PR is useless! Thanks anyway

PR #1529, opened 2022-12-07
Avoid copies of large payloads

glbrntt:
Motivation:

Messages are prefixed with a 5-byte header. Currently messages of all sizes are written into a new buffer with their header. For smaller payloads this is good: we avoid the extra allocations associated with creating HTTP/2 frames. For large payloads the cost of the copy outweighs the cost of extra allocations.

Modifications:

- For messages larger than 8KB emit an extra HTTP/2 DATA frame containing just the message header.

Result:

Better performance for large payloads.

### DIFF
--- a/Sources/GRPC/GRPCClientChannelHandler.swift
+++ b/Sources/GRPC/GRPCClientChannelHandler.swift
@@ -530,15 +530,26 @@ extension GRPCClientChannelHandler: ChannelOutboundHandler {
         allocator: context.channel.allocator
       )
       switch result {
-      case let .success(buffer):
-        // We're clear to send a message; wrap it up in an HTTP/2 frame.
-        let framePayload = HTTP2Frame.FramePayload.data(.init(data: .byteBuffer(buffer)))
+      case let .success((buffer, maybeBuffer)):
+        let frame1 = HTTP2Frame.FramePayload.data(.init(data: .byteBuffer(buffer)))
         self.logger.trace("writing HTTP2 frame", metadata: [
           MetadataKey.h2Payload: "DATA",
           MetadataKey.h2DataBytes: "\(buffer.readableBytes)",
           MetadataKey.h2EndStream: "false",
         ])
-        context.write(self.wrapOutboundOut(framePayload), promise: promise)
+        // If there's a second buffer, attach the promise to the second write.
+        let promise1 = maybeBuffer == nil ? promise : nil
+        context.write(self.wrapOutboundOut(frame1), promise: promise1)
+
+        if let maybeBuffer = maybeBuffer {
+          let frame2 = HTTP2Frame.FramePayload.data(.init(data: .byteBuffer(maybeBuffer)))
+          self.logger.trace("writing HTTP2 frame", metadata: [
+            MetadataKey.h2Payload: "DATA",
+            MetadataKey.h2DataBytes: "\(maybeBuffer.readableBytes)",
+            MetadataKey.h2EndStream: "false",
+          ])
+          context.write(self.wrapOutboundOut(frame2), promise: promise)
+        }
 
       case let .failure(writeError):
         switch writeError {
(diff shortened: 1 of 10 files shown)


Sources/GRPC/LengthPrefixedMessageWriter.swift
```diff
@@ -86,99 +86,42 @@ internal struct LengthPrefixedMessageWriter {
...
+      let compressedAndFramedPayload = try self.compress(
+        buffer: buffer,
+        using: compressor,
+        allocator: allocator
+      )
+      return (compressedAndFramedPayload, nil)
+    } else if buffer.readableBytes > Self.singleBufferSizeLimit {
+      // Buffer is larger than the limit for emitting a single buffer: create a second buffer
+      // containing just the message header.
+      var prefixed = allocator.buffer(capacity: 5)
```
Lukasa, line 100: I wonder if we can be even cleverer with this and re-use the same fixed buffer.

Sources/GRPC/GRPCClientChannelHandler.swift
```diff
@@ -530,15 +530,26 @@ extension GRPCClientChannelHandler: ChannelOutboundHandler {
...
           MetadataKey.h2DataBytes: "\(buffer.readableBytes)",
           MetadataKey.h2EndStream: "false",
         ])
-        context.write(self.wrapOutboundOut(framePayload), promise: promise)
+        // If there's a second buffer, attach the promise to the second write.
+        let promise1 = maybeBuffer == nil ? promise : nil
+        context.write(self.wrapOutboundOut(frame1), promise: promise1)
+
+        if let actuallyBuffer = maybeBuffer {
+          let frame2 = HTTP2Frame.FramePayload.data(.init(data: .byteBuffer(maybeBuffer)))
```
glbrntt, line 545:
```suggestion
          let frame2 = HTTP2Frame.FramePayload.data(.init(data: .byteBuffer(actuallyBuffer)))
```

Sources/GRPC/GRPCClientChannelHandler.swift
```diff
@@ -530,15 +530,26 @@ extension GRPCClientChannelHandler: ChannelOutboundHandler {
...
+          self.logger.trace("writing HTTP2 frame", metadata: [
+            MetadataKey.h2Payload: "DATA",
+            MetadataKey.h2DataBytes: "\(maybeBuffer.readableBytes)",
```
glbrntt, line 548:
```suggestion
            MetadataKey.h2DataBytes: "\(actuallyBuffer.readableBytes)",
```

Sources/GRPC/HTTP2ToRawGRPCServerCodec.swift
```diff
@@ -295,9 +295,16 @@ internal final class HTTP2ToRawGRPCServerCodec: ChannelInboundHandler, GRPCServe
...
 
     switch writeBuffer {
-    case let .success(buffer):
-      let payload = HTTP2Frame.FramePayload.data(.init(data: .byteBuffer(buffer)))
-      self.context.write(self.wrapOutboundOut(payload), promise: promise)
+    case let .success((buffer, maybeBuffer)):
+      if let actuallyBuffer = maybeBuffer {
+        let payload1 = HTTP2Frame.FramePayload.data(.init(data: .byteBuffer(buffer)))
+        self.context.write(self.wrapOutboundOut(payload1), promise: nil)
+        let payload2 = HTTP2Frame.FramePayload.data(.init(data: .byteBuffer(maybeBuffer)))
```
glbrntt, line 302:
```suggestion
        let payload2 = HTTP2Frame.FramePayload.data(.init(data: .byteBuffer(actuallyBuffer)))
```

Sources/GRPC/LengthPrefixedMessageWriter.swift
glbrntt: Done in https://github.com/grpc/grpc-swift/pull/1529/commits/f3f17a2ce03f811bbcb88082217cb3b8fe97aeef

Lukasa (approved):

PR #1530, opened 2022-12-12
Avoid CoWs

glbrntt:
We didn't actually avoid a CoW here. We also don't need the 'modifying' workaround; the compiler is smarter now.

### DIFF
--- a/Sources/GRPC/HTTP2ToRawGRPCStateMachine.swift
+++ b/Sources/GRPC/HTTP2ToRawGRPCStateMachine.swift
@@ -21,23 +21,6 @@ import NIOHTTP2
 struct HTTP2ToRawGRPCStateMachine {
   /// The current state.
   private var state: State = .requestIdleResponseIdle
-
-  /// Temporarily sets `self.state` to `._modifying` before calling the provided block and setting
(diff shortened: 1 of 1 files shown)


Sources/GRPC/HTTP2ToRawGRPCStateMachine.swift
```diff
@@ -21,23 +21,6 @@ import NIOHTTP2
...
-
-  /// Temporarily sets `self.state` to `._modifying` before calling the provided block and setting
-  /// `self.state` to the `State` modified by the block.
-  ///
-  /// Since we hold state as associated data on our `State` enum, any modification to that state
-  /// will trigger a copy on write for its heap allocated data. Temporarily setting the `self.state`
-  /// to `._modifying` allows us to avoid an extra reference to any heap allocated data and
-  /// therefore avoid a copy on write.
-  @inlinable
-  internal mutating func withStateAvoidingCoWs<Action>(_ body: (inout State) -> Action) -> Action {
```
FranzBusch, line 33: Well we tried 😛

FranzBusch (approved):

PR #1531, opened 2022-12-13
Delay client async writer starting

glbrntt:
Motivation:

It's possible for async streaming clients to fail and drop messages in some situations. The situation leading to this happens because streaming calls set up state and then write out headers. While setting up state the HTTP/2 stream channel is configured, when it becomes active gRPC calls outs to enable the async writer to start emitting writes. This can happen before the headers are written so if a write is already pending then it can race the headers being written. If the message is written first then the write promise is failed and the message is dropped.

Modifications:

Delay letting the async writer emit writes until the headers have been written.

Result:

Correct ordering is enforced.


FranzBusch (approved):
Very nice catch! What an annoying 🐛

PR #1537, opened 2022-12-20
Add a coalescing writer

glbrntt:
Motivation:

Creating frames in HTTP/2 isn't free: they incur one allocation for the frame payload and another for the data buffer. In addition to this HTTP/2 currently creates a promise per write, so we also have an additional write promise. On top of the allocations we have the repeated work for processing extra frames.

We can recover much of this cost by coalescing writes before a flush. For streaming RPCs this can have a fairly significant impact.

Modifications:

- Add a coalescing length prefixed writer: serialized messages are appended and length prefixing (and compression) is done lazily on calls to `next()`.
- Add an internal 'one-or-many' queue which is a FIFO queue backed by either a single element on the stack or a deque. This avoids allocating a queue on the heap for the length prefixed writer when there is only one message appended before flushing.

Result:

Nothing changes yet, but we have a writer in place which we can adopt later.

### DIFF
--- a/Sources/GRPC/CoalescingLengthPrefixedMessageWriter.swift
+++ b/Sources/GRPC/CoalescingLengthPrefixedMessageWriter.swift
@@ -1,0 +1,357 @@
+/*
+ * Copyright 2022, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


FranzBusch (approved):
LGTM! Can't wait to see how it slots into the current writing path.

Issue #1540, opened 2023-01-02
build error with grpc-swift dependency


oliverepper:
I am trying to build a project of mine that depends on grpc-swift and get a lot of these errors:

error: value of type 'ChannelPipeline.SynchronousOperations' has no member 'nioSSL_tlsVersion'
`return try self.pipeline.syncOperations.nioSSL_tlsVersion().map(GRPCTLSVersion.init)`

This happens while depending on version 1.13.1. At the moment I have no idea what I could change.

dnadoba:
You need a newer version of `swift-nio`. We forgot to update the minimum required swift version in this PR: https://github.com/grpc/grpc-swift/pull/1509

For now you can just update to the latest swift-nio version to resolve the issue manually.

However, we should update our Package.swift too.

oliverepper:
Thanks for your help! Shall I make a PR for this? Or is there anything special about the release process? If it's not fixed upstream I need to fix this in my branch, anyways. It needed updating of swift-nio-ssl, btw.

Lukasa:
A PR for this would be very welcome.

oliverepper:
For now the PR would only be the update of the version of the swift-nio-ssl dependency. I could go through all of the dependencies and try updating them but right now I have trouble at runtime. I just want to make sure that the trouble is not related. This is the error I get:

`[...]/Sources/NIOExtras/QuiescingHelper.swift:222: Fatal error: leaking promise created at (file: "/MicroSwitch/.build/checkouts/swift-nio-extras/Sources/NIOExtras/QuiescingHelper.swift", line: 222)`

I'll try updating the other nio dependencies as well.

Lukasa:
We should only need to uplift the swift-nio-ssl dependency.

The issue you're seeing there is likely a follow-on from hitting an error condition: that runtime issue is unlikely to be fixed by lifting a NIO version. Did you see that issue after lifting the swift-nio-ssl dependency?

oliverepper:
Oh. The error happens on Linux, only. MacOS is fine.

Lukasa:
This crash is likely masking an error being thrown, which is causing the promise to leak. Can you produce a complete stack at the point of the crash?

oliverepper:
I think I got it. Let me double check and if correct I will prepare the small PR for the new nio-ssl dependency.

oliverepper:
Anyways here is the trace:

```pre
[...]

PR #1541, opened 2023-01-03
update swift-nio-ssl dependency

oliverepper:
I needed to lift this version number to be able to build grpc-swift.


Lukasa (approved):

Lukasa:
Resolves #1540.

PR #1564, opened 2023-02-08
Buffer in the server pipeline configuration

glbrntt:
Motivation:

When not using TLS, the server pipeline configurator inspects the first bytes on a connection to determine whether HTTP1 or HTTP2 is being used and closes the connection if it is determined that neither are. It does this by only parsing the first packet, which may not have enough bytes to make a correct determination.

Modifications:

- Buffer bytes in the configurator.
- Parse the buffered bytes and only close if enough bytes have been received.

Result:

Better version determination.


FranzBusch (approved):
Thanks for fixing this!

PR #1570, opened 2023-02-27
Nil out the `onStart` callback after using it

glbrntt:
Motivation:

The `ClientTransport` has an `onStart` callback which is called once the stream has been established. For async calls this is used to notify the writability manager that it may now start writing. The callback references the call which holds the transport which holds the callback forming a strong retain cycle and a slow memory leak. The transport should break this cycle.

Modifications:

- nil out the `onStart` callback after it has been called.

Result:

Fixes a leak


fabianfett (approved):
Awesome fix! Thanks so much!

glbrntt:
The main-nightly failure is https://github.com/apple/swift/issues/63934

PR #1573, opened 2023-03-03
Don't trap on invalid connection state transitions

glbrntt:
Motivation:

The connection manager is quite aggresive about trapping if an invalid state is hit. It's alsmost impossible to know when states are truly unreachable so in most coses we should handle them as best as we can. If we believe them to be unreachable but cannot easily prove it then we should crash in debug mode and handle it as best as possible in release mode.

Modifications:

- Handle various state transitions more gently in the connection manager.

Result:

Gentler state handling.

### DIFF
--- a/Sources/GRPC/ConnectionManager.swift
+++ b/Sources/GRPC/ConnectionManager.swift
@@ -77,11 +77,14 @@ internal final class ConnectionManager {
     var scheduled: Scheduled<Void>
     var reason: Error
 
-    init(from state: ConnectingState, scheduled: Scheduled<Void>, reason: Error) {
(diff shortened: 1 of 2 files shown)


Lukasa (approved):

bzeeman:
Can I just say that this is not as gentle for anyone using this as it is for people writing this? Assertions left in code where developers that aren't familiar with its inner workings are less helpful as discardable errors or even a throwing function.

Lukasa:
These are intended to be loud because hitting them reveals critical bugs in either your code or ours. We believe hitting these transitions should be _impossible_, so if they aren't then we want to know about it.

PR #1583, opened 2023-04-04
Allow CORS to be configured for gRPC Web

glbrntt:
Motivation:

The WebCORS handler unconditionally sets "Access-Control-Allow-Origin" to "*" in response headers regardless of whether the request is a CORS request or whether the client sends credentials. Moreover we don't expose any knobs to control how CORS is configured.

Modifications:

- Add CORS configuration to the server and server builder
- Let the allowed origins be '.any' (i.e. '*") or '.only' (limited to the provided origins)
- Let the user configure what headers are permitted in responses.
- Let the user configure whether credentialed requests are accepted.

Result:

More control over CORS

### DIFF
--- a/Sources/GRPC/Server.swift
+++ b/Sources/GRPC/Server.swift
@@ -456,7 +456,7 @@ extension Server.Configuration {
   public struct CORS: Hashable, GRPCSendable {
     /// Determines which 'origin' header field values are permitted in a CORS request.
     public var allowedOrigins: AllowedOrigins
-    /// Sets the headaers which are permitted in a response to a CORS request.
+    /// Sets the headers which are permitted in a response to a CORS request.
     public var allowedHeaders: [String]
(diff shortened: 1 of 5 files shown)


Sources/GRPC/Server.swift
```diff
@@ -451,6 +452,55 @@ extension Server {
   }
 }
 
+extension Server.Configuration {
+  public struct CORS: Hashable, GRPCSendable {
+    /// Determines which 'origin' header field values are permitted in a CORS request.
+    public var allowedOrigins: AllowedOrigins
+    /// Sets the headaers which are permitted in a response to a CORS request.
```
Lukasa, line 459:
```suggestion
    /// Sets the headers which are permitted in a response to a CORS request.
```

Lukasa (approved):

PR #1585, opened 2023-04-11
Drop support for Swift 5.5

glbrntt:
Motivation:

Now that 5.8 has been released we can drop support for Swift 5.5.

Modifications:

- Update CI jobs
- Remove conditional branches which are no longer supported
- Remove Package@swift-5.5.swift
- Use `Sendable` in place of `GRPCSendable`
- Update docs

Result:

Swift 5.6 is the lowest supported Swift version


Sources/GRPC/AsyncAwaitSupport/GRPCSendable.swift
```diff
@@ -13,25 +13,14 @@
...
+@available(*, deprecated, renamed: "Swift.Sendable")
 public typealias GRPCSendable = Swift.Sendable
-#else
-public typealias GRPCSendable = Any
-#endif // compiler(>=5.6)
 
-#if compiler(>=5.6)
+@available(*, deprecated, message: "Use @preconcurrency Swift.Sendable")
 @preconcurrency
 public protocol GRPCPreconcurrencySendable: Sendable {}
```
FranzBusch, line 23: We should deprecate this but we can't remove the inheritance from the types sadly since this is a `public protocol` and users might have done weird things based off that inheritance

FranzBusch (changes requested):

Sources/GRPC/AsyncAwaitSupport/GRPCSendable.swift
glbrntt: We can't deprecate this without warnings in our own code but I've switched back to using it again.

FranzBusch (approved):
LGTM!

PR #1587, opened 2023-04-14
Adopt h2handler multiplexer

rnro:
Motivation:

Switch from `HTTP2StreamMultiplexer` to `HTTP2Handler.StreamMultiplexer` to benefit from improved performance due to reduced allocations.

Modifications:

- Replace all references to  `HTTP2StreamMultiplexer` with `HTTP2Handler.StreamMultiplexer`.
- `GRPCIdleHandler` is now a `NIOHTTP2StreamDelegate` to allow it to receive notifications when a stream is created or closed rather than using `UserInboundEvent`s.
- `GRPCIdleHandler` now has two states of configuration in the client case. This is required to break a cycle of dependencies which would otherwise exist at `init` because the `GRPCIdleHandler` holds a reference to the `HTTP2Handler.StreamMultiplexer` (which is a member of the `HTTP2Handler` itself) and the `HTTP2Handler` holds a reference back to the `GRPCIdleHandler` as its stream delegate.
- Several tests now insert `HTTP2Handler` into the pipeline where they used to use `HTTP2StreamMultiplexer`. This causes a few changes in behavior often around different assertions. This required some small supporting changes.
- Introduce test infrastructure `AsyncEventStreamConnectionPoolDelegate` to eliminate observed racey behavior in `EventRecordingConnectionPoolDelegate`.

Result:

Performance increase without overall behavior changes.

### DIFF
--- a/Sources/GRPC/GRPCServerPipelineConfigurator.swift
+++ b/Sources/GRPC/GRPCServerPipelineConfigurator.swift
@@ -80,7 +80,7 @@ final class GRPCServerPipelineConfigurator: ChannelInboundHandler, RemovableChan
   /// Makes an HTTP/2 handler.
   private func makeHTTP2Handler(
     for channel: Channel,
-    streamDelegate: NIOHTTP2StreamDelegate? = nil
+    streamDelegate: NIOHTTP2StreamDelegate?
   ) -> NIOHTTP2Handler {
     var connectionConfiguration = NIOHTTP2Handler.ConnectionConfiguration()
     connectionConfiguration.initialSettings = [
--- a/Sources/GRPC/GRPCServerPipelineConfigurator.swift
+++ b/Sources/GRPC/GRPCServerPipelineConfigurator.swift
@@ -78,7 +78,10 @@ final class GRPCServerPipelineConfigurator: ChannelInboundHandler, RemovableChan
   }
 
   /// Makes an HTTP/2 handler.
-  private func makeHTTP2Handler(for channel: Channel, streamDelegate: NIOHTTP2StreamDelegate? = nil) -> NIOHTTP2Handler {
+  private func makeHTTP2Handler(
+    for channel: Channel,
+    streamDelegate: NIOHTTP2StreamDelegate? = nil
+  ) -> NIOHTTP2Handler {
     var connectionConfiguration = NIOHTTP2Handler.ConnectionConfiguration()
     connectionConfiguration.initialSettings = [
       HTTP2Setting(
--- a/Sources/GRPC/GRPCServerPipelineConfigurator.swift
+++ b/Sources/GRPC/GRPCServerPipelineConfigurator.swift
@@ -78,38 +78,36 @@ final class GRPCServerPipelineConfigurator: ChannelInboundHandler, RemovableChan
   }
 
   /// Makes an HTTP/2 handler.
-  private func makeHTTP2Handler() -> NIOHTTP2Handler {
-    return .init(
-      mode: .server,
-      initialSettings: [
-        HTTP2Setting(
-          parameter: .maxConcurrentStreams,
-          value: self.configuration.httpMaxConcurrentStreams
-        ),
-        HTTP2Setting(
-          parameter: .maxHeaderListSize,
-          value: HPACKDecoder.defaultMaxHeaderListSize
-        ),
-        HTTP2Setting(
-          parameter: .maxFrameSize,
-          value: self.configuration.httpMaxFrameSize
-        ),
-        HTTP2Setting(
-          parameter: .initialWindowSize,
-          value: self.configuration.httpTargetWindowSize
-        ),
-      ]
-    )
-  }
+  private func makeHTTP2Handler(for channel: Channel, streamDelegate: NIOHTTP2StreamDelegate? = nil) -> NIOHTTP2Handler {
+    var connectionConfiguration = NIOHTTP2Handler.ConnectionConfiguration()
+    connectionConfiguration.initialSettings = [
+      HTTP2Setting(
+        parameter: .maxConcurrentStreams,
+        value: self.configuration.httpMaxConcurrentStreams
+      ),
+      HTTP2Setting(
+        parameter: .maxHeaderListSize,
+        value: HPACKDecoder.defaultMaxHeaderListSize
+      ),
+      HTTP2Setting(
+        parameter: .maxFrameSize,
+        value: self.configuration.httpMaxFrameSize
+      ),
+      HTTP2Setting(
+        parameter: .initialWindowSize,
+        value: self.configuration.httpTargetWindowSize
+      ),
+    ]
 
-  /// Makes an HTTP/2 multiplexer suitable handling gRPC requests.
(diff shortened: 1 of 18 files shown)


Sources/GRPC/GRPCServerPipelineConfigurator.swift
```diff
@@ -78,38 +78,39 @@ final class GRPCServerPipelineConfigurator: ChannelInboundHandler, RemovableChan
...
-      ]
-    )
-  }
-
-  /// Makes an HTTP/2 multiplexer suitable handling gRPC requests.
-  private func makeHTTP2Multiplexer(for channel: Channel) -> HTTP2StreamMultiplexer {
-    var logger = self.configuration.logger
+  private func makeHTTP2Handler(
+    for channel: Channel,
+    streamDelegate: NIOHTTP2StreamDelegate? = nil
```
glbrntt, line 83: For internal/private functions I tend to avoid defaults, it avoids forgetting to pass in values.
 
```suggestion
    streamDelegate: NIOHTTP2StreamDelegate?
```

Tests/GRPCTests/ClientTimeoutTests.swift
```diff
@@ -169,7 +169,7 @@ extension EmbeddedGRPCChannel: @unchecked Sendable {}
 
 private final class EmbeddedGRPCChannel: GRPCChannel {
   let embeddedChannel: EmbeddedChannel
-  let multiplexer: EventLoopFuture<HTTP2StreamMultiplexer>
+  let h2Handler: EventLoopFuture<NIOHTTP2Handler>
```
glbrntt, line 172: Let's change this so we store the `NIOHTTP2Handler.StreamMultiplexer`, we can grab it when we configure the channel below.

Tests/GRPCTests/ConnectionPool/ConnectionPoolDelegates.swift
```diff
@@ -13,9 +13,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import Dispatch
 import GRPC
 import NIOConcurrencyHelpers
 import NIOCore
+import XCTest
```
glbrntt, line 20: Looks like the new imports are unused?

Tests/GRPCTests/ConnectionManagerTests.swift
```diff
@@ -239,6 +244,24 @@ extension ConnectionManagerTests {
...
+  ///
+  /// This is useful in tests where we intentionally mis-use the channels
+  /// and call `fireChannelInactive` manually during the test but don't want
+  /// teardown to cause precondition failures due to this unexpected behavior.
+  class SwallowSecondInactiveHandler: ChannelInboundHandler {
+    typealias InboundIn = HTTP2Frame
+    typealias OutboundOut = HTTP2Frame
+
+    var seenAnInactive = false
+    public func channelInactive(context: ChannelHandlerContext) {
```
glbrntt, line 257:
```suggestion
    private var seenAnInactive = false
    func channelInactive(context: ChannelHandlerContext) {
```

Tests/GRPCTests/ConnectionManagerTests.swift
```diff
@@ -491,15 +520,16 @@ extension ConnectionManagerTests {
...
+    let readyChannelMux: EventLoopFuture<NIOHTTP2Handler.StreamMultiplexer> = self
+      .waitForStateChanges([
+        Change(from: .idle, to: .connecting),
+        Change(from: .connecting, to: .transientFailure),
+      ]) {
+        // Get a HTTP/2 stream multiplexer.
+        let readyChannelMux = manager.getHTTP2Multiplexer()
+        self.loop.run()
+        return readyChannelMux
+      }
```
glbrntt, line 532: The formatter did a really bad job here...

```suggestion
    let readyChannelMux = self.waitForStateChanges([
        Change(from: .idle, to: .connecting),
        Change(from: .connecting, to: .transientFailure),
    ]) { () -> EventLoopFuture<NIOHTTP2Handler.StreamMultiplexer> in
      // Get a HTTP/2 stream multiplexer.
      let readyChannelMux = manager.getHTTP2Multiplexer()
      self.loop.run()
      return readyChannelMux
    }
```

Tests/GRPCTests/ConnectionManagerTests.swift
```diff
@@ -592,15 +625,16 @@ extension ConnectionManagerTests {
...
```
glbrntt, line 637: Same here

```suggestion
    let readyChannelMux = self.waitForStateChanges([
      Change(from: .idle, to: .connecting),
      Change(from: .connecting, to: .transientFailure),
    ]) { () -> EventLoopFuture<NIOHTTP2Handler.StreamMultiplexer>  in 
      // Get a HTTP/2 stream multiplexer.
      let readyChannelMux = manager.getHTTP2Multiplexer()
      self.loop.run()
      return readyChannelMux
    }
```

Tests/GRPCTests/ConnectionManagerTests.swift
```diff
@@ -1169,16 +1210,26 @@ extension ConnectionManagerTests {
 
     let http2 = HTTP2Delegate()
 
+    var streamDelegate: NIOHTTP2StreamDelegate?
```
glbrntt, line 1213: We shouldn't escape the delegate from the channel provider closure here.

Instead, when we need it later on, we can grab the `GRPCIdleHandler` from the channel and call it directly.

Tests/GRPCTests/ConnectionPool/ConnectionPoolTests.swift
```diff
@@ -17,7 +17,7 @@
 import Logging
 import NIOCore
 import NIOEmbedded
-import NIOHTTP2
+@testable import NIOHTTP2
```
glbrntt, line 20: We can't do this: it gives us access to the internals of NIOHTTP2 which is verboten.

Tests/GRPCTests/ConnectionPool/ConnectionPoolTests.swift
```diff
@@ -1010,17 +1010,26 @@ extension ConnectionPool {
...
 
   internal var count: Int {
     return self.channels.count
   }
 
   internal func finish() {
     while let channel = self.channels.popLast() {
+      if !channel.isActive {
+        continue
+      }
```
glbrntt, line 1030: Let's simplify this a little:

```swift
while let state = self.channels.popLast() {
  if state.channel.isActive {
    _ = try? state.channel.finish()
  }
}
```

glbrntt (changes requested):
This looks really good already! I've left a few comments.

Tests/GRPCTests/ConnectionPool/ConnectionPoolTests.swift
rnro: Removing this required copying over a simplified version of the HTTP/2 frame encoder.

glbrntt (commented):
This more or less looks good to go. I left one nit and some of the allocation counts need updating and we'll need to get an http/2 release tagged too.

glbrntt (approved):

PR #1594, opened 2023-05-02
Allow for more CORS configuration

glbrntt:
Motivation:

We added some level of CORS configuration support in #1583. This change adds further flexibility.

Modifications:

- Add an 'originBased' mode where the value of the origin header is returned in the response head.
- Add a custom fallback where the user can specify a callback which is passed the value of the origin header and returns the value to return in the 'access-control-allow-origin' response header (or nil, if the origin is not allowed).

Result:

More flexibility for CORS.

### DIFF
--- a/Sources/GRPC/Server.swift
+++ b/Sources/GRPC/Server.swift
@@ -479,7 +479,9 @@ extension Server.Configuration.CORS {
   public struct AllowedOrigins: Hashable, Sendable {
     enum Wrapped: Hashable, Sendable {
       case all
+      case originBased
       case only([String])
+      case custom(AnyCustomCORSAllowedOrigin)
     }
 
     private(set) var wrapped: Wrapped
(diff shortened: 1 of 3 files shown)


Sources/GRPC/Server.swift
```diff
@@ -490,10 +527,25 @@ extension Server.Configuration.CORS {
...
     }
+
+    /// Provide a custom CORS origin check.
+    ///
+    /// - Parameter checkOrigin: A closure which is called with the value of the 'origin' header
+    ///     and returns the value to use in the 'access-control-allow-origin' response header,
+    ///     or `nil` if the origin is not allowed.
+    public static func custom(
+      _ checkOrigin: @escaping @Sendable (_ origin: String) -> String?
+    ) -> Self {
```
Lukasa, line 546: Given the requirement for sendability and hashability, should this take a generic type or something equivalent to enable you to maintain the invariant?

glbrntt: Hmm, we'd need to store it as an existential which means we'd get the automatic `Hashable` conformance but not the `Equatable` conformance (the stored `any P`s could be of different types). It's also mildly inconvenient for callers who are implementing that type. My lean is towards not making this generic.

Lukasa: We can get the equatable and hashable conformances using `AnyHashable` and more generally can achieve all of this using closures. The question is whether it's a good idea to keep types that don't really behave the way we want them to, I think.

Sources/GRPC/Server.swift
```diff
@@ -520,3 +535,46 @@ extension Comparable {
...
+  ///     CORS related headers should be returned.
+  func check(origin: String) -> String?
+}
+
+extension Server.Configuration.CORS.AllowedOrigins {
+  struct AnyCustomCORSAllowedOrigin: GRPCCustomCORSAllowedOrigin {
+    private var checkOrigin: @Sendable (String) -> String?
+    private let hashInto: @Sendable (inout Hasher) -> Void
+    #if swift(>=5.7)
+    private let isEqualTo: @Sendable (any GRPCCustomCORSAllowedOrigin) -> Bool
```
dnadoba, line 554: Do we get warnings if we don't use `any` with new swift compilers?

Lukasa (approved):

PR #1599, opened 2023-05-24
Allow server handlers to send response headers directly

glbrntt:
Motivation:

The async server call context allows users to set headers which are sent when the first message is sent. In many cases this is fine, however, some use cases require the headers to be sent immediately.

Modifications:

- Add `sendHeaders(_:)` to the `GRPCAsyncServerCallContext` which sends headers to the client and throws if headers have already been written or it's too late to send them.

Result:

Headers can be sent directly from a server call handler.

### DIFF
--- a/Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
+++ b/Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
@@ -30,6 +30,21 @@ public struct GRPCAsyncServerCallContext: Sendable {
     Response(contextProvider: self.contextProvider)
   }
 
+  /// Notifies the client that the RPC has been accepted for processing by the server.
+  ///
+  /// On accepting the RPC the server will send the given headers (which may be empty) along with
+  /// any transport specific headers (such the ":status" pseudo header) to the client.
+  ///
+  /// It is not necessary to call this function: the RPC is implicitly accepted when the first
+  /// response message is sent, however this may be useful when clients require an early indication
+  /// that the RPC has been accepted.
+  ///
+  /// If the RPC has already been accepted (either implicitly or explicitly) then this function is
+  /// a no-op.
+  public func acceptRPC(headers: HPACKHeaders) async {
+    await self.contextProvider.acceptRPC(headers)
+  }
+
   /// Access the ``UserInfo`` dictionary which is shared with the interceptor contexts for this RPC.
   ///
   /// - Important: While ``UserInfo`` has value-semantics, this function accesses a reference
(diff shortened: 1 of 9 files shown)


Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
```diff
@@ -747,11 +760,44 @@ extension AsyncServerHandler: AsyncServerCallContextProvider {
...
+        case .cancel:
+          return self.cancel(error: nil)
+        case .drop:
+          ()
+        }
+
+      case .intercept(.none), .drop:
+        // intercept(.none) means headers have already been sent; we should never hit this because
+        // we guard on setting the response headers above.
+        ()
```
FranzBusch, line 795: If we never should hit this should we precondition here?

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
```diff
@@ -30,6 +30,21 @@ public struct GRPCAsyncServerCallContext: Sendable {
...
+  /// On accepting the RPC the server will send the given headers (which may be empty) along with
+  /// any transport specific headers (such the ":status" pseudo header) to the client.
+  ///
+  /// It is not necessary to call this function: the RPC is implicitly accepted when the first
+  /// response message is sent, however this may be useful when clients require an early indication
+  /// that the RPC has been accepted.
+  ///
+  /// If the RPC has already been accepted (either implicitly or explicitly) then this function is
+  /// a no-op.
+  public func acceptRPC(headers: HPACKHeaders) async {
```
FranzBusch, line 44: What happens if I call `setHeaders` after using `acceptRPC`? Is `setHeaders` going to throw or just silently fail?

FranzBusch (commented):
Overall LGTM. Just two smaller comments

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerCallContext.swift
glbrntt: It'll throw.

Sources/GRPC/AsyncAwaitSupport/GRPCAsyncServerHandler.swift
glbrntt: I think it's a bit aggressive here: there's no harm in us reaching and ignoring this state.

FranzBusch: Should we call this out in the docs?

FranzBusch (approved):
LGTM!

Lukasa (approved):

PR #1605, opened 2023-05-31
Improve Sendable checking for server code

glbrntt:
Motivation:

The server handler protocol must be Sendable so that any methods on the handler are also Sendable as annotating methods as `@Sendable` does not work as expected. It follows from this that generated server interceptor factories must also be Sendable (they already are for clients). This puts the `ServerInterceptor` class in an awkward position: it must be Sendable but is not inherently thread-safe (these restrictions are documented and existed before Sendable checking was introduced to Swift). The `ClientInterceptor` has the same restrictions and is `@unchecked Sendable` so we elect to do the same for the `ServerInterceptor`.

Modifications:

- Make generated server handlers and server interceptor factories Sendable
- Make `ServerInterceptor` `@unchecked Sendable`
- Regenerate
- Fix some warnings in test code

Result:

Better Sendable checking

### DIFF
--- a/Sources/GRPC/Interceptor/ServerInterceptors.swift
+++ b/Sources/GRPC/Interceptor/ServerInterceptors.swift
@@ -42,7 +42,7 @@ import NIOCore
 /// require any extra attention. However, if work is done on a `DispatchQueue` or _other_
 /// `EventLoop` then implementers should ensure that they use `context` from the correct
 /// `EventLoop`.
-open class ServerInterceptor<Request, Response> {
+open class ServerInterceptor<Request, Response>: @unchecked Sendable {
   public init() {}
 
   /// Called when the interceptor has received a request part to handle.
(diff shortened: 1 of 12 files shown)


Sources/GRPC/Interceptor/ServerInterceptors.swift
```diff
@@ -42,7 +42,7 @@ import NIOCore
 /// require any extra attention. However, if work is done on a `DispatchQueue` or _other_
 /// `EventLoop` then implementers should ensure that they use `context` from the correct
 /// `EventLoop`.
-open class ServerInterceptor<Request, Response> {
+open class ServerInterceptor<Request, Response>: @unchecked Sendable {
```
FranzBusch, line 45: Can we actually do this? Making an `open class` @unchecked Sendable requires that all subclasses are also `Sendable` which they might not be.

FranzBusch (commented):
Overall LGTM just one comment around API breakage

Sources/GRPC/Interceptor/ServerInterceptors.swift
glbrntt: I'm not sure what other options we have: any subclasses of this _must_ be thread safe (as documented), we just didn't enforce it previously...

FranzBusch (approved):

Sources/GRPC/Interceptor/ServerInterceptors.swift
FranzBusch: Discussed this offline: Any subclass must be `Sendable` and this will just add warnings to adopters so not breaking any API

PR #1606, opened 2023-05-31
Emit 'swift-format-ignore-file' in generated files to avoid having them reformatted by swift-format

smikhanov:
Generator now emits a comment line in the generated code disabling processing of the generated file by the `swift-format` formatter. This is to allow checking in generated files into codebases that are regularly linted and reformatted using `swift-format`, without generating a change in generated files every time listing is done.


glbrntt (commented):
Thanks @smikhanov. 

Looks like this is failing the copyright header check, I think you'll need to update https://github.com/grpc/grpc-swift/blob/6b55ce088ba842522e2719d8d5404c354fd7bedf/scripts/license-check.sh#L100 so that `drop_first=9`.

glbrntt (approved):

PR #1621, opened 2023-06-23
Add support for `XcodeProjectPlugin`

denil-ct:
Carrying over the changes from https://github.com/apple/swift-protobuf/pull/1408

Can anyone point me on how to test these changes for grpc as I'm not very familiar with grpc.

### DIFF
--- a/Plugins/GRPCSwiftPlugin/plugin.swift
+++ b/Plugins/GRPCSwiftPlugin/plugin.swift
@@ -223,7 +223,7 @@ extension GRPCSwiftPlugin: BuildToolPlugin {
     guard let swiftTarget = target as? SwiftSourceModuleTarget else {
       throw PluginError.invalidTarget
     }
-    return try createBuildCommands(
+    return try self.createBuildCommands(
       pluginWorkDirectory: context.pluginWorkDirectory,
       sourceFiles: swiftTarget.sourceFiles,
       tool: context.tool
@@ -239,7 +239,7 @@ extension GRPCSwiftPlugin: XcodeBuildToolPlugin {
     context: XcodePluginContext,
     target: XcodeTarget
   ) throws -> [Command] {
-    return try createBuildCommands(
+    return try self.createBuildCommands(
       pluginWorkDirectory: context.pluginWorkDirectory,
       sourceFiles: target.inputFiles,
       tool: context.tool
--- a/Plugins/GRPCSwiftPlugin/plugin.swift
+++ b/Plugins/GRPCSwiftPlugin/plugin.swift
@@ -18,13 +18,15 @@ import Foundation
 import PackagePlugin
 
 @main
-struct GRPCSwiftPlugin: BuildToolPlugin {
+struct GRPCSwiftPlugin {
   /// Errors thrown by the `GRPCSwiftPlugin`
   enum PluginError: Error {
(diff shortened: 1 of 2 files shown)


FranzBusch (commented):
Overall looks good to me. I just realised that we should update the documentation in both repos though. We should add two things:
1. That now the config doesn't need to be at the root of the target anymore since we just search through the sources of a target
2. We should state that how it works for an Xcode project.

denil-ct:
> That now the config doesn't need to be at the root of the target anymore since we just search through the sources of a target

One thing to note here is that the target directory is calculated by deleting the last component of the config. So whatever folder the config resides in would become the target directory, and the file paths mentioned in the config should be relative to that.

FranzBusch:
> One thing to note here is that the target directory is now calculated by deleting the last component of the config. So whatever folder the config resides in would become the target directory, and the file paths mentioned in the config should be relative to that.

That's actually a very important call out! Thanks for stating this. We should clarify this in the Docc docs for the plugins. This is particularly important for the Xcode plugin since there the files can be all over the place.

glbrntt:
@denil-ct could you bring over the changes from protobuf if you get a chance? We'd like to do a release soon and it would be nice to include this change 🙂

denil-ct:
Hey @glbrntt , got a bit busy. Will make the changes by tomorrow.

glbrntt:
No worries, thank you!

denil-ct:
Hey @glbrntt , I'm facing some troubles verifying my changes.

I have this proto file
```proto
syntax = "proto3";

service NoteService {
    rpc Get (NoteRequestId) returns (Note) {}
}

message Note {
    string id = 1;
    string title = 2;
    string content = 3;
}

message NoteRequestId {
    string id = 1;
}
```

But  after the `Note.grpc.swift` file gets generated the message types like `NoteRequestId` and `Note` are not in scope.

The plugin tries to execute this command
```sh
/opt/homebrew/bin/protoc --plugin=protoc-gen-grpc-swift=/${BUILD_DIR}/${CONFIGURATION}/protoc-gen-grpc-swift" --grpc-swift_out=/Users/denilct/Library/Developer/Xcode/DerivedData/ABCApp-gkihnhqrytzfbsbcqflpkdwowxum/SourcePackages/plugins/ABCApp.output/ABCApp/GRPCSwiftPlugin -I /Users/denilct/Documents/ABCApp_2/ABCApp_2 --grpc-swift_opt=Visibility=Public --grpc-swift_opt=Server=false --grpc-swift_opt=Client=true ProtoFiles/Notes/Note.proto
```

glbrntt:
Are you running the protobuf plugin too? The grpc-swift plugin will only generate the service code, it won't generate the `Note` or `NoteRequestId`.

denil-ct:
Ah, I see. Adding the protobuf plugin fixed it. Thanks.

denil-ct:
You can review the changes. @glbrntt

glbrntt (commented):
This looks good to me but could you run `./scripts/format.sh` to run the formatter so that CI check passes?

denil-ct:
Have added the changes done by swift-format.

glbrntt (approved):

FranzBusch (approved):

glbrntt:
Thanks @denil-ct 🙏

PR #1626, opened 2023-07-05
Only schedule a close if the ping was actually sent

marhas:
If an outbound ping is blocked, I believe we should not schedule a close of the connection.

For me this caused intermittent issues with "unavailable (14): Transport became inactive" when calling the same streaming endpoint repeatedly. The last outbound ping would be blocked by shouldBlockPing because the stream had closed, but a closeConnection call is still scheduled, potentially closing the connection of a consecutive call when fired.


glbrntt:
@marhas -- thanks for the PR. I'll need you to fix up your CLA before I can review this though.

marhas:
Hello George,

I believe I already have:

￼

Is there anything else I need to do?

Thanks,
Marcel

marhas:
Thanks for reviewing and merging it so quickly!

PR #1629, opened 2023-07-13
Add note about SPM plugin known issues, provide better errors

glbrntt:


FranzBusch (approved):

denil-ct:
@FranzBusch are we not removing `importPaths` option here?

FranzBusch:
> @FranzBusch are we not removing `importPaths` option here?

This is an interesting one. I think we should remove it because I think in the current state it is not useable at all. Though it would be a potentially breaking API. @glbrntt what do you think?

glbrntt:
I don't think we should remove for API reasons. We can detect its presence and emit a warning or something though?

FranzBusch:
Yeah we should probably just warn on this or even error out. The only thing you can pass here right now are absolute paths which would be very brittle anyhow.

PR #1636, opened 2023-07-28
Add bind and connect APIs for VSOCK sockets

simonjbeaumont:
### Motivation:

The VSOCK address family facilitates communication between virtual machines and the host they are running that need a communications channel that is independent of virtual machine network configuration.

While both GRPC has support for building channels using sockets that were created out of band (`withConnectedSocket(_:)` and `withBoundSocket(_:)`, there are no APIs that facilitate the creation of VSOCK-based channels.

https://github.com/apple/swift-nio/pull/2479 adds a `VsockAddress` type and associated bootstrap APIs, but these need corresponding APIs here to be used in GRPC clients and servers.

### Modifications:

- Add `ConnectionTarget.vsockAddress` and `BindTarget.vsockAddress `.
- Add `ClientBootstrapProtocol connect(to vsockAddress:)` and `ServerBootstrapProtocol.bind(to vsockAddress:)`.
- Add `Server.bind(vsockAddress: VsockAddress)`.

### Todo:
- [ ] Depend on new version of NIO, when available.
- [ ] Decide whether to remove the `--vsock` support in the example.
- [ ] Determine what sort of tests to add in this package (there are tests in NIO, ofc).

### DIFF
--- a/Sources/Examples/HelloWorld/Client/HelloWorldClient.swift
+++ b/Sources/Examples/HelloWorld/Client/HelloWorldClient.swift
@@ -25,9 +25,6 @@ struct HelloWorld: AsyncParsableCommand {
   @Option(help: "The port to connect to")
   var port: Int = 1234
 
-  @Flag(help: "Use local vsock socket")
-  var vsock = false
-
   @Argument(help: "The name to greet")
   var name: String?
 
@@ -43,18 +40,8 @@ struct HelloWorld: AsyncParsableCommand {
     }
 
     // Configure the channel, we're not using TLS so the connection is `insecure`.
-    let target: ConnectionTarget
-    if self.vsock {
-      #if canImport(Darwin)
-      target = .vsockAddress(.init(cid: .any, port: .init(self.port)))
-      #else
-      target = .vsockAddress(.init(cid: .local, port: .init(self.port)))
-      #endif
-    } else {
-      target = .host("localhost", port: self.port)
-    }
     let channel = try GRPCChannelPool.with(
-      target: target,
+      target: .host("localhost", port: self.port),
       transportSecurity: .plaintext,
       eventLoopGroup: group
     )
--- a/Sources/Examples/HelloWorld/Client/HelloWorldClient.swift
+++ b/Sources/Examples/HelloWorld/Client/HelloWorldClient.swift
@@ -25,6 +25,9 @@ struct HelloWorld: AsyncParsableCommand {
   @Option(help: "The port to connect to")
   var port: Int = 1234
 
+  @Flag(help: "Use local vsock socket")
+  var vsock = false
+
   @Argument(help: "The name to greet")
   var name: String?
 
@@ -40,8 +43,18 @@ struct HelloWorld: AsyncParsableCommand {
     }
 
     // Configure the channel, we're not using TLS so the connection is `insecure`.
+    let target: ConnectionTarget
+    if self.vsock {
+      #if canImport(Darwin)
+      target = .vsockAddress(.init(cid: .any, port: .init(self.port)))
+      #else
+      target = .vsockAddress(.init(cid: .local, port: .init(self.port)))
+      #endif
+    } else {
+      target = .host("localhost", port: self.port)
+    }
     let channel = try GRPCChannelPool.with(
-      target: .host("localhost", port: self.port),
+      target: target,
       transportSecurity: .plaintext,
       eventLoopGroup: group
     )
--- a/Sources/GRPC/PlatformSupport.swift
+++ b/Sources/GRPC/PlatformSupport.swift
@@ -147,7 +147,7 @@ extension NIOTSConnectionBootstrap: ClientBootstrapProtocol {
   }
 
   public func connect(to vsockAddress: VsockAddress) -> EventLoopFuture<Channel> {
-    preconditionFailure("NIOTSListenerBootstrap does not support connect(to vsockAddress:)")
+    preconditionFailure("NIOTSConnectionBootstrap does not support connect(to vsockAddress:)")
   }
 }
 #endif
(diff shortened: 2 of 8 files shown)


Sources/Examples/HelloWorld/Client/HelloWorldClient.swift
```diff
@@ -25,6 +25,9 @@ struct HelloWorld: AsyncParsableCommand {
   @Option(help: "The port to connect to")
   var port: Int = 1234
 
+  @Flag(help: "Use local vsock socket")
```
glbrntt, line 28: I don't think we should include vsock in the examples: this one has an accompanying tutorial and we should keep the code in line with that. We should keep them as simple as possible: most users won't use vsock so we shouldn't expose them to it here.

Sources/GRPC/PlatformSupport.swift
```diff
@@ -144,6 +145,10 @@ extension NIOTSConnectionBootstrap: ClientBootstrapProtocol {
   public func withConnectedSocket(_ socket: NIOBSDSocket.Handle) -> EventLoopFuture<Channel> {
     preconditionFailure("NIOTSConnectionBootstrap does not support withConnectedSocket(_:)")
   }
+
+  public func connect(to vsockAddress: VsockAddress) -> EventLoopFuture<Channel> {
+    preconditionFailure("NIOTSListenerBootstrap does not support connect(to vsockAddress:)")
```
glbrntt, line 150: copy-pasta:

```suggestion
    preconditionFailure("NIOTSConnectionBootstrap does not support connect(to vsockAddress:)")
```

Tests/GRPCTests/VsockSocketTests.swift
```diff
@@ -0,0 +1,67 @@
...
+  private func vsockAvailable() -> Bool {
+    let fd: CInt
+    #if os(Linux)
+    fd = socket(AF_VSOCK, CInt(SOCK_STREAM.rawValue), 0)
+    #elseif canImport(Darwin)
+    fd = socket(AF_VSOCK, SOCK_STREAM, 0)
+    #else
+    fd = -1
+    #endif
+    guard fd != -1 else { return false }
```
glbrntt, line 63: This is a pet-peeve of mine: I think `guard !x else { ... }` is much less clear than `if x { ... }` (I appreciate they aren't exactly the same as `guard` forces you to return/throw)

```suggestion
    if fd == -1 { return false }
```

Tests/GRPCTests/VsockSocketTests.swift
```diff
@@ -0,0 +1,67 @@
...
+    assert(close(fd) == 0)
```
glbrntt, line 64: There's a subtle bug here: `assert` will only execute the expression passed to it on debug builds so for release builds we will leak the descriptor. (I have made this mistake more than once...)

```suggestion
    let rc = close(fd)
    assert(rc == 0)
```

glbrntt (commented):
Looks great aside aside from a couple of small issues.

Sources/Examples/HelloWorld/Client/HelloWorldClient.swift
simonjbeaumont: That's completely fine. I assumed this was the case and had them here purely for the purposes of the draft PR to show them in action.

Removed.

Sources/GRPC/PlatformSupport.swift
simonjbeaumont: Done.

Tests/GRPCTests/VsockSocketTests.swift
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import EchoImplementation
+import EchoModel
+@testable import GRPC
```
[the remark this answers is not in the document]
simonjbeaumont: We don't I suppose, probably copy-paste from other tests and didn't stand out as bad, given we are testing GRPC. 

Removed.

Do you prefer to only add it when necessary?

Tests/GRPCTests/VsockSocketTests.swift
simonjbeaumont: Happy to accommodate the peev and have updated to use `if`.

OOI, don't you ever like to use the `guard` and it's force-you-to-return/throw semantics? I like the compiler to help me where possible. Was it the double-negative that triggered you? Would you have been happier with `guard fd >= 0 else { ... }` or always prefer `if <exit condition> { ... }`?

Tests/GRPCTests/VsockSocketTests.swift
simonjbeaumont: Aren't tests always run in debug, otherwise `@testable import` cannot work?

simonjbeaumont (commented):
Thanks @glbrntt. Resolved all the feedback, with the exception of the comment about the use of `assert` which I responded too for more detail.

Tests/GRPCTests/VsockSocketTests.swift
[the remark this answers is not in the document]
glbrntt: I'd prefer to only add it if necessary: it makes it clear that only public API is being tested and ensures that when you mean to add public API you actually mark it as public.

Tests/GRPCTests/VsockSocketTests.swift
simonjbeaumont: Got it! And thanks for educating me on `-Xswiftc -enable-testing`.

Replaced the `assert` with a `precondition`.

glbrntt (approved):
Looks great modulo needing a NIO release. Thanks @simonjbeaumont!

glbrntt (approved):

PR #1640, opened 2023-08-15
Adopt strict concurrency checking

glbrntt:
Motivation:

In Swift 6 strict concurrency checking will be enabled by default. We can enable support now to flush out any concurrency bugs.

Modifications:

In the generator:
- Add `@Sendable` annotations to closures which were not annotated
- Wrap functions refs in closures
- Remove an incorrect comment from docs

In the runtime: various changes to internal types marking them as `Sendable` or `@unchecked Sendable`. A number or larger changes were also made:
- Require the `Request`/`Response` types `Sendable`. We didn't add this initially because we adopted `Sendable` before Swift Protobuf did. It's now been long enough since Swift Protobuf adopted `Sendable` that this is no longer an issue.
- As a result of the above, `GRPCPayload`, the serializer and deserializer are also now required to be `Sendable`. This bubbles up to quite a few functions in the public API.
- The fake transport is now `Sendable` (and still deprecated) and uses a lock to synchronize state.
- The connection pool uses loop-bound self in a number of places.

CI:
- Add strict concurrency checking to 5.8 and nightly

Result:

Passes strict concurrency checking.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -29,9 +29,8 @@ jobs:
             # No TSAN because of: https://github.com/apple/swift/issues/59068
             swift-test-flags: "-Xswiftc -strict-concurrency=complete" # "--sanitize=thread"
           - image: swift:5.8-jammy
-            swift-build-flags: "-Xswiftc -strict-concurrency=complete"
             # No TSAN because of: https://github.com/apple/swift/issues/59068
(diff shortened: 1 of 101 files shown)


glbrntt:
This has drifted quite far from `main`. Closing in favour of adopting strict concurrency in v2.

Issue #1649, opened 2023-09-16
Connection timeout is ignored when retries are set to none


afeick:
### Describe the bug

Connection timeout is ignored on last retry or when retries are set to none.

`ConnectionManager.startConnecting` uses a `ConnectionBackoffIterator` to determine the `connectTimeout` here:
[Sources/GRPC/ConnectionManager.swift#L988](https://github.com/grpc/grpc-swift/blob/84bac657e9930d26e9124bac082f26586dc2d209/Sources/GRPC/ConnectionManager.swift#L988)

Unfortunately when the `ConnectionBackoffIterator` is on its last (or only) try, it returns `nil` for the next value here:
[Sources/GRPC/ConnectionBackoff.swift#L140](https://github.com/grpc/grpc-swift/blob/84bac657e9930d26e9124bac082f26586dc2d209/Sources/GRPC/ConnectionBackoff.swift#L140)

That means the connection timeout is the bootstrap default, which in my use case is 10 seconds:
[swift-nio-transport-services/Sources/NIOTransportServices/NIOTSConnectionBootstrap.swift#L57](https://github.com/apple/swift-nio-transport-services/blob/251e9e76134570f0cc5c21e5c6ff3ccb247472ab/Sources/NIOTransportServices/NIOTSConnectionBootstrap.swift#L57)

### To reproduce

```
            let connection = ClientConnection
                .usingTLSBackedByNIOSSL(on: eventLoop)
                .withConnectionTimeout(minimum: TimeAmount.seconds(1))
                .withConnectionBackoff(retries: ConnectionBackoff.Retries.none)
                .connect(host: host, port: port)

            let client = Myservice_MyServiceNIOClient(channel: connection)
            do {
                let call = client.getData(Myservice_GetDataRequest())
                let response = try call.response.wait()
            } catch {
                logger.error("request failed: \(error)")
            }
```
### Expected behaviour

Call should timeout after 1 second, but it's taking 10 seconds.

### Additional information

This is sorta ugly, but it does resolve the issue for me locally:
```
diff --git a/Sources/GRPC/ConnectionManager.swift b/Sources/GRPC/ConnectionManager.swift
index 35ee4110..7299eea8 100644
--- a/Sources/GRPC/ConnectionManager.swift
+++ b/Sources/GRPC/ConnectionManager.swift
@@ -942,18 +942,21 @@ extension ConnectionManager {
   // states. Must be called on the `EventLoop`.
   private func startConnecting() {
[...]

PR #1650, opened 2023-09-20
Fix connection timeout when retries are set to none

afeick:
**Motivation:**

Connection timeout is ignored on last retry or when retries are set to none (issue #1649)

**Modifications:**

  Use the `minimumConnectionTimeout` value from the `ConnectionBackoff` directly instead of the `ConnectionBackoffIterator`

**Results:**

Connection timeout is honored across all connection attempts.

### DIFF
--- a/Sources/GRPC/ConnectionManager.swift
+++ b/Sources/GRPC/ConnectionManager.swift
@@ -942,18 +942,24 @@ extension ConnectionManager {
   // states. Must be called on the `EventLoop`.
   private func startConnecting() {
     self.eventLoop.assertInEventLoop()
+    var timeout: TimeAmount?
+    if let backoff = self.connectionBackoff {
+      timeout = TimeAmount.seconds(timeInterval: backoff.minimumConnectionTimeout)
+    }
     switch self.state {
     case .idle:
       let iterator = self.connectionBackoff?.makeIterator()
       self.startConnecting(
         backoffIterator: iterator,
-        muxPromise: self.eventLoop.makePromise()
+        muxPromise: self.eventLoop.makePromise(),
+        connectTimeout: timeout
       )
 
     case let .transientFailure(pending):
       self.startConnecting(
         backoffIterator: pending.backoffIterator,
-        muxPromise: pending.readyChannelMuxPromise
+        muxPromise: pending.readyChannelMuxPromise,
+        connectTimeout: timeout
       )
 
     // We shutdown before a scheduled connection attempt had started.
@@ -973,7 +979,8 @@ extension ConnectionManager {
 
   private func startConnecting(
     backoffIterator: ConnectionBackoffIterator?,
-    muxPromise: EventLoopPromise<HTTP2StreamMultiplexer>
+    muxPromise: EventLoopPromise<HTTP2StreamMultiplexer>,
+    connectTimeout: TimeAmount? = nil
   ) {
     let timeoutAndBackoff = backoffIterator?.next()
 
@@ -985,7 +992,7 @@ extension ConnectionManager {
       let channel: EventLoopFuture<Channel> = self.channelProvider.makeChannel(
         managedBy: self,
         onEventLoop: self.eventLoop,
-        connectTimeout: timeoutAndBackoff.map { .seconds(timeInterval: $0.timeout) },
+        connectTimeout: connectTimeout,
         logger: self.logger
       )
 


glbrntt (changes requested):
Apologies for the delay in taking a look at this.

This is slightly more awkward than I first realised. As you pointed out there are two cases we need to handle:
- never retrying, and
- the final retry

Dealing with never retrying is fine: we can do as you've proposed by getting the `minimumConnectionTimeout` in the `idle` case and using that instead of the value from the iterator. That works because the value from the iterator will be the same as `minimumConnectionTimeout`.

For the final retry it's a bit difficult: we want the iterator to provide the connect timeout but not the backoff. The API for the iterator will always return both to us unfortunately and we can't change that without breaking API. Two options come to mind:

1. Accept that the connect timeout for the last retry is still wrong and that we're only fixing the case where we never retry. (This is okay: strictly an improvement on what we have now.)
2. Return a sentinel value from the iterator and pick that up in `startConnecting` and set the `Reconnect` to `none` in that case.

I'm happy with either.

afeick:
> This is slightly more awkward than I first realised.

Yeah, I was a little surprised you were OK with the hacky fix I posted in the issue report 😂  

My initial thought for a proper fix was to refactor Iterator to make the timeout public and remove it from the `Element` since the timeout never changes. I think that's what you're describing in option 2, right?

>     2. Return a sentinel value from the iterator and pick that up in `startConnecting` and set the `Reconnect` to `none` in that case.

I think another option is to have the `backoff` value in [the Element](https://github.com/grpc/grpc-swift/blob/84bac657e9930d26e9124bac082f26586dc2d209/Sources/GRPC/ConnectionBackoff.swift#L102) be optional (`TimeInterval?`). Then `next()` could always return an element with a timeout, but backoff could be `nil` if there are no more retries. That should work with [the reconnect check](https://github.com/grpc/grpc-swift/blob/84bac657e9930d26e9124bac082f26586dc2d209/Sources/GRPC/ConnectionManager.swift#L1000), and it seems like a simpler fix. I'm just not sure if it's ideal for an `IteratorProtocol` to always return a next value.

afeick:
> This is slightly more awkward than I first realised.

Same. I missed that [the timeout can increase with the backoff](https://github.com/grpc/grpc-swift/blob/84bac657e9930d26e9124bac082f26586dc2d209/Sources/GRPC/ConnectionBackoff.swift#L163). Not sure how I didn't see that because I was looking for where it might change, given the variable name **minimum**ConnectionTimeout.

glbrntt:
> > This is slightly more awkward than I first realised.
> 
> Same. I missed that [the timeout can increase with the backoff](https://github.com/grpc/grpc-swift/blob/84bac657e9930d26e9124bac082f26586dc2d209/Sources/GRPC/ConnectionBackoff.swift#L163). Not sure how I didn't see that because I was looking for where it might change, given the variable name **minimum**ConnectionTimeout.

Yeah, it's only really an issue for the final value though; the solution I outlined above should be sufficient here.

Sources/GRPC/ConnectionBackoff.swift
```diff
@@ -135,7 +135,12 @@ public class ConnectionBackoffIterator: IteratorProtocol {
     case let .limited(limit) where limit > 0:
       self.connectionBackoff.retries.limit = .limited(limit - 1)
 
-    // limit must be <= 0, no new element.
+    // limit is reached, return an element with only a timeout.
+    case let .limited(limit) where limit == 0:
+      self.connectionBackoff.retries.limit = .limited(limit - 1)
+      return self.makeElement(backoff: nil)
```
afeick, line 141: @glbrntt What do you think about using the Element.backoff as the sentinel value for the iterator?

Tests/GRPCTests/ConnectionBackoffTests.swift
```diff
@@ -79,19 +79,19 @@ class ConnectionBackoffTests: GRPCTestCase {
     for limit in [1, 3, 5] {
       let backoff = ConnectionBackoff(retries: .upTo(limit))
       let values = Array(backoff)
-      XCTAssertEqual(values.count, limit)
+      XCTAssertEqual(values.count, limit+1)
```
afeick, line 82: The value count would make more sense if the `ConnectionBackoff` took a value called `tries` or `attempts` instead of `retries`, but I didn't think it was a good idea to make a breaking change like that.

Sources/GRPC/ConnectionBackoff.swift
glbrntt: It changes the `Element` type which is an API breaking change so we can't do it unfortunately. Otherwise it'd be a good solution.

glbrntt:
This got fixed in https://github.com/grpc/grpc-swift/pull/1777/

Issue #1409, opened 2022-05-23
unavailable (14): Transport became inactive


ripplek:
### Describe the bug

1. I am using the following:
1.1. iOS (Swift)
1.2. Cocoapods
```
  - gRPC-Swift (1.7.3):
    - CGRPCZlib (= 1.7.3)
    - Logging (< 2.0.0, >= 1.4.0)
    - SwiftNIO (< 3.0.0, >= 2.32.0)
    - SwiftNIOExtras (< 2.0.0, >= 1.4.0)
    - SwiftNIOHTTP2 (< 2.0.0, >= 1.19.2)
    - SwiftNIOSSL (< 3.0.0, >= 2.14.0)
    - SwiftNIOTransportServices (< 2.0.0, >= 1.11.1)
    - SwiftProtobuf (< 2.0.0, >= 1.9.0)
  - SwiftNIO (2.38.0):
    - SwiftNIOCore (= 2.38.0)
    - SwiftNIOEmbedded (= 2.38.0)
    - SwiftNIOPosix (= 2.38.0)
  - SwiftNIOConcurrencyHelpers (2.38.0):
    - CNIOAtomics (= 2.38.0)
  - SwiftNIOCore (2.38.0):
    - CNIOLinux (= 2.38.0)
    - SwiftNIOConcurrencyHelpers (= 2.38.0)
  - SwiftNIOEmbedded (2.38.0):
    - _NIODataStructures (= 2.38.0)
    - SwiftNIOCore (= 2.38.0)
  - SwiftNIOExtras (1.10.2):
    - SwiftNIO (< 3, >= 2.32.0)
  - SwiftNIOFoundationCompat (2.38.0):
    - SwiftNIO (= 2.38.0)
    - SwiftNIOCore (= 2.38.0)
  - SwiftNIOHPACK (1.20.1):
    - SwiftNIO (< 3, >= 2.35.0)
    - SwiftNIOConcurrencyHelpers (< 3, >= 2.35.0)
    - SwiftNIOCore (< 3, >= 2.35.0)
    - SwiftNIOHTTP1 (< 3, >= 2.35.0)
  - SwiftNIOHTTP1 (2.38.0):
    - CNIOHTTPParser (= 2.38.0)
    - SwiftNIO (= 2.38.0)
    - SwiftNIOConcurrencyHelpers (= 2.38.0)
    - SwiftNIOCore (= 2.38.0)
  - SwiftNIOHTTP2 (1.20.1):
[...]

Issue #1148, opened 2021-03-15
Channel level interceptors


jongarate:
### Description

The [current](https://github.com/grpc/grpc-swift/blob/main/docs/interceptors-tutorial.md) interceptor implementation sits on top of each service client. This forces the implementation on the client side of the gRPC service to arrange a factory with an interceptor instance per available service method.

When the backend supports multiple services (which is usually the case), forces the development on the client side to define interceptors per-service/per-method, resulting in a lot of boilerplate code and cumbersomeness.

### Suggested solution

Just like in Android, it would be helpful to have an option to define interceptors at the `Channel` level, so that at least basic exchange data can be logged/observed (e.g. Headers, etc...) by hooking to a single point in the whole architecture.

glbrntt:
This is quite cumbersome at the moment and worth improving. 

Adding it at the channel-level has a few difficulties, mainly that it's not easy to resolve which interceptors to use when they exist at the channel level and in the interceptor factory. Should the interceptors from the factory override or be added to those defined on the channel?

Another approach might be to change the generated interceptor factory: it could provide a generic factory method for 'default' interceptors and default the implementation of each factory method to call the default. That way you'd only have to implement one method per service and you retain full control over configuring interceptors on a per-method basis.

jongarate:
To my view, channel-level interceptors would sit higher architecturally speaking than the method specific ones, which could be further extended in more detail in the method-level ones if found necessary. They should essentially coexist as their scope and purpose is quite different.

Providing a generic factory for default interceptors would certainly be an improvement, and I guess more feasible in the short run. However, it would still require attaching them to each client explicitly, which hurts maintainability in the long run.

VorobyevS:
For base factory, temp solution. Resolving unnecessary recreations for same requests. U may improve it for your self.

```swift
typealias InterceptorEntity = (SwiftProtobuf.Message
                               & SwiftProtobuf._MessageImplementationBase
                               & SwiftProtobuf._ProtoNameProviding)

class GPRCBaseInterceptorFactory {
    private final class InterceptorsLocator {
        private var interceptors = [String: [Any]]()
        
        func register<T: InterceptorEntity, D: InterceptorEntity>(_ interceptors: [ClientInterceptor<T, D>]) {
            let description = String(describing: T.self) + String(describing: D.self)
            self.interceptors[description] = interceptors
        }
        
[...]

PR #1652, opened 2023-09-22
Add proposal for v2 stub design

glbrntt:


Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
```diff
@@ -0,0 +1,1100 @@
...
+        // ...
+      }
+    }
+  }
+}
+
+// MARK: - Supporting types
+
+/// A sink for values which are produced over time.
+public protocol Writer<Element: Sendable>: Sendable {
```
FranzBusch, line 342: We probably want to support `~Copyable` types here and all the other protocols/generic types as well and have a conditional `Copyable` conformance. I don't think the latter is expressible yet but we have to keep it in mind.

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
```diff
@@ -0,0 +1,1100 @@
...
+  /// Write a sequence of elements.
+  ///
+  /// Writes may suspend if the sink is unable to accept writes.
+  func write(contentsOf elements: some Sequence<Element>) async throws
+}
+
+extension Writer {
+  /// Write a single element.
+  public func write(_ element: Element) async throws {
```
FranzBusch, line 351: Can we also have a `write` method that takes an `AsyncSequence`?

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
```diff
@@ -0,0 +1,1100 @@
...
+}
+
+// MARK: - Supporting types
+
+/// A sink for values which are produced over time.
+public protocol Writer<Element: Sendable>: Sendable {
+  /// Write a sequence of elements.
+  ///
+  /// Writes may suspend if the sink is unable to accept writes.
+  func write(contentsOf elements: some Sequence<Element>) async throws
```
FranzBusch, line 346: Do we need a `finish` method?

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
```diff
@@ -0,0 +1,1100 @@
...
+  }
+
+  /// The error code.
+  public var code: Code
+
+  /// A message describing the error.
+  public var message: String
+
+  /// Metadata associated with the error.
+  public var metadata: Metadata
```
FranzBusch, line 395: Interesting. Are we expecting to put the detailed error trailer in here or just all trailers>

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
```diff
@@ -0,0 +1,1100 @@
...
+  ) async throws -> ServerResponse.Stream<EchoResponse>
+
+  func update(
+    request: ServerRequest.Stream<EchoResponse>
+  ) async throws -> ServerResponse.Stream<EchoResponse>
+}
+
+// Generated conformance to `BindableService`.
+extension EchoServiceStreamingProtocol {
+  public func bind(to router: inout RPCRouter) {
```
FranzBusch, line 464: Is `bind` a term of are here in gRPC? I personally think `register(with router:)` might sound better.

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
```diff
@@ -0,0 +1,1100 @@
...
+public typealias EchoResponse = ...
+
+/// The generated base protocol for the "Echo" service providing each method
+/// in a fully streamed form.
+///
+/// This protocol should typically not be implemented, instead you should
+/// implement ``EchoServiceProtocol`` which refines this protocol. However, if
+/// you require more granular control over your RPCs then they may implement
+/// this protocol, or methods from this protocol, instead.
+public protocol EchoServiceStreamingProtocol: BindableService, Sendable {
```
FranzBusch, line 444: I find the name `BindableService` interesting. Especially with `Service` from `ServiceLifecycle`. What do you think about naming it `GRPCService`?

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
```diff
@@ -0,0 +1,1100 @@
...
+      print("'Expand' failed with error '\(error.code)' and metadata '\(error.metadata)'")
+    } catch {
+      print("'Expand' failed with error '\(error)'")
+    }
+  }
+
+  // Note: there is no generated 'default' handler, the function body defines
+  // the lifetime of the RPC and the RPC is cancelled once the closure exits.
+  // Therefore escaping the message sequence would result in the sequence
+  // throwing an error. It must be consumed from within the handler.
```
FranzBusch, line 778: In the future, we should see if can adopt `~Escapable` here and let the compiler make sure this is enforced.

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
```diff
@@ -0,0 +1,1100 @@
...
+  let request = ClientRequest.Stream<Echo.Response>(metadata: ["foo": "bar"]) { writer in
+    for text in ["foo", "bar", "baz"] {
+      try await writer.write(Echo.Request(text: text))
+    }
+  }
+
+  // Execute the RPC:
+  try await echo.collect(request: request) { response in
+    // (Same as for the server streaming "Expand")
+  }
```
FranzBusch, line 792: I find this spelling very interesting and a bit surprising. The problem that I see is that there are now two separate scopes - one for the request writer and one for the response. If you want to implement business logic where you write a certain request based on a response you received this becomes very hard and needs a layer of indirection.

From previous experience, I found it helpful to have the bidirectional streaming closure provide both the `responseStream` and the `requestWriter`.

Just some pseudo code for what I mean

```
try await someClient.bidistream(metadata: ...) { requestWriter, responseStream in
    try await requestWriter.write(...)
    
    let firstResponse = try await responseStream.first
    
    if firstResponse.someProperty {
        // different message written
    } else {
        // different message written
    }
}
```

FranzBusch (commented):
Great work. Left some comments inline!

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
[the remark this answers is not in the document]
glbrntt: It's nice to have the markdown rendered somewhere. I agree most adopters won't care but I also don't think that's an issue, they don't have to read it!

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
```diff
@@ -0,0 +1,1100 @@
...
+
+    /// Create a new single client request.
+    public init(message: Message, metadata: Metadata = [:]) {
+      // ...
+    }
+  }
+
+  /// A request created by the client containing a message producer.
+  public struct Stream<Message: Sendable>: Sendable {
+    public typealias Producer = @Sendable (any Writer<Message>) async throws -> Void
```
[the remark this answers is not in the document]
glbrntt: Yes; we should do that. I will update this in the next version of this doc.

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
```diff
@@ -0,0 +1,1100 @@
...
+  public struct Stream<Message: Sendable>: Sendable {
+    /// Metadata received from the client at the begining of the RPC.
+    public var metadata: Metadata
+
+    /// An `AsyncSequence` of messages received from the client.
+    public var messages: any AsyncSequence<Message>
+
+    // Note: the above syntax does not compilee as `AsyncSequence` does not have a
+    //   primary associated type. However the shape of API is illustrative and
+    //   the exact type is to be determined.
```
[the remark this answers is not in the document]
glbrntt: Right, I expect it will end up like the writer to be honest as we'll no doubt also need to deal with existentials.

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
glbrntt: Internally we'll likely need a version with `finish`. In the public API we won't need it: the `finish` is implicit when the producer function returns. 

One thing I noticed while prototyping is that when wrapping writers in interceptors it's very tempting to call `finish` after calling the producer which can lead to multiple finish calls because the producer usually writes `finish` too. If you're e.g. logging in an interceptor you end up with duplicated logs. Removing it entirely and having it implicit in the API felt more natural.

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
glbrntt: All trailers. Initially I didn't have `metadata` in the error type at all and ended up with a wrapper struct pairing an error with metadata which also conformed to `Error`. It was quite unpleasant to work with (do I catch `RPCError` or `RPCErrorWithMetadata`? Or both?).

Ultimately we expect this to represent the final state of a failed RPC where we'd expect to have metadata as well. That's not always the case because the final error can be synthesised on the client side but using an empty `Metadata` in that case is simpler than having to deal with two different error types anyway.

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
glbrntt: I'm open to bikeshedding the naming. I believe grpc-java uses `BindableService` but I'm not wedded to that. I would be tempted to drop the "GRPC" and just call it `Service`: all generated code should be fully qualified anyway so would spell it `GRPC.Service` so I don't think ambiguity would be an issue.

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
glbrntt: Not especially, it fits with `BindableService` though. The naming isn't really important -- users are very unlikely to call this directly. We'd probably have API on the server/router to add a service which would just call through to this.

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
glbrntt: Is there a timeline for this?

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
glbrntt: I see this a _good_ thing: if your input stream depends on the output stream then you should make the dependency explicit via an additional `AsyncStream` or similar. Splitting them up makes it harder to accidentally deadlock.

It does make some use cases harder, but they are still possible.

FranzBusch: IMO, this is really important. I have myself written a bunch of tests where driving both the requests and responses in tandem is important. This is mostly the case when you implement something like a distributed state machine over a bidirectional streaming RPC.

Having to go through an intermediate stream can work but has two downsides:
1. It is currently not implementable since we don't have a backpressure `AsyncSequence` that we could use here.
2. It becomes less performant since we now have another layer of indirection.

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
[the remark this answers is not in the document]
FranzBusch: Maybe we should just put it into a separate section in the index page so that it is all grouped together.

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
FranzBusch: I agree. This seems like a nice API, I was just confused by the document because it says `associated with the error` and I wasn't sure what that meant.

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
FranzBusch: Agreed we can bikeshed later I think we should not use common types in our libraries. We have made a similar decision in `Kafka`, `Redis` etc. where we use `XXXClient` instead of just `Client`. `Service` is particularly dangerous since `ServiceLifecycle` uses it and it is probably quite common to have both in the same scope.

Since we have `RPCError` maybe just `RPCService`?

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
FranzBusch: This was part of the `BufferView` proposal.

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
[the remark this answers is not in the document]
glbrntt: I don't think we even need to link to them; they should just appear as a separate folder in the sidebar (see: https://swiftpackageindex.com/apple/swift-openapi-generator/0.2.2/documentation/swift-openapi-generator)

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
glbrntt: Beyond the issues mentioned above, the request and response being tied together has an impact on the interceptor API: it effectively takes the same shape as the function which isn't at all ergonomic to use. The knock on from this is that the client API diverges even further from the server API.

I'm confident that this is the correct spelling for the base API: it separates the input from the output and makes it hard for users to misuse the API. I

> It is currently not implementable since we don't have a backpressure AsyncSequence that we could use here.

We also don't have this API right now! There are backpressure `AsyncSequence`s which do work: `AsyncChannel` (sure, a bit slow as the buffer size is 1, but it works and is probably suitable for most uses cases)  and NIOs producer. There are others in the works too.

> It becomes less performant since we now have another layer of indirection.

True, but if your two streams depend on each other then the RTT will likely dominate anyway. 

>  I have myself written a bunch of tests where driving both the requests and responses in tandem is important.

If requested enough we can add a utility or sugar for this. I don't think it should be the base API though.

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
glbrntt: For reference: https://github.com/atrick/swift-evolution/blob/bufferview-roadmap/visions/language-support-for-BufferView.md

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
glbrntt: I don't think `Service` would be an issue for the reasons given above but I'm also not opposed to `RPCService`.

Jerry-Carter:
I have reviewed the [proposal](https://github.com/glbrntt/grpc-swift/blob/gb-stub-proposal/Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md) and [motivation](https://forums.swift.org/t/grpc-swift-plans-for-v2/67361) document.

First, let me strongly endorse the motivation.  I concur that "the ecosystem is trending towards using common types and patterns such as Swift HTTP Types 7 and is adopting Swift concurrency in place of EventLoopFuture based APIs" and, as a user, would love to see gRPC-Swift stay in the vanguard.  The support for async in 1.x was great to see and 2.x should be even better.

Second, I have reviewed the API.  There are no real surprises.  GRPC is conceptually fairly simple and the propose API for 2.x is appropriately straightforward and clear.

Third, in any rewrite, I would like to see [channel level interceptors](https://github.com/grpc/grpc-swift/issues/1148) added.

Fourth, thought should be given to enhancing gRPC-Swift 2.x by attempting to automatically recover from network issues.  I have found gRPC-Swift to work very well between devices with stable connections, but for IOT or mobile clients, the recovery code that currently needs to be written and tested in the client would seem better moved into the library.  See [1409](https://github.com/grpc/grpc-swift/issues/1409) and similar open issues.

Finally, the [Using ~Copyable writers](https://github.com/grpc/grpc-swift/issues/1409) section is missing a word:  "Initially it may either send metadata or it may **return** a status as its first and only value."

Thanks, George, Franz, et. al, for doing a great job with 1.x.  I look forward to migrating to 2.x ASAP.

glbrntt:
> I have reviewed the [proposal](https://github.com/glbrntt/grpc-swift/blob/gb-stub-proposal/Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md) and [motivation](https://forums.swift.org/t/grpc-swift-plans-for-v2/67361) document.

Thanks for taking the time to review this, it's great to get feedback from the community 🙏

> First, let me strongly endorse the motivation. I concur that "the ecosystem is trending towards using common types and patterns such as Swift HTTP Types 7 and is adopting Swift concurrency in place of EventLoopFuture based APIs" and, as a user, would love to see gRPC-Swift stay in the vanguard. The support for async in 1.x was great to see and 2.x should be even better.

Agreed. I think this is a really exciting time for the Swift and Swift on Server.

> Second, I have reviewed the API. There are no real surprises. GRPC is conceptually fairly simple and the propose API for 2.x is appropriately straightforward and clear.

:+1:

> Third, in any rewrite, I would like to see [channel level interceptors](https://github.com/grpc/grpc-swift/issues/1148) added.

Don't worry – this is very high up our list. The v1 interceptors are a bugbear of mine which I look forward to rectifying.

> Fourth, thought should be given to enhancing gRPC-Swift 2.x by attempting to automatically recover from network issues. I have found gRPC-Swift to work very well between devices with stable connections, but for IOT or mobile clients, the recovery code that currently needs to be written and tested in the client would seem better moved into the library. See [1409](https://github.com/grpc/grpc-swift/issues/1409) and similar open issues.

Agreed; this is a real weakness of v1. Supporting [gRFC A6](https://github.com/grpc/proposal/blob/master/A6-client-retries.md) out-of-the-box _should_ help somewhat here. However the transport layer will also need to become more resilient in the face of changing network conditions.

> Finally, the [Using ~Copyable writers](https://github.com/grpc/grpc-swift/issues/1409) section is missing a word: "Initially it may either send metadata or it may **return** a status as its first and only value."

Thanks. I've fixed that up.
 
> Thanks, George, Franz, et. al, for doing a great job with 1.x. I look forward to migrating to 2.x ASAP.

🙏

FranzBusch (approved):

Sources/GRPC/Docs.docc/Proposals/0001-stub-api.md
```diff
@@ -0,0 +1,1135 @@
...
+An example of using the Echo client follows. Some functions highlight the
+"sugared" API built on top of the more verbose lower-level API:
+
+```swift
+func get(echo: some EchoClientProtocol) async throws {
+  // Make the request:
+  let request = ClientRequest.Single(message: Echo.Request(text: "foo"))
+
+  // Execute the RPC (most verbose API):
+  await echo.get(request: request) { response in
```
vykut, line 708: Hi, 

Why is the API modeled in such a way that the consumer would need to pass a closure in order to use the `response`?
Wouldn't it be more straightforward to just do 
```
let response = echo.get(request: request) // response: ClientResponse.Single<EchoResponse>
switch response.result {
  ...
}
```

glbrntt: This is different to a "traditional" completion handler: here the `echo.get` won't return until the handler has run which means we _know_ that after `echo.get` returns, all resources associated with the call have been freed. 

For unary RPCs this isn't actually an issue because once there's a single response the RPC is done, you can just return the `response` from the closure. We may actually default the implementation of the closure to do this in generated code.

For RPCs which stream requests and responses we can't do this without resorting to unstructured concurrency which we absolutely want to avoid: cancellation is not propagated to unstructured tasks which makes the underlying code and code which uses these APIs difficult to reason about.

vykut: Even for the stream requests, the `body` closure will be executed only once. 
In my mind, the same syntax should work for stream requests as well, i.e
```
let response = try await echo.expand(request: request)
for try await message in response.messages {
  ...
}
```

glbrntt: The the important part here is this:

> For RPCs which stream requests and responses we can't do this **without resorting to unstructured concurrency** which we absolutely want to avoid: cancellation is not propagated to unstructured tasks which makes the underlying code and code which uses these APIs difficult to reason about.

The `echo.expand` call is hiding a `TaskGroup` which runs a number of tasks required to actually execute the RPC. The response handler is running as one of those tasks.

vykut: I see. Now that makes sense. 
Thanks for your explanation!

PR #1654, opened 2023-09-29
Tool that generates serialised file descriptor protos

stefanadranca:
Motivation:

In order to implement the reflection service we need to be able to generate serialised file descriptor protos of the services and messages that a server offers.

Modifications:

Added an option for protoc-gen-grpc-swift that enables generating a binary file containing a base64 encoded representation of the  serialized file descriptor proto of the specified proto file.

Result:

This change enables the generation of serialised file descriptor protos that will be useful in the implementation of the reflection service.


Sources/protoc-gen-grpc-swift/main.swift
```diff
@@ -138,16 +147,32 @@ func main(args: [String]) throws {
...
-        fileDescriptor: fileDescriptor,
-        fileNamingOption: options.fileNaming,
-        generatedFiles: &generatedFiles
-      )
-      let grpcGenerator = Generator(fileDescriptor, options: options)
-      var grpcFile = Google_Protobuf_Compiler_CodeGeneratorResponse.File()
-      grpcFile.name = grpcFileName
-      grpcFile.content = grpcGenerator.code
+        var binaryFile = Google_Protobuf_Compiler_CodeGeneratorResponse.File()
+        if (options.reflectionService) {
```
glbrntt, line 151: We're currently only generating reflection data if `!fileDescriptor.services.isEmpty` but we need to generate it for files which only contain messages as well.

Sources/protoc-gen-grpc-swift/options.swift
```diff
@@ -142,6 +143,13 @@ final class GeneratorOptions {
         } else {
           throw GenerationError.invalidParameterValue(name: pair.key, value: pair.value)
         }
+    
+      case "ReflectionService":
```
glbrntt, line 147: "ReflectionData" more accurately describes what we're generating.

Tests/GRPCTests/Codegen/Serialization/SerializationTests.swift
```diff
@@ -0,0 +1,70 @@
+//
+//  File.swift
+//  
+//
+//  Created by Stefana Dranca on 26/09/2023.
+//
+
```
glbrntt, line 7: This needs to be a proper copyright header. Look at the other files in the project for a template.

Tests/GRPCTests/Codegen/Serialization/SerializationTests.swift
```diff
@@ -0,0 +1,70 @@
...
+import GRPC
+import Foundation
+import SwiftProtobuf
+import SwiftProtobufPluginLibrary
```
glbrntt, line 11: Do we really need this import?

Tests/GRPCTests/Codegen/Serialization/SerializationTests.swift
```diff
@@ -0,0 +1,70 @@
...
+        XCTAssertEqual(syntax, "proto3")
+        
+        let package = self.fileDescriptorProto.package
+        XCTAssertEqual(package, "echo")
+    }
+    
+    func testMessages() {
+        let messages = self.fileDescriptorProto.messageType
+        XCTAssertEqual(messages.count, 2)
+        for msg in messages {
```
glbrntt, line 40: nit: `message`

Tests/GRPCTests/Codegen/Serialization/SerializationTests.swift
```diff
@@ -0,0 +1,70 @@
...
+    override func setUp() {
+        super.setUp()
+        let binaryFileURL = URL(fileURLWithPath: #filePath)
+            .deletingLastPathComponent().appendingPathComponent("echo.binary-grpc.txt")
+        let base64EncodedData = try! Data(contentsOf: binaryFileURL)
+        let binaryData = Data(base64Encoded: base64EncodedData)!
+        self.fileDescriptorProto = try! Google_Protobuf_FileDescriptorProto(serializedData: binaryData)
+    }
+
+    func testBinaryFile() throws {
```
glbrntt, line 26: nit: we're really testing the file metadata here: `testFileDescriptorMetadata`

Tests/GRPCTests/Codegen/Serialization/SerializationTests.swift
```diff
@@ -0,0 +1,70 @@
...
+        XCTAssertEqual(name, "echo.proto")
+        
+        let syntax = self.fileDescriptorProto.syntax
+        XCTAssertEqual(syntax, "proto3")
+        
+        let package = self.fileDescriptorProto.package
+        XCTAssertEqual(package, "echo")
+    }
+    
+    func testMessages() {
```
glbrntt, line 37: `testFileDescriptorMessages`

Tests/GRPCTests/Codegen/Serialization/SerializationTests.swift
```diff
@@ -0,0 +1,70 @@
...
+        XCTAssertEqual(messages.count, 2)
+        for msg in messages {
+            XCTAssert((msg.name == "EchoRequest") || (msg.name == "EchoResponse"))
+            XCTAssertEqual(msg.field.count, 1)
+            XCTAssertEqual(msg.field.first!.name, "text")
+            XCTAssert(msg.field.first!.hasNumber)
+        }
+    }
+        
+    func testService() {
```
glbrntt, line 48: `testFileDescriptorServices`

Tests/GRPCTests/Codegen/Serialization/SerializationTests.swift
```diff
@@ -0,0 +1,70 @@
...
+        let services = self.fileDescriptorProto.service
+        XCTAssertEqual(services.count, 1)
```
glbrntt, line 50: We should assert the number of methods as well

Tests/GRPCTests/Codegen/Serialization/SerializationTests.swift
```diff
@@ -0,0 +1,70 @@
...
+        for method in self.fileDescriptorProto.service.first!.method {
+            switch method.name {
+            case "Get":
```
glbrntt, line 53: We should be checking the input and output type for each method here.

Makefile
```diff
@@ -117,6 +117,20 @@ ${NORMALIZATION_GRPC}: ${NORMALIZATION_PROTO} ${PROTOC_GEN_GRPC_SWIFT}
...
+${SERIALIZATION_GRPC}: ${ECHO_PROTO} ${PROTOC_GEN_GRPC_SWIFT}
+	protoc $< \
+		--proto_path=$(dir $<) \
+		--plugin=${PROTOC_GEN_GRPC_SWIFT} \
+		--grpc-swift_opt=ReflectionService=true \
+		--grpc-swift_out=$(dir ${SERIALIZATION_PATH})
+
+# Generates protobufs and gRPC client and server for the Serialization test
+.PHONY:
+test-serialization: ${SERIALIZATION_GRPC}
```
glbrntt, line 132: I don't think this will work correctly... 

`${SERIALIZATION_GRPC}` is the "target", which is the name of the object to make, but `SERIALIZATION_GRPC` isn't defined anywhere. It should be defined as "Tests/GRPCTests/Codegen/Serialization/whatever-the-generated-file-is-named".

`${ECHO_PROTO}` and `${PROTOC_GEN_GRPC_SWIFT}` are pre-requisites: i.e. targets which must exist before this target can be created. This part is right: we want echo.proto to exist and the grpc-swift plugin for protoc.

In the options we only specify `ReflectionService=true`, but we also don't want to generate the client/server code using this rule so we should set `Client=false` and `Server=false` too.

We should also think about naming a bit more: all the other rules for generating code start with `generate-`. I think we should name the rule `generate-reflection-data`.

glbrntt (changes requested):

Makefile
glbrntt: This might be useful: https://makefiletutorial.com

Sources/protoc-gen-grpc-swift/main.swift
stefanadranca: Should this also be enabled for generating the .pb file?

glbrntt: We don't generate `.pb.swift` files, `protoc-gen-swift` does.  _But_ we should generate reflection data for descriptors which only contain messages.

Sources/protoc-gen-grpc-swift/main.swift
```diff
@@ -145,35 +140,40 @@ func main(args: [String]) throws {
 
   // Only generate output for services.
   for name in request.fileToGenerate {
-    if let fileDescriptor = descriptorSet.fileDescriptor(named: name),
-       !fileDescriptor.services.isEmpty {
+    if let fileDescriptor = descriptorSet.fileDescriptor(named: name) {
+      if (!fileDescriptor.services.isEmpty || !fileDescriptor.messages.isEmpty) {
```
glbrntt, line 144: The only thing we need to check here is `options.reflectionService`.

Sources/protoc-gen-grpc-swift/main.swift
```diff
@@ -65,14 +65,9 @@ func outputFileName(
   component: String,
   fileDescriptor: FileDescriptor,
   fileNamingOption: FileNaming,
-  binaryFile: Bool
+  extensionType: String
```
glbrntt, line 68: Just `extension` is fine

Sources/protoc-gen-grpc-swift/main.swift
```diff
@@ -65,14 +65,9 @@ func outputFileName(
...
 ) -> String {
-    let ext: String
-    if binaryFile {
-        ext = "." + component + ".txt"
-    } else {
-        ext = "." + component + ".swift"
-    }
+  let ext = "." + component + extensionType
```
glbrntt, line 70: The "." isn't usually included in the extension, so instead of passing it in `extensionType` we should be adding it here.

Sources/protoc-gen-grpc-swift/options.swift
```diff
@@ -143,13 +143,13 @@ final class GeneratorOptions {
...
-      case "ReflectionService":
-          if let value = Bool(pair.value) {
-              self.reflectionService = value
-          } else {
-              throw GenerationError.invalidParameterValue(name: pair.key, value: pair.value)
-          }
+
+      case "ReflectionData":
+        if let value = Bool(pair.value) {
+          self.reflectionService = value
```
glbrntt, line 149: Let's call the property `generateReflectionData`, this aligns with the "Client" and "Server" options mapping to `generateClient` and `generateServer` properties.

Makefile
```diff
@@ -117,19 +117,20 @@ ${NORMALIZATION_GRPC}: ${NORMALIZATION_PROTO} ${PROTOC_GEN_GRPC_SWIFT}
...
-# For serialization we'll set the ReflectionService option to true.
-${SERIALIZATION_GRPC}: ${ECHO_PROTO} ${PROTOC_GEN_GRPC_SWIFT}
+# For serialization we'll set the ReflectionData option to true.
+${SERIALIZATION_GRPC_REFLECTION}: ${ECHO_PROTO} ${PROTOC_GEN_GRPC_SWIFT}
 	protoc $< \
 		--proto_path=$(dir $<) \
 		--plugin=${PROTOC_GEN_GRPC_SWIFT} \
-		--grpc-swift_opt=ReflectionService=true \
+		--grpc-swift_opt=Client=false,Server=false,ReflectionData=true \
 		--grpc-swift_out=$(dir ${SERIALIZATION_PATH})
```
glbrntt, line 129: There's no need to define the `SERIALIZATION_PATH` above, you can get the directory of the file we want to generate:

```suggestion
		--grpc-swift_out=$(dir ${SERIALIZATION_GRPC_REFLECTION})
```

Sources/protoc-gen-grpc-swift/main.swift
stefanadranca: I tried, but since it is a keyword it doesn't work

stefanadranca: would fileExtension be better?

glbrntt: You can wrap it in backticks

PR #1656, opened 2023-10-02
Add RPCError and Status

glbrntt:
Motivation:

One mistake we made in v1 was having the status also be an error. This leads to interesting situations where a status with code 'ok' can be treated as an error.

Here, we split them into two objects.

Modifications:

- Add a new GRPCCore target
- Add RPCError and Status types and tests

Result:

RPCError and status are separate types.

### DIFF
--- a/Sources/GRPCCore/RPCError.swift
+++ b/Sources/GRPCCore/RPCError.swift
@@ -18,6 +18,8 @@
 ///
 /// See also ``Status``.
 public struct RPCError: @unchecked Sendable, Hashable, Error {
+  // @unchecked because it relies on heap allocated storage and 'isKnownUniquelyReferenced'
+
   private var storage: Storage
   private mutating func ensureStorageIsUnique() {
     if !isKnownUniquelyReferenced(&self.storage) {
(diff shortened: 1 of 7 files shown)


Sources/GRPCCore/RPCError.swift
```diff
@@ -0,0 +1,237 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/// An error representing the outcome of an RPC.
+///
+/// See also ``Status``.
+public struct RPCError: @unchecked Sendable, Hashable, Error {
```
FranzBusch, line 20: Can we add a comment why we use `@unchecked` here

Sources/GRPCCore/Status.swift
```diff
@@ -0,0 +1,262 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/// A status object represents the outcome of an RPC.
+///
+/// Each ``Status`` is composed of a ``Status/code-swift.property`` and ``Status/message``. Each
```
FranzBusch, line 19: The first docc link is very interesting :D

glbrntt: ``Status``? 🤨

FranzBusch: I mean this ```Status/code-swift.property```

glbrntt: Off-by-one error ;) 

Yeah, docc symbols aren't case sensitive so `Code`-the-type and `code`-the-property are ambiguous otherwise.

PR #1665, opened 2023-10-06
Add client request and response objects

glbrntt:


Sources/GRPCCore/Call/ClientRequest.swift
```diff
@@ -0,0 +1,116 @@
...
+
+extension ClientRequest {
+  /// A request created by the client for a stream of messages.
+  ///
+  /// This is used for client-streaming and bidirectional-streaming RPCs.
+  ///
+  /// See ``ClientRequest/Single`` for single-message requests and ``ServerRequest/Stream`` for the
+  /// servers representation of a streaming-message request.
+  public struct Stream<Message: Sendable>: Sendable {
+    public typealias Producer = @Sendable (RPCWriter<Message>) async throws -> Void
```
FranzBusch, line 71: Do we want to use this typealias? I know that previously we were very unhappy with how IDEs treated them.

Sources/GRPCCore/Call/ClientRequest.swift
```diff
@@ -0,0 +1,116 @@
...
+    /// Caller-specified metadata sent to the server at the start of the RPC.
+    ///
+    /// Both gRPC Swift and its transport layer may insert additional metadata. Keys prefixed with
+    /// "grpc-" are prohibited and may result in undefined behaviour. Transports may also insert
+    /// their own metadata, you should avoid using key names which may clash with transport specific
+    /// metadata. Note that transports may also impose limits in the amount of metadata which may
+    /// be sent.
+    public var metadata: Metadata
+
+    /// The message to send to the server.
```
FranzBusch, line 45: Extreme ultra nit: Above you say `metadata sent to the server` and here you say `message to send to the server`. Do we want to align this.

Sources/GRPCCore/Call/ClientResponse.swift
```diff
@@ -0,0 +1,436 @@
...
+      return RPCAsyncSequence.throwing(error)
+    }
+  }
+}
+
+// MARK: - Conversion
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension ClientResponse.Single {
+  @_spi(Testing)
```
FranzBusch, line 374: In the future we should probably switch to `package` access for those, but that is gated behind Swift versions.

FranzBusch (approved):
Just a few nits

Sources/GRPCCore/Call/ClientResponse.swift
glbrntt: Agreed. I think it's best to use `_spi` for now since it isn't version specific. We can revisit this when we get close the releasing or when 5.9 is our minimum supported version.

Sources/GRPCCore/Call/ClientRequest.swift
glbrntt: Removed.

PR #1669, opened 2023-10-09
Implemented file-by-filename and list-services requests 

stefanadranca:
Motivation:

The two requests are necessary for implementing the Reflection Service.

Modifications:

Implemented the ReflectionServiceData struct that stores the useful information from the file descriptor protos received when initiaizing the service and provides the data used for constructing the server responses for each type of request. Also imlemented the functions that create the server responses for the two requests. The requests and the functionality of the ReflectionServiceData struct's methods are tested in the "Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift" file.

Result:

The Reflection Service will support the file-by-filename and list-services requests.


Tests/GRPCTests/GRPCReflectionServiceTests/reflection.grpc.swift
```diff
@@ -0,0 +1,209 @@
+//
```
glbrntt, line 1: Can you add a `Generated` directory in `GRPCReflectionServiceTests` and generate the file in there instead? It's more obvious when you're looking through the code where the generated and non-generated code is.

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,241 @@
...
+        self.protos.remove(at: originalIndex!)
+      }
+
+      for serviceName in services {
+        XCTAssert(self.servicesNames.contains(serviceName))
+        self.servicesNames.removeAll { $0 == serviceName }
+      }
+    } catch {
+      XCTFail(error.localizedDescription)
+    }
```
glbrntt, line 205: You don't need to do this: just let the test throw an error

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,241 @@
...
+      let method = Google_Protobuf_MethodDescriptorProto.with {
+        $0.name = "testMethod" + String(id)
+        $0.inputType = inputMessage.name
+        $0.outputType = outputMessage.name
+      }
+      let serviceDescriptor = Google_Protobuf_ServiceDescriptorProto.with {
+        $0.method = [method]
+        $0.name = "service" + String(id)
+      }
+      self.servicesNames.append("service" + String(id))
```
glbrntt, line 62: We shouldn't be modifying our state here: it's not all obvious from the name of the function that it has side effects which can lead to bugs. For example your test could create the descriptors twice and end up with two copies a service in `serviceNames`.

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,241 @@
...
+      let inputMessage = Google_Protobuf_DescriptorProto.with {
+        $0.name = "inputMessage"
+        $0.field = [Google_Protobuf_FieldDescriptorProto.with {
+          $0.name = "inputField"
+          $0.type = .bool
+        }]
+      }
+
+      let outputMessage = Google_Protobuf_DescriptorProto.with {
+        $0.name = "inputMessage"
```
glbrntt, line 46: presumably should be `outputMessage`

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,241 @@
...
+    XCTAssertEqual(receivedServices?.count, 5)
+    for serviceName in receivedServices! {
+      XCTAssertTrue(self.servicesNames.contains(serviceName))
+    }
+  }
+
+  func testReflectionServiceData() throws {
+    self.createFileDescriptorProto()
+    do {
+      let registry = try ReflectionServiceData(fileDescriptorProtos: self.protos)
```
glbrntt, line 180: You should be testing each registry method in separate tests, it makes them easier to follow and debug when they fail.

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,241 @@
...
+      }
+    } catch {
+      XCTFail(error.localizedDescription)
+    }
+  }
+
+  func testGetSerializedFileDescriptorProtos() throws {
+    self.createFileDescriptorProto()
+    do {
+      let registry = try ReflectionServiceData(fileDescriptorProtos: self.protos)
```
glbrntt, line 211: You need more test cases than this covering different dependency graphs. We also need a test for getting the file descriptor protos of a filename we don't know about.

glbrntt (changes requested):

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -0,0 +1,183 @@
...
+ */
+
+import Foundation
+import GRPC
+import SwiftProtobuf
+
+public struct ReflectionServiceData: Sendable {
+  public struct protoData: Sendable {
+    fileprivate var serializedFileDescriptorProto: Data = .init()
+    fileprivate var dependency: [String] = []
```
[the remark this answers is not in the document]
stefanadranca: is the array of name of files imported by the proto, similar to the dependency field from Google_Protobuf_FileDescriptorProto

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -0,0 +1,183 @@
...
+      let protoData = self.fileDescriptorData[currentFileName]
+      if (protoData != nil) {
+        let serializedFileDescriptorProto = protoData!.serializedFileDescriptorProto
+        if (!protoData!.dependency.isEmpty) {
+          toVisit
+            .append(
+              contentsOf: self.fileDescriptorData[currentFileName]!.dependency
+                .filter { name in
+                  return !visited.contains(name)
+                }
```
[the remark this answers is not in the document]
stefanadranca: I'll switch to using a Set so the complexity will become O(d)

Sources/GRPCReflectionService/Server/ReflectionService.swift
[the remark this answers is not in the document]
glbrntt: Okay, you should make that clearer with a comment or a more descriptive property name 😉

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
stefanadranca: This is the test that checks the init, that's why it's checking both the dictionary and the service names list. Should these be done separately? I used the getters since the attributes are private

stefanadranca: Should I switch to internal as the access level for the attributes and not use the getters anymore?

stefanadranca: Should I switch to internal as the access level for the attributes and not use the getters anymore?

glbrntt: > This is the test that checks the init, that's why it's checking both the dictionary and the service names list. Should these be done separately? I used the getters since the attributes are private

Yes. Typically you should start by testing the API of the type you're testing. There are two APIs which need testing here:
1. `getServices()`
2. `getfileDescriptorData()`

At the moment all the work to prepare values for these functions is done in `init`, but that's an implementation detail that could change.

glbrntt: > Should I switch to internal as the access level for the attributes and not use the getters anymore?

Yes

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -0,0 +1,176 @@
...
+  private let protoRegistry: ReflectionServiceData
+
+  public init(fileDescriptorProtos: [Google_Protobuf_FileDescriptorProto]) throws {
+    self.protoRegistry = try ReflectionServiceData(
+      fileDescriptor: fileDescriptorProtos
+    )
+  }
+
+  internal func findFileByFileName(
+    fileName: String,
```
glbrntt, line 102: We can drop the label of `fileName` here because it's evident from the name of the function (otherwise it's `findFileByFileName(fileName: fileName, ...)` which is three `fileName`s in a row...)

```suggestion
  internal func findFileByFileName(
    _ fileName: String,
```

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,362 @@
...
+import XCTest
+
+@testable import GRPCReflectionService
+
+final class GRPCReflectionServiceTests: GRPCTestCase {
+  private var group: MultiThreadedEventLoopGroup?
+  private var server: Server?
+  private var channel: GRPCChannel?
+
+  private func generateProto(name: String, id: Int) -> Google_Protobuf_FileDescriptorProto {
```
glbrntt, line 31: This is _much_ clearer!

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,362 @@
...
+        fileDependencies[0].dependency.append(fileDescriptorProto.name)
+      }
+      fileDependencies.append(fileDescriptorProto)
+    }
+    return fileDependencies
+  }
+
+  private func makeProtosWithComplexDependencies() -> [Google_Protobuf_FileDescriptorProto] {
+    var protos: [Google_Protobuf_FileDescriptorProto] = []
+    protos.append(generateProto(name: "foo", id: 0))
```
glbrntt, line 88:
```suggestion
    protos.append(self.generateProto(name: "foo", id: 0))
```

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,362 @@
...
+    for id in 1 ... 10 {
+      var root = protos.count
```
glbrntt, line 90: `root` isn't used for anything in this block? Is that intentional?

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,362 @@
...
+      protos[id - 1].dependency.append(fileDescriptorProtoA.name)
+      protos[id - 1].dependency.append(fileDescriptorProtoB.name)
+      protos.append(fileDescriptorProtoA)
+      protos.append(fileDescriptorProtoB)
+    }
+    return protos
+  }
+
+  private func getServicesNamesFromProtos(protos: [Google_Protobuf_FileDescriptorProto]) -> [String]
+  {
```
glbrntt, line 103:
```suggestion
  private func getServicesNamesFromProtos(
    protos: [Google_Protobuf_FileDescriptorProto]
  ) -> [String] {
```

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,362 @@
...
+    var servicesNames: [String] = []
+    for proto in protos {
+      servicesNames.append(contentsOf: proto.service.map { $0.name })
+    }
+    return servicesNames
```
glbrntt, line 108: You can use a `flatMap` here:

```swift
return protos.flatMap { fileDescriptor in 
  fileDescriptor.service.map { $0.name }
}
```

or just `protos.flatMap { $0.service.map { $0.name } }` for short

---

This would actually be nicer as an extension on `Sequence` where the element is `Google_Protobuf_FileDescriptorProto`:

```swift
extension Sequence where Element == Google_Protobuf_FileDescriptorProto {
  var serviceNames: [String] {
    self.flatMap { $0.service.map { $0.name } }
  }
}
```

So then in your tests you can just call `protos.serviceNames`

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,362 @@
...
+    serviceReflectionInfo.requestStream.finish()
+    var iterator = serviceReflectionInfo.responseStream.makeAsyncIterator()
+    let message = try await iterator.next()
+    let receivedServices = message?.listServicesResponse.service.map { $0.name }
+    XCTAssertEqual(receivedServices?.count, 4)
+
+    let servicesNames = self.getServicesNamesFromProtos(protos: self.makeProtosWithDependencies())
+    for serviceName in receivedServices! {
+      XCTAssertTrue(servicesNames.contains(serviceName))
+    }
```
glbrntt, line 196: It would be better to compare the sorted names of services: if `XCTAssertTrue` fails then the message will just say it was false but expected true. We lose the information about the expected and actual service names.

If we compare the equality of two sorted arrays and they're different we can see what's missing.

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,362 @@
...
+
+  func testReflectionServiceDataFileDescriptorDataByFilename() throws {
+    var protos = self.makeProtosWithDependencies()
+    let registry = try ReflectionServiceData(fileDescriptor: protos)
+
+    let registryFileDescriptorData = registry.fileDescriptorDataByFilename
+
+    for (fileName, protoData) in registryFileDescriptorData {
+      let serializedFiledescriptorData = protoData.serializedFileDescriptorProto
+      let dependecies = protoData.dependencyNames
```
glbrntt, line 207: nit spelling: dependencies

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,362 @@
...
+
+      let originalIndex = protos.firstIndex(where: { $0.name == fileName })
+      XCTAssertNotNil(originalIndex)
+
+      let originalProto = protos[originalIndex!]
+      XCTAssertEqual(originalProto.name, fileName)
+      XCTAssertEqual(try originalProto.serializedData(), serializedFiledescriptorData)
+      XCTAssertEqual(originalProto.dependency, dependecies)
+
+      protos.remove(at: originalIndex!)
```
glbrntt, line 217: Rather than banging `originalIndex`, arrange the code so that you don't have to bang it:

```swift
guard let index = protos.firstIndex(where: { $0.name == fileName }) else {
  return XCTFail("....")
}

let proto = protos[index]
protos.remove(at: index)

// test things...

```

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,362 @@
...
+  func testReflectionServiceServicesNames() throws {
+    let protos = self.makeProtosWithDependencies()
+    var servicesNames = self.getServicesNamesFromProtos(protos: protos)
+    let registry = try ReflectionServiceData(fileDescriptor: protos)
+    let registryServices = registry.serviceNames
+    for serviceName in registryServices {
+      XCTAssert(servicesNames.contains(serviceName))
+      servicesNames.removeAll { $0 == serviceName }
+    }
+    XCTAssert(servicesNames.isEmpty)
```
glbrntt, line 231: We should just sort the names and compare arrays of names for equality

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,362 @@
...
+  func testGetSerializedFileDescriptorProtos() throws {
+    var protos = self.makeProtosWithDependencies()
+    let registry = try ReflectionServiceData(fileDescriptor: protos)
+    let serializedFileDescriptorProtos =
+      try registry
+      .getSerializedFileDescriptorProtos(fileName: "bar1.proto")
+    let fileDescriptorProtos = try serializedFileDescriptorProtos.map {
+      try Google_Protobuf_FileDescriptorProto(serializedData: $0)
+    }
+    // Tests that the functions returns all the tranzitive dependencies, with their services and
```
glbrntt, line 243:
```suggestion
    // Tests that the functions returns all the transitive dependencies, with their services and
```

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,362 @@
...
+    // methods, together with the initial proto, as serialized data.
+    XCTAssertEqual(fileDescriptorProtos.count, 4)
+    for fileDescriptorProto in fileDescriptorProtos {
+      XCTAssert(protos.contains(fileDescriptorProto))
+      let protoIndex = protos.firstIndex(of: fileDescriptorProto)
```
glbrntt, line 248: `guard` that the index isn't `nil` so we can avoid all `!` below

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,362 @@
...
+      for service in fileDescriptorProto.service {
+        XCTAssert(protos[protoIndex!].service.contains(service))
+        let serviceIndex = protos[protoIndex!].service.firstIndex(of: service)
```
glbrntt, line 251: Same here, use `guard`

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,362 @@
...
+      try registry.getSerializedFileDescriptorProtos(fileName: "bar1.proto")
+    ) { error in
+      XCTAssertEqual(
+        error as! GRPCStatus,
+        GRPCStatus(
+          code: .notFound,
+          message: "The provided file or a dependency of the provided file could not be found."
+        )
+      )
+    }
```
glbrntt, line 360: There's an interesting edge case here that you'll need to look into. I'm actually not sure if this is a valid test (it might be).

There are some protobuf messages whose generated code is distributed with the language they are for. For example SwiftProtobuf comes with a copy of `Google_Protobuf_Any` (https://github.com/apple/swift-protobuf/blob/main/Sources/SwiftProtobuf/Google_Protobuf_Any%2BExtensions.swift). If you depend on that type in the definition of another proto message then you don't need to provide the `.proto` file when generating Swift code (I believe this is true for other languages too).

So I don't know if the transitive dependencies need to include these "standard" messages or not. I would start by looking at the file descriptor proto for a message uses e.g. `Google_Protobuf_Any` and seeing what the dependencies are.

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -0,0 +1,184 @@
...
+ * limitations under the License.
+ */
+
+import DequeModule
+import Foundation
+import GRPC
+import SwiftProtobuf
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+public class ReflectionService: CallHandlerProvider {
```
glbrntt, line 23: This should be a `struct`

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -0,0 +1,184 @@
...
+  public var serviceName: Substring
```
glbrntt, line 24: This should just call through to the reflection service: 

```swift
public var serviceName: String {
  self.reflectionService.serviceName
}
```

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -0,0 +1,184 @@
...
+
+  internal var fileDescriptorDataByFilename: [String: FileDescriptorProtoData]
+  internal var serviceNames: [String]
+
+  internal init(fileDescriptors: [Google_Protobuf_FileDescriptorProto]) throws {
+    self.serviceNames = []
+    self.fileDescriptorDataByFilename = [:]
+    do {
+      for fileDescriptorProto in fileDescriptors {
+        let protoDataObj = FileDescriptorProtoData(
```
glbrntt, line 54: no need to suffix this with `Obj`, `protoData` is fine here

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -0,0 +1,184 @@
...
+          dependencyFileNames: fileDescriptorProto.dependency
+        )
+        self.fileDescriptorDataByFilename[fileDescriptorProto.name] = protoDataObj
+        self.serviceNames.append(contentsOf: fileDescriptorProto.service.map { $0.name })
+      }
+    } catch {
+      throw GRPCStatus(
+        code: .invalidArgument,
+        message:
+          "The \(fileDescriptors[self.fileDescriptorDataByFilename.count].name) could not be serialized."
```
glbrntt, line 65: This is quite indirect, we can move the scope of the `do-catch` around the code which actually requires it:

```swift
for fileDescriptorProto in fileDescriptors {
  let serializedFileDescriptorProto: Data
  do {
    serializedFileDescriptorProto = try fileDescriptorProto.serializedData()
  } catch {
    // ...
  }

  let protoDataObj = FileDescriptorProtoData(
    serializedFileDescriptorProto: serializedFileDescriptorProto,
    dependencyFileNames: fileDescriptorProto.dependency
  )
  self.fileDescriptorDataByFilename[fileDescriptorProto.name] = protoDataObj
  self.serviceNames.append(contentsOf: fileDescriptorProto.service.map { $0.name })
}
```

glbrntt: Generally speaking try to reduce the scope of do-catch blocks

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -0,0 +1,184 @@
...
+      )
+    }
+  }
+
+  public func getSerializedFileDescriptorProtos(fileName: String) throws -> [Data] {
```
glbrntt, line 70: This should be `internal`

glbrntt: I also wonder if this is a useful name: it's not clear what protos it returns.

Something like `serialisedFileDescriptorProtosForDependenciesOfFile(named fileName: String)` is more verbose but at least tells us we're returning the dependencies of the file

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -0,0 +1,184 @@
...
+    while let currentFileName = toVisit.popFirst() {
+      if let protoData = self.fileDescriptorDataByFilename[currentFileName] {
+        let serializedFileDescriptorProto = protoData.serializedFileDescriptorProto
+        toVisit.append(
+          contentsOf: protoData.dependencyFileNames
+            .filter { name in
+              return !visited.contains(name)
+            }
+        )
+        serializedFileDescriptorProtos.append(serializedFileDescriptorProto)
```
glbrntt, line 85:
```suggestion
        toVisit.append(
          contentsOf: protoData.dependencyFileNames
            .filter { name in
              return !visited.contains(name)
            }
        )
        
        let serializedFileDescriptorProto = protoData.serializedFileDescriptorProto
        serializedFileDescriptorProtos.append(serializedFileDescriptorProto)
```

Sources/GRPCReflectionService/Server/ReflectionService.swift
glbrntt: It also needs to conform to `Sendable`

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,384 @@
...
+
+  func testGetSerializedFileDescriptorProtosDependencyNotProto() throws {
+    var protos = self.makeProtosWithDependencies()
+    protos[0].dependency.append("invalidDependency")
+    let registry = try ReflectionServiceData(fileDescriptors: protos)
+    XCTAssertThrowsError(
+      try registry.getSerializedFileDescriptorProtos(fileName: "bar1.proto")
+    ) { error in
+      XCTAssertEqual(
+        error as! GRPCStatus,
```
glbrntt, line 370: No need to `as!` here, `as?` is fine, the equality check will fail if the error isn't a `GRPCStatus`. At the moment it will crash which is bad!

```suggestion
        error as? GRPCStatus,
```

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,384 @@
...
+        for messageType in fileDescriptorProto.messageType {
+          XCTAssert(protos[protoIndex].messageType.contains(messageType))
+        }
+      }
+      protos.removeAll { $0 == fileDescriptorProto }
+    }
+    XCTAssert(protos.isEmpty)
+  }
+
+  func testGetSerializedFileDescriptorProtosDependencyComplexGraph() throws {
```
glbrntt, line 281: The tests below here still use `!` for indexes which are avoidable. These need fixing up.

glbrntt (changes requested):

Sources/GRPCReflectionService/Server/ReflectionService.swift
stefanadranca: But CallHandlerProvider is a class-only protocol as far as I can tell (it conforms to AnyObject)

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,408 @@
...
+    let message = try await iterator.next()
+    let receivedServices = message?.listServicesResponse.service.map { $0.name }.sorted()
+    XCTAssertEqual(receivedServices?.count, 4)
+
+    let servicesNames = self.getServicesNamesFromProtos(
+      protos: self.makeProtosWithDependencies()
+    ).sorted()
+
+    for index in 0 ... servicesNames.count - 1 {
+      XCTAssertEqual(receivedServices![index], servicesNames[index])
```
glbrntt, line 194: another `!` here – we can avoid this one too when we get the `message`: 

```swift
guard let message = try await iterator.next() else {
  return XCTFail(...)
}
```

glbrntt, line 194: No need to compare them one-by-one: just compare the two arrays.

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,408 @@
...
+
+    for (fileName, protoData) in registryFileDescriptorData {
+      let serializedFiledescriptorData = protoData.serializedFileDescriptorProto
+      let dependencyFileNames = protoData.dependencyFileNames
+
+      guard let index = protos.firstIndex(where: { $0.name == fileName }) else {
+        return XCTFail(
+          """
+          Could not find the file descriptor proto of \(fileName)
+          in the original file descriptor protos list.
```
glbrntt, line 212: Best to keep messages as a single lines.

The `"""` format will put new lines where you've put (i.e. this will be split over two lines). We should leave splitting the text over new lines to whatever tool (e.g. Xcode) is running the test.

We can avoid the newlines here by doing:

```suggestion
          Could not find the file descriptor proto of \(fileName) \
          in the original file descriptor protos list.
```

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,408 @@
...
+
+  func testReflectionServiceServicesNames() throws {
+    let protos = self.makeProtosWithDependencies()
+    let servicesNames = self.getServicesNamesFromProtos(protos: protos).sorted()
+    let registry = try ReflectionServiceData(fileDescriptors: protos)
+    let registryServices = registry.serviceNames.sorted()
+    XCTAssertEqual(registryServices.count, servicesNames.count)
+    for index in 0 ... servicesNames.count - 1 {
+      XCTAssertEqual(registryServices[index], servicesNames[index])
+    }
```
glbrntt, line 235:
```suggestion
    XCTAssertEqual(registryServices, servicesNames)
```

Tests/GRPCTests/GRPCReflectionServiceTests/GRPCReflectionServiceTests.swift
```diff
@@ -0,0 +1,408 @@
...
+        }
+        let originalMethods = protos[protoIndex].service[serviceIndex].method
+        for method in service.method {
+          XCTAssert(originalMethods.contains(method))
+        }
+        for messageType in fileDescriptorProto.messageType {
+          XCTAssert(protos[protoIndex].messageType.contains(messageType))
+        }
+      }
+      protos.removeAll { $0 == fileDescriptorProto }
```
glbrntt, line 276: Newlines help readability, it usually helps to put one between each "block" of code

```suggestion
      guard let protoIndex = protos.firstIndex(of: fileDescriptorProto) else {
        return XCTFail(
          """
          Could not find the file descriptor proto of \(fileDescriptorProto.name)
          in the original file descriptor protos list.
          """
        )
      }
      
      for service in fileDescriptorProto.service {
        guard let serviceIndex = protos[protoIndex].service.firstIndex(of: service) else {
          return XCTFail(
            """
            Could not find the \(service.name) in the service
            list of the \(fileDescriptorProto.name) file descriptor proto.
            """
          )
        }
        
        let originalMethods = protos[protoIndex].service[serviceIndex].method
        for method in service.method {
          XCTAssert(originalMethods.contains(method))
        }
        
        for messageType in fileDescriptorProto.messageType {
          XCTAssert(protos[protoIndex].messageType.contains(messageType))
        }
      }
      
      protos.removeAll { $0 == fileDescriptorProto }
```

glbrntt (commented):
This is nearly there!

glbrntt (approved):
Great, I think this looks good now!

PR #1675, opened 2023-10-16
Added support for file-containing-symbol reflection request.

stefanadranca:
Motivation:

The file-containing-symbol request is part of the Reflection Service and enables users to find the proto file containing a symbol they specify and its transitive dependencies.

Modifications:

Added a dictionary of the fully qualified names of symbols and their corresponding file names, in the ReflectionServiceData struct. Added the function that creates the server response, after getting the file name corresponding to the symbol name and getting its tranaitive dependencies. Also, split the tests into Integration and Unit tests, and added tests for the new request.

Result:

Users of the Reflection Service will be able to get from the server the proto file that contains the symbols they are specifying in the request and its transitive dependencies.


Tests/GRPCTests/GRPCReflectionServiceTests/ReflectionServiceIntegrationTests.swift
```diff
@@ -0,0 +1,180 @@
...
+
+@testable import GRPCReflectionService
+
+final class ReflectionServiceIntegrationTests: GRPCTestCase {
+  private var group: MultiThreadedEventLoopGroup?
+  private var server: Server?
+  private var channel: GRPCChannel?
+  private var protos: [Google_Protobuf_FileDescriptorProto] {
+    makeProtosWithDependencies()
+  }
```
glbrntt, line 32: Because this is a computed property, each time we call it we call `makeProtosWithDependencies()` but the protos returned from it will never change. This can simply be `private let protos = makeProtosWithDependencies()`.

Tests/GRPCTests/GRPCReflectionServiceTests/ReflectionServiceIntegrationTests.swift
```diff
@@ -0,0 +1,180 @@
...
+
+  private var independentProto: Google_Protobuf_FileDescriptorProto {
```
glbrntt, line 34: same here

Tests/GRPCTests/GRPCReflectionServiceTests/ReflectionServiceIntegrationTests.swift
```diff
@@ -0,0 +1,180 @@
...
+import GRPC
+import GRPCReflectionService
+import NIOPosix
+import SwiftProtobuf
+import XCTest
+
+@testable import GRPCReflectionService
+
+final class ReflectionServiceIntegrationTests: GRPCTestCase {
+  private var group: MultiThreadedEventLoopGroup?
```
glbrntt, line 27: We never use this

Tests/GRPCTests/GRPCReflectionServiceTests/ReflectionServiceUnitTests.swift
```diff
@@ -23,222 +23,146 @@ import XCTest
...
-
-  override func tearDown() {
-    if let channel = self.channel {
-      XCTAssertNoThrow(try channel.close().wait())
-    }
-    if let server = self.server {
-      XCTAssertNoThrow(try server.close().wait())
+    for proto in protos {
+      let qualifiedSymbolNames = proto.qualifiedSymbolNames
+      symbolsCount = symbolsCount + qualifiedSymbolNames.count
```
glbrntt, line 76: nit: you can use `+=`

Tests/GRPCTests/GRPCReflectionServiceTests/Utils.swift
```diff
@@ -0,0 +1,117 @@
...
+ * limitations under the License.
+ */
+
+import Foundation
+import GRPC
+import GRPCReflectionService
+import NIOPosix
+import SwiftProtobuf
+
+public func generateProto(name: String, id: Int) -> Google_Protobuf_FileDescriptorProto {
```
glbrntt, line 23: None of these functions need to be public

Tests/GRPCTests/GRPCReflectionServiceTests/Utils.swift
```diff
@@ -0,0 +1,117 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+import GRPC
+import GRPCReflectionService
+import NIOPosix
+import SwiftProtobuf
```
glbrntt, line 21: Which of these imports do we actually need?

Tests/GRPCTests/GRPCReflectionServiceTests/Utils.swift
glbrntt: This name doesn't tell me anything about the proto it generates nor what the name/id are for

glbrntt (changes requested):

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -67,6 +70,21 @@ internal struct ReflectionServiceData: Sendable {
...
+        )
+        if let oldValue = oldValue {
+          throw GRPCStatus(
+            code: .alreadyExists,
+            message:
+              "The \(qualifiedSybolName) symbol from \(fileDescriptorProto.name) already exists in \(oldValue)."
+          )
+        } else {
+          self.fileNameBySymbol[qualifiedSybolName] = fileDescriptorProto.name
+        }
```
glbrntt, line 86: You don't need this `else` branch, it's covered by `updateValue`

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -99,6 +117,16 @@ internal struct ReflectionServiceData: Sendable {
     }
     return serializedFileDescriptorProtos
   }
+
+  internal func nameOfFileContainingSymbol(named symbolName: String) throws -> String {
```
glbrntt, line 121: For "basic" failure cases it makes more sense to return an optional. An error thrown from here could be anything and it's not clear to the caller why it might thrown. If this function doesn't throw and we return `nil` instead then it's quite obvious that no file was found which contained the symbol. It should be in the service where we throw a `notFound` error.

glbrntt (changes requested):

glbrntt (changes requested):

glbrntt (approved):

PR #1677, opened 2023-10-18
Implemented file-containing-extension request for Reflection Service

stefanadranca:
Motivation:

The file-containing-extension request enables users to get the file descriptor protos of the proto file containing the extension they are looking for and its transitive dependencies.

Modifications:

- Created a new struct (ExtensionDescriptor) to represent the type that is extended and the field number of each extension that exists inside the protos passed to the Reflection Service.
- Added a <ExtensionDescriptor, FileName> dictionary inside the ReflectionServiceData registry to store the extensions, avoid duplicates and retrieve nicely the file name of the proto that contains the requested extension. The file name is then used to get the serialized file descriptor protos of the proto containing the extension and its transitive dependencies.
- Added integration and unit tests.

Result:

The Reflection Service can now be used to get the serialized file descriptor protos of the proto containing the provided extension and its transitive dependencies.

### DIFF
--- a/Sources/GRPCReflectionService/Server/ReflectionService.swift
+++ b/Sources/GRPCReflectionService/Server/ReflectionService.swift
@@ -43,15 +43,21 @@ internal struct ReflectionServiceData: Sendable {
     internal var serializedFileDescriptorProto: Data
     internal var dependencyFileNames: [String]
   }
+  private struct ExtensionDescriptor: Sendable, Hashable {
+    internal let extendeeTypeName: String
+    internal let fieldNumber: Int32
+  }
 
   internal var fileDescriptorDataByFilename: [String: FileDescriptorProtoData]
   internal var serviceNames: [String]
   internal var fileNameBySymbol: [String: String]
+  private var fileNameByExtensionDescriptor: [ExtensionDescriptor: String]
 
   internal init(fileDescriptors: [Google_Protobuf_FileDescriptorProto]) throws {
     self.serviceNames = []
     self.fileDescriptorDataByFilename = [:]
     self.fileNameBySymbol = [:]
+    self.fileNameByExtensionDescriptor = [:]
 
     for fileDescriptorProto in fileDescriptors {
       let serializedFileDescriptorProto: Data
@@ -70,6 +76,8 @@ internal struct ReflectionServiceData: Sendable {
       )
       self.fileDescriptorDataByFilename[fileDescriptorProto.name] = protoData
       self.serviceNames.append(contentsOf: fileDescriptorProto.service.map { $0.name })
+
+      // Populating the <symbol, file name> dictionary.
       for qualifiedSybolName in fileDescriptorProto.qualifiedSymbolNames {
         let oldValue = self.fileNameBySymbol.updateValue(
           fileDescriptorProto.name,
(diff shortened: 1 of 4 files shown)


Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -119,6 +149,15 @@ internal struct ReflectionServiceData: Sendable {
   internal func nameOfFileContainingSymbol(named symbolName: String) -> String? {
     return self.fileNameBySymbol[symbolName]
   }
+
+  internal func nameOfFileContainingExtension(
+    extendeeTypeName extendeeName: String,
```
glbrntt, line 154:
```suggestion
    named extendeeName: String,
```

Tests/GRPCTests/GRPCReflectionServiceTests/Utils.swift
```diff
@@ -116,3 +129,9 @@ extension Sequence where Element == Google_Protobuf_EnumDescriptorProto {
     self.map { $0.name }
   }
 }
+
+extension Sequence where Element == Google_Protobuf_FieldDescriptorProto {
+  var names: [String] {
+    self.map { $0.name }
+  }
+}
```
glbrntt, line 137: This extension is so simple that I don't think it's worth defining, especially as it's only used in one place.

Tests/GRPCTests/GRPCReflectionServiceTests/ReflectionServiceUnitTests.swift
```diff
@@ -329,4 +329,86 @@ final class ReflectionServiceUnitTests: GRPCTestCase {
...
+  }
+
+  func testNameOfFileContainingExtensionsInvalidTypeName() throws {
+    let protos = makeProtosWithDependencies()
+    let registry = try ReflectionServiceData(fileDescriptors: protos)
+    let registryFileName = registry.nameOfFileContainingExtension(
+      extendeeTypeName: "InvalidType",
+      fieldNumber: 2
+    )
+    XCTAssertEqual(registryFileName, nil)
```
glbrntt, line 377: nit: use `XCTAssertNil(registryFileName)` (and elsewhere)

Tests/GRPCTests/GRPCReflectionServiceTests/ReflectionServiceIntegrationTests.swift
```diff
@@ -174,4 +174,56 @@ final class ReflectionServiceIntegrationTests: GRPCTestCase {
...
+      } else {
+        XCTAssert(
+          dependentProtos.contains(fileDescriptorProto),
+          """
+          The \(fileDescriptorProto.name) is not a dependency of the \
+          proto file containing the \"extensionInputMessage1\" symbol.
+          """
+        )
+      }
+    }
```
glbrntt, line 227: This isn't a good test because `receivedData` could be an empty array which is clearly not right! If it's non-empty it's possible that it contains only `fileToFind` or only `dependentProtos`. In all cases the test would pass but shouldn't.

glbrntt (changes requested):

glbrntt (approved):

PR #1680, opened 2023-10-20
Support all_extension_numbers_of_type reflection requests

stefanadranca:
Motivation:

The reflection service should provide the possibility for users to request the list with all the field numbers of
the extensions of a type.

Modifications:

- Implemented the dictionary that stores the arrays of integers representing the field numbers of the extensions
for each type that has extensions.
- Implemented the methods of the Reflection Service that create the specific response with the array of integers
representing the field numbers or an empty array for the case that the type doesn't have any extensions (but is a valid type).
- Implemented integration and Unit tests.

Result:

The Reflection Service will enable users to find all the extension field numbers for a specific type they requested.

### DIFF
--- a/Sources/GRPCReflectionService/Server/ReflectionService.swift
+++ b/Sources/GRPCReflectionService/Server/ReflectionService.swift
@@ -51,13 +51,21 @@ internal struct ReflectionServiceData: Sendable {
   internal var fileDescriptorDataByFilename: [String: FileDescriptorProtoData]
   internal var serviceNames: [String]
   internal var fileNameBySymbol: [String: String]
+<<<<<<< Updated upstream
   private var fileNameByExtensionDescriptor: [ExtensionDescriptor: String]
 
+=======
+  internal var fieldNumbersByType: [String: Set<Int32>]
+>>>>>>> Stashed changes
   internal init(fileDescriptors: [Google_Protobuf_FileDescriptorProto]) throws {
     self.serviceNames = []
     self.fileDescriptorDataByFilename = [:]
     self.fileNameBySymbol = [:]
+<<<<<<< Updated upstream
     self.fileNameByExtensionDescriptor = [:]
+=======
+    self.fieldNumbersByType = [:]
+>>>>>>> Stashed changes
 
     for fileDescriptorProto in fileDescriptors {
       let serializedFileDescriptorProto: Data
@@ -91,6 +99,7 @@ internal struct ReflectionServiceData: Sendable {
           )
         }
       }
+<<<<<<< Updated upstream
 
       // Populating the <extension descriptor, file name> dictionary.
       for `extension` in fileDescriptorProto.extension {
@@ -113,6 +122,15 @@ internal struct ReflectionServiceData: Sendable {
           )
         }
       }
+=======
+        
+         for `extension` in fileDescriptorProto.extension {
+            if self.fieldNumbersByType[`extension`.extendee] == nil {
+                self.fieldNumbersByType[`extension`.extendee] = Set<Int32>()
+            }
+            self.fieldNumbersByType[`extension`.extendee]?.insert(`extension`.number)
+        }
+>>>>>>> Stashed changes
     }
   }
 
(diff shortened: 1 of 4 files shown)


Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -112,8 +119,28 @@ internal struct ReflectionServiceData: Sendable {
...
           )
         }
+        if self.fieldNumbersByType[typeName] == nil {
+          self.fieldNumbersByType[typeName] = []
+        }
+        let numberPosition = self.fieldNumbersByType[typeName]!.count
+        self.fieldNumbersByType[typeName]?.insert(
+          `extension`.number,
+          at: numberPosition
+        )
```
glbrntt, line 129: A couple of notes:
1. Don't use `insert` if you want to add an element to the end of the array, use `append`
2. We don't need to check whether there's a value in the dictionary and then insert an empty array. Instead we can use https://developer.apple.com/documentation/swift/dictionary/subscript(_:default:)

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -112,8 +119,28 @@ internal struct ReflectionServiceData: Sendable {
...
+      )
+    }
+  }
+
+  internal static func extractTypeNameFrom(fullyQualifiedName name: String) -> String {
+    var nameCopy = name
+    if nameCopy.first == "." {
+      nameCopy.removeFirst()
     }
+    return nameCopy
```
glbrntt, line 143: We don't need a whole function for this, it's just a one liner:`String(name.drop(while: { $0 == "." }))`

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -151,12 +178,20 @@ internal struct ReflectionServiceData: Sendable {
...
   internal func nameOfFileContainingExtension(
-    named extendeeName: String,
+    extendeeName: String,
     fieldNumber number: Int32
   ) -> String? {
     let key = ExtensionDescriptor(extendeeTypeName: extendeeName, fieldNumber: number)
     return self.fileNameByExtensionDescriptor[key]
   }
+
+  internal func extensionsFieldNumbersOfType(named typeName: String) -> [Int32]? {
```
glbrntt, line 188: Optional collections are rarely the right thing to do. Are we sure that we should return `[Int32]?` here vs `[Int32]`?

glbrntt (changes requested):

Sources/GRPCReflectionService/Server/ReflectionService.swift
stefanadranca: I was following the pattern used for the other requests too, meaning that the functions of the registry could return nil, when the thing we are requesting for doesn't exist and the Reflection Service itself "decides" whether or not it should throw an error / send an error response. Should I change this?

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -151,12 +165,21 @@ internal struct ReflectionServiceData: Sendable {
...
   internal func nameOfFileContainingExtension(
-    named extendeeName: String,
+    extendeeName: String,
     fieldNumber number: Int32
   ) -> String? {
     let key = ExtensionDescriptor(extendeeTypeName: extendeeName, fieldNumber: number)
     return self.fileNameByExtensionDescriptor[key]
   }
+
+  // Returns nil if the type has no extensions or if it doesn't exist.
```
glbrntt, line 175: This behaviour is surprising: `nil` usually means the thing we were looking for didn't exist (e.g. consider `dictionary["foo"]`). Having it also return `nil` when the type exists but doesn't have any extensions isn't a good API because it places the onus on the caller to work out whether the type exists or not.

In this case we should probably just do the work upfront and ensure that `fieldNumbersByType` contains the right information, i.e. empty arrays when the type exists but has no extensions.

glbrntt: This is still not right: the only reason we store `messageTypeNames` is to check whether a message exists in this function, it isn't used anywhere else. 

When we create and populate `fieldNumbersByType` we should do so such it has a value for every type we know about, even if the value is an empty array. It means we do the work upfront and the semantics are clear: if a value exists then the type exists, if no value exists then we don't know about the type.

glbrntt (changes requested):

glbrntt (approved):

PR #1681, opened 2023-10-23
Vendor in NIOs lock

glbrntt:
Motivation:

For v2 we need a lock, and since we want to avoid taking on a NIO dependency in the core library we can vendor in NIOs lock type instead.

Motivation:

- Vendor in NIOs lock type and locked value box, make them internal
- Update NOTICES.txt

Result:

We have a lock type.

### DIFF
--- a/NOTICES.txt
+++ b/NOTICES.txt
@@ -21,7 +21,7 @@ This product uses scripts derived from SwiftNIO's integration testing
 framework: 'test_01_allocation_counts.sh', 'run-nio-alloc-counter-tests.sh' and
 'test_functions.sh'.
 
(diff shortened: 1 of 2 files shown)


gjcairo (approved):
LGTM, just a tiny typo, I believe.

dnadoba:
What will the minimum platform and swift version be for v2?
The standard library is in the process of getting its own lock and locked value box.

glbrntt:
> What will be the minimum platform and swift version for v2? The standard library is in the process of getting its own lock and locked value box.

Unclear yet, but that's useful info. We can revisit this at a later date.

PR #1682, opened 2023-10-24
Added support for Error Responses from the Reflection Service

stefanadranca:
Motivation:

Whenever an error occurs during the exchange between the server implementing the Reflection Service and a client, the client should receive an Error Message.

Modifications:

Replaced throwing errors with creating an Error Reponse and sending it back to the client. Also, changed the APIs to use the Result type and added new extensions for the Reflection_ServerReflectionResponse initialisation. Also modified the tests.

Result:

When an error occurs within the Reflection Service, the clients will be sent an Error Response.


Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -222,45 +242,44 @@ internal final class ReflectionServiceProvider: Reflection_ServerReflectionAsync
...
-        message: "The provided symbol could not be found."
-      )
+  ) -> Reflection_ServerReflectionResponse {
+    let nameOfFileContainingSymbolResult = self.protoRegistry.nameOfFileContainingSymbol(
+      named: symbolName
+    )
+    switch nameOfFileContainingSymbolResult {
+    case .success(let fileName):
+      return findFileByFileName(fileName, request: request)
+    case .failure(let gRPCStatus):
```
glbrntt, line 252: just call this `status`, no need to spell it out as `gRPCStatus`

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -313,34 +339,63 @@ internal final class ReflectionServiceProvider: Reflection_ServerReflectionAsync
...
   init(
     request: Reflection_ServerReflectionRequest,
-    listServicesResponse: Reflection_ListServiceResponse
+    result: Result<[Data], GRPCStatus>
   ) {
-    self = .with {
-      $0.validHost = request.host
-      $0.originalRequest = request
-      $0.listServicesResponse = listServicesResponse
+    switch result {
```
glbrntt, line 355: We have a bunch of repeated code in all of these `init`s. We can simplify this quite a lot and remove a lot of the redundancy and avoid switching in a bunch of places.

The thing that all responses have in common is the `OneOf_MessageResponse` and the original request, so we can work backwards from that. Our starting point in most cases is `Result<X, GRPCStatus>` (and if not, that's very easy to get to).

The first step we can do is getting from `X` to `Reflection_ServerReflectionResponse.OneOf_MessageResponse`, that's quite easy, we don't need helpers for that, e.g.:

```swift
let name = "Foo"
let result = self.registry.extensionsFieldNumbersOfType(named: name).map { fieldNumbers in
  Reflection_ExtensionNumberResponse.with { 
    $0.fieldNumbers = fieldNumbers
    $0.typeName = name
  }
}
```

The next step is getting from `GRPCStatus` as the `Failure` type of the `Result` to `Reflection_ServerReflectionResponse.OneOf_MessageResponse` as the `Success` type. We can do this with an extension on `Result`:

```swift
extension Result where Success == Reflection_ServerReflectionResponse.OneOf_MessageResponse, Failure == GRPCStatus {
  func recover() -> Result<Reflection_ServerReflectionResponse.OneOf_MessageResponse, Never> {
    self.flatMapError { status in
      let error = Reflection_ErrorResponse.with {
        $0.errorCode = Int32(status.code.rawValue)
        $0.errorMessage = status.message ?? ""
      }

      return .success(.errorResponse(error))
    }
  }
}
```

Now we can quite easily wrangle `Result<X, GRPCStatus>` into `Result< Reflection_ServerReflectionResponse.OneOf_MessageResponse, Never>`. 

The next bit is mapping `Reflection_ServerReflectionResponse.OneOf_MessageResponse` into a `Reflection_ServerReflectionResponse`, another extension helps us out here:

```swift
extension Result where Success == Reflection_ServerReflectionResponse.OneOf_MessageResponse {
  func attachRequest(_ request: Reflection_ServerReflectionRequest) -> Result<Reflection_ServerReflectionResponse, Failure> {
    self.map { message in
      Reflection_ServerReflectionResponse.with {
        $0.validHost = request.host
        $0.originalRequest = request
        $0.messageResponse = message
      }
    }
  }
}
```

We can actually combine the above two functions into one which also unwraps the response result into a response:

```swift
extension Result where Success == Reflection_ServerReflectionResponse.OneOf_MessageResponse, Failure == GRPCStatus {
  func makeResponse(request: Reflection_ServerReflectionRequest) -> Reflection_ServerReflectionResponse {
    let result = self.recover().attachingRequest(request)
    // Safe to '!' as the failure type is 'Never'.
    return try! result.get()
  }
}
```

Going back to the original example, this is all we need:

```swift
let name = "Foo"
let result = self.registry.extensionsFieldNumbersOfType(named: name).map { fieldNumbers in
  Reflection_ExtensionNumberResponse.with { 
    $0.fieldNumbers = fieldNumbers
    $0.typeName = name
  }
}

return result.makeResponse(request: request)
```

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -147,37 +147,59 @@ internal struct ReflectionServiceData: Sendable {
...
-        message: "The provided type is invalid."
+      return .failure(
+        GRPCStatus(
+          code: .invalidArgument,
+          message: "The provided type is invalid."
+        )
       )
     }
-    return fieldNumbers
+    return .success((typeName, fieldNumbers))
```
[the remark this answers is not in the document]
stefanadranca: We need it in the Server Response, so if we want to have an init for it with only a result type and the initial request as parameters we need to include it in the result. If not we would need a third parameter for the init

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -433,3 +405,41 @@ extension Google_Protobuf_FileDescriptorProto {
...
+
+extension Result
+where Success == Reflection_ServerReflectionResponse.OneOf_MessageResponse, Failure == GRPCStatus {
+  func makeResponse(
+    request: Reflection_ServerReflectionRequest
+  ) -> Reflection_ServerReflectionResponse {
+    let result = self.recover().attachRequest(request)
+    // Safe to '!' as the failure type is 'Never'.
+    return try! result.get()
+  }
```
glbrntt, line 444: Move this one up to where `recover()` is defined, the extension has the same requirements

glbrntt (changes requested):

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -215,53 +246,53 @@ internal final class ReflectionServiceProvider: Reflection_ServerReflectionAsync
     }
     return Reflection_ServerReflectionResponse(
       request: request,
-      listServicesResponse: listServicesResponse
+      messageResponse: Reflection_ServerReflectionResponse.OneOf_MessageResponse
+        .listServicesResponse(
+          listServicesResponse
+        )
```
glbrntt, line 252: Let's use type inference here:
```suggestion
      messageResponse: .listServicesResponse(listServicesResponse)
```

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -283,28 +314,37 @@ internal final class ReflectionServiceProvider: Reflection_ServerReflectionAsync
...
           named: typeName,
           request: request
         )
         try await responseStream.send(response)
 
       default:
-        throw GRPCStatus(code: .unimplemented)
+        let response = Reflection_ServerReflectionResponse(
+          request: request,
+          messageResponse: Reflection_ServerReflectionResponse.OneOf_MessageResponse.errorResponse(
```
glbrntt, line 340: nit: use type-inference here

```suggestion
          messageResponse: .errorResponse(
```

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -378,3 +396,32 @@ extension Google_Protobuf_FileDescriptorProto {
...
+  func recover() -> Result<Reflection_ServerReflectionResponse.OneOf_MessageResponse, Never> {
+    self.flatMapError { status in
+      let error = Reflection_ErrorResponse.with {
+        $0.errorCode = Int32(status.code.rawValue)
+        $0.errorMessage = status.message ?? ""
+      }
+      return .success(.errorResponse(error))
+    }
+  }
+  func makeResponse(
```
glbrntt, line 410: nit: leave a blank line between functions

```suggestion
  }

  func makeResponse(
```

Tests/GRPCTests/GRPCReflectionServiceTests/ReflectionServiceUnitTests.swift
```diff
@@ -117,134 +117,199 @@ final class ReflectionServiceUnitTests: GRPCTestCase {
   func testNameOfFileContainingSymbolEnum() throws {
     let protos = makeProtosWithDependencies()
     let registry = try ReflectionServiceData(fileDescriptors: protos)
-    let fileName = registry.nameOfFileContainingSymbol(named: "packagebar2.enumType2")
-    XCTAssertEqual(fileName, "bar2.proto")
+    let nameOfFileContainingSymbolResult = registry.nameOfFileContainingSymbol(
+      named: "packagebar2.enumType2"
+    )
+    switch nameOfFileContainingSymbolResult {
```
glbrntt, line 123: No need to change these now, but it would've been much easier to just do:

```swift
XCTAssertEqual(try nameOfFileContainingSymbolResult.get(), "bar2.proto")
```

Tests/GRPCTests/GRPCReflectionServiceTests/ReflectionServiceUnitTests.swift
```diff
@@ -117,134 +117,199 @@ final class ReflectionServiceUnitTests: GRPCTestCase {
   func testNameOfFileContainingSymbolEnum() throws {
```
glbrntt, line 117: It would be great to add some integration tests which validate we get appropriate error responses too. Feel free to do it in a separate PR if you prefer.

glbrntt (changes requested):
This looks much better! I left a couple of nits but this is basically good to go otherwise.

glbrntt (approved):
Nice one!

PR #1689, opened 2023-10-25
Add a retry throttle

glbrntt:
Motivation:

To implement retries and hedging, transports need to be able to throttle attempts to svoid overloading servers. The uses a token based system where successful requests, ones which end with non-retryable status code, add tokens and those which fail remove tokens from the system. Successful requests add 1 token, failed requests remove `tokenRatio` tokens (typically less than 1).

Modification:

- Implement a retry throttle
- Add a requirement to the `ClientTransport`

Result:

Retries can be throttled.


Package.swift
```diff
@@ -46,6 +46,10 @@ let packageDependencies: [Package.Dependency] = [
     url: "https://github.com/apple/swift-nio-extras.git",
     from: "1.4.0"
   ),
+  .package(
+    url: "https://github.com/apple/swift-atomics.git",
+    from: "1.1.0"
```
gjcairo, line 51: There was a 1.2.0 release last week to support Swift 5.9 - should we depend on that version instead?

Sources/GRPCCore/Transport/RetryThrottle.swift
```diff
@@ -0,0 +1,124 @@
...
+/// by ``tokenRatio`` and failed attempts decrement the available tokens by one. In the context
+/// of throttling, a failed attempt is one where the server terminates the RPC with a status code
+/// which is retryable or non fatal (as defined by ``RetryPolicy/retryableStatusCodes`` and
+/// ``HedgingPolicy/nonFatalStatusCodes``) or when the client receives a pushback response from
+/// the server.
+///
+/// See also [gRFC A6: client retries](https://github.com/grpc/proposal/blob/master/A6-client-retries.md).
+public struct RetryThrottle: Sendable {
+  // Note: only three figures after the decimal point from the original token ratio are used so
+  //   all computation is done a scaled number of tokens (tokens * 1000). This allows us to all
```
gjcairo, line 33:
```suggestion
  //   all computation is done a scaled number of tokens (tokens * 1000). This allows us to do all
```

Sources/GRPCCore/Transport/RetryThrottle.swift
```diff
@@ -0,0 +1,124 @@
...
+  /// The maximum number of tokens the throttle may hold.
+  public var maximumTokens: Int {
+    self.scaledMaximumTokens / 1000
+  }
+
+  /// The number of tokens the throttle currently has.
+  ///
+  /// If this value is less than or equal to half of ``maximumTokens`` then RPCs will not be retried
+  /// and hedging will be disabled.
+  public var tokens: Double {
```
gjcairo, line 68: Nit: maybe `availableTokens` is clearer?

Sources/GRPCCore/Transport/RetryThrottle.swift
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  /// The maximum number of tokens the throttle may hold.
+  public var maximumTokens: Int {
+    self.scaledMaximumTokens / 1000
+  }
+
+  /// The number of tokens the throttle currently has.
+  ///
+  /// If this value is less than or equal to half of ``maximumTokens`` then RPCs will not be retried
```
gjcairo, line 66:
```suggestion
  /// If this value is not above the retry threshold, then RPCs will not be retried
```

Sources/GRPCCore/Transport/RetryThrottle.swift
```diff
@@ -0,0 +1,124 @@
...
+    self.scaledRetryThreshold = scaledTokens / 2
+    self.scaledTokenRatio = scaledTokenRatio
+    self.scaledTokensAvailable = LockedValueBox(scaledTokens)
+  }
+
+  /// Records a success, adding a token to the throttle.
+  @usableFromInline
+  func recordSuccess() {
+    self.scaledTokensAvailable.withLockedValue { value in
+      value = Swift.min(self.scaledMaximumTokens, value &+ self.scaledTokenRatio)
```
gjcairo, line 110: Nit: is the `Swift.` namespace prefix needed?

Sources/GRPCCore/Transport/RetryThrottle.swift
```diff
@@ -0,0 +1,124 @@
...
+    }
+  }
+
+  /// Records a failure, removing tokens from the throttle. Returns whether retries will now be
+  /// throttled.
```
gjcairo, line 115: Nit, but I guess we prefer
```suggestion
  /// Records a failure, removing tokens from the throttle. 
  /// - Returns: Whether retries will now be throttled.
```

Package.swift
glbrntt: Good spot. It's actually not needed at all, I started implementing this with atomics but realised you can't clamp atomically so ended up with a lock...

Sources/GRPCCore/Transport/RetryThrottle.swift
glbrntt: Hmm, could it be confused with the number of tokens which are available to get (i.e. max tokens minus current tokens)?

Sources/GRPCCore/Transport/RetryThrottle.swift
```diff
@@ -0,0 +1,124 @@
...
+/// the server.
+///
+/// See also [gRFC A6: client retries](https://github.com/grpc/proposal/blob/master/A6-client-retries.md).
+public struct RetryThrottle: Sendable {
+  // Note: only three figures after the decimal point from the original token ratio are used so
+  //   all computation is done a scaled number of tokens (tokens * 1000). This allows us to do all
+  //   computation in integer space.
+
+  /// The number of tokens available, multiplied by 1000.
+  private let scaledTokensAvailable: LockedValueBox<Int>
```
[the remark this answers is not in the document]
glbrntt: Yeah, good question. So each client transport needs to instantiate a retry throttle and then just hold on to it. The call layer needs access and mutate the throttle from across threads so it needs to be thread safe. IMO it doesn't make sense for the transport to have to do the locking which is why it's here. I don't think users would reach for a lock anyway: there are no public mutating functions for them to call.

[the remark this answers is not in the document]
FranzBusch: I guess all of this makes a lot more sense when you see the larger picture. Right now the type and the semantics seem a bit confusing to me but I will see in the follow up PRs how it is used :)

FranzBusch (approved):

Sources/GRPCCore/Transport/RetryThrottle.swift
gjcairo: Yes right... okay that's fine then, I think this is clear enough.

PR #1691, opened 2023-10-30
Add benchmarks for Metadata collection

gjcairo:
## Motivation

This PR adds benchmarks (perf tests) for the Metadata collection introduced in https://github.com/grpc/grpc-swift/pull/1683

## Modifications

Benchmarks added

## Result

Benchmarks for Metadata collection are now present.


Performance/Benchmarks/Benchmarks/GRPCSwiftBenchmark/Benchmarks.swift
```diff
@@ -14,24 +14,128 @@
...
+  Benchmark("Metadata - Add string") { benchmark in
+    for _ in benchmark.scaledIterations {
+      var metadata = Metadata()
+      for i in 0..<1000 {
+          metadata.addString("\(i)", forKey: "\(i)")
+      }
+    }
+  }
+  
+  Benchmark("Metadata - Remove string") { benchmark in
```
glbrntt, line 45: `removeAllValues(forKey:)` doesn't care about the type being removed. We can remove the binary equivalent to this test and just call it "Remove all values for key" (or something along those lines)

Performance/Benchmarks/Benchmarks/GRPCSwiftBenchmark/Benchmarks.swift
```diff
@@ -14,24 +14,128 @@
...
+      var metadata = Metadata()
+      for i in 0..<1000 {
+          metadata.addString("\(i)", forKey: "\(i)")
+      }
+      
+      benchmark.startMeasurement()
+      for i in 0..<1000 {
+          metadata.removeAllValues(forKey: "\(i)")
+      }
+    }
```
glbrntt, line 56: I think this test is probably more interesting/useful if we measure adding 1000 values for a single key then removing all values for that one key.

glbrntt (changes requested):
Couple of small comments, looks good otherwise. Presumably we need to add a baseline for these too.

Performance/Benchmarks/Benchmarks/GRPCSwiftBenchmark/Benchmarks.swift
gjcairo: Yes I realised this as I was generating the thresholds - copy pasta :D

Package.swift
```diff
@@ -461,6 +461,11 @@ extension Product {
     name: grpcProductName,
     targets: [grpcTargetName]
   )
+    
+  static let grpcCore: Product = .library(
+    name: "GRPCCore",
+    targets: ["GRPCCore"]
+  )
```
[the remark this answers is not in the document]
gjcairo: Renaming the product to `_GRPCCore` works well.

Performance/Benchmarks/Benchmarks/GRPCSwiftBenchmark/Benchmarks.swift
```diff
@@ -14,24 +14,128 @@
...
+  Benchmark("Metadata - Add binary") { benchmark in
+    let value: [UInt8] = [1, 2, 3]
+    benchmark.startMeasurement()
+    for _ in benchmark.scaledIterations {
+      var metadata = Metadata()
+      for i in 0..<1000 {
+          metadata.addBinary(value, forKey: "\(i)")
+      }
+    }
+  }
```
dnadoba, line 68: Benchmark.init has an overload with a trailing setup closure which is executed once and can return some setup data: https://swiftpackageindex.com/ordo-one/package-benchmark/1.11.2/documentation/benchmark/benchmark/init(_:configuration:closure:setup:teardown:)-8kbjd

Performance/Benchmarks/Benchmarks/GRPCSwiftBenchmark/Benchmarks.swift
```diff
@@ -14,24 +14,119 @@
...
+      
+      benchmark.startMeasurement()
+      for i in 0..<1000 {
+          metadata.addBinary(value, forKey: "\(i)")
+      }
+      benchmark.stopMeasurement()
+    }
+  }
+  
+  Benchmark("Metadata_Remove_values_for_key") { benchmark in
```
glbrntt, line 58: Seems like we only need underscores in the filename:
- https://github.com/apple/swift-certificates/blob/main/Benchmarks/Benchmarks/CertificatesBenchmark/Benchmarks.swift
- https://github.com/apple/swift-certificates/tree/main/Benchmarks/Thresholds/5.9

Can we remove them here?

gjcairo: This is what I had originally (files with underscores, benchmark names with spaces) but the check against the thresholds would always pass even when it shouldn't - I could only make it fail once I started using underscores in the names too.

glbrntt (approved):

PR #1692, opened 2023-10-30
Additional keepalive diagnostics

glbrntt:
Motivation:

It's useful to know the state of keepalive when debugging connection issues. gRPC doesn't emit any logs around this at the moment which makes debugging difficult.

Modifications:

Add additional logs to the idle handler when:
- the idle timeout task is scheduled, cancelled, and fires
- the keepalive timer is scheduled
- the scheduled close timer is fired
- the connection is closed
- GOAWAY frames are sent (already logs on receive)
- PING frames are sent and received

Result:

Better visibility into connection lifecycle

### DIFF
--- a/Sources/GRPC/GRPCIdleHandler.swift
+++ b/Sources/GRPC/GRPCIdleHandler.swift
@@ -98,19 +98,6 @@ internal final class GRPCIdleHandler: ChannelInboundHandler {
     )
   }
 
-  private func sendGoAway(lastStreamID streamID: HTTP2StreamID) {
-    guard let context = self.context else {
-      return
-    }
-
-    let frame = HTTP2Frame(
-      streamID: .rootStream,
-      payload: .goAway(lastStreamID: streamID, errorCode: .noError, opaqueData: nil)
-    )
-
-    context.writeAndFlush(self.wrapOutboundOut(frame), promise: nil)
-  }
-
   private func perform(operations: GRPCIdleHandlerStateMachine.Operations) {
     // Prod the connection manager.
     if let event = operations.connectionManagerEvent, let manager = self.mode.connectionManager {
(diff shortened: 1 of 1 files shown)


Sources/GRPC/GRPCIdleHandler.swift
```diff
@@ -186,8 +187,9 @@ internal final class GRPCIdleHandler: ChannelInboundHandler {
       ()
 
     case .ack:
-      // NIO's HTTP2 handler acks for us so this is a no-op.
-      ()
+      // NIO's HTTP2 handler acks for us so this is a no-op. Log so it doesn't appear that we are
+      // ignoring pings.
+      self.stateMachine.logger.debug("sending PING frame", metadata: ["ack": "true"])
```
Lukasa, line 192: Should we be adding this metadata key to the list?

Sources/GRPC/GRPCIdleHandler.swift
```diff
@@ -137,11 +124,17 @@ internal final class GRPCIdleHandler: ChannelInboundHandler {
...
       switch idleTask {
       case let .cancel(task):
+        self.stateMachine.logger.debug("idle timeout task cancelled")
         task.cancel()
 
       case .schedule:
         if self.idleTimeout != .nanoseconds(.max), let context = self.context {
+          self.stateMachine.logger.debug(
+            "scheduling idle timeout task",
+            metadata: ["delay_ms": "\(self.idleTimeout.milliseconds)"]
```
Lukasa, line 134: Should we be adding this metadata key to the list?

Sources/GRPC/GRPCIdleHandler.swift
```diff
@@ -210,6 +218,11 @@ internal final class GRPCIdleHandler: ChannelInboundHandler {
       return
     }
 
+    self.stateMachine.logger.debug(
+      "scheduled keepalive pings",
+      metadata: ["interval_ms": "\(delay.milliseconds)"]
```
Lukasa, line 223: Should we be adding this metadata key to the list?

Sources/GRPC/GRPCIdleHandler.swift
```diff
@@ -318,6 +332,7 @@ internal final class GRPCIdleHandler: ChannelInboundHandler {
     case let .settings(.settings(settings)):
       self.perform(operations: self.stateMachine.receiveSettings(settings))
     case let .ping(data, ack):
+      self.stateMachine.logger.debug("received PING frame", metadata: ["ack": "\(ack)"])
```
Lukasa, line 335: Same here regarding metadata key.

PR #1699, opened 2023-11-01
Added ReflectionService initialiser using file paths

stefanadranca:
Motivation:

Initialising the Reflection Service using file descriptor protos is too laborious for users, so simply passing the file paths of the binary files containing the serialised file descriptor protos and letting the Service extract the data from those paths is more convenient.

Modifications:

Added a new initializer that takes in file paths and extracts and converts to file descriptor proto format the Data from the files. Also modified the integration tests to create temporary files containing the serialized file descriptor protos.

Result:

GRPC users can add the ReflectionService to their Server in a more convenient way.


Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -26,6 +26,10 @@ public final class ReflectionService: CallHandlerProvider, Sendable {
     self.reflectionService.serviceName
   }
 
+  public init(protoFilePaths: [String]) throws {
```
glbrntt, line 29: We aren't loading proto files (i.e. `.proto`) we're loading serialised file descriptor protos:

```swift
public init(serializedFileDescriptorProtoFilePaths: [String]) throws {
```

glbrntt: This also needs documentation

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -207,6 +211,27 @@ internal struct ReflectionServiceData: Sendable {
 internal final class ReflectionServiceProvider: Grpc_Reflection_V1_ServerReflectionAsyncProvider {
   private let protoRegistry: ReflectionServiceData
 
+  internal init(protoFilePaths: [String]) throws {
+
```
glbrntt, line 215: nit: unnecessary blank line
```suggestion
```

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -207,6 +211,27 @@ internal struct ReflectionServiceData: Sendable {
...
+        URL(filePath: $0, directoryHint: .notDirectory)
+      } else {
+        URL(fileURLWithPath: $0)
+      }
+      #endif
+    }
+    let binaryData = binaryFilesURLs.map { Data(base64Encoded: try! Data(contentsOf: $0)) }
+    let fileDescriptorProtos = binaryData.map {
+      try! Google_Protobuf_FileDescriptorProto(serializedData: $0!)
+    }
```
glbrntt, line 230: We can't `try!` these, they may legitimately fail and this would cause a crash.

We should just do the basic thing:

```swift
var fileDescriptorProtos = [Google_Protobuf_FileDescriptorProto]()
for path in paths {
  // ...  
  fileDescriptorProtos.append(...)
}

```

Tests/GRPCTests/GRPCReflectionServiceTests/ReflectionServiceIntegrationTests.swift
```diff
@@ -33,9 +33,28 @@ final class ReflectionServiceIntegrationTests: GRPCTestCase {
...
+    let fileDescriptorProtos = self.protos + [self.independentProto]
+    for fileDescriptorProto in fileDescriptorProtos {
+      let data = try fileDescriptorProto.serializedData()
+        .base64EncodedData()
+      #if os(Linux)
+      let temporaryDirectory = "/tmp/"
+      #else
+      let temporaryDirectory = fileManager.temporaryDirectory.path()
+      #endif
+      let filePath = temporaryDirectory + fileDescriptorProto.name
```
glbrntt, line 48: I don't think this is sufficiently unique (especially on linux): if we ran these tests in parallel we could end up removing files which another test is expecting to be there. One workaround which is probably fine is to add some randomness to the name.

Tests/GRPCTests/GRPCReflectionServiceTests/ReflectionServiceIntegrationTests.swift
```diff
@@ -33,9 +33,28 @@ final class ReflectionServiceIntegrationTests: GRPCTestCase {
   )
 
   private func setUpServerAndChannel() throws {
-    let reflectionServiceProvider = try ReflectionService(
-      fileDescriptors: self.protos + [self.independentProto]
-    )
+    var filePaths: [String] = []
```
glbrntt, line 36: We actually don't need to do this in the integration tests: we can do it in the unit tests because the actual work is done by the registry rather than the service. This also means we can do this in a separate test as well.

glbrntt (changes requested):

Tests/GRPCTests/GRPCReflectionServiceTests/ReflectionServiceIntegrationTests.swift
glbrntt: Oh, my mistake. Thinking about this some more, I think we can implement this as a couple of helpers on the `ReflectionService` which would be called by the `init`:

```swift
extension ReflectionService {
  static func readSerializedFileDescriptorProto(atPath path: String) throws -> Google_Protobuf_FileDescriptorProto { 
    // ...
  }

  static func readSerializedFileDescriptorProtos(atPaths paths: [String]) throws -> [Google_Protobuf_FileDescriptorProto] {     
    // ... 
  }
}
```

This way we can test the loading independently of the registry and operation of the service. I.e. we can test that these functions can load and deserialize the serialized descriptors correctly.

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -26,6 +26,18 @@ public final class ReflectionService: CallHandlerProvider, Sendable {
     self.reflectionService.serviceName
   }
 
+  ///  Receives as parameter an array of strings representing the file paths to the binary files containing
+  ///  the serialised file descriptor protos describing the services supported by the Server and
+  ///  discovarable through Server Reflection.
+  public init(serializedFileDescriptorProtoFilePaths: [String]) throws {
```
glbrntt, line 32: We should use a shorter label internally:

```suggestion
  public init(serializedFileDescriptorProtoFilePaths filePaths: [String]) throws {
```

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -425,3 +437,43 @@ where Success == Grpc_Reflection_V1_ServerReflectionResponse.OneOf_MessageRespon
...
+      fileURL = URL(fileURLWithPath: path)
+    }
+    #endif
+    let binaryData = try Data(contentsOf: fileURL)
+    guard let serializedData = Data(base64Encoded: binaryData) else {
+      throw GRPCStatus(
+        code: .invalidArgument,
+        message:
+          """
+          The \(fileURL.lastPathComponent) file contents could not be transformed \
```
glbrntt, line 462: We should include the whole file path. The user could have two files with the same name in different directories which would make this error unclear.

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -425,3 +437,43 @@ where Success == Grpc_Reflection_V1_ServerReflectionResponse.OneOf_MessageRespon
...
+          """
+      )
+    }
+    return try Google_Protobuf_FileDescriptorProto(serializedData: serializedData)
+  }
+
+  static func readSerializedFileDescriptorProtos(
+    atPaths paths: [String]
+  ) throws -> [Google_Protobuf_FileDescriptorProto] {
+    var fileDescriptorProtos = [Google_Protobuf_FileDescriptorProto]()
```
glbrntt, line 473: We can reserve the capacity of this array as we know how bit it'll be (this lets us avoid reallocating the array on append which could happen multiple times).
```suggestion
    var fileDescriptorProtos = [Google_Protobuf_FileDescriptorProto]()
    fileDescriptorProtos.reserveCapacity(paths.count)

```

glbrntt (changes requested):

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -26,6 +26,23 @@ public final class ReflectionService: CallHandlerProvider, Sendable {
     self.reflectionService.serviceName
   }
 
+  /// Creates a `ReflectionService` by loading serialized reflection data created by `protoc-gen-grpc-swift`.
+  ///
+  /// You can generate serialized reflection data using the `protoc-gen-grpc-swift` plugin for `protoc` by
+  /// setting the to `True`. The paths provided should be absolute or relative to the current working directory.
```
glbrntt, line 32: "setting the to `True`" ...?

glbrntt (changes requested):
Doc needs fixing but looks fine otherwise

glbrntt (approved):

PR #1703, opened 2023-11-01
Adding support for v1alpha reflection

stefanadranca:
Motivation:

v1alpha reflection is also used by GRPC users, so we should provide support for it.

Modifications:

- Generated the .grpc and .pb files for the v1alpha reflection.
- Added an extension for the ReflectionService that represents the version, that is set by the users.
- Created an enum to represent the 2 possible Reflection Service Providers and changed the methods that were calling provider's methods to switch between the possible cases.

Result:

Users will be able to set the version for the Reflection Service that they use, v1alpha becoming a possibility.


Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -21,9 +21,15 @@ import SwiftProtobuf
 
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
 public final class ReflectionService: CallHandlerProvider, Sendable {
-  private let reflectionService: ReflectionServiceProvider
+  private let reflectionServiceProvider: ProviderType
```
glbrntt, line 24: This naming doesn't need to be so verbose: we're within the `ReflectionService` so we can just use "provider".

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -21,9 +21,15 @@ import SwiftProtobuf
...
+
   public var serviceName: Substring {
-    self.reflectionService.serviceName
+    switch self.reflectionServiceProvider {
+    case .v1Provider(let reflectionV1Provider):
```
glbrntt, line 28: Same with this naming: we're switching over providers so each case doesn't need to have "provider" in the name, "v1" and "v1Alpha" are fine. Same goes for the associated value labels, `provider` is fine there too.

```swift
switch self.provider {
case .v1(let provider):
  // ...
case .v1Alpha(let provider):
  // ...
}
```

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -35,24 +41,46 @@ public final class ReflectionService: CallHandlerProvider, Sendable {
   /// - Parameter filePaths: The paths to files containing serialized reflection data.
   ///
   /// - Throws: When a file can't be read from disk or parsed.
-  public init(serializedFileDescriptorProtoFilePaths filePaths: [String]) throws {
+  public init(serializedFileDescriptorProtoFilePaths filePaths: [String], version: Version) throws {
```
glbrntt, line 44: nit: `version` needs documenting

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -412,35 +283,23 @@ extension Google_Protobuf_FileDescriptorProto {
...
+    private init(_ wrapped: Wrapped) { self.wrapped = wrapped }
 
-  func makeResponse(
-    request: Grpc_Reflection_V1_ServerReflectionRequest
-  ) -> Grpc_Reflection_V1_ServerReflectionResponse {
-    let result = self.recover().attachRequest(request)
-    // Safe to '!' as the failure type is 'Never'.
-    return try! result.get()
+    public static var v1: Self { Self(.v1) }
+    public static var v1Alpha: Self { Self(.v1Alpha) }
```
glbrntt, line 297: These need docs

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -412,35 +283,23 @@ extension Google_Protobuf_FileDescriptorProto {
...
-  {
-    self.flatMapError { status in
-      let error = Grpc_Reflection_V1_ErrorResponse.with {
-        $0.errorCode = Int32(status.code.rawValue)
-        $0.errorMessage = status.message ?? ""
-      }
-      return .success(.errorResponse(error))
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension ReflectionService {
+  public struct Version: Sendable {
```
glbrntt, line 288: This should  be `Hashable` as well

Sources/GRPCReflectionService/Server/ReflectionService.swift
```diff
@@ -412,35 +283,23 @@ extension Google_Protobuf_FileDescriptorProto {
...
 
-extension Result
-where Success == Grpc_Reflection_V1_ServerReflectionResponse.OneOf_MessageResponse {
-  func attachRequest(
-    _ request: Grpc_Reflection_V1_ServerReflectionRequest
-  ) -> Result<Grpc_Reflection_V1_ServerReflectionResponse, Failure> {
-    self.map { message in
-      Grpc_Reflection_V1_ServerReflectionResponse(request: request, messageResponse: message)
-    }
+  private enum ProviderType {
```
glbrntt, line 300: `Provider` is fine here

Tests/GRPCTests/GRPCReflectionServiceTests/ReflectionServiceV1AlphaIntegrationTests.swift
```diff
@@ -0,0 +1,349 @@
...
+
+    guard let service = receivedFileDescriptorProto.service.first else {
+      return XCTFail("The received file descriptor proto doesn't have any services.")
+    }
+    guard let method = service.method.first else {
+      return XCTFail("The service of the received file descriptor proto doesn't have any methods.")
+    }
+    XCTAssertEqual(method.name, "testMethod1")
+    XCTAssertEqual(message.fileDescriptorResponse.fileDescriptorProto.count, 4)
+  }
```
glbrntt, line 104: I don't love that we've essentially copied-and-pasted all the tests. It's going to be difficult to maintain going forward.

I think we can be a bit smarter here by having each case iterate over the two versions and having some convenience functions to switch between v1 and v1 alpha. This obviously won't compile but hopefully gives an idea of what we can do:

```suggestion
  func testFileByFileName() async throws {
    for version in versions {
      try self.setUpServerAndChannel(version: version)

      let request = V1Request.with { 
        $0.host = "127.0.0.1"
        $0.fileByFilename = "bar1.proto"
      }

      let response: V1Response?

      switch version {
      case .v1:
        let client = V1Client(channel: self.channel!)
        let call = client.makeServerReflectionInfoCall()
        try await call.send(request)
        call.finish()
        var iterator = serviceReflectionInfo.responseStream.makeAsyncIterator()
        response = try await iterator.next()
      case .v2:
        let client = V1Client(channel: self.channel!)
        let call = client.makeServerReflectionInfoCall()
        try await call.send(V1AlphaRequest(request)) // convert from v1 to alpha
        call.finish()
        response = try await iterator.next().map { V1AlphaResponse($0) } // convert from v1 to alpha
      }

      guard let message = response else {
        return XCTFail("Could not get a response message.")
      }

      let receivedFileDescriptorProto =
        try Google_Protobuf_FileDescriptorProto(
          serializedData: (message.fileDescriptorResponse
            .fileDescriptorProto[0])
        )

      XCTAssertEqual(receivedFileDescriptorProto.name, "bar1.proto")
      XCTAssertEqual(receivedFileDescriptorProto.service.count, 1)

      guard let service = receivedFileDescriptorProto.service.first else {
        return XCTFail("The received file descriptor proto doesn't have any services.")
      }
      guard let method = service.method.first else {
        return XCTFail("The service of the received file descriptor proto doesn't have any methods.")
      }
      XCTAssertEqual(method.name, "testMethod1")
      XCTAssertEqual(message.fileDescriptorResponse.fileDescriptorProto.count, 4)
    }

```

glbrntt (changes requested):

glbrntt (changes requested):
Looks good! You need to avoid leaking some resources in the tests now that we do two iterations.

Tests/GRPCTests/GRPCReflectionServiceTests/ReflectionServiceIntegrationTests.swift
```diff
@@ -66,284 +68,278 @@ final class ReflectionServiceIntegrationTests: GRPCTestCase {
...
+
   func testFileByFileName() async throws {
-    try self.setUpServerAndChannel()
-    let client = Grpc_Reflection_V1_ServerReflectionAsyncClient(channel: self.channel!)
-    let serviceReflectionInfo = client.makeServerReflectionInfoCall()
-    try await serviceReflectionInfo.requestStream.send(
-      .with {
+    try await self.forEachVersion {
+      channel,
+      version in
```
glbrntt, line 121:
```suggestion
    try await self.forEachVersion { channel, version in 
```

Tests/GRPCTests/GRPCReflectionServiceTests/ReflectionServiceIntegrationTests.swift
```diff
@@ -66,284 +68,278 @@ final class ReflectionServiceIntegrationTests: GRPCTestCase {
...
   func testListServices() async throws {
-    try self.setUpServerAndChannel()
-    let client = Grpc_Reflection_V1_ServerReflectionAsyncClient(channel: self.channel!)
-    let serviceReflectionInfo = client.makeServerReflectionInfoCall()
-
-    try await serviceReflectionInfo.requestStream.send(
-      .with {
+    try await self.forEachVersion {
+      channel,
+      version in
```
glbrntt, line 155:
```suggestion
    try await self.forEachVersion { channel, version in 
```

glbrntt (approved):

PR #1705, opened 2023-11-02
availability guard usages of async stream

rnro:


Sources/GRPCCore/Streaming/Internal/AsyncStream+MakeStream.swift
```diff
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-#if swift(<5.9)
+//#if swift(<5.9)
```
glbrntt, line 17: The `#if` needs to stay -- `makeStream` was added in 5.9, this is a backport.

glbrntt (changes requested):

Sources/GRPCCore/Streaming/Internal/AsyncStream+MakeStream.swift
rnro: Oops, I didn't mean to commit that, I was using it so I could see the call-sites

glbrntt (approved):

PR #1708, opened 2023-11-03
Add support for retries

glbrntt:
Motivation:

The `ClientRPCExecutor` currently ignores retry and hedging policies. This change adds support for retries.

Modifications:

- Add a retry executor and wire it up to the client rpc executor
- Add a few missing state transitions to the broadcasts sequence

Result:

RPC can be retried under certain conditions

### DIFF
--- a/Sources/GRPCCore/Call/Client/Internal/ClientRPCExecutor+RetryExecutor.swift
+++ b/Sources/GRPCCore/Call/Client/Internal/ClientRPCExecutor+RetryExecutor.swift
@@ -96,7 +96,7 @@ extension ClientRPCExecutor.RetryExecutor {
       if let timeout = self.timeout {
         group.addTask {
           let result = await Result {
-            try await Task.sleep(for: timeout, clock: .continuous)
+            try await Task.sleep(until: .now.advanced(by: timeout), clock: .continuous)
           }
           return .timedOut(result)
         }
@@ -257,10 +257,13 @@ extension ClientRPCExecutor.RetryExecutor {
               // If the delay is overridden with server pushback then reset the iterator for the
               // next retry.
               delayIterator = delaySequence.makeIterator()
-              try? await Task.sleep(for: delayOverride, clock: .continuous)
+              try? await Task.sleep(until: .now.advanced(by: delayOverride), clock: .continuous)
             } else {
               // The delay iterator never terminates.
-              try? await Task.sleep(for: delayIterator.next()!, clock: .continuous)
+              try? await Task.sleep(
+                until: .now.advanced(by: delayIterator.next()!),
+                clock: .continuous
+              )
             }
 
             break loop  // from the while loop so another attempt can be started.
--- a/Sources/GRPCCore/Call/Client/Internal/ClientRPCExecutor+RetryExecutor.swift
+++ b/Sources/GRPCCore/Call/Client/Internal/ClientRPCExecutor+RetryExecutor.swift
@@ -1,0 +1,303 @@
+/*
+ * Copyright 2023, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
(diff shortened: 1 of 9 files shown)


Sources/GRPCCore/Call/Client/Internal/ClientRPCExecutor+RetryExecutor.swift
```diff
@@ -0,0 +1,303 @@
...
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+extension ClientRPCExecutor.RetryExecutor {
+  @inlinable
+  func execute<R: Sendable>(
+    request: ClientRequest.Stream<Input>,
+    method: MethodDescriptor,
+    responseHandler: @Sendable @escaping (ClientResponse.Stream<Output>) async throws -> R
+  ) async throws -> R {
+    // There's quite a lot going on here...
```
gjcairo, line 74: Thanks for all the comments on this method, it makes things a lot easier to follow.

Sources/GRPCCore/Streaming/Internal/BroadcastAsyncSequence.swift
```diff
@@ -682,16 +686,18 @@ struct _BroadcastSequenceStateMachine<Element: Sendable>: Sendable {
...
+          return .resume(continuation, .failure(CancellationError()))
+        }
       }
 
       @inlinable
       mutating func cancel(
         _ id: _BroadcastSequenceStateMachine<Element>.Subscriptions.ID
       ) -> OnCancelSubscription {
-        let (removed, continuation) = self.subscriptions.removeSubscriber(withID: id)
-        assert(removed)
```
gjcairo, line 694: Are we removing this because it would sometimes fail or just because it seemed unnecessary to assert it?

gjcairo (commented):
LGTM - I think the comments make the logic pretty straightforward to follow. I wonder if it would make sense to break each task up in small methods to make it a bit more palatable, but I don't have strong opinions to be honest.

Sources/GRPCCore/Call/Client/Internal/ClientRPCExecutor+RetryExecutor.swift
```diff
@@ -0,0 +1,303 @@
...
+    // retries may be skipped if the throttle is applied.
+    let result = await withTaskGroup(
+      of: _RetryExecutorTask<R>.self,
+      returning: Result<R, Error>.self
+    ) { group in
+      // Add a task to limit the overall execution time of the RPC.
+      if let timeout = self.timeout {
+        group.addTask {
+          let result = await Result {
+            try await Task.sleep(for: timeout, clock: .continuous)
```
FranzBusch, line 99: Thinking about iOS use-cases here this might not be what people expect. If you execute a request and the user backgrounds your app you might still be interested in the result later but it might have timedout on the continuous clocked whereas it would still be okay on the suspending one.
It all depends on how the underlying transport handles this of course

Sources/GRPCCore/Call/Client/Internal/ClientRPCExecutor+RetryExecutor.swift
```diff
@@ -0,0 +1,303 @@
...
+                if retry.stream.isKnownSafeForNextSubscriber {
+                  return .retry(retryDelayOverride)
+                }
+              }
+
+              // Not retrying or not safe to retry.
+              let result = await Result {
+                // Check for cancellation; the RPC may have timed out in which case we should skip
+                // the response handler.
+                try Task.checkCancellation()
```
FranzBusch, line 217: This seems a bit odd to me. Why do we have to explicitly check here?

FranzBusch (commented):
Overall looks good to me. Just two minor comments

Sources/GRPCCore/Call/Client/Internal/ClientRPCExecutor+RetryExecutor.swift
glbrntt: In gRPC timeouts are propagated to the server (we don't do that here, yet) which will also cancel the RPC so I'm not it makes sense to make the change. We also can't do background networking anyway so if you background the app your RPC is toast anyway.

Sources/GRPCCore/Call/Client/Internal/ClientRPCExecutor+RetryExecutor.swift
glbrntt: Because nothing above is throwing: in the case of cancellation we will have a response of some kind (it may be failed already, and if it is failed the failure is constrained to `RPCError`). This is our last chance to jump into a `CancellationError` before reaching into user code.

Sources/GRPCCore/Call/Client/Internal/ClientRPCExecutor+RetryExecutor.swift
FranzBusch: > We also can't do background networking anyway so if you background the app your RPC is toast anyway.

Is that really true? If you were to build a transport on top of `URLSession` one could do background networking with the right entitlements.

FranzBusch (approved):

PR #1713, opened 2023-11-12
Add in-process client transport

gjcairo:
## Motivation
We want to have a basic in-process transport implementation, to be used for example for testing purposes.

## Modification
- Added a new `InProcessClientTransport`.
- To allow for multiple configurations to be used for each method, a new `ClientRPCExecutionConfigurationCollection` was also added.
- Finally, `ClientTransport/openStream` was replaced by `ClientTransport/withStream`.

## Result
We now have an in-process implementation of `ClientTransport`.

### DIFF
--- a/Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
+++ b/Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
@@ -20,7 +20,10 @@ public struct ClientRPCExecutionConfigurationCollection: Sendable, Hashable {
   private let defaultConfiguration: ClientRPCExecutionConfiguration
 
   public init(
-    defaultConfiguration: ClientRPCExecutionConfiguration = ClientRPCExecutionConfiguration(executionPolicy: nil, timeout: nil)
+    defaultConfiguration: ClientRPCExecutionConfiguration = ClientRPCExecutionConfiguration(
+      executionPolicy: nil,
+      timeout: nil
+    )
   ) {
     self.elements = [:]
     self.defaultConfiguration = defaultConfiguration
@@ -35,7 +38,7 @@ public struct ClientRPCExecutionConfigurationCollection: Sendable, Hashable {
       serviceLevelDescriptor.method = ""
       return self.elements[serviceLevelDescriptor, default: self.defaultConfiguration]
     }
-    
+
     set {
       if descriptor.service.isEmpty {
         preconditionFailure("Method descriptor's service cannot be empty.")
--- a/Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
+++ b/Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
@@ -18,20 +18,25 @@
 public struct ClientRPCExecutionConfigurationCollection: Sendable {
   private var elements: [MethodDescriptor: ClientRPCExecutionConfiguration]
   private let defaultConfiguration: ClientRPCExecutionConfiguration
-  
+
   public init(defaultConfiguration: ClientRPCExecutionConfiguration) {
     self.elements = [:]
     self.defaultConfiguration = defaultConfiguration
   }
-  
-  public mutating func addConfiguration(_ configuration: ClientRPCExecutionConfiguration, forMethod descriptor: MethodDescriptor) {
+
+  public mutating func addConfiguration(
+    _ configuration: ClientRPCExecutionConfiguration,
+    forMethod descriptor: MethodDescriptor
+  ) {
     self.elements[descriptor] = configuration
   }
-  
-  public func getConfiguration(forMethod descriptor: MethodDescriptor) -> ClientRPCExecutionConfiguration {
+
+  public func getConfiguration(
+    forMethod descriptor: MethodDescriptor
+  ) -> ClientRPCExecutionConfiguration {
     self.elements[descriptor, default: self.defaultConfiguration]
   }
-  
+
   public subscript(_ descriptor: MethodDescriptor) -> ClientRPCExecutionConfiguration {
     self.getConfiguration(forMethod: descriptor)
   }
--- a/Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
+++ b/Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
@@ -15,6 +15,15 @@
  */
 
 @available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+
+/// A collection of ``ClientRPCExecutionConfiguration``s, mapped to specific methods or services.
+///
+/// When creating a new instance, you must provide a default configuration to be used when getting
+/// a configuration for a method that has not been given a specific override.
+/// Use ``setDefaultConfiguration(_:forService:)`` to set a specific override for a whole
+/// service.
+///
+/// Use the subscript to get and set configurations for methods.
 public struct ClientRPCExecutionConfigurationCollection: Sendable, Hashable {
   private var elements: [MethodDescriptor: ClientRPCExecutionConfiguration]
   private let defaultConfiguration: ClientRPCExecutionConfiguration
@@ -40,10 +49,28 @@ public struct ClientRPCExecutionConfigurationCollection: Sendable, Hashable {
     }
 
     set {
-      if descriptor.service.isEmpty {
-        preconditionFailure("Method descriptor's service cannot be empty.")
-      }
+      precondition(
+        !descriptor.service.isEmpty,
+        "Method descriptor's service cannot be empty."
+      )
+      
       self.elements[descriptor] = newValue
     }
+  }
+  
+  /// Set a default configuration for a service.
+  ///
+  /// If getting a configuration for a method that's part of a service, and the method itself doesn't have an
+  /// override, then this configuration will be used instead of the default configuration passed when creating
+  /// this instance of ``ClientRPCExecutionConfigurationCollection``.
+  ///
+  /// - Parameters:
+  ///   - configuration: The default configuration for the service.
+  ///   - service: The name of the service for which this override applies.
+  mutating public func setDefaultConfiguration(
+    _ configuration: ClientRPCExecutionConfiguration,
+    forService service: String
+  ) {
+    self[MethodDescriptor(service: service, method: "")] = configuration
   }
 }
--- a/Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
+++ b/Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
@@ -15,32 +15,24 @@
  */
 
 @available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
-struct ClientRPCExecutionConfigurationCollection: ExpressibleByDictionaryLiteral {
+public struct ClientRPCExecutionConfigurationCollection: Sendable {
   private var elements: [MethodDescriptor: ClientRPCExecutionConfiguration]
   private let defaultConfiguration: ClientRPCExecutionConfiguration
   
-  init(defaultConfiguration: ClientRPCExecutionConfiguration) {
+  public init(defaultConfiguration: ClientRPCExecutionConfiguration) {
     self.elements = [:]
     self.defaultConfiguration = defaultConfiguration
   }
   
-  init(dictionaryLiteral elements: (Key, Value)...) {
-    elements.forEach({ (key, value) in
-      self.elements[key] = value
-    })
-  }
-  
-  mutating func addConfiguration(_ configuration: ClientRPCExecutionConfiguration, forMethod descriptor: MethodDescriptor) {
+  public mutating func addConfiguration(_ configuration: ClientRPCExecutionConfiguration, forMethod descriptor: MethodDescriptor) {
     self.elements[descriptor] = configuration
   }
   
-  func getConfiguration(forMethod descriptor: MethodDescriptor) -> ClientRPCExecutionConfiguration {
-    self.elements[descriptor] ?? self.defaultConfiguration
+  public func getConfiguration(forMethod descriptor: MethodDescriptor) -> ClientRPCExecutionConfiguration {
+    self.elements[descriptor, default: self.defaultConfiguration]
   }
   
-  subscript(_ descriptor: MethodDescriptor) -> ClientRPCExecutionConfiguration {
+  public subscript(_ descriptor: MethodDescriptor) -> ClientRPCExecutionConfiguration {
     self.getConfiguration(forMethod: descriptor)
   }
-  
-  
 }
--- a/Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
+++ b/Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
@@ -67,7 +67,7 @@ public struct ClientRPCExecutionConfigurationCollection: Sendable, Hashable {
   /// - Parameters:
   ///   - configuration: The default configuration for the service.
   ///   - service: The name of the service for which this override applies.
-  mutating public func setDefaultConfiguration(
+  public mutating func setDefaultConfiguration(
     _ configuration: ClientRPCExecutionConfiguration,
     forService service: String
   ) {
--- a/Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
+++ b/Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
@@ -53,11 +53,11 @@ public struct ClientRPCExecutionConfigurationCollection: Sendable, Hashable {
         !descriptor.service.isEmpty,
         "Method descriptor's service cannot be empty."
       )
-      
+
       self.elements[descriptor] = newValue
     }
   }
-  
+
   /// Set a default configuration for a service.
   ///
   /// If getting a configuration for a method that's part of a service, and the method itself doesn't have an
--- a/Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
+++ b/Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
@@ -15,29 +15,32 @@
  */
 
 @available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
-public struct ClientRPCExecutionConfigurationCollection: Sendable {
+public struct ClientRPCExecutionConfigurationCollection: Sendable, Hashable {
   private var elements: [MethodDescriptor: ClientRPCExecutionConfiguration]
   private let defaultConfiguration: ClientRPCExecutionConfiguration
 
-  public init(defaultConfiguration: ClientRPCExecutionConfiguration) {
+  public init(
+    defaultConfiguration: ClientRPCExecutionConfiguration = ClientRPCExecutionConfiguration(executionPolicy: nil, timeout: nil)
+  ) {
     self.elements = [:]
     self.defaultConfiguration = defaultConfiguration
   }
 
-  public mutating func addConfiguration(
-    _ configuration: ClientRPCExecutionConfiguration,
-    forMethod descriptor: MethodDescriptor
-  ) {
-    self.elements[descriptor] = configuration
-  }
-
-  public func getConfiguration(
-    forMethod descriptor: MethodDescriptor
-  ) -> ClientRPCExecutionConfiguration {
-    self.elements[descriptor, default: self.defaultConfiguration]
-  }
-
   public subscript(_ descriptor: MethodDescriptor) -> ClientRPCExecutionConfiguration {
-    self.getConfiguration(forMethod: descriptor)
+    get {
+      if let methodLevelOverride = self.elements[descriptor] {
+        return methodLevelOverride
+      }
+      var serviceLevelDescriptor = descriptor
+      serviceLevelDescriptor.method = ""
+      return self.elements[serviceLevelDescriptor, default: self.defaultConfiguration]
+    }
+    
+    set {
+      if descriptor.service.isEmpty {
+        preconditionFailure("Method descriptor's service cannot be empty.")
+      }
+      self.elements[descriptor] = newValue
+    }
   }
 }
--- a/Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
+++ b/Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2023, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+struct ClientRPCExecutionConfigurationCollection: ExpressibleByDictionaryLiteral {
+  private var elements: [MethodDescriptor: ClientRPCExecutionConfiguration]
+  private let defaultConfiguration: ClientRPCExecutionConfiguration
+  
+  init(defaultConfiguration: ClientRPCExecutionConfiguration) {
+    self.elements = [:]
+    self.defaultConfiguration = defaultConfiguration
+  }
+  
+  init(dictionaryLiteral elements: (Key, Value)...) {
+    elements.forEach({ (key, value) in
+      self.elements[key] = value
+    })
+  }
+  
+  mutating func addConfiguration(_ configuration: ClientRPCExecutionConfiguration, forMethod descriptor: MethodDescriptor) {
+    self.elements[descriptor] = configuration
+  }
+  
+  func getConfiguration(forMethod descriptor: MethodDescriptor) -> ClientRPCExecutionConfiguration {
+    self.elements[descriptor] ?? self.defaultConfiguration
+  }
+  
+  subscript(_ descriptor: MethodDescriptor) -> ClientRPCExecutionConfiguration {
+    self.getConfiguration(forMethod: descriptor)
+  }
+  
+  
+}
--- a/Sources/GRPCCore/Transport/InProcessClientTransport.swift
+++ b/Sources/GRPCCore/Transport/InProcessClientTransport.swift
@@ -203,11 +203,6 @@ public struct InProcessClientTransport: ClientTransport {
     maybeContinuation?.finish()
   }
 
-  private enum WithStreamResult {
-    case success(_ streamID: Int)
-    case pending(AsyncStream<Void>)
-  }
-
   /// Opens a stream using the transport, and uses it as input into a user-provided closure.
   ///
   /// - Important: The opened stream is closed after the closure is finished.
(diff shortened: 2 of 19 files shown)


Sources/GRPCCore/Transport/InProcessClientTransport.swift
```diff
@@ -0,0 +1,153 @@
...
+  /// are no longer required by the caller who signals this by calling ``close()``, or by cancelling the
+  /// task this function runs in.
+  ///
+  /// - Parameter lazily: This parameter is ignored in this implementation.
+  public func connect(lazily: Bool) async throws {
+    try self.state.withLockedValue { state in
+      switch state {
+      case .unconnected(let server):
+        state = .connected(server)
+        _ = server.listen()
```
glbrntt, line 60: I'm not sure we should be calling `listen()`, this is the responsibility of the server.

Sources/GRPCCore/Transport/InProcessClientTransport.swift
```diff
@@ -0,0 +1,153 @@
...
+      outbound: request.writer
+    )
+
+    let serverStream = RPCStream(
+      descriptor: descriptor,
+      inbound: request.stream,
+      outbound: response.writer
+    )
+
+    let error: RPCError? = try self.state.withLockedValue { state in
```
glbrntt, line 109: At the moment we have two error paths: this is throwing _and_ returns an `RPCError?`. We should only have `RPCError?` here, I think. For non `RPCError`s thrown by `acceptStream` we can return an `.unknown` status code.

Alternatively we can return a `Result<Void, RPCError>` here which is roughly the same as `RPCError?`

Sources/GRPCCore/Transport/InProcessClientTransport.swift
```diff
@@ -0,0 +1,153 @@
...
+
+  public typealias Inbound = RPCAsyncSequence<RPCResponsePart>
+  public typealias Outbound = RPCWriter<RPCRequestPart>.Closable
+
+  public var retryThrottle: RetryThrottle
+
+  private let executionConfigurations: ClientRPCExecutionConfigurationCollection
+  private let state: LockedValueBox<State>
+
+  public init(
```
glbrntt, line 34: Does the server transport support multiple clients?

Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
```diff
@@ -0,0 +1,43 @@
...
+    forMethod descriptor: MethodDescriptor
+  ) {
+    self.elements[descriptor] = configuration
+  }
+
+  public func getConfiguration(
+    forMethod descriptor: MethodDescriptor
+  ) -> ClientRPCExecutionConfiguration {
+    self.elements[descriptor, default: self.defaultConfiguration]
+  }
```
glbrntt, line 38: I think we can remove both of these in favour of the subscript API.

Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+public struct ClientRPCExecutionConfigurationCollection: Sendable {
+  private var elements: [MethodDescriptor: ClientRPCExecutionConfiguration]
+  private let defaultConfiguration: ClientRPCExecutionConfiguration
+
+  public init(defaultConfiguration: ClientRPCExecutionConfiguration) {
```
glbrntt, line 22: We should default the default configuration so that it has no timeout and no execution policy.

Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
```diff
@@ -0,0 +1,43 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+public struct ClientRPCExecutionConfigurationCollection: Sendable {
+  private var elements: [MethodDescriptor: ClientRPCExecutionConfiguration]
```
glbrntt, line 19: I think we need to be able to support setting a default configuration at a per-service level as well. We can do that by having setting the method name to `""` in the `MethodDescriptor`. That way we can look up for a specific method and if that's `nil` look up the service specific configuration and then fallback to the default configuration.

Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+public struct ClientRPCExecutionConfigurationCollection: Sendable {
```
glbrntt, line 18: I really don't like this name (I'm also not a huge fan of `ClientRPCExecutionConfiguration`). 

What we're really shadowing with these types is ServiceConfig and MethodConfig:
- https://github.com/grpc/grpc/blob/master/doc/service_config.md
- https://github.com/grpc/grpc-proto/blob/master/grpc/service_config/service_config.proto

At the moment `ClientRPCExecutionConfiguration` is a subset of MethodConfig which we'll likely need to expand and this type is a container for that.

I'm wondering if when we add a `Client` this should be nested within that type as `MethodConfigurationRegistry` or something and turn `ClientRPCExecutionConfiguration` into `Client.MethodConfiguration`.

WDYT?

glbrntt, line 18: Also this can be `Hashable` and needs docs...

Tests/GRPCCoreTests/Call/Client/ClientRPCExecutionConfigurationCollectionTests.swift
```diff
@@ -0,0 +1,63 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import GRPCCore
+import XCTest
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+final class ClientRPCExecutionConfigurationCollectionTests: XCTestCase {
+  func testCollection() {
```
glbrntt, line 21: Can we split this into smaller tests? One to check we get a specific configuration for a known method and another which gets the default configuration?

Tests/GRPCCoreTests/Call/Client/Internal/ClientRPCExecutorTests.swift
```diff
@@ -150,12 +150,12 @@ final class ClientRPCExecutorTests: XCTestCase {
...
-      XCTAssertThrowsRPCError(try response.message) {
-        XCTAssertEqual($0.code, .aborted)
-      }
+    await XCTAssertThrowsRPCErrorAsync({
+      try await tester.unary(
+        request: ClientRequest.Single(message: [1, 2, 3], metadata: ["foo": "bar"])
+      ) { _ in }
+    }) {
+      XCTAssertEqual($0.code, .aborted)
     }
```
glbrntt, line 159: nit: use multiple trailing closure syntax (might be `errorHandler` instead of `onError`, don't remember off the top of my head)

```suggestion
    await XCTAssertThrowsRPCErrorAsync {
      try await tester.unary(
        request: ClientRequest.Single(message: [1, 2, 3], metadata: ["foo": "bar"])
      ) { _ in }
    } onError: {
      XCTAssertEqual($0.code, .aborted)
    }
```

Tests/GRPCCoreTests/Call/Client/Internal/ClientRPCExecutorTests.swift
```diff
@@ -169,14 +169,14 @@ final class ClientRPCExecutorTests: XCTestCase {
...
 
-    try await tester.clientStreaming(
-      request: ClientRequest.Stream(metadata: ["foo": "bar"]) {
-        try await $0.write([1, 2, 3])
-      }
-    ) { response in
-      XCTAssertThrowsRPCError(try response.message) {
-        XCTAssertEqual($0.code, .aborted)
-      }
+    await XCTAssertThrowsRPCErrorAsync({
```
glbrntt, line 172: nit: same here, use multiple trailing closure syntax

Tests/GRPCCoreTests/Test Utilities/Transport/AnyTransport.swift
```diff
@@ -21,20 +21,20 @@ struct AnyClientTransport: ClientTransport, Sendable {
   typealias Outbound = RPCWriter<RPCRequestPart>.Closable
 
   private let _retryThrottle: @Sendable () -> RetryThrottle
-  private let _openStream: @Sendable (MethodDescriptor) async throws -> RPCStream<Inbound, Outbound>
+  private let _withStream:
+    @Sendable (MethodDescriptor, (RPCStream<Inbound, Outbound>) async throws -> Any) async throws ->
+      Any
```
glbrntt, line 26:
```suggestion
  private let _withStream: @Sendable (
    _ method: MethodDescriptor, 
    _ body: (RPCStream<Inbound, Outbound>) async throws -> Any
  ) async throws -> Any
```

Sources/GRPCCore/Transport/InProcessClientTransport.swift
```diff
@@ -0,0 +1,153 @@
...
+  /// Connections may be added or removed over time as required by the implementation and the
+  /// demand for streams by the client.
+  ///
+  /// Implementations of this function will typically create a long-lived task group which
+  /// maintains connections. The function exits when all open streams have been closed and new connections
+  /// are no longer required by the caller who signals this by calling ``close()``, or by cancelling the
+  /// task this function runs in.
+  ///
+  /// - Parameter lazily: This parameter is ignored in this implementation.
+  public func connect(lazily: Bool) async throws {
```
glbrntt, line 55: This returns quickly but should only return when the connection is closed and all streams are handled.

You'll need to manage a bit more state here and I suspect the way to go is to do the connect and create and store a continuation which indicates when the `connect` can return.

glbrntt (changes requested):
Good start. I left a bunch of feedback.

Sources/GRPCCore/Transport/InProcessClientTransport.swift
gjcairo: Hm, it just felt like the right thing to do if we call `stopListening()` on `close()` - although this really only makes sense if there can only be one client per server, really. You think we should avoid calling both start and stop listening from the client?

Sources/GRPCCore/Transport/InProcessClientTransport.swift
gjcairo: Currently, yes, there's nothing on the server that prevents more than one client from using the same server

Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
gjcairo: I agree, the name is bad. I'm unsure about renaming `ClientRPCExecutionConfiguration` to `MethodConfiguration` though, because, with the change you suggested in another comment to fall back to a service configuration by setting the descriptor's method name to `""`, the `MethodConfiguration` could also be referring to a service configuration, which sounds wrong. 

I think nesting the types inside the `Client` is fine though. But maybe it would make sense to have the same hierarchy `ServiceConfiguration`s containing `MethodConfiguration`s, like in the service_config proto file you shared?

Sources/GRPCCore/Transport/InProcessClientTransport.swift
glbrntt: I do, yes. Users shouldn't use these types directly, they should use them via their client and server which will have `run()` methods which will call through to these methods under the hood. Coupling the client and server in this way feels a bit wrong _and_ it puts us in a weird spot with multiple client transports as you noted.

Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
glbrntt: > I agree, the name is bad. I'm unsure about renaming `ClientRPCExecutionConfiguration` to `MethodConfiguration` though, because, with the change you suggested in another comment to fall back to a service configuration by setting the descriptor's method name to `""`, the `MethodConfiguration` could also be referring to a service configuration, which sounds wrong.

The configuration returned for a method name of `""` isn't service configuration: it would be the default method configuration for the service.

The service configuration is different still: it includes things like the retry throttle configuration _and_ any method configuration.

> I think nesting the types inside the `Client` is fine though. But maybe it would make sense to have the same hierarchy `ServiceConfiguration`s containing `MethodConfiguration`s, like in the service_config proto file you shared?

Yes, I think that'd be sensible.

gjcairo: > it would be the default method configuration for the service
Fair enough :D

Should I leave the names as they are in this PR and make the whole change when I nest the types under `Client`, then?

glbrntt: Yeah, we can leave them for now.

Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
```diff
@@ -0,0 +1,49 @@
...
+      }
+      var serviceLevelDescriptor = descriptor
+      serviceLevelDescriptor.method = ""
+      return self.elements[serviceLevelDescriptor, default: self.defaultConfiguration]
+    }
+
+    set {
+      if descriptor.service.isEmpty {
+        preconditionFailure("Method descriptor's service cannot be empty.")
+      }
```
glbrntt, line 45: This can just be a precondition: `precondition(!descriptor.service.isEmpty, "...")`

Sources/GRPCCore/Call/Client/ClientRPCExecutionConfigurationCollection.swift
```diff
@@ -0,0 +1,49 @@
...
+    self.defaultConfiguration = defaultConfiguration
+  }
+
+  public subscript(_ descriptor: MethodDescriptor) -> ClientRPCExecutionConfiguration {
+    get {
+      if let methodLevelOverride = self.elements[descriptor] {
+        return methodLevelOverride
+      }
+      var serviceLevelDescriptor = descriptor
+      serviceLevelDescriptor.method = ""
```
glbrntt, line 38: I think we need a way to set a service level configuration too:  `setDefaultConfiguration(_:forService:)`

Sources/GRPCCore/Transport/InProcessClientTransport.swift
```diff
@@ -0,0 +1,282 @@
...
+      // finished - whichever happens first.
+      // The continuation will be finished when `close()` is called and there
+      // are no more open streams.
+    }
+  }
+
+  /// Signal to the transport that no new streams may be created.
+  ///
+  /// Existing streams may run to completion naturally but calling ``withStream(descriptor:_:)``
+  /// should result in an ``RPCError`` with code ``RPCError/Code/failedPrecondition`` being thrown.
```
glbrntt, line 117: We're implementing this here so we can be more concrete: 

```suggestion
  /// will result in an ``RPCError`` with code ``RPCError/Code/failedPrecondition`` being thrown.
```

Sources/GRPCCore/Transport/InProcessClientTransport.swift
```diff
@@ -0,0 +1,282 @@
...
+  }
+
+  /// Opens a stream using the transport, and uses it as input into a user-provided closure.
+  ///
+  /// - Important: The opened stream is closed after the closure is finished.
+  ///
+  /// Transport implementations should throw an ``RPCError`` with the following error codes:
+  /// - ``RPCError/Code/failedPrecondition`` if the transport is closing or has been closed.
+  /// - ``RPCError/Code/unavailable`` if it's temporarily not possible to create a stream and it
+  ///   may be possible after some backoff period.
```
glbrntt, line 148: We can be more concrete here and talk about _this_ implementation.

Sources/GRPCCore/Transport/InProcessClientTransport.swift
```diff
@@ -0,0 +1,282 @@
...
+      switch state {
+      case .connected(let serverTransport, let openStreams, let signalEndConnection):
+        do {
+          try serverTransport.acceptStream(serverStream)
+          state = .connected(serverTransport, openStreams + 1, signalEndConnection)
+        } catch let acceptStreamError as RPCError {
+          return .failure(acceptStreamError)
+        } catch {
+          return .failure(RPCError(code: .unknown, message: "Unknown error: \(error)."))
+        }
```
glbrntt, line 187: (No need to change what you already have but an FYI in case you find it useful in the future)

We have an extension on `Result` to cast the error to a given type (or create one if the error is not of that type). I find it pretty handy when using result types.

```swift
return await Result { 
  ... 
}.castError(to: RPCError.self) {
  RPCError(...)
}
```

Sources/GRPCCore/Transport/InProcessClientTransport.swift
```diff
@@ -0,0 +1,282 @@
...
+            signalEndConnection?.finish()
+          }
+        }
+      }
+    }
+
+    let userResult = try await closure(clientStream)
+
+    serverStream.outbound.finish()
+    clientStream.outbound.finish()
```
glbrntt, line 268: Should we be doing this in the `defer` instead?

Sources/GRPCCore/Transport/InProcessClientTransport.swift
```diff
@@ -0,0 +1,282 @@
...
+        )
+      }
+    }
+
+    for await _ in stream {
+      // This for-await loop will exit (and thus `connect(lazily:)` will return)
+      // only when the task is cancelled, or when the stream's continuation is
+      // finished - whichever happens first.
+      // The continuation will be finished when `close()` is called and there
+      // are no more open streams.
```
glbrntt, line 110: Hmm, in the case of cancellation I don't think we'll actually close the open streams?

Sources/GRPCCore/Transport/InProcessClientTransport.swift
```diff
@@ -0,0 +1,282 @@
...
+/// when the containing task is cancelled.
+///
+/// To execute requests using this client, use ``withStream(descriptor:_:)``. If this function is
+/// called before ``connect(lazily:)`` is called, then any streams will remain pending and the call will
+/// block until ``connect(lazily:)`` is called or the task is cancelled.
+///
+/// - SeeAlso: ``ClientTransport``
+public struct InProcessClientTransport: ClientTransport {
+  private enum State: Sendable {
+    case unconnected(
```
glbrntt, line 39: nit: prefer using state `struct`s, they're a bit better for readability (e.g. if you had two properties of the same type in an enum you could mix them up). Less important here are there aren't many states, but you can also encode valid transitions by having `init`s which take valid previous states.

glbrntt (changes requested):

Sources/GRPCCore/Transport/InProcessClientTransport.swift
gjcairo: Yeah you're right, good catch

Tests/GRPCCoreTests/Transport/InProcessClientTransportTests.swift
```diff
@@ -0,0 +1,330 @@
...
+      maximumAttempts: 10,
+      initialBackoff: .seconds(1),
+      maximumBackoff: .seconds(1),
+      backoffMultiplier: 1.0,
+      retryableStatusCodes: [.unavailable]
+    )
+    let client = InProcessClientTransport(
+      server: .init(),
+      executionConfigurations: .init(defaultConfiguration: .init(retryPolicy: retryPolicy))
+    )
```
glbrntt, line 33: This is in every test and adds quite a lot of noise, can we factor it out to a helper? `private func makeClient(configuration: Configuration? = nil, server: InProcessServerTransport = InProcessServerTransport())`

Tests/GRPCCoreTests/Transport/InProcessClientTransportTests.swift
```diff
@@ -0,0 +1,330 @@
...
+
+    XCTAssertNoThrow(client.close())
+  }
+
+  func testCloseWhenClosed() {
+    func testCloseWhenUnconnected() {
```
glbrntt, line 124: What's going on here? 🙃

Tests/GRPCCoreTests/Transport/InProcessClientTransportTests.swift
```diff
@@ -0,0 +1,330 @@
...
+      executionConfigurations: .init(defaultConfiguration: .init(retryPolicy: retryPolicy))
+    )
+
+    try await withThrowingTaskGroup(of: Void.self) { group in
+      group.addTask {
+        try await client.connect(lazily: false)
+      }
+      group.addTask {
+        try await Task.sleep(for: .milliseconds(100))
+      }
```
glbrntt, line 161: Why do we need this sleep? Can't we just connect and then close?

Tests/GRPCCoreTests/Transport/InProcessClientTransportTests.swift
```diff
@@ -0,0 +1,330 @@
...
+        try await client.connect(lazily: false)
+      }
+
+      group.addTask {
+        try await client.withStream(descriptor: .init(service: "test", method: "test")) { stream in
+          try await stream.outbound.write(.message([1]))
+          for try await response in stream.inbound {
+            guard case .message(let message) = response else {
+              XCTFail()
+              fatalError()
```
glbrntt, line 251: Let's avoid crashing in tests

Tests/GRPCCoreTests/Transport/InProcessClientTransportTests.swift
```diff
@@ -0,0 +1,330 @@
...
+            }
+            receivedMessages.withLockedValue({ $0.append(message) })
```
glbrntt, line 253: I think this over complicates things a bit, we can just collect all inbound messages (`stream.inbound.collect()`) and check we receive the messages we expected. You'll need to close the outbound stream too.

Tests/GRPCCoreTests/Transport/InProcessClientTransportTests.swift
```diff
@@ -0,0 +1,330 @@
...
+        }
+      }
+
+      group.addTask {
+        for try await stream in server.listen() {
+          try await stream.outbound.write(
+            contentsOf: stream.inbound.map({ requestPart in
+              guard case .message(let message) = requestPart else {
+                XCTFail()
+                fatalError()
```
glbrntt, line 264: Let's avoid this fatal error

Tests/GRPCCoreTests/Transport/InProcessClientTransportTests.swift
```diff
@@ -0,0 +1,330 @@
...
+      group.addTask {
+        try await Task.sleep(for: .milliseconds(100))
+        client.close()
+        await XCTAssertThrowsRPCErrorAsync {
+          try await client.withStream(descriptor: .init(service: "test", method: "test")) { _ in }
+        } errorHandler: { error in
+          XCTAssertEqual(error.code, .failedPrecondition)
+          XCTAssertEqual(error.message, "The client transport is closed.")
+        }
+
```
glbrntt, line 282: I don't think this part is necessary, it has its own test elsewhere.

Tests/GRPCCoreTests/Transport/InProcessClientTransportTests.swift
```diff
@@ -0,0 +1,330 @@
...
+      group.cancelAll()
+    }
+
+    await XCTAssertThrowsRPCErrorAsync {
+      try await client.connect(lazily: false)
+    } errorHandler: { error in
+      XCTAssertEqual(error.code, .failedPrecondition)
+      XCTAssertEqual(error.message, "Can't connect to server, transport is closed.")
+    }
+
```
glbrntt, line 297: Same with this part: it's not necessary as it has its own test elsewhere

glbrntt (changes requested):
This looks great, there are a couple of places where you call out while holding a lock which need fixing up before we can merge this though.

Tests/GRPCCoreTests/Transport/InProcessClientTransportTests.swift
gjcairo: Botched copy pasting 😓

Tests/GRPCCoreTests/Transport/InProcessClientTransportTests.swift
gjcairo: Connect won't return until `closed` is called so we must do this in separate tasks. The sleep is there to make sure that there isn't a race and the state change in `connect` happens before we call `close`

Sources/GRPCCore/Transport/InProcessClientTransport.swift
```diff
@@ -0,0 +1,350 @@
...
+    // occurred and all open streams must be closed at this point.
+    self.state.withLockedValue { state in
+      switch state {
+      case .unconnected:
+        ()
+      case .connected(let connectedState):
+        state = .closed(.init())
+        for (clientStream, serverStream) in connectedState.openStreams.values {
+          clientStream.outbound.finish()
+          serverStream.outbound.finish()
```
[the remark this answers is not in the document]
gjcairo: In the `withStream` function, we create both the client and server streams. The server will get the server-side stream via the `acceptStream` call, but we are currently closing both the client and server outbounds when the user-provided closure is done executing.

If we close the client stream only, the closure should finish and that will cause the server stream to also be closed via the `defer` block in `withStream` - is this what you're suggesting, or am I misunderstanding your assumption?

Sources/GRPCCore/Transport/InProcessClientTransport.swift
```diff
@@ -0,0 +1,350 @@
...
+  ///
+  /// If you want to forcefully cancel all active streams then cancel the task running ``connect(lazily:)``.
+  public func close() {
+    self.state.withLockedValue { state in
+      switch state {
+      case .unconnected:
+        state = .closed(.init())
+      case .connected(let connectedState):
+        if connectedState.openStreams.count == 0 {
+          connectedState.signalEndContinuation.finish()
```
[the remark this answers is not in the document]
gjcairo: Is this because it can be a pretty slow operation and we try to have as small a critical region as possible, or something else?

Sources/GRPCCore/Transport/InProcessClientTransport.swift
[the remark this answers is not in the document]
glbrntt: Simpler than that: when we call `acceptStream` we're handing ownership of those streams over to the server. The only time we should close them is when `acceptStream` throws.

Sources/GRPCCore/Transport/InProcessClientTransport.swift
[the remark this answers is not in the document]
glbrntt: Partially to minimise the critical region but mostly because making out calls (i.e. anything which could cause unknown side effects) can be very difficult to reason about. The recent calling continuations while holding a lock is an example of this: we _thought_ it was a safe outcall to make but ended up causing weird deadlock bugs. It's best to just avoid it entirely.

Tests/GRPCCoreTests/Transport/InProcessClientTransportTests.swift
glbrntt: 🤦‍♂️

Sources/GRPCCore/Transport/InProcessClientTransport.swift
```diff
@@ -0,0 +1,361 @@
...
+
+    defer {
+      clientStream.outbound.finish()
+
+      let maybeEndContinuation = self.state.withLockedValue { state in
+        switch state {
+        case .unconnected:
+          fatalError("Invalid state")
+        case .connected(var connectedState):
+          connectedState.openStreams.removeValue(forKey: connectedState.nextStreamID - 1)
```
glbrntt, line 336: There's a bug here: `nextStreamID - 1` isn't guaranteed to be the ID of the stream we opened in this call. Another thread can acquire the lock and open a stream between the stream being opened by this function and the defer being executed, meaning we can remove the wrong stream. 

The state machine should return the ID of the stream when the server accepts it.

glbrntt (changes requested):

Sources/GRPCCore/Transport/InProcessClientTransport.swift
```diff
@@ -0,0 +1,333 @@
...
+          // The state cannot be unconnected at this point, because if we made
+          // it this far, it's because the transport was connected.
+          // Once connected, it's impossible to transition back to unconnected,
+          // so this is an invalid state.
+          fatalError("Invalid state")
+        case .connected(var connectedState):
+          connectedState.openStreams.removeValue(forKey: streamID)
+          state = .connected(connectedState)
+        case .closed(let closedState):
+          if closedState.openStreams.count == 1 {
```
glbrntt, line 311: I think I found another bug: in `close()`, if we're in the `connected` state then we always transition to the `closed` state, but we never remove streams while in the `closed` state here. If we have multiple streams open and then call `close()` then `connect(lazily:)` won't return (until cancelled).

glbrntt (changes requested):

Sources/GRPCCore/Transport/InProcessClientTransport.swift
gjcairo: Good catch - the problem was we weren't removing streams at the end of `withStream` when in `closed` state.
I've added a test covering this as well.

glbrntt (approved):
Nice one!

PR #1714, opened 2023-11-13
Add support for hedging

glbrntt:
Motivation:

The `ClientRPCExecutor` currently ignores hedging policies. This change adds support for them.

Modifications:

- Add a hedging executor and wire it up to the client rpc executor

Result:

RPC can be hedged under certain conditions


Sources/GRPCCore/Call/Client/Internal/ClientRPCExecutor+HedgingExecutor.swift
```diff
@@ -0,0 +1,499 @@
...
+  var nextAttempt: ScheduledAttempt?
+
+  @usableFromInline
+  struct ScheduledAttempt {
+    @usableFromInline
+    let task: Task<Void, Error>
+
+    @inlinable
+    init(after delay: Duration, _ body: @Sendable @escaping () -> Void) {
+      self.task = Task {
```
FranzBusch, line 317: I think we can make this work without an unstructured task. If I understand this correctly we are just using this `Task` here because we need a handle to the different attempts since we have to cancel everything except one (simply speaking). We can models this in a structured pattern where we spawn a child task per request and inside that child task we setup another task group where one child task is doing the request and the other child task is waiting for a cancellation signal. If we want to cancel we can just trigger that signal and the request attempt will be cancelled.
We should be able to even abstract this away and provide an extension on `TaskGroup` where we have a `addTask(_ :()) -> TaskHandle`. In the end that is just setting up the nested task group and returns a continuation like structure.

glbrntt: I played around with this idea a bit. You end up having to have an `AsyncStream` to provide the signal in addition to the extra overhead from the task group and extra task.

It works quite well for queuing up the next attempt but not so well for handling the different attempts. 

The reason is that each attempt signals whether it should be processed or discarded. If it's processed we need to continue running its task but cancel all the others. Returning a `TaskHandle` doesn't let us do this directly. We need a way to identify each attempt and linking that back to its handle. We also need to include that ID when we signal when an attempt should be processed and then cancel all handles which don't have that ID. It's doable but to me it felt harder to follow and more error prone.

Sources/GRPCCore/Call/Client/Internal/ClientRPCExecutor.swift
```diff
@@ -75,8 +75,22 @@ enum ClientRPCExecutor {
...
-      fatalError()
+    case .hedge(let policy):
+      let hedging = HedgingExecutor(
+        transport: transport,
+        policy: policy,
+        timeout: configuration.timeout,
+        interceptors: interceptors,
+        serializer: serializer,
+        deserializer: deserializer,
+        bufferSize: 64  // TODO: the client should have some control over this.
```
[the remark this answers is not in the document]
glbrntt: Yes

Tests/GRPCCoreTests/Call/Client/Internal/ClientRPCExecutorTests+Hedging.swift
```diff
@@ -0,0 +1,206 @@
...
+      // Set a long delay to reduce the risk of racing the second attempt and checking the number
+      // of streams being opened.
+      configuration: .hedge(delay: .seconds(5), nonFatalCodes: [])
+    ) { response in
+      XCTAssertRejected(response) { error in
+        XCTAssertEqual(error.code, .aborted)
+      }
+    }
+
+    // The first response is fatal. Others
```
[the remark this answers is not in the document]
glbrntt: Fixed, thanks.

Sources/GRPCCore/Call/Client/Internal/ClientRPCExecutor+HedgingExecutor.swift
```diff
@@ -0,0 +1,499 @@
...
+      self.deserializer = deserializer
+      self.bufferSize = bufferSize
+    }
+  }
+}
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+extension ClientRPCExecutor.HedgingExecutor {
+  @inlinable
+  func execute<R>(
```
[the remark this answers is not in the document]
glbrntt: Good catch, fixed.

Tests/GRPCCoreTests/Call/Client/Internal/ClientRPCExecutorTests+Hedging.swift
```diff
@@ -0,0 +1,206 @@
...
+          try await $0.write([0])
+        },
+        configuration: .hedge(delay: .seconds(60), nonFatalCodes: [.unavailable])
+      ) { response in
+        switch response.accepted {
+        case .success:
+          XCTFail("Expected RPC to fail")
+        case .failure(let error):
+          XCTAssertEqual(error.code, .unavailable)
+        }
```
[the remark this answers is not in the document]
glbrntt: Thanks, updated to use it.

Sources/GRPCCore/Call/Client/Internal/ClientRPCExecutor+HedgingExecutor.swift
FranzBusch: I understand your reasoning here but I am very concerned about using unstructured tasks here. Nothing is holding us accountable that we clean them up and they might be a source of bugs where we miss this. I am especially thinking about long term maintenance of this code. IMO, we should try everything possible to spell constructs in a structured way and bring any ergonomic deficiencies up with the compiler folks. 

It would be very valuable to understand what construct we need from the language to make this pattern easy. Is it enough if `addTask` would return us a `ChildTask` which we could store in a dictionary with an ID and the `ChildTask` allows us to call `.cancel()` and `.value()`?

glbrntt: > IMO, we should try everything possible to spell constructs in a structured way and bring any ergonomic deficiencies up with the compiler folks.

I agree, I just don't think returning a task handle is the right approach for racing the RPCs.

> It would be very valuable to understand what construct we need from the language to make this pattern easy. Is it enough if addTask would return us a ChildTask which we could store in a dictionary with an ID and the ChildTask allows us to call .cancel() and .value()?

I think returning a handle which can be cancelled is useful but wouldn't solve all problems here. Being able to get the value is weird though (and also not useful here).

FWIW I think we can make the RPC racing structured by using the broadcast async sequence to signal which attempt gets to survive. This, with some shared state, allows one attempt to effectively cancel the rest.

Sources/GRPCCore/Internal/TaskGroup+CancellableTask.swift
```diff
@@ -0,0 +1,82 @@
...
+  /// Adds a child task to the group which is individually cancellable.
+  ///
+  /// - Parameter operation: The task to add to the group.
+  /// - Returns: A handle which can be used to cancel the task without cancelling the rest of
+  ///     the group.
+  @inlinable
+  mutating func addCancellableTask(
+    _ operation: @Sendable @escaping () async -> ChildTaskResult
+  ) -> CancellableTaskHandle {
+    let signal = AsyncStream.makeStream(of: Void.self)
```
FranzBusch, line 28: Just a note for future us, we probably want to use a bare continuation (probably even unsafe) here in the future to avoid all the allocs that `AsyncStream` does

FranzBusch (approved):
Looks great now. Thanks for the refactoring to use structured concurrency!

Sources/GRPCCore/Internal/TaskGroup+CancellableTask.swift
glbrntt: Agreed. I deliberately avoided that for now to avoid having to worry about double resumptions etc. but we should definitely do this in the future, I think there are plenty of allocations we can save...!

gjcairo (approved):

PR #1716, opened 2023-11-22
Created the CodeGenerationRequest 

stefanadranca:
Created the CodeGenerationRequest which represents the services, dependencies and trivia from an IDL file, and the IDL itself through its specific serializer and deserializer.

Motivation:

The CodeGenerationRequest is part of the new code generator library for grpc-swift. It represents services described by IDL and is the first stage in the process of code generation.

Modifications:

Implemented the CodeGenerationRequest struct and added the new target and library in manifest.

Result:

The code generators will be able to use a general representation for the services described in different IDLs.


Sources/GRPCCodeGen/layers/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,84 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+/// Describes the services, dependencies and trivia from an IDL file,
+/// and the IDL itself through its specific serializer and deserializer.
+struct CodeGenerationRequest {
+  /// The name of the source file containing the IDL.
```
glbrntt, line 22: Does this include the extension of the file?

Sources/GRPCCodeGen/layers/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,84 @@
...
+  var fileName: String
+
+  /// The swift imports that this file depends on.
```
glbrntt, line 25:
```suggestion
  /// The Swift imports that the generated file should depends on.
```

Sources/GRPCCodeGen/layers/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,84 @@
...
+  var leadingTrivia: String
+
+  /// An array of service descriptors.
+  var services: [ServiceDescriptor]
+
+  /// String representation of the serializer call.
+  var lookupSerializer: (String) -> String
+
+  /// String representation of the deserializer call.
+  var lookupDeserializer: (String) -> String
```
glbrntt, line 39: I don't think the documentation for these really explains what they are for.

Sources/GRPCCodeGen/layers/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,84 @@
...
+      enum Kind: String {
+        case `typealias` = "typealias"
+        case `struct` = "struct"
+        case `class` = "class"
+        case `enum` = "enum"
+        case `protocol` = "protocol"
+        case `let` = "let"
+        case `var` = "var"
+        case `func` = "func"
+      }
```
glbrntt, line 54: This needs to be a struct backed by an enum so we can evolve it. We also don't need it to be raw representable as a `String`.

Sources/GRPCCodeGen/layers/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,84 @@
...
+    }
+    var symbol: Symbol?
+    var module: String
+  }
+
+  /// Represents a service described in an IDL file.
+  struct ServiceDescriptor {
+    /// Description of the service from comments
+    /// above the description from the IDL file.
+    var docs: String
```
glbrntt, line 67: nit: `documentation`

Sources/GRPCCodeGen/layers/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,84 @@
...
+
+    /// Represents a method described in an IDL file.
+    struct MethodDescriptor {
+      var name: String
+      var isInputStreaming: Bool
+      var isOutputStreaming: Bool
+      var inputType: String
+      var ouputType: String
+      var docs: String
+    }
```
glbrntt, line 82: This needs documentation

glbrntt (changes requested):

Sources/GRPCCodeGen/layers/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,84 @@
+/*
```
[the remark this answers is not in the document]
stefanadranca: I was thinking we can have all the representations in a directory (CodeGenerationRequest, StructuredSwiftSyntax and SourceFile). Would "Representations" be a better name, or should I not create the directory?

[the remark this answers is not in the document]
glbrntt: I tend to split them into directories based on the "area" when there are (or will be) many files. In this case I'm not sure there will be enough files so I probably wouldn't bother. One thing that can be useful though is to but `internal` things in an `Internal` directory.

Sources/GRPCCodeGen/layers/CodeGenerationRequest.swift
glbrntt: Yeah it should be mentioned in the documentation. I think we should keep the extension: it's often useful to have a comment at the top of generated code saying which file it was generated from.

Sources/GRPCCodeGen/layers/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,84 @@
...
+import Foundation
+
+/// Describes the services, dependencies and trivia from an IDL file,
+/// and the IDL itself through its specific serializer and deserializer.
+struct CodeGenerationRequest {
+  /// The name of the source file containing the IDL.
+  var fileName: String
+
+  /// The swift imports that this file depends on.
+  var dependencies: [Dependency]?
```
gjcairo, line 26: Do we need this to be optional? An empty array can still represent no dependencies.

Sources/GRPCCodeGen/layers/CodeGenerationRequest.swift
gjcairo: You could maybe also use `- SeeAlso:` to link to `Dependency`?

Sources/GRPCCodeGen/layers/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,84 @@
...
+
+  /// Any comments at the top of the file including
+  /// documentation and copyright headers.
+  var leadingTrivia: String
+
+  /// An array of service descriptors.
```
gjcairo, line 32: Consider linking to `ServiceDescriptor`

Sources/GRPCCodeGen/layers/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,84 @@
...
+        case `let` = "let"
+        case `var` = "var"
+        case `func` = "func"
+      }
+
+      var kind: Kind
+      var name: String
+    }
+    var symbol: Symbol?
+    var module: String
```
gjcairo, line 60: Some docs are missing for these props

Sources/GRPCCodeGen/layers/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  /// Represents a service described in an IDL file.
+  struct ServiceDescriptor {
+    /// Description of the service from comments
+    /// above the description from the IDL file.
+    var docs: String
+
+    var name: String
```
gjcairo, line 69: Nit, docs are missing

gjcairo (changes requested):

Sources/GRPCCodeGen/layers/CodeGenerationRequest.swift
stefanadranca: You're right, it doesn't need to be optional

gjcairo (approved):

glbrntt (changes requested):

Sources/GRPCCodeGen/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,224 @@
...
+  /// as they will be added by default in the generated file.
+  ///
+  /// - SeeAlso: ``Dependency``.
+  public var dependencies: [Dependency]
+
+  /// Any comments at the top of the file such as documentation and copyright headers.
+  /// They will be placed at the top of the generated file.
+  public var leadingTrivia: String
+
+  /// An array of service descriptors.
```
glbrntt, line 33: This doesn't tell me anything more than `[ServiceDescriptor]` tells me (i.e. what the thing _is_). If you say "A description of each service to generate." then it tells the user what it's used for.

Sources/GRPCCodeGen/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,224 @@
...
+  ///
+  /// - SeeAlso: ``ServiceDescriptor``.
+  public var services: [ServiceDescriptor]
+
+  /// Closure that receives the message type and returns a string representation of the
+  /// serializer call for that message type. The result is inserted in the string representing
+  /// the generated code, where clients or servers serialize their output.
+  ///
+  /// For example: `lookupSerializer: {"ProtobufSerializer<\($0)>()"}`.
```
glbrntt, line 42: Docs should start with a single sentence summary and add more detail in other paragraphs. Tools like Xcode and DocC show the first paragraph more prominently so it's quite important for it to be concise.

We can also be more specific about the requirements here: the serializer type _must_ conform to `MessageSerializer`.

```suggestion
  /// Closure that receives a message type as a `String` and returns a code snippet to 
  /// initialize a `MessageSerializer` for that type as a `String`.
  ///
  /// The result is inserted in the generated code, where clients serialize RPC inputs and
  /// servers serialize RPC outputs.
  ///
  /// For example, to serialize Protobuf messages you could specify a serializer as:
  /// ```swift
  /// request.lookupSerializer = { messageType in
  ///   "ProtobufSerializer<\(messageType)>()"
  /// }
  /// ```
```

Sources/GRPCCodeGen/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,224 @@
...
+  public var lookupSerializer: (_ messageType: String) -> String
+
+  /// Closure that receives the message type and returns a string representation of the
+  /// deserializer call for that message type. The result is inserted in the string representing
+  /// the generated code, where clients or servers deserialize their input.
+  ///
+  /// For example: `lookupDeserializer: {"ProtobufDeserializer<\($0)>()"}`.
```
glbrntt, line 49: This needs a similar treatment to the one above, but the deserializer must conform to `MessageDeserializer`.

Sources/GRPCCodeGen/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,224 @@
...
+  public var lookupDeserializer: (_ messageType: String) -> String
+
+  public init(
+    fileName: String,
+    dependencies: [Dependency] = [],
+    leadingTrivia: String,
```
glbrntt, line 55: Generally parameters should be specified in this order:
1. required
2. optional
3. closures

In this case, I think we should probably just remove the defaults for `dependencies` and `services`. It's highly unlikely that there will be no dependencies or services.

We should also invert the order of `dependencies` and `leadingTrivia`, if you're reading a file top-to-bottom the leading trivial comes at the very top followed by the dependencies.

Sources/GRPCCodeGen/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,224 @@
...
+      public var kind: Kind
+
+      /// The imported item's symbol / name.
+      public var name: String
+
+      public init(kind: Kind, name: String) {
+        self.kind = kind
+        self.name = name
+      }
+      /// Represents the imported item's kind.
```
glbrntt, line 94: nit: leave a space between these

Sources/GRPCCodeGen/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,224 @@
...
+    public var module: String
+
+    public init(item: Item? = nil, module: String) {
+      self.item = item
+      self.module = module
+    }
+
+    /// Represents an item imported from a module.
+    public struct Item {
+      /// The keyword that specifies the item's kind (e.g.  `func`, `struct`).
```
glbrntt, line 84:
```suggestion
      /// The keyword that specifies the item's kind (e.g. `func`, `struct`).
```

Sources/GRPCCodeGen/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,224 @@
...
+      public var kind: Kind
+
+      /// The imported item's symbol / name.
```
glbrntt, line 87:
```suggestion
      /// The name of the imported item.
```

Sources/GRPCCodeGen/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,224 @@
...
+    public var name: String
+
+    /// The service namespace.
+    ///
+    /// For `.proto` files it is the package name.
+    public var namespace: String
+
+    /// Array of descriptors for the methods of a service.
+    ///
+    /// - SeeAlso: ``MethodDescriptor``.
```
glbrntt, line 172: Same comment as on `services`, this doesn't tell me anything more than `[MethodDescriptor]`.

Sources/GRPCCodeGen/CodeGenerationRequest.swift
```diff
@@ -0,0 +1,224 @@
...
+    public var methods: [MethodDescriptor]
+
+    public init(
+      documentation: String,
+      name: String,
+      namespace: String,
+      methods: [MethodDescriptor] = []
```
glbrntt, line 179: I don't think we should default this, It'd be very strange to generate a service with no methods!

glbrntt (changes requested):

glbrntt (approved):

PR #1717, opened 2023-11-22
Add the router and service protocol

glbrntt:
Motivation:

The transport layer accepts streams from the client and transforms them into requests which are handled by services. Services, written by users, need a way to tell the server which streams should be handled by which methods.

Modifications:

- Add the RPC router which routes streams to "handlers"
- Add the RegistrableRPCService protocol which services can conform to to register routes with a router.

The `RegistrableRPCService` is typically not implemented by users and instead will be implemented in the generated code. However, it builds on public API offered by the router.

Result:

We have interfaces which allow streams to be routed to a service handler.

### DIFF
--- a/Sources/GRPCCore/Call/Server/RPCRouter.swift
+++ b/Sources/GRPCCore/Call/Server/RPCRouter.swift
@@ -1,0 +1,135 @@
+/*
+ * Copyright 2023, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/// Stores and provides handlers for RPCs.
+///
+/// The router stores a handler for each RPC it knows about. Each handler encapsulate the business
+/// logic for the RPC which is typically implemented by service owners. To register a handler you
+/// can call ``registerHandler(forMethod:deserializer:serializer:handler:)``. You can check whether
+/// the router has a handler for a method with ``hasHandler(forMethod:)`` or get a list of all
+/// methods with handlers registered by calling ``methods``. You can also remove the handler for a
+/// given method by calling ``removeHandler(forMethod:)``.
+///
+/// In most cases you won't need to interact with the router directly. Instead you should register
+/// your services with ``Server/Services-swift.struct/register(_:)`` which will in turn register
+/// each method with the router.
+///
+/// You may wish to not serve all methods from your service in which case you can either:
+///
+/// 1. Remove individual methods by calling ``removeHandler(forMethod:)``, or
+/// 2. Implement ``RegistrableRPCService/registerMethods(with:)`` to register only the methods you
+///    want to be served.
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+public struct RPCRouter: Sendable {
+  @usableFromInline
+  struct RPCHandler: Sendable {
+    @usableFromInline
+    let _fn:
+      @Sendable (
+        _ stream: RPCStream<RPCAsyncSequence<RPCRequestPart>, RPCWriter<RPCResponsePart>.Closable>,
+        _ interceptors: [any ServerInterceptor]
+      ) async -> Void
+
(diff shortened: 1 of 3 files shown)


Sources/GRPCCore/Call/Server/RPCRouter.swift
```diff
@@ -0,0 +1,135 @@
...
+/// 2. Implement ``RegistrableRPCService/registerMethods(with:)`` to register only the methods you
+///    want to be served.
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+public struct RPCRouter: Sendable {
+  @usableFromInline
+  struct RPCHandler: Sendable {
+    @usableFromInline
+    let _fn:
+      @Sendable (
+        _ stream: RPCStream<RPCAsyncSequence<RPCRequestPart>, RPCWriter<RPCResponsePart>.Closable>,
```
FranzBusch, line 42:
```suggestion
        _ stream: RPCStream<RPCAsyncSequence<RPCRequestPart>,
        _ writer: RPCWriter<RPCResponsePart>.Closable>,
```

Sources/GRPCCore/Call/Server/RPCRouter.swift
```diff
@@ -0,0 +1,135 @@
...
+  /// Returns the number of methods registered with the router.
+  public var count: Int {
+    self.handlers.count
+  }
+
+  /// Returns whether a handler exists for a given method.
+  ///
+  /// - Parameter descriptor: A descriptor of the method.
+  /// - Returns: Whether a handler exists for the method.
+  public func hasHandler(forMethod descriptor: MethodDescriptor) -> Bool {
```
FranzBusch, line 97: Do those things have to be `public`?

Sources/GRPCCore/Call/Server/RPCRouter.swift
```diff
@@ -0,0 +1,135 @@
...
+      handler: handler
+    )
+  }
+
+  /// Removes any handler registered for the specified method.
+  ///
+  /// - Parameter descriptor: A descriptor of the method to remove a handler for.
+  /// - Returns: Whether a handler was removed.
+  @discardableResult
+  public mutating func removeHandler(forMethod descriptor: MethodDescriptor) -> Bool {
```
FranzBusch, line 132: Same here. Does this need to be public?

Sources/GRPCCore/Call/Server/RPCRouter.swift
glbrntt: This was right as is: `RPCStream` is a type which has an async sequence and a writer (and a descriptor of the method)

Sources/GRPCCore/Call/Server/RPCRouter.swift
glbrntt: If you don't want to offer all methods in a service then this is one way to remove the ones you don't want (the other being to reimplement the service protocol).

Sources/GRPCCore/Call/Server/RPCRouter.swift
glbrntt: I think there's value in having `methods` (e.g. logging what methods your server serves). You can obviously just query `methods` instead of using this function but it's then O(n) rather than O(1) which is a bit sad and the main reason to add this.

gjcairo: Wouldn't this be solved if `methods` was a `Set` instead of an `Array`?

Sources/GRPCCore/Call/Server/RPCRouter.swift
FranzBusch: Oh right. I got so confused by the parameters

Sources/GRPCCore/Call/Server/RPCRouter.swift
FranzBusch: I see. I am wondering if we should expose this publicly right away. Do we have concrete use-cases in mind where want to do this or did we get a feature request for this before?

gjcairo (approved):

Sources/GRPCCore/Call/Server/RPCRouter.swift
glbrntt: No, while the lookup would be O(1) creating the set would still be O(n).

Sources/GRPCCore/Call/Server/RPCRouter.swift
glbrntt: One concrete use case is the interoperability tests. One test uses the generated client to call a method which isn't implemented on the server. At the moment we achieve this by patching the generated code.

glbrntt: I also think having these methods is very low burden: they're unlikely to change and they still make sense if we change the internals of the router.

stefanadranca (approved):

PR #1718, opened 2023-11-23
Add new Timeout type

gjcairo:
## Motivation
We need an encodable/decodable `Timeout` type to represent gRPC timeouts.

## Modification
Added a new `Timeout` type that can be encoded to and from a `String` with [the format specified by gRPC](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests).

## Result
We now have a way to represent `Timeout`s.


Sources/GRPCCore/Timeout.swift
```diff
@@ -0,0 +1,154 @@
...
+  /// The wire encoding of this timeout as described in the gRPC protocol.
+  /// See: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md.
+  public let wireEncoding: String
+  public let duration: Duration
+
+  public var description: String {
+    return self.wireEncoding
+  }
+
+  public init?(stringLiteral value: String) {
```
gjcairo, line 36: I originally wanted to have this type be `ExpressibleByStringLiteral`, but I cannot have a failing nor throwing `init`, so I just added this `init` without the protocol conformance. If there's something you would do differently or a workaround here, let me know @glbrntt :D

Sources/GRPCCore/Timeout.swift
```diff
@@ -0,0 +1,154 @@
...
+  internal init(amount: Int64, unit: TimeoutUnit) {
+    precondition(0 ... Timeout.maxAmount ~= amount)
+
+    self.duration = Duration(amount: amount, unit: unit)
+    self.wireEncoding = "\(amount)\(unit.rawValue)"
+  }
+
+  /// Create a timeout by rounding up the timeout so that it may be represented in the gRPC
+  /// wire format.
+  internal init(rounding amount: Int64, unit: TimeoutUnit) {
```
gjcairo, line 63: I decided to carry over this `init` because it's present in `GRPCTimeout`, but I'm not sure if we want something like this now - we're not (yet) using it anywhere at least.

Sources/GRPCCore/Timeout.swift
```diff
@@ -0,0 +1,154 @@
...
+  /// See: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md.
+  public let wireEncoding: String
+  public let duration: Duration
+
+  public var description: String {
+    return self.wireEncoding
+  }
+
+  public init?(stringLiteral value: String) {
+    guard 2 ... 8 ~= value.count else {
```
dnadoba, line 37: I would recommend using pattern matching instead of calling the pattern matching operator directly
```suggestion
    guard case 2...8 = value.count else {
```

Sources/GRPCCore/Timeout.swift
```diff
@@ -0,0 +1,154 @@
...
+ * limitations under the License.
+ */
+import Dispatch
+
+/// A timeout for a gRPC call.
+///
+/// Timeouts must be positive and at most 8-digits long.
+/// See "Timeout" in https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+public struct Timeout: CustomStringConvertible, Equatable {
```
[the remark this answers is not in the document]
glbrntt: Gus and I talked about these changes offline earlier, `Timeout` is fine as it'll be an `internal` type anyway.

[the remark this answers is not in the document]
FranzBusch: Ah if it is internal then the name is fine. Right now it is public 😅

Sources/GRPCCore/Timeout.swift
```diff
@@ -0,0 +1,154 @@
...
+  /// The largest amount of any unit of time which may be represented by a gRPC timeout.
+  internal static let maxAmount: Int64 = 99_999_999
+
+  /// The wire encoding of this timeout as described in the gRPC protocol.
+  /// See: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md.
+  public let wireEncoding: String
+  public let duration: Duration
```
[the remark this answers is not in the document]
gjcairo: This type is only an internal representation, as George mentioned in another comment. However we have decided to expose `Duration`s, so this is kept this way, but the API has changed slightly.

Sources/GRPCCore/Timeout.swift
```diff
@@ -0,0 +1,154 @@
...
+/// Timeouts must be positive and at most 8-digits long.
+/// See "Timeout" in https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+public struct Timeout: CustomStringConvertible, Equatable {
+  /// The largest amount of any unit of time which may be represented by a gRPC timeout.
+  internal static let maxAmount: Int64 = 99_999_999
+
+  /// The wire encoding of this timeout as described in the gRPC protocol.
+  /// See: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md.
+  public let wireEncoding: String
```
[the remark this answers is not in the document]
gjcairo: This is all internal now

Sources/GRPCCore/Timeout.swift
```diff
@@ -0,0 +1,154 @@
...
+  internal static func minutes(_ minutes: Int64) -> Duration {
+    return Self.init(secondsComponent: 60 * minutes, attosecondsComponent: 0)
+  }
+
+  /// Construct a `Duration` given a number of hours represented as an `Int64`.
+  ///
+  ///       let d: Duration = .hours(3)
+  ///
+  /// - Returns: A `Duration` representing a given number of hours.
+  internal static func hours(_ hours: Int64) -> Duration {
```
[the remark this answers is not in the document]
gjcairo: Hmm okay, I'm using these methods in tests though, and currently using `@testable import`. Should I mark these as `@_spi(Testing)` instead? Or is there a way I can access file private methods from tests?

Sources/GRPCCore/Timeout.swift
gjcairo: Is this for readability or is there some other benefit to doing this?

glbrntt (changes requested):

Sources/GRPCCore/Timeout.swift
dnadoba: @gjcairo just for readability. 
@glbrntt fine with this as well.

Sources/GRPCCore/Timeout.swift
```diff
@@ -0,0 +1,207 @@
...
+
+  /// The largest amount of any unit of time which may be represented by a gRPC timeout.
+  static let maxAmount: Int64 = 99_999_999
+
+  /// The wire encoding of this timeout as described in the gRPC protocol.
+  /// See "Timeout" in https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
+  let wireEncoding: String
+
+  @usableFromInline
+  let duration: Duration
```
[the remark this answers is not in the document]
gjcairo: I considered getting the `duration` to be something that we'll want to do relatively often, so avoiding the extra allocations for each time we get it may have been something we were looking for. I can change it into a computed property though.

Sources/GRPCCore/Timeout.swift
```diff
@@ -0,0 +1,226 @@
...
+  /// the target unit.
+  @usableFromInline
+  init(duration: Duration) {
+    let (seconds, attoseconds) = duration.components
+
+    if seconds == 0 {
+      // There is no seconds component, so only pay attention to the attoseconds.
+      // Try converting to nanoseconds first, and continue rounding up if the
+      // max amount of digits is exceeded.
+      let nanoseconds = Int64(round(Double(attoseconds) / 1e+9))
```
glbrntt, line 102: Hmm, we'll only over round up one nanosecond. Do we care about that?

Sources/GRPCCore/Timeout.swift
```diff
@@ -0,0 +1,226 @@
...
+      // we will have to use minutes or hours.
+      // We can also ignore attoseconds, since we won't have enough precision
+      // anyways to represent the (at most) one second that the attoseconds
+      // component can express.
+      self.init(rounding: seconds, unit: .seconds)
+    } else {
+      // We can't convert seconds to nanoseconds because that would take us
+      // over the 8 digit limit (1 second = 1e+9 nanoseconds).
+      // We can however, try converting to microseconds or milliseconds.
+      let nanoseconds = Int64(Double(attoseconds) / 1e+9)
```
glbrntt, line 115: nit: we round when doing this conversion on L102 but not here. Let's pick one and be consistent.

glbrntt (approved):
One nit, looks good otherwise!

PR #1721, opened 2023-11-24
Add new Server object

glbrntt:
Motivation:

The server is the main entry point for running gRPC services. It combines a set of transports, services, and interceptors.

Modifications:

- Add the 'Server' and 'ServerError'

Result:

Can piece together transports, interceptors and services

### DIFF
--- a/Sources/GRPCCore/Server.swift
+++ b/Sources/GRPCCore/Server.swift
@@ -1,0 +1,428 @@
+/*
+ * Copyright 2023, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Atomics
+
+/// A gRPC server.
+///
+/// The server accepts connections from clients and listens on each connection for new streams
+/// which are initiated by the client. Each stream maps to a single RPC. The server routes accepted
+/// streams to a service to handle the RPC or rejects them with an appropriate error if no service
+/// can handle the RPC.
+///
+/// A ``Server`` may listen with multiple transports (for example, HTTP/2 and in-process) and route
+/// requests from each transport to the same service instance. You can also use "interceptors",
+/// to implement cross-cutting logic which apply to all accepted RPCs. Example uses of interceptors
+/// include request filtering, authentication, and logging. Once requests have been intercepted
+/// they are passed to a handler which in turn returns a response to send back to the client.
+///
+/// ## Creating and configuring a server
+///
+/// The following example demonstrates how to create and configure a server.
+///
+/// ```swift
+/// var server = Server()
+///
+/// // Create and add an in-process transport.
+/// let inProcessTransport = InProcessServerTransport()
+/// server.transports.add(inProcessTransport)
+///
+/// // Create and register the 'Greeter' and 'Echo' services.
+/// server.services.register(GreeterService())
+/// server.services.register(EchoService())
+///
+/// // Create and add some interceptors.
+/// server.interceptors.add(StatsRecordingServerInterceptors())
+/// ```
+///
+/// ## Starting and stopping the server
+///
+/// Once you have configured the server call ``run()`` to start it. Calling ``run()`` starts each
+/// of the servers transports. A ``ServerError`` is thrown if any of the transports can't be
+/// started.
+///
+/// ```swift
+/// // Start running the server.
+/// try await server.run()
+/// ```
+///
+/// The ``run()`` method won't return until the server has finished handling all requests. You can
+/// signal to the server that it should stop accepting new requests by calling ``stopListening()``.
+/// This allows the server to drain existing requests gracefully. To stop the server more abruptly
+/// you can cancel the task running your server. If your services require additional resources that
+/// need their lifecycles managed you should consider using [Swift Service
+/// Lifecycle](https://github.com/swift-server/swift-service-lifecycle).
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+public struct Server: Sendable {
+  typealias Stream = RPCStream<ServerTransport.Inbound, ServerTransport.Outbound>
+
+  /// A collection of ``ServerTransport`` implementations that the server uses to listen
+  /// for new requests.
+  public var transports: Transports
+
+  /// The services registered which the server is serving.
+  public var services: Services
+
+  /// A collection of ``ServerInterceptor`` implementations which are applied to all accepted
+  /// RPCs.
+  ///
+  /// RPCs are intercepted in the order that interceptors are added. That is, a request received
+  /// from the client will first be intercepted by the first added interceptor followed by the
+  /// second, and so on.
+  public var interceptors: Interceptors
+
+  /// The state of the server.
+  private enum State {
+    /// The server hasn't been started yet. Can transition to `starting` or `stopped`.
+    case notStarted
+    /// The server is starting but isn't accepting requests yet. Can transition to `running`
+    /// and `stopping`.
+    case starting
+    /// The server is running and accepting RPCs. Can transition to `stopping`.
+    case running
+    /// The server is stopping and no new RPCs will be accepted. Existing RPCs may run to
+    /// completion. May transition to `stopped`.
+    case stopping
+    /// The server has stopped, no RPCs are in flight and no more will be accepted. This state
+    /// is terminal.
+    case stopped
+  }
+
+  private let state: LockedValueBox<State>
+
+  /// Creates a new server with no resources.
+  ///
+  /// You can add resources to the server via ``transports-swift.property``,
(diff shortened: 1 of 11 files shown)


Sources/GRPCCore/Server.swift
```diff
@@ -0,0 +1,428 @@
...
+/// // Start running the server.
+/// try await server.run()
+/// ```
+///
+/// The ``run()`` method won't return until the server has finished handling all requests. You can
+/// signal to the server that it should stop accepting new requests by calling ``stopListening()``.
+/// This allows the server to drain existing requests gracefully. To stop the server more abruptly
+/// you can cancel the task running your server. If your services require additional resources that
+/// need their lifecycles managed you should consider using [Swift Service
+/// Lifecycle](https://github.com/swift-server/swift-service-lifecycle).
```
FranzBusch, line 67: I love the callout but I think we should be even more specific here. Not only when you need handling additional resources but also when you want to add support to gracefully shutdown the server. WDYT?

Sources/GRPCCore/Server.swift
```diff
@@ -0,0 +1,428 @@
...
+    case running
+    /// The server is stopping and no new RPCs will be accepted. Existing RPCs may run to
+    /// completion. May transition to `stopped`.
+    case stopping
+    /// The server has stopped, no RPCs are in flight and no more will be accepted. This state
+    /// is terminal.
+    case stopped
+  }
+
+  private let state: LockedValueBox<State>
```
FranzBusch, line 104: Interesting. I didn't expect to have a struct with reference semantics in the end. For `ServiceLifecycle` we decided to use an actor for the `ServiceGroup`. Here I would have actually expected a `class`. Especially since you can do weird things like this
```
var server = Server()
server.transport = [SomeTransport()]
async let _ = server.run()
server.transports = [OtherTransport()]
async let _ = server.run()
```

IMO here having a class for the `Server` makes more sense.

Sources/GRPCCore/Server.swift
```diff
@@ -0,0 +1,428 @@
...
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+public struct Server: Sendable {
+  typealias Stream = RPCStream<ServerTransport.Inbound, ServerTransport.Outbound>
+
+  /// A collection of ``ServerTransport`` implementations that the server uses to listen
+  /// for new requests.
+  public var transports: Transports
+
+  /// The services registered which the server is serving.
+  public var services: Services
```
FranzBusch, line 77: I am a bit unsure about making those vars. In the `ServiceGroup` I had the same decision and went with expecting everything in the inits. The reason for this was that you really expect everything to be setup before `run()` is called and allowing users to change them after `run()` is called is a bit weird.

Sources/GRPCCore/Server.swift
```diff
@@ -0,0 +1,428 @@
...
+  /// A collection of ``ServerInterceptor`` implementations which are applied to all accepted
+  /// RPCs.
+  ///
+  /// RPCs are intercepted in the order that interceptors are added. That is, a request received
+  /// from the client will first be intercepted by the first added interceptor followed by the
+  /// second, and so on.
+  public var interceptors: Interceptors
+
+  /// The state of the server.
+  private enum State {
```
FranzBusch, line 88: Can we move the transports & interceptors into the `State` so we release them as early as possible?

Sources/GRPCCore/Server.swift
```diff
@@ -0,0 +1,428 @@
...
+    defer {
+      self.state.withLockedValue { $0 = .stopped }
+    }
+
+    var listeners: [RPCAsyncSequence<Stream>] = []
+    listeners.reserveCapacity(self.transports.values.count)
+
+    for transport in self.transports.values {
+      do {
+        let listener = try await transport.listen()
```
FranzBusch, line 164: Interesting. I did not expect this. I expected that we had to create a child task per transport and each transport has a `run` method as well? In which task do transports run here?

Sources/GRPCCore/Server.swift
glbrntt: IMO these make the API easier to use, passing everything in `init` is quite awkward when there are potentially many things to pass in and is awkward w.r.t. API evolution. Sure, you can wrap everything up in a config, but that just adds more indirection for users.

I don't see changing things after `run()` to be an issue to be honest, trying to make changes after `run()` has been called just makes no sense at all so it's highly unlikely that someone would fall foul to this. We also document quite clearly that this isn't supported.

Sources/GRPCCore/Server.swift
glbrntt: Yeah, probably should be a `class` tbh. I started with a `struct` and didn't think to back out that change when I added the state management.

Sources/GRPCCore/Server.swift
glbrntt: Can you be more specific? We support graceful shutdown here...

Sources/GRPCCore/Server.swift
glbrntt: We do have a child task per transport (see `handleRequests`). We start them all in the same task so we can ensure they all start successfully.

Sources/GRPCCore/Server.swift
FranzBusch: Hm, I am not fully convinced yet. Let's imagine the following

```
var server = Server()
server.transports = [NIOH2Transport(port: 1234)]
async let _ = try await server.run()
server.transports = [NIOH2Transport(port: 5677])
async let _ try await server.run()
```

There really isn't a reason the second run should fail but it does because of the shared state in the struct. At least personally, I find these semantics confusing. I do get the reason why we want a struct here but then maybe we shouldn't share the state across `run` invocations?

Sources/GRPCCore/Server.swift
FranzBusch: Maybe we can give a general recommendation that we recommend running the `server` in a `ServiceGroup` since that gives you orchestration of multiple services and support for listening to signals to trigger graceful shutdown.

Sources/GRPCCore/Server.swift
FranzBusch: Just to clarify this. Does that mean that `beginListening` is expected to bind the port? Because so far we always followed the `run` pattern where inside the `run()` method call we bind the ports and then consume the inbound connection/data.

Tests/GRPCCoreTests/ServerTests.swift
```diff
@@ -0,0 +1,441 @@
...
+      task.cancel()
+      try await task.value
+
+      group.cancelAll()
+    }
+  }
+
+  func testTestRunServerWithNoTransport() async throws {
+    let server = Server()
+    try await server.run()
```
gjcairo, line 328: Would it be reasonable to throw if we try to run a server without a transport configured?

Sources/GRPCCore/Server.swift
```diff
@@ -0,0 +1,428 @@
...
+/// server.services.register(EchoService())
+///
+/// // Create and add some interceptors.
+/// server.interceptors.add(StatsRecordingServerInterceptors())
+/// ```
+///
+/// ## Starting and stopping the server
+///
+/// Once you have configured the server call ``run()`` to start it. Calling ``run()`` starts each
+/// of the servers transports. A ``ServerError`` is thrown if any of the transports can't be
```
gjcairo, line 54:
```suggestion
/// of the server's transports. A ``ServerError`` is thrown if any of the transports can't be
```

Sources/GRPCCore/Server.swift
```diff
@@ -0,0 +1,428 @@
...
+        return true
+      }
+    }
+
+    // If the server is stopping then notify the transport and then consume them: there may be
+    // streams opened at a lower level (e.g. HTTP/2) which are already open and need to be consumed.
+    if isStopping {
+      for transport in self.transports.values {
+        transport.stopListening()
+      }
```
gjcairo, line 203: We're already calling `stopListening()` in `rejectRequests(_:)` - we should be able to get rid of this

Sources/GRPCCore/Server.swift
```diff
@@ -0,0 +1,428 @@
...
+            }
+          }
+        }
+      }
+    }
+  }
+
+  /// Signal to the server that it should stop listening for new requests.
+  ///
+  /// By calling this function you indicate to clients that they mustn't make start new requests
```
gjcairo, line 322:
```suggestion
  /// By calling this function you indicate to clients that they mustn't start new requests
```

gjcairo (changes requested):

Sources/GRPCCore/Server.swift
glbrntt: `listen()` is expected to bind the port. `listen()` is on the transport which is a lower-level API which users aren't expected to interact with. The `run()` on the server which we do expect users to interact with _does_ do the bind and consume.

Sources/GRPCCore/Server.swift
glbrntt: Good catch, this was leftover from a refactor.

Tests/GRPCCoreTests/ServerTests.swift
glbrntt: Good question, I thought about this too but decided against it at the time because it was easier to explain in a sentence what the function does. I think I've changed my mind (again) though: the most likely reason for not configuring a transport is a bug, we should make that more obvious so let's throw an error.

Sources/GRPCCore/Server.swift
glbrntt: I'm a bit hesitant to document specifics here. I think the best solution is to have a DocC article which runs through using lifecycle with grpc-swift (which we should link to from these docs). It also means there's a single place which needs to be kept up-to-date.

gjcairo (approved):
Left a tiny nit about a doc but looking good 👍🏼

PR #1725, opened 2023-11-28
Update StructuredSwiftRepresentation

stefanadranca:
Update StructuredSwiftRepresentation to the latest version from OpenAPI generator.

Motivation:

The latest version of the Code Renderer uses the Expression type for the `left` property of a VariableDescription, so we need to also update the StructuredSwiftRepresentation accordingly.

Result:

We will have the latest version of StructuredSwiftRepresentation for the new CodeGenLib.

### DIFF
--- a/Sources/GRPCCodeGen/Internal/StructuredSwiftRepresentation.swift
+++ b/Sources/GRPCCodeGen/Internal/StructuredSwiftRepresentation.swift
@@ -65,7 +65,7 @@ struct ImportDescription: Equatable, Codable {
 /// A description of an access modifier.
 ///
 /// For example: `public`.
-public enum AccessModifier: String, Sendable, Equatable, Codable {
+internal enum AccessModifier: String, Sendable, Equatable, Codable {
 
   /// A declaration accessible outside of the module.
   case `public`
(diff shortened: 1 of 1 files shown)


Sources/GRPCCodeGen/Internal/StructuredSwiftRepresentation.swift
```diff
@@ -66,11 +65,14 @@ struct ImportDescription: Equatable, Codable {
 /// A description of an access modifier.
 ///
 /// For example: `public`.
-enum AccessModifier: String, Equatable, Codable {
+public enum AccessModifier: String, Sendable, Equatable, Codable {
```
glbrntt, line 68: We should keep this `internal`

glbrntt (changes requested):

Sources/GRPCCodeGen/Internal/StructuredSwiftRepresentation.swift
stefanadranca: Should all of the structs and enums from this file be internal?

glbrntt: They should be no more visible than `internal`, fine if some are `fileprivate`/`private`. There shouldn't be any `public` types though.

PR #1726, opened 2023-11-28
Added TypeName and TypeUsage from OpenAPI Generator.

stefanadranca:
Motivation:

The types are used by the code renderer that will be brought over from OpenAPI Generator to the new gRPC Swift CodeGenLib.

Modifications:

Copied the TypeName, TypeUsage and some TypeName extensions that will be used in the context of gRPC, and modified TypeName so it represents only the Swift path of a type.

Result:

We will be able to bring over the Code Renderer and we will have types representations that can be used by the service translator.

### DIFF
--- a/Sources/GRPCCodeGen/Internal/TypeName.swift
+++ b/Sources/GRPCCodeGen/Internal/TypeName.swift
@@ -34,45 +34,26 @@ import Foundation
 /// Use the type name to define a type, see also `TypeUsage` when referring
 /// to a type.
 struct TypeName: Hashable {
-
-  /// Describes a single component of the Swift path.
-  struct Component: Hashable {
-    /// The name of the Swift path component.
-    var swift: String?
-  }
-
   /// A list of components that make up the type name.
-  private let components: [Component]
-
-  /// The list of Swift path components.
-  var swiftKeyPathComponents: [String] { components.compactMap(\.swift) }
+  private let components: [String]
 
   /// Creates a new type name with the specified list of components.
   /// - Parameter components: A list of components for the type.
-  init(components: [Component]) {
-    precondition(!components.compactMap(\.swift).isEmpty, "TypeName Swift key path cannot be empty")
+  init(components: [String]) {
+    precondition(!components.isEmpty, "TypeName path cannot be empty")
     self.components = components
-  }
-
-  /// Creates a new type name with the specified list of Swift path
-  /// components.
-  ///
-  /// - Parameter swiftKeyPath: A list of Swift path components for the type.
-  init(swiftKeyPath: [String]) {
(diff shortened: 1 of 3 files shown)


Sources/GRPCCodeGen/Internal/TypeName.swift
```diff
@@ -0,0 +1,127 @@
...
+  /// - Parameter name: The name of the type.
+  /// - Returns: A TypeName representing the specified type within the Swift module.
+  static func swift(_ name: String) -> TypeName { TypeName(swiftKeyPath: ["Swift", name]) }
+
+  /// Returns a type name for a type with the specified name in the
+  /// Foundation module.
+  /// - Parameter name: The name of the type.
+  /// - Returns: A TypeName representing the specified type within the Foundation module.
+  static func foundation(_ name: String) -> TypeName {
+    TypeName(swiftKeyPath: ["Foundation", name])
```
stefanadranca, line 125: This extension is from another file - `Builtins.swift`. Should I signal this in any way?

Sources/GRPCCodeGen/Internal/TypeName.swift
```diff
@@ -0,0 +1,127 @@
...
+struct TypeName: Hashable {
+
+  /// Describes a single component of the Swift path.
+  struct Component: Hashable {
+    /// The name of the Swift path component.
+    var swift: String?
+  }
+
+  /// A list of components that make up the type name.
+  private let components: [Component]
```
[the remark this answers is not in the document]
glbrntt: No, we should just use `String`s here which means we can simplify more of this type too.

Sources/GRPCCodeGen/Internal/TypeName.swift
glbrntt: No, this is fine.

glbrntt (changes requested):

Sources/GRPCCodeGen/Internal/TypeName.swift
```diff
@@ -0,0 +1,107 @@
...
+  /// Returns a type name by appending the specified components to the
+  /// current type name.
+  ///
+  /// In other words, returns a type name for a child type.
+  /// - Precondition: At least one of the components must be non-nil.
+  /// - Parameters:
+  ///   - component: The name of the Swift type component.
+  /// - Returns: A new type name.
+  func appending(component: String? = nil) -> Self {
+    precondition(component != nil, "The type name must be non-nil.")
```
glbrntt, line 67: This is a weird combination, we should just make the `String` non-optional

glbrntt (changes requested):
One nit, looks good otherwise

glbrntt (approved):

PR #1727, opened 2023-11-29
Add new GRPCClient object

gjcairo:
## Motivation

Clients are the highest level API to make requests to a gRPC server.

## Modifications

Add a `GRPCClient`

## Result

We now have a high level client that can be used by the generated stubs to start RPCs to gRPC servers.


Sources/GRPCCore/GRPCClient.swift
```diff
@@ -0,0 +1,823 @@
...
+        }
+      }
+    }
+  }
+
+  /// A ``MethodConfigurationRegistry`` containing ``MethodConfiguration``s for calls
+  /// made from this ``Client``.
+  ///
+  /// - Note: These configurations will override those configurations set in the ``ClientTransport``.
+  public var methodConfigurationOverrides: MethodConfigurationRegistry {
```
glbrntt, line 93: These aren't just overrides, they're the default configuration too. I _think_ we probably need two of these: one for overrides one for defaults. 

I think we should stick these in a small wrapper type like the interceptors too:

```swift
client.methodConfiguration.overrides[foo] = ...
client.methodConfiguration.defaults[bar] = ...
```

I think this means that the storage for these doesn't need to offer a default value either, the client can just create the empty config if we didn't find one.

---

Naming here get's awkward: we'd have `GRPCClient.MethodConfiguration` as the wrapper containing the collections of what-used-to-be-called `ClientRPCExecutionConfiguration` for the `overrides`/`defaults`.

Maybe we should keep the existing name and nest the registry thing within `MethodConfiguration`?

Sources/GRPCCore/GRPCClient.swift
```diff
@@ -0,0 +1,823 @@
...
+        storage.state = .stopping
+      case .stopping, .stopped:
+        ()
+      }
+    }
+
+    self.transport.close()
+  }
+
+  /// Start a unary RPC.
```
glbrntt, line 208: This doesn't just start the RPC, it finishes it too. We can just say "executes" though.

Sources/GRPCCore/GRPCClient.swift
```diff
@@ -0,0 +1,823 @@
...
+/// You can also use ``ClientInterceptor``s to implement cross-cutting logic which apply to all
+/// RPCs. Example uses of interceptors include authentication and logging.
+///
+/// ## Creating and configuring a client
+///
+/// The following example demonstrates how to create and configure a server.
+///
+/// ```swift
+/// // Create and add an in-process transport.
+/// let inProcessTransport = InProcessClientTransport()
```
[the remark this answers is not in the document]
FranzBusch: Out of interest are you planning to have an `InProcessClientTransport` and `InProcessServerTransport` or just one `InProcessTransport` that implements both?

Sources/GRPCCore/GRPCClient.swift
```diff
@@ -0,0 +1,823 @@
...
+/// ```
+///
+/// ## Starting and stopping the client
+///
+/// Once you have configured the client, call ``run()`` to start it. Calling ``run()`` connects to the given
+/// transport.
+///
+/// ```swift
+/// // Start running the client.
+/// try await client.run()
```
FranzBusch, line 59: Since this method won't return IMO it might be worth slightly extending the example to call run in a child task and use the client to execute an example request

Sources/GRPCCore/GRPCClient.swift
```diff
@@ -0,0 +1,823 @@
...
+/// Lifecycle](https://github.com/swift-server/swift-service-lifecycle).
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+public final class GRPCClient: Sendable {
+  /// A collection of ``ClientInterceptor`` implementations which are applied to all accepted
+  /// RPCs.
+  ///
+  /// RPCs are intercepted in the order that interceptors are added. That is, a request sent from the client to
+  /// the server will first be intercepted by the first added interceptor followed by the second, and so on.
+  /// For responses from the server, they'll be applied in the opposite order.
+  public var interceptors: Interceptors {
```
FranzBusch, line 76: What happens if I change interceptors while a request is currently making its way through the interceptor pipeline?

Sources/GRPCCore/GRPCClient.swift
```diff
@@ -0,0 +1,823 @@
...
+    /// The client is stopping and no new RPCs will be sent. Existing RPCs may run to
+    /// completion. May transition to `stopped`.
+    case stopping
+    /// The client has stopped, no RPCs are in flight and no more will be accepted. This state
+    /// is terminal.
+    case stopped
+  }
+
+  /// Underlying storage for the client.
+  private struct Storage {
```
FranzBusch, line 121: This is the opposite of what we normally do. We put the storage into the associated values of the `State` enum. The advantage of that is that once we transition to `stopped` we free all of the retained storage.

Sources/GRPCCore/GRPCClient.swift
```diff
@@ -0,0 +1,823 @@
...
+      self.interceptors = Interceptors()
+      self.methodConfigurationOverrides = MethodConfigurationRegistry()
+      self.state = .notStarted
+    }
+  }
+
+  private let storage: LockedValueBox<Storage>
+
+  /// The transport which provides a bidirectional communication channel with the server.
+  private let transport: ClientTransport
```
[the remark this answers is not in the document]
FranzBusch: Maybe you wanna enable `ExistentialAny` for your new targets?

Sources/GRPCCore/GRPCClient.swift
```diff
@@ -0,0 +1,823 @@
...
+
+  /// Creates a new client with no resources.
```
FranzBusch, line 138: I find `resources` a bit weird naming wise. Like what resource?

Sources/GRPCCore/GRPCClient.swift
```diff
@@ -0,0 +1,823 @@
...
+        )
+      }
+    }
+
+    // When we exit this function we must have stopped.
+    defer {
+      self.storage.withLockedValue { $0.state = .stopped }
+    }
+
+    try await withThrowingTaskGroup(of: Void.self) { group in
```
FranzBusch, line 180: Why do we use a task group with a single task here? :D

Sources/GRPCCore/GRPCClient.swift
```diff
@@ -0,0 +1,823 @@
...
+      try await group.next()
+    }
+  }
+
+  /// Close the client.
+  ///
+  /// The transport will be closed: this means that it will be given enough time to wait for in-flight RPCs to
+  /// finish executing, but no new RPCs will be accepted.
+  /// You can cancel the task executing ``run()`` if you want to immediately stop all work.
+  public func close() {
```
FranzBusch, line 193: IMO `close` is the wrong name here and we should follow the lifecycle naming of `triggerGracefulShutdown`

Sources/GRPCCore/GRPCClient.swift
```diff
@@ -0,0 +1,823 @@
...
+          code: .clientIsNotRunning,
+          message: "Client must be running to make an RPC: call run() first."
+        )
+      case .stopping, .stopped:
+        throw ClientError(
+          code: .clientIsStopped,
+          message: "Client has been stopped. Can't make any more RPCs."
+        )
+      }
+    }
```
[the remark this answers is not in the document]
FranzBusch: Yes agreed. We just need to implement the bi-di version and can derive the other method from it

Sources/GRPCCore/GRPCClient.swift
```diff
@@ -0,0 +1,823 @@
...
+
+    return configurationOverrides[descriptor]
+  }
+}
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+extension GRPCClient {
+  /// A collection of interceptors providing cross-cutting functionality to each accepted RPC.
+  public struct Interceptors: Sendable {
```
FranzBusch, line 440: Do we ever want to actually make this fully generic? So you can have a `GRPCCLient` with a generic interceptor pipeline which hopefully gets specialised. @glbrntt

Sources/GRPCCore/GRPCClient.swift
```diff
@@ -0,0 +1,823 @@
...
+    public mutating func add(_ interceptor: some ClientInterceptor) {
+      self.values.append(interceptor)
+    }
+  }
+}
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+extension GRPCClient {
+  /// The execution policy for an RPC.
+  public enum ExecutionPolicy: Hashable, Sendable {
```
FranzBusch, line 460: Can we use a struct for this instead?

Sources/GRPCCore/GRPCClient.swift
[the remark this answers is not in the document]
glbrntt: Separate types. You could have them as a single object but it gets a bit weird as they have separate state management.

Sources/GRPCCore/GRPCClient.swift
glbrntt: Why do you think it's the wrong name?

Sources/GRPCCore/GRPCClient.swift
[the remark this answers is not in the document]
gjcairo: We already have both (`InProcessClientTransport` and `InProcessServerTransport`)

Sources/GRPCCore/GRPCClient.swift
glbrntt: I don't think the generics carry their weight at the interceptor level tbh. Having a client generic over a transport and a bunch of interceptors hurts ergonomics quite badly.

Sources/GRPCCore/GRPCClient.swift
glbrntt: We discussed this one previously (it's just been renamed from elsewhere)

Sources/GRPCCore/GRPCClient.swift
gjcairo: You can't modify interceptors once the Client is running (and you cannot execute RPCs unless the Client is running either) so this cannot happen.

Sources/GRPCCore/GRPCClient.swift
glbrntt: Not always. We tend to do that when the state management and dependencies are complicated. That's not the case here.

Sources/GRPCCore/GRPCClient.swift
glbrntt: Nothing. There's a state check _and_ we grab the array of interceptors before executing the request.

Sources/GRPCCore/GRPCClient.swift
gjcairo: I borrowed the naming from the docs of the `GRPCServer`. In this case, it refers to the `Interceptors` and the `MethodConfiguration`s. Happy to change this to a different word, or add a "(i.e.: `Interceptors` and `MethodConfiguration`s)" to make it clearer :D

Sources/GRPCCore/GRPCClient.swift
FranzBusch: I see but I also find it a bit weird. Like you are calling `interceptors = []` after calling run and that call just gets dropped and nothing happens. Why don't we just make them a part of the init and not allow modifying them afterwards?

Sources/GRPCCore/GRPCClient.swift
FranzBusch: Sure but I still think there is benefits to that approach like freeing up the resources as soon as possible. Right now we are going to retain the interceptors for longer than we need.

Sources/GRPCCore/GRPCClient.swift
FranzBusch: > Why do you think it's the wrong name?

At least coming from a NIO world or an FD world `close` means "close right now". However, from the documentation here it sounds way more like a graceful shutdown than an immediate close. Furthermore, we are kinda trying to establish the pattern of `gracefulShutdown` with `ServiceLifecycle` so I think aligning on a common name is beneficial for the ecosystem.

PR #1728, opened 2023-11-29
Brought CodeRenderer from OpenAPI Generator.

stefanadranca:
Motivation:

Provides an API which takes Structured Swift Representation of a file and generates an object containing the generated code as a String.

Modifications:

- Brought over RendererProtocol and TextBasedRenderer.
- Brought over unit tests in Test_TextBasedRenderer.
- Fixed TypeName bugs that produced errors while building.

Result:

The CodeGenLib will have a code renderer that transforms `StructuredSwiftRepresentation` into `SourceFile`.


Tests/GRPCTests/GRPCCodeGen/Test_TextBasedRenderer.swift
```diff
@@ -0,0 +1,781 @@
+/*
```
glbrntt, line 1: `GRPCTests` are tests for v1. Since `GPRCCodeGen` is a new target for v2, it should have its own corresponding test target `GRPCCodeGenTests`.

Package.swift
```diff
@@ -217,6 +217,7 @@ extension Target {
       .nioTransportServices,
       .logging,
       .reflectionService,
+      .grpcCodeGen
```
glbrntt, line 220: This should be removed

Package.swift
```diff
@@ -534,6 +542,7 @@ let package = Package(
     .protocGenGRPCSwift,
     .grpcSwiftPlugin,
     .reflectionService,
+    .grpcCodeGen,
```
glbrntt, line 545: nit: move this down to the `// v2` section below `.grpcCore`.

Tests/GRPCCodeGenTests/Test_TextBasedRenderer.swift
```diff
@@ -0,0 +1,781 @@
+/*
```
glbrntt, line 1: re: filename, I know you copy-pasted from OpenAPI, but naming is generally `<Foo>Tests.swift`, this should be `TextBasedRendererTests.swift`. We're also trying to keep the hierarchy of library and test code the same so it should be `GRPCCodeGenTests/Internal/Renderer/TextBasedRendererTests.swift`.

glbrntt (approved):

PR #1729, opened 2023-11-30
Add Client object

glbrntt:
Motivation:

Clients are the highest level API to make requests to a gRPC server.

Modifications:

Add a `GRPCClient`

Result:

We now have a high level client that can be used by the generated stubs to start RPCs to gRPC servers.


glbrntt:
This builds on top of and supersedes https://github.com/grpc/grpc-swift/pull/1727

Sources/GRPCCore/ClientError.swift
```diff
@@ -0,0 +1,148 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/// A runtime error thrown by the client.
+///
+/// In contrast to ``RPCError``, the ``ClientError`` represents errors which happen at a scope
+/// wider than an individual RPC. For example, attempting to start a client which is already
+/// stopped would result in a ``ClientError``.
+public struct ClientError: Error, Hashable, @unchecked Sendable {
```
FranzBusch, line 22: NIT: Should we name this `GRPCClientError` if the type is called `GRPCClient`?

Sources/GRPCCore/GRPCClient.swift
```diff
@@ -0,0 +1,473 @@
...
+/// more abruptly you can cancel the task running your client. If your application requires
+/// additional resources that need their lifecycles managed you should consider using [Swift Service
+/// Lifecycle](https://github.com/swift-server/swift-service-lifecycle).
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+public struct GRPCClient: Sendable {
+  /// The transport which provides a bidirectional communication channel with the server.
+  private let transport: any ClientTransport
+
+  /// The configuration used by the client.
+  public let configuration: Configuration
```
FranzBusch, line 105: Do we really want to make this public. Users cannot change anything anymore on it. Do you have a use-case in mind where people might want to access this?

Sources/GRPCCore/MethodConfigurations.swift
```diff
@@ -0,0 +1,92 @@
...
+  /// Configuration is hierarchical and can be set per-method, per-service
+  /// (``setDefaultConfiguration(_:forService:)``) and globally (``setDefaultConfiguration(_:)``).
+  /// This subscript sets the per-method configuration but retrieves a configuration respecting
+  /// the hierarchy. If no per-method configuration is present, the per-service configuration is
+  /// checked and returned if present. If the per-service configuration isn't present then the
+  /// global configuration is returned, if present.
+  ///
+  /// - Parameters:
+  ///  - descriptor: The ``MethodDescriptor`` for which to get or set a ``MethodConfiguration``.
+  public subscript(_ descriptor: MethodDescriptor) -> MethodConfiguration? {
```
FranzBusch, line 45: Does it make sense to subscript for a service in the future? Then I think making the subscript argument label external makes sense.
```suggestion
  public subscript(descriptor: MethodDescriptor) -> MethodConfiguration? {
```

Sources/GRPCCore/GRPCClient.swift
glbrntt: Visibility, i.e. being able to see what the configuration is.

FranzBusch (approved):

PR #1731, opened 2023-12-04
Drop the '.txt' extension from reflection data

glbrntt:
Motivation:

Generated reflection data has a ".txt" extension. This isn't quite right as the contents is base64 encdoded data. Moreover the generated reflection data is intended as an implementation detail for the reflection service.

Modifications:

- Change the generated reflection data extension from ".grpc.reflection.txt" to ".grpc.reflection".
- Fix a bug in the reflection server example where resources couldn't be loaded because no subdirectory was provided

Result:

Generated reflection data files don't have a ".txt" extension


stefanadranca (approved):
Looks good to me!👍

PR #1733, opened 2023-12-05
[CodeGenLib] Translator for enums containing type aliases and static properties

stefanadranca:
Motivation:

In the generated code we want to have:

- enums for each namespace containing enums for all their services
- enums for each service (inside the corresponding namespace enum) containing
enums for all their methods, an array of all corresponding MethodDescriptors and type aliases for streaming and non-streaming protocols
- enums for each method (inside the corresponding service enum) containing
type aliases for the input and output types and a method descriptor (GRPCCore)

Modifications:

- Created the Translator protocol defining the 'translate()' function.
- Implemented the IDLToStructuredSwiftTranslator struct which conforms to the Translator protocol and creates the StructuredSwiftRepresentation for a CodeGenerationRequest calling the Specialized translators' functions.
- Created the SpecializedTranslator protocol.
- Implemented the TypeAliasTranslator, the first of the 3 SpecializedTranslators that takes in a CodeGenerationRequest object and creates all the enums, type aliases and properties mentioned in the Motivation, in StructuredSwiftRepresentation format.
- Created CodeGenError.
- Wrote SnippetTests.

Result:

The generated code will contain useful type aliases for message types and
protocol names, organized in namespace, service and method specific enums.


Sources/GRPCCodeGen/Internal/Translator/Translator.swift
```diff
@@ -0,0 +1,35 @@
...
+  /// - Parameters:
+  ///   - codeGenerationRequest: The IDL described RPCs representation.
+  ///   - visibility: The access control level for the generated code.
+  ///   - client: Whether or not client code should be generated from the IDL described RPCs representation.
+  ///   - server: Whether or not server code should be generated from the IDL described RPCs representation.
+  /// - Returns: A structured Swift representation of the generated code.
+  /// - Throws: An error if there are issues translating the codeGenerationRequest.
+  func translate(
+    codeGenerationRequest: CodeGenerationRequest,
+    //    visibility: Visibility,
```
stefanadranca, line 31: the Visibility type will be defined in the SourceGenerator API, which is not implemented yet.

Sources/GRPCCodeGen/Internal/Translator/TypeAliasTranslator.swift
```diff
@@ -0,0 +1,209 @@
...
+    return codeBlocks
+  }
+}
+
+extension TypeAliasTranslator {
+  func translateNamespaceEnumDeclaration(
+    for namespace: String,
+    containing services: [CodeGenerationRequest.ServiceDescriptor]
+  ) -> Declaration {
+    var namespaceEnumDescription = EnumDescription(name: namespace)
```
glbrntt, line 77: nit: we can shorten this name (and others) to `namespaceEnum`

Sources/GRPCCodeGen/Internal/Translator/IDLToStructuredSwiftTranslator.swift
```diff
@@ -0,0 +1,43 @@
...
+  let typeAliasTranslator = TypeAliasTranslator()
+
+  func translate(
+    codeGenerationRequest: CodeGenerationRequest,
+    client: Bool,
+    server: Bool
+  ) -> StructuredSwiftRepresentation {
+    var topComment = Comment.doc(codeGenerationRequest.leadingTrivia)
+    var imports: [ImportDescription] = [
+      ImportDescription(moduleName: "SwiftProtobuf"),
```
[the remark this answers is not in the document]
glbrntt: That's right, `SwiftProtobuf` and `WhateverModuleWeAddProtobufSerializersToo` will come as imports in the code gen request.

Sources/GRPCCodeGen/Internal/Translator/TypealiasTranslator.swift
```diff
@@ -0,0 +1,200 @@
...
+      let codeBlockItem = CodeBlockItem.declaration(namespaceEnumDeclaration)
+      codeBlocks.append(CodeBlock(item: codeBlockItem))
+    }
+
+    return codeBlocks
+  }
+}
+
+extension TypealiasTranslator {
+  func makeNamespaceTypealiasesEnum(
```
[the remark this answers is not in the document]
glbrntt: Also given this is the typealias translator, we can drop that from all of these method names

Tests/GRPCCodeGenTests/Internal/Translator/SnippetBasedTranslatorTests.swift
```diff
@@ -0,0 +1,200 @@
...
+    let expectedSwift =
+      """
+      enum echo {
+          enum Echo {
+              enum Collect {
+                  typealias Input = Echo_EchoRequest
+                  typealias Output = Echo_EchoResponse
+                  static let descriptor = MethodDescriptor(
+                      service: echo.Echo,
+                      method: Collect
```
glbrntt, line 105: This isn't correct, these should be strings:
```suggestion
                      service: "echo.Echo",
                      method: "Collect"
```

Tests/GRPCCodeGenTests/Internal/Translator/SnippetBasedTranslatorTests.swift
```diff
@@ -0,0 +1,200 @@
...
+      lookupSerializer: {
+        "ProtobufSerializer<\($0)>()"
+      },
+      lookupDeserializer: {
+        "ProtobufDeserializer<\($0)>()"
+      }
+    )
+  }
+
+  func testTypealiasTranslate() throws {
```
glbrntt, line 95: This test is much bigger than is necessary, it's also not sufficient as a test.

Here are a handful of tests I thought of and what each is checking.
- One service in a namespace with one method. ("basic" case)
- One service with an _empty_ namespace and one method. (Checks that an empty namespace is treated as no namespace)
- One service in a namespace with two methods. (Checks order of methods is preserved). 
- One service in a namespace with no methods. (A valid edge case)
- Two services in the same namespace each with no methods where the services are in reverse alphabetical order in the request. (Checks that the services are ordered alphabetically within a namespace)
- Two services, both with empty namespaces, each with no methods. (Checks that the services are ordered alphabetically even when there's no namespace.)
- Two services in different namespaces each with no methods where the namespaces are in reverse alphabetical order in the request. (Checks that the namespaces are ordered alphabetically)
- Two services, one in a namespace, one not in a namespace, each with no methods. (Checks the order of namespace vs. no namespace)

There's also a handful of additional validation tests we can do:
- Two services with the same name and no namespace throws an error
- Two services with the same name and same namespace throws an error
- Two methods with the same name in a service throws an error

Tests/GRPCCodeGenTests/Internal/Translator/SnippetBasedTranslatorTests.swift
```diff
@@ -0,0 +1,200 @@
...
+      """
+      XCTAssertEqualWithDiff failed (click for diff)
+      \(try diff(expected: expected, actual: actual))
+      """,
+      file: file,
+      line: line
+    )
+  }
+
+  func _assertTypealiasTranslation(
```
glbrntt, line 189: This shouldn't need to be prefixed with an underscore. It can also be `private`.

glbrntt (changes requested):

Tests/GRPCCodeGenTests/Internal/Translator/SnippetBasedTranslatorTests.swift
```diff
@@ -0,0 +1,487 @@
...
+      documentation: "Documentation for ServiceA",
+      name: "ServiceA",
+      namespace: "namespaceA",
+      methods: []
+    )
+    let expectedSwift =
+      """
+      enum namespaceA {
+          enum ServiceA {
+              static let methods: [MethodDescriptor] = []
```
glbrntt, line 183: We should still generate an empty `Methods` enum

Tests/GRPCCodeGenTests/Internal/Translator/SnippetBasedTranslatorTests.swift
```diff
@@ -0,0 +1,487 @@
...
+      methods: []
+    )
+    let expectedSwift =
+      """
+      enum BService {
+          static let methods: [MethodDescriptor] = []
+          typealias StreamingServiceProtocol = BServiceServiceStreamingProtocol
+          typealias ServiceProtocol = BServiceServiceProtocol
+      }
+      enum anamespace {
```
glbrntt, line 327: This highlights another potential issue: a service with no namespace can't have the same name as another namespace in the request. We'll need a test for this too.

Tests/GRPCCodeGenTests/Internal/Translator/SnippetBasedTranslatorTests.swift
```diff
@@ -0,0 +1,487 @@
...
+    let serviceB = ServiceDescriptor(
+      documentation: "Documentation for BService",
+      name: "AService",
+      namespace: "",
+      methods: []
+    )
+
+    let codeGenerationRequest = self.makeCodeGenerationRequest(services: [serviceA, serviceB])
+    let translator = TypealiasTranslator()
+    self.assertThrowsError(try translator.translate(from: codeGenerationRequest)) {
```
glbrntt, line 358: Just copy this from the core-tests: https://github.com/grpc/grpc-swift/blob/79865359b894e8334e0d3b520294ceb2dfc7bcfd/Tests/GRPCCoreTests/Test%20Utilities/XCTest%2BUtilities.swift#L40-L51

Tests/GRPCCodeGenTests/Internal/Translator/SnippetBasedTranslatorTests.swift
```diff
@@ -0,0 +1,487 @@
...
+      name: "AService",
+      namespace: "",
+      methods: []
+    )
+    let serviceB = ServiceDescriptor(
+      documentation: "Documentation for BService",
+      name: "AService",
+      namespace: "",
+      methods: []
+    )
```
glbrntt, line 354: You don't need this, just pass `serviceA` twice

glbrntt (changes requested):

Sources/GRPCCodeGen/Internal/Translator/TypealiasTranslator.swift
```diff
@@ -0,0 +1,266 @@
...
+    /// If the namespace is empty, the service enums are independent CodeBlocks.
+    if namespace.isEmpty {
+      return namespaceEnum.members.map {
+        CodeBlock(item: .declaration($0))
+      }
+    }
+    return [CodeBlock(item: .declaration(.enum(namespaceEnum)))]
+  }
+
+  private func makeServiceEnum(
```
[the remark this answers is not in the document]
glbrntt: This is much simpler now!

glbrntt (changes requested):

glbrntt (approved):

PR #1738, opened 2023-12-08
Add a convenience method for creating in-process client/server

glbrntt:
Motivation:

To use the in-process transport users must create a server and then a client depending on that server. It rarely makes sense to create one without the other and I have written more or less the same extension in three places to do this for me and return a pair. This should just be made into API.

Modifications:

- Add a `InProcessTransport.makePair()` convenience method
- Replace use of old helpers

Result:

It's easier to create in-process transport.

### DIFF
--- a/Sources/GRPCInProcessTransport/InProcessTransport.swift
+++ b/Sources/GRPCInProcessTransport/InProcessTransport.swift
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2023, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


Sources/GRPCInProcessTransport/InProcessTransport.swift
```diff
@@ -0,0 +1,43 @@
...
+  /// and a client using that server transport.
+  ///
+  /// - Parameters:
+  ///   - methodConfiguration: Method specific configuration used by the client transport to
+  ///       determine how RPCs should be executed.
+  ///   - retryThrottle: The retry throttle the client transport uses to determine whether a call
+  ///       should be retried.
+  /// - Returns: A tuple containing the connected server and client in-process transports.
+  @available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+  public static func makePair(
```
FranzBusch, line 31: Optional nit: I would prefer `makeClientAndServerTransport`

FranzBusch (approved):

PR #1744, opened 2023-12-12
allow setting protoc-gen-grpc-swift via config

Lutzifer:
I ran into the same issue described in https://github.com/apple/swift-protobuf/issues/1506 where protoc-gen-grpc-swift is missing some frameworks it needs to run.

Until this is solve(able/d), this PR allows the user to set the path to protoc-gen-grpc-swift via the configuration, e.g. to use protoc-gen-grpc-swift installed via mint.

Is this something that could be merged? if so, I can update the docs and the tests for this.

- [ ] TODO: update docs
- [ ] TODO: update tests

### DIFF
--- a/Plugins/GRPCSwiftPlugin/plugin.swift
+++ b/Plugins/GRPCSwiftPlugin/plugin.swift
@@ -80,6 +80,11 @@ struct GRPCSwiftPlugin {
     ///
     /// If this is not set, SPM will try to find the tool itself.
     var protocPath: String?
+    
+    /// The path to the `protoc-gen-grpc-swift` binary.
+    ///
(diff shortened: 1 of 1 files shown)


glbrntt:
This change doesn't seem right to me at all. `protoc-gen-grpc-swift` is a dependency of the build plugin so should be built and provided by the build system. 

If the build system has issues building the dependency then please file a bug against the build system. In the meantime you workaround this by generating your code manually.

Lutzifer:
fine for me, I'll keep using my branch until the build system is fixed then.

Haven't had the time to setup a minimal example to provide a bug report, yet.

PR #1746, opened 2023-12-19
Update available to fix build for non-macOS Apple platforms

simonjbeaumont:
## Motivation

The package currently doesn't build for non-macOS Apple platforms, e.g. iOS, because of missing `@available` annotations, mostly in tests.

## Modifications

- Add missing `@available` annotations.
- Use `#if os(macOS) || os(Linux)` in test utils that require `Foundation.Process`. Ideally we'd use `#if canImport(Foundation.Process)`, but the version of `swift-format` used by this project doesn't understand it.

## Result

Code and tests can build for, and run on, other platforms, e.g. iOS.

### DIFF
--- a/Tests/GRPCCoreTests/Test Utilities/RPCAsyncSequence+Utilities.swift
+++ b/Tests/GRPCCoreTests/Test Utilities/RPCAsyncSequence+Utilities.swift
@@ -16,7 +16,6 @@
 import GRPCCore
 
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
-@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
 extension RPCAsyncSequence {
   static func elements(_ elements: Element...) -> Self {
     return .elements(elements)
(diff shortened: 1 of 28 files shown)


Tests/GRPCCoreTests/Test Utilities/RPCAsyncSequence+Utilities.swift
```diff
@@ -15,6 +15,8 @@
  */
 import GRPCCore
 
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
```
glbrntt, line 19: copy-pasta-pasta?

```suggestion
@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
```

glbrntt (approved):
Looks good modulo one duplicated availability guard, thanks Si!

Tests/GRPCCoreTests/Test Utilities/RPCAsyncSequence+Utilities.swift
simonjbeaumont: Yum—nom nom.

PR #1748, opened 2023-12-20
[CodeGenLib] Translator for Client Code

stefanadranca:
Motivation:
In the generated client code for a service described in a Source IDL file we want to have:
- a protocol declaring the methods for a client, with the serializer and deserializer as parameters.
- an extension for the protocol defining the methods of the client, without the serializer and deserializer as parameters,
that are calling the already declared methods, providing the default Protobuf serializer and deserializer.
- a struct representing the generated client that has a GRPCClient as a property and conforms to the Client Protocol.

Modifications:
- Implemented the ClientCodeTranslator struct that contains all the methods for creating the `StructuredSwiftRepresentation` for the client code.
- Added a new struct in the `StructuredSwiftRepresentation` that represents closure signatures, and a new `ParameterDescription` property that is set only when the parameter is a closure signature.
- Added and modified render functions for the `StructuredSwiftRepresentation` additions.
- Implemented snippet tests.

Result:
Client code can now be generated for a CodeGenerationRequest object.


Sources/GRPCCodeGen/Internal/StructuredSwiftRepresentation.swift
```diff
@@ -503,7 +514,12 @@ enum FunctionKind: Equatable, Codable {
   /// A function or a method. Can be static.
   ///
   /// For example `foo()`, where `name` is `foo`.
-  case function(name: String, isStatic: Bool)
+  case function(
+    name: String,
+    isStatic: Bool,
+    genericType: String? = nil,
+    conformances: [String] = []
```
stefanadranca, line 521: We need these to be able to represent generic functions

Sources/GRPCCodeGen/Internal/StructuredSwiftRepresentation.swift
```diff
@@ -483,7 +488,13 @@ struct ParameterDescription: Equatable, Codable {
...
   ///
   /// For example, in `bar baz: String = "hi"`, `type` is `String`.
-  var type: ExistingTypeDescription
+  var type: ExistingTypeDescription? = nil
+
+  /// The description of the closure parameter.
+  ///
+  /// The parameter doesn't have a type, but is a closure.
+  /// For example: `(String) async throws -> Int`.
+  var closureDescription: ClosureSignatureDescription? = nil
```
stefanadranca, line 497: I have decided to have a different variable that signals if the parameter is a closure (and also its description) because the ExistingTypeDescription rendering doesn't allow more complex types.

Sources/GRPCCodeGen/Internal/StructuredSwiftRepresentation.swift
glbrntt: This isn't quite right: the "conformance" here applies to the generic type, not the function. Also functions can have multiple generic types. 

It looks like the best way to represent this would be by modifying the `FunctionSignatureDescription` to add: a `generics: [ExistingTypeDescription]` property and an optional `var whereClause: WhereClause?`.

That way we can express the `Sendable` requirement as a where clause:

```swift
func foo<R>() -> R where R: Sendable { 
 // ...
}
```

glbrntt (changes requested):

Sources/GRPCCodeGen/Internal/Translator/ClientCodeTranslator.swift
```diff
@@ -0,0 +1,414 @@
...
+    return parameters
+  }
+  private func clientRequestParameter(
+    for method: CodeGenerationRequest.ServiceDescriptor.MethodDescriptor,
+    in service: CodeGenerationRequest.ServiceDescriptor
+  ) -> ParameterDescription {
+    let clientRequestType =
+      method.isInputStreaming
+      ? ExistingTypeDescription.member(["ClientRequest", "Stream"])
+      : ExistingTypeDescription.member(["ClientRequest", "Single"])
```
glbrntt, line 223: nit: minimise branching

```swift
let requestType = method.isInputStreaming ? "Stream" : "Single"
let clientRequestType = ExistingTypeDescription.member(["ClientRequest", requestType])
```

Sources/GRPCCodeGen/Internal/Translator/ClientCodeTranslator.swift
```diff
@@ -0,0 +1,414 @@
...
+    if method.isInputStreaming && method.isOutputStreaming {
+      return "bidirectionalStreaming"
+    }
+    if method.isInputStreaming && !method.isOutputStreaming {
+      return "clientStreaming"
+    }
+    if !method.isInputStreaming && method.isOutputStreaming {
+      return "serverStreaming"
+    }
+    return "unary"
```
glbrntt, line 342: There's a neater way of doing this which ensures you cover all possibilities:

```swift
switch (method.isInputStreaming, method.isOutputStreaming) {
case (true, true):
  return "bidi..."
case (true, false):
  return "client..."
case (false, true):
  return "server..."
case (false, false):
  return "unary"
}
```

Sources/GRPCCodeGen/Internal/Translator/ClientCodeTranslator.swift
```diff
@@ -0,0 +1,414 @@
...
+    return .function(
+      signature: .init(
+        kind: .initializer,
+        parameters: [.init(label: "client", type: .member(["GRPCCore", "GRPCClient"]))]
+      ),
+      body: [CodeBlock(item: .expression(initializerBody))]
+    )
+  }
+
+  private func getGRPCMethodName(
```
glbrntt, line 330: "grpc method name" is misleading here, it sounds like the name of the RPC, `clientMethod` is a bit clearer. Also we should just pass in two parameters here: input and output streaming

glbrntt (changes requested):

glbrntt (approved):

PR #1750, opened 2023-12-21
[CodeGenLib] Typealias translator configuration for client and server

stefanadranca:
Typealias translator configuration for client and server type aliases

Motivation:

The typealias translator should generate typealiases for the client and server protocols only if the associated code (client/server code) will be generated. Also, at the moment the typealias translator doesn't generate the `ClientProtocol` typealias.

Modifications:

- added a new initialiser for the `TypealiasTranslator` that sets the client and server property.
- Based on the client and server properties' values, the typealias translator generates or not the protocol typealiases.
- created method for client protocol typalias generation.
- modified tests, so they include the client code.
- added new tests for generating only client or server code.

Result:

The typealias translator will generate typealiases for client and server protocols only when the associated code is also generated, getting rid of the possibility of an error (when the protocol doesn't exist).

### DIFF
--- a/Sources/GRPCCodeGen/Internal/Translator/TypealiasTranslator.swift
+++ b/Sources/GRPCCodeGen/Internal/Translator/TypealiasTranslator.swift
@@ -53,6 +53,14 @@
 /// A ``CodeGenerationRequest`` can contain multiple namespaces, so the TypealiasTranslator will create a ``CodeBlock``
 /// for each namespace.
 struct TypealiasTranslator: SpecializedTranslator {
+  let client: Bool
+  let server: Bool
+
+  init(client: Bool, server: Bool) {
+    self.client = client
+    self.server = server
+  }
+
   func translate(from codeGenerationRequest: CodeGenerationRequest) throws -> [CodeBlock] {
     var codeBlocks: [CodeBlock] = []
     let services = codeGenerationRequest.services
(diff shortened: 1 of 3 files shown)


Sources/GRPCCodeGen/Internal/Translator/TypealiasTranslator.swift
```diff
@@ -307,4 +323,19 @@ extension TypealiasTranslator {
...
   }
+
+  private func makeClientProtocolTypealias(
+    for service: CodeGenerationRequest.ServiceDescriptor
+  ) -> Declaration {
+    let namespacedPrefix: String
+    if service.namespace.isEmpty {
+      namespacedPrefix = service.name
+    } else {
+      namespacedPrefix = "\(service.namespace)_\(service.name)"
```
stefanadranca, line 334: After merging the server code translator PR, I will use the service property from the extension to get the namespacedPrefix

Tests/GRPCCodeGenTests/Internal/Translator/SnippetBasedTranslatorTests.swift
```diff
@@ -58,13 +58,122 @@ final class SnippetBasedTranslatorTests: XCTestCase {
...
+    )
+    let expectedSwift =
+      """
+      enum namespaceA {
+          enum ServiceA {
+              enum Methods {}
+              static let methods: [MethodDescriptor] = []
+              typealias StreamingServiceProtocol = namespaceA_ServiceAServiceStreamingProtocol
+              typealias ServiceProtocol = namespaceA_ServiceAServiceProtocol
+              typealias ClientProtocol = namespaceA_ServiceAClientProtocol
```
glbrntt, line 89: We can add a `typealias` for the `Client` as well

glbrntt (changes requested):

Tests/GRPCCodeGenTests/Internal/Translator/SnippetBasedTranslatorTests.swift
stefanadranca: for the struct?

glbrntt: yes

glbrntt (approved):

PR #1752, opened 2024-01-02
[CodeGenLib] Translating dependencies into StructuredSwiftRepresentation

stefanadranca:
Motivation:

Dependency representations within CodeGeneratorRequest (Dependency) and StructuredSwiftRepresentation (ImportDescription) don't match, as each contains fields that the other one doesn't have. The CodeGenerationRequest representation of a dependency doesn't contain the preconcurrency requirement and spi, while the StructuredSwiftRepresentation doesn't allow importing items (func, class etc.). We need to add the missing fields to both representations in  order to translate the imports, without losing information.

Modifications:

- Added the preconcurrency and spi parameters to the CodeGenerationRequest Dependency struct (and the PreconcurrencyRequirement struct that encapsulates the possible requirement types).
- Added the item property to the StructuredSwiftRepresentation ImportDescription struct (and the Item struct backed up by an enum representing the possible cases).
- Created the function that translates a CodeGenerationRequest.Dependency into a StructuredSwiftRepresentation.ImportDescription.
- Added the item rendering in the TextBasedRenderer.
- Added tests for the TextBasedRenderer and the IDLToStructuredSwiftRepresentationTranslator that check the imports.

Result:

Imports can now be translated and added in the generated code.


Sources/GRPCCodeGen/Internal/Translator/IDLToStructuredSwiftTranslator.swift
```diff
@@ -49,6 +49,40 @@ struct IDLToStructuredSwiftTranslator: Translator {
...
 
+extension IDLToStructuredSwiftTranslator {
+  private func translateImport(
+    dependency: CodeGenerationRequest.Dependency
+  ) throws -> ImportDescription {
+    var importDescription = ImportDescription(moduleName: dependency.module)
+    if let item = dependency.item {
+      if let matchedKind = ImportDescription.Kind.allCases.first(where: {
+        "\($0)" == item.kind.value.description
+      }) {
```
gjcairo, line 60: `ImportDescription.Kind` has a raw value of `String`. This means that it provides an `init?(rawValue:)` initialiser. You can take advantage of it to avoid iterating over all cases as you're doing here - if the provided raw value doesn't match an enum case, the initialiser will return nil:
```suggestion
      if let matchedKind = ImportDescription.Kind(rawValue: item.kind.value.description) {
```

Sources/GRPCCodeGen/CodeGenerationRequest.swift
```diff
@@ -109,7 +125,7 @@ public struct CodeGenerationRequest {
       /// Represents the imported item's kind.
       public struct Kind {
         /// Describes the keyword associated with the imported item.
-        internal enum Value {
+        internal enum Value: CustomStringConvertible {
```
gjcairo, line 128: A bit of a nit, but `ImportDescription.Kind`, which basically mirrors this type, has a `String` raw value instead of conforming to `CustomStringConvertible`. Is there a reason why this enum conforms to `CustomStringConvertible` instead of doing the same as `ImportDescription.Kind`?

Sources/GRPCCodeGen/Internal/Translator/IDLToStructuredSwiftTranslator.swift
```diff
@@ -49,6 +49,40 @@ struct IDLToStructuredSwiftTranslator: Translator {
...
+          message: "Invalid kind name for import: \(item.kind.value.description)"
+        )
+      }
+    }
+    if let spi = dependency.spi {
+      importDescription.spi = spi
+    }
+
+    let preconcurrency = dependency.preconcurrency.value
+    switch preconcurrency {
```
gjcairo, line 74: Nit, but as you're not using `preconcurrency` anywhere else, you can switch directly over the value without declaring the let before:
```suggestion
    switch dependency.preconcurrency.value {
```

gjcairo (changes requested):
Looking good, left only a few small comments.

Sources/GRPCCodeGen/CodeGenerationRequest.swift
stefanadranca: Not really, I will use raw values instead in both cases

gjcairo (approved):

Sources/GRPCCodeGen/CodeGenerationRequest.swift
```diff
@@ -88,9 +88,25 @@ public struct CodeGenerationRequest {
...
 
-    public init(item: Item? = nil, module: String) {
+    /// The name of the private interface for an `@_spi` import.
+    ///
+    /// For example, if `spi` was "Secret" and the module name was "Foo" then the import
+    /// would be `@_spi(Secret) import Foo`.
+    var spi: String? = nil
+
+    /// Requirements for the `@preconcurrency` attribute.
+    var preconcurrency: PreconcurrencyRequirement = .notRequired
```
glbrntt, line 98: Presumably these are meant to be `public`?

glbrntt: We should avoid double-defaulting them, prefer defaulting them in the `init`

glbrntt (changes requested):

glbrntt (approved):

PR #1753, opened 2024-01-03
[CodeGenLib] Add validation step for input

stefanadranca:
Add validation step for input in IDLToStructuredSwiftTranslator

Motivation:

The IDLToStructuredSwiftTranslator should discover errors in the CodeGenerationRequest object, before the specialized translators start their work. This way, if the input is not correct no translator will be started.

Modifications:

Moved the validation functions into a IDLToStructuredSwiftTranslator extension and the associated tests in a separate test file.

Result:

The input validation is not done by each individual specialzed translator, but by the main translator, before transformations take place.


Sources/GRPCCodeGen/Internal/Translator/IDLToStructuredSwiftTranslator.swift
```diff
@@ -49,6 +50,87 @@ struct IDLToStructuredSwiftTranslator: Translator {
...
+      if namespaces.contains(service.name) {
+        throw CodeGenError(
+          code: .nonUniqueServiceName,
+          message: """
+            Services with no namespace must not have the same names as the namespaces. \
+            \(service.name) is used as a name for a service with no namespace and a namespace.
+            """
+        )
+      }
+    }
```
gjcairo, line 83: Very minor nit, but you can skip the `, default: []]` subscript and instead make this whole block conditional to it being an empty key:

```suggestion
    if let noNamespaceServices = servicesByNamespace[""] {
        let namespaces = servicesByNamespace.keys
        for service in noNamespaceServices {
          if namespaces.contains(service.name) {
            throw CodeGenError(
              code: .nonUniqueServiceName,
              message: """
                Services with no namespace must not have the same names as the namespaces. \
                \(service.name) is used as a name for a service with no namespace and a namespace.
                """
            )
          }
      }
    }
```

gjcairo (approved):
LGTM, just left a very small comment

PR #1754, opened 2024-01-07
[CodeGenLib] Handle casing for service and method names

stefanadranca:
Motivation:

The method and service names set by the user in the CodeGenerationRequest might not have the right casing for the Swift identifiers that will be used in the generated code (method names and enum names). We need to make sure that the method name used in the descriptors is the one from the input, but that the method names from the generated code start with a lowercase letter. The service names must start with a capital letter. The Swift identifiers use CamelCase so if the input names are in snake case they are transformed into camel case.

Modifications:

- Brought over methods from SwiftProtobuf that change the casing of a String,matching the Swift identifiers format (camel case).
- Brought over tests.
- Added a new property to the CodeGenerationRequest MethodDescriptor and ServiceDescriptor to represent the name of the Swift identifier associated with them.
- Changed the properties used in the translators from the name to the Swift identifier name (syntaxName) where needed.

Result:

Enums representing the services will have identifiers starting with a capital letter and method names will start with a lowercase letter.

### DIFF
--- a/Tests/GRPCCodeGenTests/Internal/Translator/TypealiasTranslatorSnippetBasedTests.swift
+++ b/Tests/GRPCCodeGenTests/Internal/Translator/TypealiasTranslatorSnippetBasedTests.swift
@@ -44,7 +44,7 @@ final class TypealiasTranslatorSnippetBasedTests: XCTestCase {
       enum namespaceA {
           enum ServiceA {
               enum Methods {
-                  enum MethodA {
+                  enum methodA {
                       typealias Input = NamespaceA_ServiceARequest
                       typealias Output = NamespaceA_ServiceAResponse
                       static let descriptor = MethodDescriptor(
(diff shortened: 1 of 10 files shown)


Tests/GRPCCodeGenTests/Internal/Translator/TypealiasTranslatorSnippetBasedTests.swift
```diff
@@ -44,7 +44,7 @@ final class TypealiasTranslatorSnippetBasedTests: XCTestCase {
       enum namespaceA {
```
stefanadranca, line 44: Should this one start with an uppercase letter?

Tests/GRPCCodeGenTests/Internal/Translator/TypealiasTranslatorSnippetBasedTests.swift
```diff
@@ -44,7 +44,7 @@ final class TypealiasTranslatorSnippetBasedTests: XCTestCase {
...
           enum ServiceA {
               enum Methods {
-                  enum MethodA {
+                  enum methodA {
```
stefanadranca, line 47: should this one start with an uppercase letter?

glbrntt (changes requested):
I don't think this is the right fix for the problem: I think we need to make a distinction between the _generated_ service/method names and the _actual_ service/method names in the code generation request.

This then gives the author of the code generator the flexibility to choose how their code generator should work. More concretely, the protobuf code generator would handle the casing and just plug the appropriately cased names into the request.

PR #1756, opened 2024-01-10
Add client and server tracing interceptors

gjcairo:
## Motivation

With v2 of gRPC, we want to provide tracing interceptors for both the client and server sides.

## Modifications

This change adds tracing interceptors implementations for both the client and the server, using `swift-distributed-tracing`.

## Result

We've got tracing interceptors.


Sources/GRPCInterceptors/ClientTracingInterceptor.swift
```diff
@@ -0,0 +1,63 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import GRPCCore
+import Tracing
+
+/// A client interceptor that injects tracing information into the request.
+///
+/// The tracing information will be taken from the current `ServiceContext`, and injected into the request's
```
glbrntt, line 22: Prefer writing in the active voice for docs, e.g. "Tracing information is taken from the current `ServiceContext`"

Sources/GRPCInterceptors/ClientTracingInterceptor.swift
```diff
@@ -0,0 +1,63 @@
...
+/// metadata, to be picked up by the server-side ``ServerTracingInterceptor``.
```
glbrntt, line 23: Total nit, but metadata is a gRPC concept but the `ServerTracingInterceptor` is specific to grpc-swift. I think the latter part of the sentence should be a separate sentence.

Sources/GRPCInterceptors/ClientTracingInterceptor.swift
```diff
@@ -0,0 +1,63 @@
...
+/// For more information, refer to the documentation for `swift-distributed-tracing`.
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+public struct ClientTracingInterceptor: ClientInterceptor {
+
+  /// Create a new instance of a ``ClientTracingInterceptor``.
+  public init() {}
+
+  /// This interceptor will inject as the request's metadata whatever `ServiceContext` key/value pairs
```
glbrntt, line 31: nit: "key/value" should be "key-value"

Tests/GRPCCoreTests/Call/TracingInterceptorTests.swift
```diff
@@ -0,0 +1,77 @@
...
+import XCTest
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+final class ClientTracingInterceptorTests: TracingTestCase {
+  func testClientInterceptor() async throws {
+    var serviceContext = ServiceContext.topLevel
+    let traceIDString = UUID().uuidString
+    serviceContext.traceID = traceIDString
+    try await ServiceContext.withValue(serviceContext) {
+      let tester = ClientRPCExecutorTestHarness(server: .echo)
```
glbrntt, line 29: I don't think we actually need this at all for the tests: we can just construct the interceptor and call it directly with a stream request and have the `next` function validate the metadata is injected and return an empty response or something similar.

Tests/GRPCCoreTests/Call/TracingTestsUtils.swift
```diff
@@ -0,0 +1,62 @@
+/*
```
glbrntt, line 1: Can these tests go in the interceptor testing module?

Tests/GRPCCoreTests/Call/TracingTestsUtils.swift
```diff
@@ -0,0 +1,62 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Tracing
+import XCTest
+
+class TracingTestCase: XCTestCase {
+  override class func setUp() {
+    InstrumentationSystem.bootstrap(TestTracer())
```
glbrntt, line 22: Does this blow up like logging/metrics?

glbrntt (changes requested):

Package.swift
```diff
@@ -131,9 +135,11 @@ extension Target.Dependency {
   )
   static let dequeModule: Self = .product(name: "DequeModule", package: "swift-collections")
   static let atomics: Self = .product(name: "Atomics", package: "swift-atomics")
+  static let tracing: Self = .product(name: "Tracing", package: "swift-distributed-tracing")
 
   static let grpcCore: Self = .target(name: "GRPCCore")
   static let grpcInProcessTransport: Self = .target(name: "GRPCInProcessTransport")
+  static let grpcInterceptors: Self = .target(name: "GRPCInterceptors")
```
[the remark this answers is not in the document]
gjcairo: Hmm well, we'll likely have tracing, logging and metrics interceptors. What other kinds of extras would end up in an extras package? I agree it may be too small a unit, but my only concern with an `extras` module is that it will end up being a bag for random things if it's not well defined. Of course this seems to have worked well (?) in our other repos, so maybe this is a non-issue :)

Package.swift
```diff
@@ -247,10 +262,16 @@ extension Target {
   static let grpcInProcessTransportTests: Target = .testTarget(
     name: "GRPCInProcessTransportTests",
     dependencies: [
-      .grpcCore,
```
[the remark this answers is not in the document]
gjcairo: Oops, no, this was a mistake.

Sources/GRPCInterceptors/ClientTracingInterceptor.swift
```diff
@@ -0,0 +1,70 @@
...
+      ClientResponse.Stream<Output>
+  ) async throws -> ClientResponse.Stream<Output> where Input: Sendable, Output: Sendable {
+    var request = request
+    let tracer = InstrumentationSystem.tracer
+    let serviceContext = ServiceContext.current ?? .topLevel
+    
+    tracer.inject(
+      serviceContext,
+      into: &request,
+      using: ClientRequestInjector()
```
FranzBusch, line 49: Two questions here:
1. Should we create the `ClientRequestInjector` only once in the init of this struct?
2. Is there a case where the user might want to provide their own injector here?

Sources/GRPCInterceptors/ServerTracingInterceptor.swift
```diff
@@ -0,0 +1,70 @@
...
+    next: @Sendable (ServerRequest.Stream<Input>, ServerInterceptorContext) async throws ->
+      ServerResponse.Stream<Output>
+  ) async throws -> ServerResponse.Stream<Output> where Input: Sendable, Output: Sendable {
+    var serviceContext = ServiceContext.topLevel
+    let tracer = InstrumentationSystem.tracer
+      
+    tracer.extract(
+      request,
+      into: &serviceContext,
+      using: ServerRequestExtractor()
```
FranzBusch, line 48: Same questions about this as with the client injector.

Tests/GRPCCoreTests/Call/TracingTestsUtils.swift
gjcairo: You mean if it fails when called more than once? If so, yes, and that's why I had to do it here, so that it's called just once.

Sources/GRPCInterceptors/ClientTracingInterceptor.swift
```diff
@@ -0,0 +1,70 @@
...
+    )
+    
+    return try await tracer.withSpan(context.descriptor.fullyQualifiedMethod, context: serviceContext, ofKind: .client) { span in
+      span.addEvent("Sending request")
+      let response = try await next(request, context)
+      span.addEvent("Received response")
```
FranzBusch, line 55: So this is interesting because we might be streaming here. We should definitely rename this to something like `Received response end` or similar. Next, should we provide a mode of this interceptor that emits events on every single message?

Sources/GRPCInterceptors/ServerTracingInterceptor.swift
```diff
@@ -0,0 +1,70 @@
...
+    )
+    
+    return try await tracer.withSpan(context.descriptor.fullyQualifiedMethod, context: serviceContext, ofKind: .server) { span in
+      span.addEvent("Received request")
+      let response = try await next(request, context)
+      span.addEvent("Sending response")
```
FranzBusch, line 54: Similar comment as above on the client.

Sources/GRPCInterceptors/ClientTracingInterceptor.swift
```diff
@@ -0,0 +1,70 @@
...
+    let tracer = InstrumentationSystem.tracer
+    let serviceContext = ServiceContext.current ?? .topLevel
+    
+    tracer.inject(
+      serviceContext,
+      into: &request,
+      using: ClientRequestInjector()
+    )
+    
+    return try await tracer.withSpan(context.descriptor.fullyQualifiedMethod, context: serviceContext, ofKind: .client) { span in
```
FranzBusch, line 52: Are there other attributes that we might want to add to the span from the request?

Sources/GRPCInterceptors/ClientTracingInterceptor.swift
gjcairo: 1. 👍 - had to change the `Carrier` type to be `Metadata` to avoid generics hell.
2. I don't think so, really - these injectors are provided as part of the basic interceptor we present here. If users need more control I think they could just create their own interceptors.

Sources/GRPCInterceptors/ServerTracingInterceptor.swift
gjcairo: See above

Tests/GRPCCoreTests/Call/TracingTestsUtils.swift
```diff
@@ -0,0 +1,143 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Tracing
+import XCTest
+
+class TracingTestCase: XCTestCase {
```
FranzBusch, line 20:
```suggestion
final class TracingTestCase: XCTestCase {
```

FranzBusch (approved):
LGTM! One missing `final` nit

Sources/GRPCInterceptors/ClientTracingInterceptor.swift
gjcairo: Can't think of anything else that would be universal to all requests

Sources/GRPCInterceptors/ClientTracingInterceptor.swift
```diff
@@ -0,0 +1,98 @@
...
+ * limitations under the License.
+ */
+
+import GRPCCore
+import Tracing
+
+/// A client interceptor that injects tracing information into the request.
+///
+/// The tracing information is taken from the current `ServiceContext`, and injected into the request's
+/// metadata. I twill then be picked up by the server-side ``ServerTracingInterceptor``.
```
glbrntt, line 23:
```suggestion
/// metadata. It will then be picked up by the server-side ``ServerTracingInterceptor``.
```

Sources/GRPCInterceptors/ClientTracingInterceptor.swift
```diff
@@ -0,0 +1,98 @@
...
+///
+/// For more information, refer to the documentation for `swift-distributed-tracing`.
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+public struct ClientTracingInterceptor: ClientInterceptor {
+  private let injector: ClientRequestInjector
+  private let emitEventOnEachWrite: Bool
+
+  /// Create a new instance of a ``ClientTracingInterceptor``.
+  public init(emitEventOnEachWrite: Bool = false) {
```
glbrntt, line 32: Can you document this parameter?

Sources/GRPCInterceptors/ClientTracingInterceptor.swift
```diff
@@ -0,0 +1,98 @@
...
+    ) { span in
+      span.addEvent("Request started")
+
+      if self.emitEventOnEachWrite {
+        let wrappedProducer = request.producer
+        request.producer = { writer in
+          let eventEmittingWriter = HookedWriter(
+            wrapping: writer,
+            beforeEachWrite: {
+              span.addEvent("Sending request part")
```
glbrntt, line 71: I think we should add events for receiving response parts as well if we're doing them for request parts.

Sources/GRPCInterceptors/HookedWriter.swift
```diff
@@ -0,0 +1,40 @@
...
+    wrapping other: some RPCWriterProtocol<Element>,
+    beforeEachWrite: @Sendable @escaping () -> Void,
+    afterEachWrite: @Sendable @escaping () -> Void
+  ) {
+    self.writer = other
+    self.beforeEachWrite = beforeEachWrite
+    self.afterEachWrite = afterEachWrite
+  }
+
+  public func write(contentsOf elements: some Sequence<Element>) async throws {
```
glbrntt, line 35:
```suggestion
  init(
    wrapping other: some RPCWriterProtocol<Element>,
    beforeEachWrite: @Sendable @escaping () -> Void,
    afterEachWrite: @Sendable @escaping () -> Void
  ) {
    self.writer = other
    self.beforeEachWrite = beforeEachWrite
    self.afterEachWrite = afterEachWrite
  }

  func write(contentsOf elements: some Sequence<Element>) async throws {
```

Sources/GRPCInterceptors/ServerTracingInterceptor.swift
```diff
@@ -0,0 +1,117 @@
...
+///
+/// The extracted tracing information is made available to user code via the current `ServiceContext`.
+/// For more information, refer to the documentation for `swift-distributed-tracing`.
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+public struct ServerTracingInterceptor: ServerInterceptor {
+  private let extractor: ServerRequestExtractor
+  private let emitEventOnEachWrite: Bool
+
+  /// Create a new instance of a ``ServerTracingInterceptor``.
+  public init(emitEventOnEachWrite: Bool = false) {
```
glbrntt, line 30: Can you document this parameter please?

Sources/GRPCInterceptors/ServerTracingInterceptor.swift
```diff
@@ -0,0 +1,117 @@
...
+                wrapping: writer,
+                beforeEachWrite: {
+                  span.addEvent("Sending response part")
+                },
+                afterEachWrite: {
+                  span.addEvent("Sent response part")
+                }
+              )
+
+              let wrappedResult = try await wrappedProducer(
```
glbrntt, line 82: We should catch this error, add an event to the span, and then rethrow it

Sources/GRPCInterceptors/ServerTracingInterceptor.swift
```diff
@@ -0,0 +1,117 @@
...
+                RPCWriter(wrapping: eventEmittingWriter)
+              )
+              span.addEvent("Sent response end")
+              return wrappedResult
+            }
+          } else {
+            success.producer = { writer in
+              let wrappedResult = try await wrappedProducer(writer)
```
glbrntt, line 90: Same here re: the error

Tests/GRPCCoreTests/Call/Client/Internal/ClientRPCExecutorTestSupport/ClientRPCExecutorTestHarness.swift
```diff
@@ -69,11 +69,13 @@ struct ClientRPCExecutorTestHarness {
   func unary(
     request: ClientRequest.Single<[UInt8]>,
     configuration: MethodConfiguration? = nil,
+    interceptors: [any ClientInterceptor] = [],
```
glbrntt, line 72: can you back out these changes as they're no longer needed?

Tests/GRPCInterceptorsTests/TracingTestsUtilities.swift
```diff
@@ -0,0 +1,177 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import GRPCCore
+import Tracing
+
+class TestTracer: Tracer {
```
glbrntt, line 20: nit final
```suggestion
final class TestTracer: Tracer {
```

Tests/GRPCInterceptorsTests/TracingInterceptorTests.swift
```diff
@@ -0,0 +1,280 @@
...
+      XCTAssertEqual(message, ["response"])
+
+      let tracer = InstrumentationSystem.tracer as! TestTracer
+      XCTAssertEqual(
+        tracer.latestSpanEvents.map { $0.name },
+        [
+          "Request started",
+          "Received response end",
+        ]
+      )
```
glbrntt, line 77: Nice!

Tests/GRPCInterceptorsTests/TracingTestsUtilities.swift
```diff
@@ -0,0 +1,177 @@
...
+  typealias Span = TestSpan
+
+  private var latestTestSpan: TestSpan?
```
glbrntt, line 23: Hmm, isn't this used unsafely across different threads? I.e. in the tests we read out `latestSpanEvents` but in the interceptors we start spans which creates this.

Package.swift
```diff
@@ -274,10 +288,18 @@ extension Target {
...
-      .grpcInProcessTransport,
+      .grpcInProcessTransport
     ]
   )
   
+  static let grpcInterceptorsTests: Target = .testTarget(
+    name: "GRPCInterceptorsTests",
+    dependencies: [
+      .grpcCore,
+      .grpcInterceptors
```
glbrntt, line 299: I think this needs` .tracing` too

glbrntt (changes requested):
This looks in pretty good shape, I mostly left nits but there are a couple of things that I think do need fixing before we can merge this.

Sources/GRPCInterceptors/ClientTracingInterceptor.swift
gjcairo: Fair point, I agree. I've added this - it was not as straightforward as I would've liked :D

Tests/GRPCInterceptorsTests/TracingTestsUtilities.swift
gjcairo: Yes, but my understanding is all tests inside an `XCTestCase` are run sequentially (i.e. no tests inside the same `XCTestCase` will be run in parallel), so this should be safe, as we only start a span once per test. I know this may be a bit of an abuse, but I thought it made it unnecessary to make it thread safe by using a lock.

gjcairo: I'm not sure it's circumstantial - the order in which they're run [is explained in the docs](https://developer.apple.com/documentation/xctest/xctestcase/). However I'm okay with making this thread-safe, just pushed the changes.

glbrntt: I meant in the sense that we've lined up a set of conditions in which it is safe but they could easily change, e.g.:
- we could run multiple RPCs concurrently in one test
- XCTest could change how it runs tests
- we could migrate to swift-testing which may have different guarantees

glbrntt (approved):
Looks good! Left one small question.

Sources/GRPCInterceptors/OnFinishAsyncSequence.swift
```diff
@@ -0,0 +1,57 @@
...
+
+    mutating func next() async throws -> Element? {
+      let elem = try await self.iterator.next()
+
+      if elem == nil {
+        self.onFinish?()
+        self.onFinish = nil
+      }
+
+      return elem as? Element
```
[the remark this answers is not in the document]
gjcairo: Because the wrapped iterator is `any AsyncIteratorProtocol`, so calling `next()` on the wrapped iterator will return `Any?`.

PR #1757, opened 2024-01-11
Create new HTTP2 transport modules

gjcairo:
## Motivation

We will write an HTTP2 transport for gRPC v2. We need to create some new modules to organise our code.

## Modifications

Created new targets and test targets:
- GRPCHTTP2Core[Tests]
- GRPCHTTP2TransportNIOPosix[Tests]
- GRPCHTTP2TransportNIOTransportServices[Tests]

## Result

Modules in place so we can start writing the H2 transport.


stefanadranca (approved):

Sources/GRPCHTTP2TransportNIOTransportServices/README.md
```diff
@@ -0,0 +1 @@
+# gRPC HTTP2 Transport NIO Transport Services
```
glbrntt, line 1: We don't have a corresponding module for this one (or its tests) in `Package.swift`.

glbrntt (changes requested):

Package.swift
```diff
@@ -251,6 +277,27 @@ extension Target {
...
+  
+  static let grpcHTTP2TransportNIOPosixTests: Target = .testTarget(
+    name: "GRPCHTTP2TransportNIOPosixTests",
+    dependencies: [
+      .grpcHTTP2TransportNIOPosix
+    ]
+  )
+  
+  static let grpcHTTP2TransportNIOTransportServicesTests: Target = .testTarget(
+    name: "GRPCHTTP2TransportNIOTransportServicesTests",
```
glbrntt, line 296: That's quite the name... 😬 we may want to revisit these in the future.

glbrntt (approved):

PR #1759, opened 2024-01-15
[CodeGenLib]  Differentiate type names from identifying names

stefanadranca:
Motivation:

The names of the namespaces/service/methods used in the type names from the generated code can be different from the names identifying the namespaces/service/methods in the IDL. Methods also have a third name - for the function signatures.

Modifications:

- Created a Name struct that has 3 properties: the base name, the upper case name and the lower case name
- Added the new names properties in the ServiceDescriptor and MethodDescriptor from CodeGenerationRequest
- Modified the existing tests accordingly
- Added new validation methods verifying the generated names (or signature name)
- Added tests for the new checks

Result:

Users will be able to set an identifying name, a generated upper case and a generated lower case name for the namespaces, services and methods.

### DIFF
--- a/Sources/GRPCCodeGen/CodeGenerationRequest.swift
+++ b/Sources/GRPCCodeGen/CodeGenerationRequest.swift
@@ -220,12 +220,16 @@ public struct CodeGenerationRequest {
     /// Documentation from comments above the IDL service description.
     public var documentation: String
 
-    /// Service name in different formats.
+    /// The service name in different formats.
+    ///
+    /// All properties of this object must be unique for each service from within a namespace.
     public var name: Name
 
     /// The service namespace in different formats.
     ///
-    /// For `.proto` files it is the package name.
+    /// All different services from within the same namespace must have
+    /// the same ``Name`` object as this property.
+    /// For `.proto` files the base name of this object is the package name.
     public var namespace: Name
 
     /// A description of each method of a service.
@@ -251,6 +255,9 @@ public struct CodeGenerationRequest {
       public var documentation: String
 
       /// Method name in different formats.
+      ///
+      /// All properties of this object must be unique for each method
+      /// from within a service.
       public var name: Name
 
       /// Identifies if the method is input streaming.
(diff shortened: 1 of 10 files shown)


Sources/GRPCCodeGen/Internal/Translator/IDLToStructuredSwiftTranslator.swift
```diff
@@ -109,21 +117,23 @@ extension IDLToStructuredSwiftTranslator {
...
+              Services with no namespace must not have the same generated names as the namespaces. \
+              \(service.generatedName) is used as a generated name for a service with no namespace and a namespace.
               """
           )
         }
       }
     }
 
-    // Check that service names are unique within each namespace.
+    // Check that service names and service generated names are unique within each namespace.
```
glbrntt, line 132: We only call this function with the descriptors grouped by the generated namespace, why do we still need to check the non-generated service names?

I think the only modifications you need to make to your existing checks should be to use the generated names instead of the current ones _and_ to add one more check which ensures that all descriptors are unique.

glbrntt (changes requested):

Sources/GRPCCodeGen/Internal/Translator/IDLToStructuredSwiftTranslator.swift
stefanadranca: So, for the methods, I think it is important to check that the upper case names are unique within a service (because they are enum names inside the service enum), but also that the lowerCase names are unique within a service, as they will be used for function declaration in the same protocols. So, in this case I think it's useful to have 3 checks (the third one is for the descriptor).  But for the services I will only check the upperCase names and the descriptors (that they are unique).

Sources/GRPCCodeGen/CodeGenerationRequest.swift
```diff
@@ -282,4 +282,39 @@ public struct CodeGenerationRequest {
...
+    /// `<service_namespace_baseName>.<service_baseName>.<method_baseName>`.
+    public let base: String
+
+    /// The `generatedUpperCase` name is used as part of generated type names, which begin with a capital letter
+    /// and are (partially) using CamelCase. It is using UpperCamelCase in order to follow the Swift naming conventions.
+    ///
+    /// It is used in the generated code as an enum name and as part of protocol names.
+    /// For example, in the generated server code the name of the service protocol
+    /// follows this pattern:
+    /// `<service_namespace_generatedUpperCaseName>_<service_generatedUpperCaseName>ServiceProtocol`.
```
glbrntt, line 306: We don't need to be so specific, we're setting ourselves up to have outdated docs (because we may change the pattern and forget about the docs).

In this case the relevant info is that this name is used in generated code and is expected to be the upper-camel-case version of the base name.

You can then provide an abstract example, e.g. if `base` is "echo" then `generatedUpperCase` could be "Echo".

Sources/GRPCCodeGen/CodeGenerationRequest.swift
```diff
@@ -282,4 +282,39 @@ public struct CodeGenerationRequest {
...
+  ///
+  /// The base name, the generatedUpperCase (and the generatedLowerCase) must be unique for methods
+  /// from within the same service and for services within the same namespace.
+  public struct Name: Hashable {
+    /// The base name is the name used for the namespace/service/method in the IDL file, so it should follow
+    /// the specific casing of the IDL.
+    ///
+    /// The base name is also used in the descriptors that identify a specific method or service :
+    /// `<service_namespace_baseName>.<service_baseName>.<method_baseName>`.
+    public let base: String
```
glbrntt, line 298: Properties on structs should be `var` unless you have a good reason for them not to be.

Making them a `let` only makes using the type harder:

```swift
var foo = Foo(bar: "bar")
foo.bar = "baz"  // Not possible if 'bar' is a 'let'
foo = Foo(bar: "baz")  // Possible, but more annoying to do than the above, especially if there are multiple properties
```

Sources/GRPCCodeGen/Internal/Translator/IDLToStructuredSwiftTranslator.swift
```diff
@@ -112,87 +115,165 @@ extension IDLToStructuredSwiftTranslator {
...
   }
 
-  // Verify method names are unique for the service.
+  // Verify method names are unique within a service.
   private func checkMethodNamesAreUnique(
     in service: CodeGenerationRequest.ServiceDescriptor
   ) throws {
-    let methodNames = service.methods.map { $0.name }
-    var seenNames = Set<String>()
+    // Check that the method descriptors are unique, by checking that the base names
```
glbrntt, line 170: This function does the same thing three times but with different information. Can you refactor this to avoid the duplication?

Sources/GRPCCodeGen/Internal/Translator/IDLToStructuredSwiftTranslator.swift
```diff
@@ -112,87 +115,165 @@ extension IDLToStructuredSwiftTranslator {
...
+    if let duplicate = descriptors.hasDuplicates() {
+      throw CodeGenError(
+        code: .nonUniqueServiceName,
+        message: """
+          Services must have unique descriptors. \
+          \(duplicate) is the descriptor of at least two different services.
+          """
+      )
     }
   }
```
glbrntt, line 240: There's no need to create an intermediate array here, we can just do this directly which is a bit simpler:

```swift
var descriptors: Set<String> = []
for service in services {
  let name = service.namespace.base.isEmpty ? service.name.base : "\(service.namespace.base).\(service.name.base)"
  let (inserted, _) = descriptors.insert(name)
  if !inserted { 
    throw ...
  }
}
```

glbrntt (changes requested):

glbrntt (approved):

PR #1760, opened 2024-01-16
Add server connection state machine

glbrntt:
Motivation:

The server pipeline needs a channel handler which manages the lifecycle of TCP connections. This includes: policing keepalive pings sent by the client, managing graceful shutdown, shutting down idle connections, and shutting down connections which have lived for too long. Much of this logic can be abstracted into a state machine. This change add that state machine.

Modifications:

- Add a server connection handler state machine which tracks the graceful shutdown state and polices keepalive pings sent by the client.
- Replace a few READMEs with empty Swift files to suppress a few build warnings.

Result:

State machine for managing server connections.

### DIFF
--- a/Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionHandler+StateMachine.swift
+++ b/Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionHandler+StateMachine.swift
@@ -1,0 +1,344 @@
+/*
+ * Copyright 2024, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import NIOCore
+import NIOHTTP2
+
+extension ServerConnectionHandler {
+  /// Tracks the state of TCP connections at the server.
+  ///
+  /// The state machine manages the state for the graceful shutdown procedure as well as policing
+  /// client-side keep alive.
+  struct StateMachine {
+    /// Current state.
+    private var state: State
+
+    /// Opaque data sent to the client in a PING frame after emitting the first GOAWAY frame
+    /// as part of graceful shutdown.
+    private let goAwayPingData: HTTP2PingData
+
+    /// Create a new state machine.
+    ///
+    /// - Parameters:
+    ///   - allowKeepAliveWithoutCalls: Whether the client is permitted to send keep alive pings
+    ///       when there are no active calls.
+    ///   - minPingReceiveIntervalWithoutCalls: The minimum time interval required between keep
+    ///       alive pings when there are no active calls.
+    ///   - goAwayPingData: Opaque data sent to the client in a PING frame when the server
+    ///       initiates graceful shutdown.
+    init(
+      allowKeepAliveWithoutCalls: Bool,
+      minPingReceiveIntervalWithoutCalls: TimeAmount,
+      goAwayPingData: HTTP2PingData = HTTP2PingData(withInteger: .random(in: .min ... .max))
+    ) {
+      let keepAlive = KeepAlive(
+        allowWithoutCalls: allowKeepAliveWithoutCalls,
+        minPingReceiveIntervalWithoutCalls: minPingReceiveIntervalWithoutCalls
+      )
+
+      self.state = .active(State.Active(keepAlive: keepAlive))
+      self.goAwayPingData = goAwayPingData
+    }
+
+    /// Record that the stream with the given ID has been opened.
+    mutating func streamOpened(_ id: HTTP2StreamID) {
+      switch self.state {
+      case .active(var state):
+        state.lastStreamID = id
+        let (inserted, _) = state.openStreams.insert(id)
+        assert(inserted, "Can't open stream \(Int(id)), it's already open")
+        self.state = .active(state)
+
+      case .closing(var state):
+        state.lastStreamID = id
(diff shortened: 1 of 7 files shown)


Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionHandler+StateMachine.swift
```diff
@@ -0,0 +1,349 @@
...
+          hasOpenStreams: !state.openStreams.isEmpty
+        )
+
+        if tooManyPings {
+          onPing = .enhanceYourCalmThenClose(state.lastStreamID)
+          self.state = .closed
+        } else {
+          onPing = .sendAck
+          self.state = .closing(state)
+        }
```
stefanadranca, line 156: These two cases are pretty similar, except for the state that is set in the else branch. Could the duplication be avoided or do we want to see clearly the steps for each case?

stefanadranca (approved):

Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionHandler+StateMachine.swift
```diff
@@ -0,0 +1,349 @@
...
+        onPing = .none
+      }
+
+      return onPing
+    }
+
+    enum OnPingAck: Equatable {
+      /// Send a GOAWAY frame with no error and the given last stream ID, optionally closing the
+      /// connection immediately afterwards.
+      case sendGoAway(HTTP2StreamID, Bool)
```
gjcairo, line 168: Can we name the properties here? Mostly because of the boolean, I think it makes the code easier to follow when using it.

Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionHandler+StateMachine.swift
```diff
@@ -0,0 +1,349 @@
...
+
+    /// The maximum number of "bad" pings sent by the client the server tolerates before closing
+    /// the connection.
+    private let maxPingStrikes: Int
+
+    /// The number of "bad" pings sent by the client. This can be reset when the server sends
+    /// DATA or HEADERS frames.
+    ///
+    /// Ping strikes account for pings being occasionally being used for purposes other than keep
+    /// alive (a low number of strikes is therefore expected an okay).
```
gjcairo, line 251: Nit:
```suggestion
    /// alive (a low number of strikes is therefore expected and okay).
```

Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionHandler+StateMachine.swift
```diff
@@ -0,0 +1,349 @@
...
+        interval = .hours(2)
+      }
+
+      // If there's no last ping time then the first is acceptable.
+      let isAcceptablePing = self.lastValidPingTime.map { $0 + interval <= time } ?? true
+      let tooManyPings: Bool
+
+      if isAcceptablePing {
+        self.lastValidPingTime = time
+        tooManyPings = false
```
gjcairo, line 296: I assume the bad pings don't have to be consecutive, right? Otherwise, we should reset the ping strikes to 0.

Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionHandler+StateMachine.swift
```diff
@@ -0,0 +1,349 @@
...
+    init(allowWithoutCalls: Bool, minPingReceiveIntervalWithoutCalls: TimeAmount) {
+      self.allowWithoutCalls = allowWithoutCalls
+      self.minPingReceiveIntervalWithoutCalls = minPingReceiveIntervalWithoutCalls
+      self.maxPingStrikes = 2
+      self.pingStrikes = 0
+      self.lastValidPingTime = nil
+    }
+
+    /// Reset ping strikes and the time of the last valid ping.
+    mutating func reset() {
```
gjcairo, line 271: We're never calling this, which makes me think we haven't yet implemented the ping strike reset logic when receiving `DATA`/`HEADERS` frames - correct?

Tests/GRPCHTTP2CoreTests/Server/Connection/ServerConnectionHandler+StateMachineTests.swift
```diff
@@ -0,0 +1,202 @@
...
+    state.streamOpened(1)
+    state.streamOpened(2)
+    XCTAssertEqual(state.streamClosed(2), .none)
+  }
+
+  func testOpenAndCloseStreamWhenClosed() {
+    var state = self.makeStateMachine()
+    state.markClosed()
+    XCTAssertEqual(state.streamClosed(1), .none)
+  }
```
gjcairo, line 53: Looks like you forgot to open the stream before closing it.

Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionHandler+StateMachine.swift
glbrntt: It's somewhat unavoidable here as the associated state types are different. IMO deduplicating here would also result in less readable code.

gjcairo (changes requested):

Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionHandler+StateMachine.swift
glbrntt: Correct. This will have to be done when the connection handler is done as it requires input from the stream channels as well (which is a bit annoying...). There's effectively nothing to do in the state machine beyond calling reset though.

Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionHandler+StateMachine.swift
glbrntt: Yes, that's my understanding from the various gRFCs.

Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionHandler+StateMachine.swift
glbrntt: I'll add a test that the reset actually works though as that's missing right now.

gjcairo (approved):

PR #1762, opened 2024-01-18
Add a handler for managing connections on the server

glbrntt:
Motivation:

Servers must manage connections created by clients. Part of this is gracefully closing connections (by sending GOAWAY frames and ratcheting down the last stream ID) in response to various conditions: the client sending too many pings, the connection being idle too long, the connection existing for longer than some configured limit, etc.

A previous change added a state machine which handles much of this behaviour. This change adds a channel handler which builds on top of that state machine.

Modifications:

- Add a channel handler for managing connections on the server.

Result:

We have a handler in place which can manage connections on the server.


Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionManagementHandler.swift
```diff
@@ -0,0 +1,473 @@
...
+
+    default:
+      ()
+    }
+
+    context.fireUserInboundEventTriggered(event)
+  }
+
+  func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+    self.inReadLoop = true
```
FranzBusch, line 319: Adding this comment here but it applies to more of this PR. Why do we have some state management inside a state machine but that some state managed as vars of this handler. It isn't clear to me where that separation goes and it feels to me like we should have a single state machine for this handler that handles all the state.

glbrntt: Layering, mostly. Most of the state within the handler that isn't the state machine are triggers which prod the state machine. Storing timers etc., back in the state machine complicates the state machine more than it helps. As for the flushing etc, it's not related to connection management, it's its own separate set of state.

Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionManagementHandler.swift
```diff
@@ -0,0 +1,473 @@
...
+/// Some of the behaviours are described in:
+/// - [gRFC A8](https://github.com/grpc/proposal/blob/master/A8-client-side-keepalive.md), and
+/// - [gRFC A9](https://github.com/grpc/proposal/blob/master/A9-server-side-conn-mgt.md).
+final class ServerConnectionManagementHandler: ChannelDuplexHandler {
+  typealias InboundIn = HTTP2Frame
+  typealias InboundOut = HTTP2Frame
+  typealias OutboundIn = HTTP2Frame
+  typealias OutboundOut = HTTP2Frame
+
+  /// The `EventLoop` of the `Channel` this handler exists it.
```
stefanadranca, line 46: nit:
```suggestion
  /// The `EventLoop` of the `Channel` this handler exists in.
```

stefanadranca (approved):
Looks good to me!

Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionManagementHandler.swift
```diff
@@ -0,0 +1,473 @@
...
+      ()  // Only interested in PING frames, ignore the rest.
+    }
+
+    context.fireChannelRead(data)
+  }
+
+  func channelReadComplete(context: ChannelHandlerContext) {
+    while self.flushPending {
+      self.flushPending = false
+      context.flush()
```
gjcairo, line 344: This is more of a NIO question but want to double check: `context.flush()` will start by calling `self.flush()` in this case, right? Which would in turn call `maybeFlush(context:)`, which can reset `flushPending` to be `true` again. 
I just want to make sure I understand why we're checking for `flushPending` in a `while` loop.

gjcairo (approved):
LGTM, just left a nit and a question.

Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionManagementHandler.swift
glbrntt: No, `context.flush()` will call `flush(context:)` on the _next_ channel handler. 

We're calling it in a loop here in case a flush re-entrantly triggers another write which could set `flushPending` back to true.

PR #1765, opened 2024-01-20
Fix no output when Client=False,TestClient=True

severnt:
Up until 1.19.0, passing `Client=False` and `TestClient=True` would result in a file being generated. After commit https://github.com/grpc/grpc-swift/commit/09c46b0a6582652ebddb9bd6967ca2eb8fc67073 - this no longer happens.


glbrntt (approved):
Thank you!

Sources/protoc-gen-grpc-swift/main.swift
```diff
@@ -156,7 +156,7 @@ func main(args: [String]) throws {
         binaryFile.content = serializedFileDescriptorProto
         response.file.append(binaryFile)
       }
-      if !fileDescriptor.services.isEmpty && (options.generateClient || options.generateServer) {
+      if !fileDescriptor.services.isEmpty && (options.generateClient || options.generateServer || options.generateTestClient) {
```
glbrntt, line 159: Oh sorry, the formatter doesn't like this, could you run `./scripts/format.sh -f`?

glbrntt (changes requested):

Sources/protoc-gen-grpc-swift/main.swift
severnt: Looks like the formatter script isn't happy with `-f`.. I'll put up a separate PR for that.

glbrntt: Ah, right, `-f` isn't required but the script tells you to pass it 🙈

glbrntt (approved):

PR #1766, opened 2024-01-22
Add zlib compressor and decompressor

glbrntt:
Motivation:

v2 should support compression.

Modifications:

- Add a Zlib compressor and decompressor supporting deflate and gzip.
- This is loosely based on the code we had in v1 but refactored to fit slightly different requirements.

Result:

Can compress and decompress messages using deflate/gzip

### DIFF
--- a/Sources/GRPCHTTP2Core/Compression/Zlib.swift
+++ b/Sources/GRPCHTTP2Core/Compression/Zlib.swift
@@ -38,8 +38,7 @@ extension Zlib {
   /// Creates a new compressor for the given compression format.
   ///
   /// This compressor is only suitable for compressing whole messages at a time. Callers
-  /// must ``initialize()`` the compressor before using it and must ``reset()`` it between
-  /// subsequent calls to ``compress(_:into:)``.
+  /// must ``initialize()`` the compressor before using it.
   struct Compressor {
     private var stream: z_stream
     private let method: Method
@@ -57,6 +56,12 @@ extension Zlib {
       self.isInitialized = true
     }
 
+    static func initialized(_ method: Method) -> Self {
+      var compressor = Compressor(method: method)
+      compressor.initialize()
+      return compressor
+    }
+
     /// Compresses the data in `input` into the `output` buffer.
     ///
     /// - Parameter input: The complete data to be compressed.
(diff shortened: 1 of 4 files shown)


Sources/GRPCHTTP2Core/Compression/Zlib.swift
```diff
@@ -0,0 +1,423 @@
...
+    }
+  }
+}
+
+extension Zlib {
+  /// Creates a new compressor for the given compression format.
+  ///
+  /// This compressor is only suitable for compressing whole messages at a time. Callers
+  /// must ``initialize()`` the compressor before using it and must ``reset()`` it between
+  /// subsequent calls to ``compress(_:into:)``.
```
gjcairo, line 42: Is there a reason why we couldn't simplify this API by calling `stream/deflateInit` in the initialiser, and doing the logic of `reset()` before calling `deflate` in `compress(_:into:)`? It would make it impossible to make mistakes (by forgetting to call these methods) while using the compressor.
Same question about the `Decompressor` below.

glbrntt: Unfortunately not, initialising in the `init` results in a `Z_STREAM_ERROR` when you call compress (same goes for the decompress). The same is true if you use a static function to create and then initialize and return the initialized compressor/decompressor. I don't understand _why_ that's the case though. In v1 we _did_ do this in `init` but the compressor/decompressor were both classes but I'd rather save the allocation.

W.r.t. `reset()`, that seems like a reasonable idea. I'm not sure what the performance implications are but the docs say it does not free or reallocate internal state so _should_ be okay to do.

gjcairo: Ah, that's unfortunate about the `init`.

W.r.t `reset()` - if the performance was to be affected by running it always in the `defer` block, we could instead add another flag (`isDirty` or something like that, signalling that we've already called `(de)compress`) and if that's set at the beginning of the `reset()` call, then we reset just before doing the rest of the logic. 
That way we'd be saving the cost of _one_ `reset()` (the one that would otherwise happen after the last (de)compression) - but well, I don't know if we care enough about this for now :D

gjcairo (approved):

PR #1767, opened 2024-01-22
Fix format.sh -f printing usage

severnt:
* Passing `-f` now continues to format.
* Prints usage if both or neither of `-f` and `-l` are passed.

### DIFF
--- a/scripts/format.sh
+++ b/scripts/format.sh
@@ -29,9 +29,13 @@ function usage() {
   echo >&2 "  -l   Lint check without formatting the source code"
 }
 
+format=false
 lint=false
-while getopts ":lh" opt; do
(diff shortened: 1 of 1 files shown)


scripts/format.sh
```diff
@@ -29,9 +29,13 @@ function usage() {
   echo >&2 "  -l   Lint check without formatting the source code"
 }
 
+format=false
```
glbrntt, line 32: I think we should default this to true. Many folks (myself included) are in the habit of just running `./scripts/format.sh`. `-f` becomes a bit of a no-op (which is fine)

glbrntt (changes requested):

glbrntt (approved):
Thank you!

PR #1768, opened 2024-01-22
Add `GRPCMessageFramer`

gjcairo:
## Motivation
We need a helper to frame gRPC messages (i.e. prepend the data with compression and message length information) for the HTTP2 transport implementation we're carrying out.
The changes were inspired by the existing `CoalescingLengthPrefixedMessageWriter` in GRPC v1, but slightly simplified to always coalesce the bytes - this is because we'll deal with `UInt8` arrays instead of `ByteBuffers` internally, and we'll need to pay for the copy into a `ByteBuffer` anyways.

## Modifications
- Added a `GRPCFramer`. This first draft doesn't include compression logic: that's being added in https://github.com/grpc/grpc-swift/pull/1771.
- Copied over the `OneOrManyQueue` from `GRPC`.

## Result
We've got a usable gRPC framer now.


Sources/GRPCHTTP2Core/Server/Connection/GRPCFramer.swift
```diff
@@ -0,0 +1,86 @@
...
+    if self.pendingMessages.isEmpty {
+      // Nothing pending: exit early.
+      return nil
+    }
+
+    var requiredCapacity = 0
+    for message in self.pendingMessages {
+      // TODO: Maybe we should add some break condition here, e.g. a max buffer size
+      // or max number of messages to include in the same buffer, but I'm unsure what
+      // this number should be.
```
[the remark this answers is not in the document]
gjcairo: Do we know, from previous experience, a value that works well as a threshold here?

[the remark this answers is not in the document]
glbrntt: The default flow control window size is 64kB (I think). The default http/2 configuration is to have 100 concurrent streams max, if each is full then we arrive at ~6MB per connection which seems reasonable.

Sources/GRPCHTTP2Core/Server/Connection/GRPCFramer.swift
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+import NIOCore
+
+/// A ``GRPCFramer`` helps with the framing of gRPC data frames:
+/// - It prepends data with the required metadata (compression flag and message length).
+/// - It compresses messages using the specified compression algorithm (if configured).
+/// - It coalesces multiple messages (appended into the `Framer` by calling ``append(_:compress:)``)
+/// into a single `ByteBuffer`.
+struct GRPCFramer {
```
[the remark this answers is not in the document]
gjcairo: Maybe `GRPCMessageFramer` (notice the final `r`)? I'm not sure I'd call a buffer of multiple concatenated frames a `GRPCMessageFrame`.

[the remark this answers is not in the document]
glbrntt: Oh that was a typo, I meant `GRPCMessageFramer` 😅

glbrntt (changes requested):

Sources/GRPCHTTP2Core/GRPCMessageFramer.swift
```diff
@@ -0,0 +1,104 @@
...
+struct GRPCMessageFramer {
+  /// Length of the gRPC message header (1 compression byte, 4 bytes for the length).
+  static let metadataLength = 5
+
+  /// Maximum size the `writeBuffer` can be when concatenating multiple frames.
+  /// This limit will not be considered if only a single message/frame is written into the buffer, meaning
+  /// frames with messages over 64KB can still be written.
+  /// - Note: This is expressed as the power of 2 closer to 6KB (i.e., 6KiB) because `ByteBuffer`
+  /// reserves capacity in powers of 2. This way, we can take advantage of the whole buffer.
+  static let maximumWriteBufferLength = 16_384
```
glbrntt, line 33: I'm a bit confused by the numbers here, the max length is 16 KiB but the docs talk about 64 KiB and 6KiB, can you clear this up?

Sources/GRPCHTTP2Core/GRPCMessageFramer.swift
```diff
@@ -0,0 +1,104 @@
...
+  /// - Throws: If an error is encountered, such as a compression failure, an error will be thrown.
+  mutating func next() throws -> ByteBuffer? {
+    if self.pendingMessages.isEmpty {
+      // Nothing pending: exit early.
+      return nil
+    }
+
+    self.writeBuffer.clear()
+    defer {
+      if self.writeBuffer.capacity > Self.maximumWriteBufferLength {
```
glbrntt, line 67: Can you add a comment here?

Sources/GRPCHTTP2Core/GRPCMessageFramer.swift
```diff
@@ -0,0 +1,104 @@
...
+  /// If there are pending messages to be framed, a `ByteBuffer` will be returned with the framed data.
+  /// Data may also be compressed (if configured) and multiple frames may be coalesced into the same `ByteBuffer`.
+  /// - Throws: If an error is encountered, such as a compression failure, an error will be thrown.
+  mutating func next() throws -> ByteBuffer? {
+    if self.pendingMessages.isEmpty {
+      // Nothing pending: exit early.
+      return nil
+    }
+
+    self.writeBuffer.clear()
```
glbrntt, line 65: This doesn't seem necessary as we do it on line 78 with the right capacity

Sources/GRPCHTTP2Core/GRPCMessageFramer.swift
```diff
@@ -0,0 +1,104 @@
...
+    defer {
+      if self.writeBuffer.capacity > Self.maximumWriteBufferLength {
+        self.writeBuffer = ByteBuffer()
+      }
+    }
+
+    var messagesToCoalesce = 0
```
glbrntt, line 72: We don't need to track this anymore since we always do all buffered messages

Sources/GRPCHTTP2Core/GRPCMessageFramer.swift
```diff
@@ -0,0 +1,104 @@
...
+    var requiredCapacity = 0
+    for message in self.pendingMessages {
+      messagesToCoalesce += 1
+      requiredCapacity += message.bytes.count + Self.metadataLength
+    }
+    self.writeBuffer.clear(minimumCapacity: requiredCapacity)
+
+    while messagesToCoalesce > 0 {
```
glbrntt, line 80: This can be simplified now as well: `while let message = self.pendingMessages.pop()`

Tests/GRPCHTTP2CoreTests/GRPCMessageFramerTests.swift
```diff
@@ -0,0 +1,117 @@
...
+    XCTAssertFalse(compressed)
+    XCTAssertEqual(length, 128)
+    XCTAssertEqual(buffer.readSlice(length: Int(length)), ByteBuffer(repeating: 42, count: 128))
+    XCTAssertEqual(buffer.readableBytes, 0)
+
+    // No more bufers.
+    XCTAssertNil(try framer.next())
+  }
+
+  func testLargeWrite() throws {
```
glbrntt, line 38: I don't think this test makes sense anymore

glbrntt (changes requested):

Tests/GRPCHTTP2CoreTests/GRPCMessageFramerTests.swift
gjcairo: I sort of agree - I thought maybe it was worth keeping, to make sure that even when writing beyond our internal buffer's size limit, we'd emit all of the framed messages in a single buffer. But I can get rid of it, I don't have a strong opinion and agree it's not very useful.

glbrntt (approved):
LGTM!

PR #1771, opened 2024-01-24
Add compression logic to `GRPCMessageFramer`

gjcairo:
## Motivation
In https://github.com/grpc/grpc-swift/pull/1768 we added a `GRPCMessageFramer` to frame messages with gRPC's metadata. This type should also take care of compressing the payloads, if so configured.

## Modifications
This PR adds compression logic to the `GRPCMessageFramer`.

## Result
If set up to do so, payloads will now be compressed by the `GRPCMessageFramer`.

### DIFF
--- a/Sources/GRPCHTTP2Core/GRPCMessageFramer.swift
+++ b/Sources/GRPCHTTP2Core/GRPCMessageFramer.swift
@@ -81,16 +81,15 @@ struct GRPCMessageFramer {
   }
 
   private mutating func encode(_ message: [UInt8], compressor: Zlib.Compressor?) throws {
-    if compressor != nil {
+    if let compressor {
       self.writeBuffer.writeInteger(UInt8(1))  // Set compression flag
 
       // Write zeroes as length - we'll write the actual compressed size after compression.
       let lengthIndex = self.writeBuffer.writerIndex
       self.writeBuffer.writeInteger(UInt32(0))
 
-      // This force-unwrap is safe, because we know `compressor` is not `nil`.
-      let writtenBytes = try compressor!.compress(message, into: &self.writeBuffer)
-
+      // Compress and overwrite the payload length field with the right length.
+      let writtenBytes = try compressor.compress(message, into: &self.writeBuffer)
       self.writeBuffer.setInteger(UInt32(writtenBytes), at: lengthIndex)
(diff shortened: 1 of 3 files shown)


Sources/GRPCHTTP2Core/GRPCMessageFramer.swift
```diff
@@ -0,0 +1,119 @@
...
+
+  private var pendingMessages: OneOrManyQueue<[UInt8]>
+
+  private var writeBuffer: ByteBuffer
+
+  private var compressor: Zlib.Compressor?
+
+  /// Create a new ``GRPCMessageFramer``.
+  /// - Parameter compressor: An optional compressor to use when compressing messages.
+  /// - Important: The `compressor` must have been `initialized()`.
```
glbrntt, line 43: Hmm, the caller is also responsible for ending it (I assume) but the semantics here aren't clear because the framer holds on to it. I wonder if we should just pass it in to the `next` to make it clearer that the framer doesn't own the compressor. WDYT? (I appreciate that this requires bringing back the `PendingMessage`...)

gjcairo: Ah well, this doc is outdated, forgot to update it - we don't need to `initialize` the compressor anymore, just end it. But yes, I agree the semantics aren't clear because we hold on to it... I think passing it in `next()` is not a bad idea. I don't think we need the `PendingMessage` wrapper though: we can make it an optional argument on `next()`, and compress if present, not compress if not.

Sources/GRPCHTTP2Core/GRPCMessageFramer.swift
```diff
@@ -72,27 +69,35 @@ struct GRPCMessageFramer {
...
+      try self.encode(message, compressor: compressor)
     }
 
     return self.writeBuffer
   }
 
-  private mutating func encode(_ message: PendingMessage) throws {
-    if message.compress {
+  private mutating func encode(_ message: [UInt8], compressor: Zlib.Compressor?) throws {
+    if compressor != nil {
```
glbrntt, line 84: I think this can just be `if let` now?

glbrntt (approved):
Looks good modulo one nit

PR #1775, opened 2024-01-25
Better waiter errors with NIOTS

glbrntt:
Motivation:

When establishing a connection using NIOTS, Network.framework may enter a 'waiting' state because of a transient error. This is surfaced as a user inbound event.

In many cases the connect call won't resolve until the connect timeout passes (defaults to 20 seconds). Attempts to start a call on this connection during this time will fail with a timeout at the connection pool layer (where the max wait time defaults to 5 seconds) and users will see a 'deadline exceeded' error without any more context. However, we can provide more information by passing up the transient error from Network.framework to the connection pool.

This isn't currently possible as events flow up on state changes, in this case from connecting to transient failure when the connect times out.

Modifications:

- Catch the `WaitingForConnectivity` event and bubble up its error though the idle handler to the connection pool

Result:

Better errors on connect timeouts in some situations

### DIFF
--- a/Tests/GRPCTests/GRPCNetworkFrameworkTests.swift
+++ b/Tests/GRPCTests/GRPCNetworkFrameworkTests.swift
@@ -24,6 +24,7 @@ import NIOCore
 import NIOPosix
 import NIOSSL
 import NIOTransportServices
+import GRPCSampleData
 import Security
 import XCTest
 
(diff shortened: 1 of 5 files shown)


Tests/GRPCTests/GRPCNetworkFrameworkTests.swift
```diff
@@ -208,6 +209,63 @@ final class GRPCNetworkFrameworkTests: GRPCTestCase {
...
+    ) {
+      configure(&$0)
+    }
+
+    let echo = Echo_EchoAsyncClient(channel: client)
+    do {
+      let _ = try await echo.get(.with { $0.text = "ignored" })
+    } catch let error as GRPCConnectionPoolError {
+      XCTAssertEqual(error.code, .deadlineExceeded)
+      print(error)
```
gjcairo, line 238: We probably don't want this `print`

gjcairo (approved):
LGTM, just one nit

Tests/GRPCTests/GRPCNetworkFrameworkTests.swift
glbrntt: We definitely didn't want that `print`

PR #1776, opened 2024-01-25
Add `GRPCMessageDeframer`

gjcairo:
## Motivation
We need a helper to deframe gRPC messages for the HTTP2 transport implementation we're carrying out.
This is the counterpart to https://github.com/grpc/grpc-swift/pull/1768

## Modifications
Added a `GRPCMessageDeframer`. This implementation conforms to NIO's `NIOSingleStepByteToMessageDecoder` to reuse the logic of stitching together multiple frames.

## Result
We've got a usable gRPC deframer now.

### DIFF
--- a/Sources/GRPCHTTP2Core/GRPCMessageDeframer.swift
+++ b/Sources/GRPCHTTP2Core/GRPCMessageDeframer.swift
@@ -54,14 +54,31 @@ struct GRPCMessageDeframer: NIOSingleStepByteToMessageDecoder {
 
     // Store the current reader index in case we don't yet have enough
     // bytes in the buffer to decode a full frame, and need to reset it.
+    // The force-unwraps for the compression flag and message length are safe,
+    // because we've checked just above that we've got at least enough bytes to
+    // read all of the metadata.
     let originalReaderIndex = buffer.readerIndex
     let isMessageCompressed = buffer.readInteger(as: UInt8.self)! == 1
+    let messageLength = buffer.readInteger(as: UInt32.self)!
 
-    guard var message = buffer.readLengthPrefixedSlice(as: UInt32.self) else {
-      // We have moved the reader index, but we don't have enough bytes to
-      // read the full message payload, so reset it to its previous, original
-      // position for now, and return. We'll try decoding again, once more
-      // bytes become available in our buffer.
+    if messageLength > self.maximumPayloadSize {
+      throw RPCError(
+        code: .resourceExhausted,
+        message: """
+          Message has exceeded the configured maximum payload size \
+          (max: \(self.maximumPayloadSize), actual: \(messageLength))
+          """
+      )
+    }
+
+    guard var message = buffer.readSlice(length: Int(messageLength)) else {
+      // `ByteBuffer/readSlice(length:)` returns nil when there are not enough
+      // bytes to read the requested length. This can happen if we don't yet have
+      // enough bytes buffered to read the full message payload.
+      // By reading the metadata though, we have already moved the reader index,
+      // so we must reset it to its previous, original position for now,
+      // and return. We'll try decoding again, once more bytes become available
+      // in our buffer.
       buffer.moveReaderIndex(to: originalReaderIndex)
       return nil
     }
@@ -76,13 +93,6 @@ struct GRPCMessageDeframer: NIOSingleStepByteToMessageDecoder {
       }
       return try decompressor.decompress(&message, limit: self.maximumPayloadSize)
     } else {
-      if message.readableBytes > self.maximumPayloadSize {
-        throw RPCError(
-          code: .resourceExhausted,
-          message:
-            "Message has exceeded the configured maximum payload size (max: \(self.maximumPayloadSize), actual: \(message.readableBytes))"
-        )
-      }
       return Array(buffer: message)
     }
   }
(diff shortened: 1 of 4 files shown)


Sources/GRPCHTTP2Core/GRPCMessageDeframer.swift
```diff
@@ -0,0 +1,90 @@
...
+      // position for now, and return. We'll try decoding again, once more
+      // bytes become available in our buffer.
+      buffer.moveReaderIndex(to: originalReaderIndex)
+      return nil
+    }
+
+    if isMessageCompressed {
+      guard let decompressor = self.decompressor else {
+        // We cannot decompress the payload - discard the message.
+        return nil
```
glbrntt, line 72: We need to throw an error here: the other side has sent a compressed message but we don't have a compressor. Since compression is negotiated it means that either we, or the remote peer, has a bug.

Tests/GRPCHTTP2CoreTests/GRPCMessageDeframerTests.swift
```diff
@@ -0,0 +1,209 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import NIOCore
+import XCTest
+
+@testable import GRPCHTTP2Core
+
+final class GRPCMessageDeframerTests: XCTestCase {
```
glbrntt, line 22: FYI NIOTestUtils has a helper we should use: https://github.com/apple/swift-nio/blob/635b2589494c97e48c62514bc8b37ced762e0a62/Sources/NIOTestUtils/ByteToMessageDecoderVerifier.swift#L54

Tests/GRPCHTTP2CoreTests/Test Utilities/XCTest+Utilities.swift
```diff
@@ -28,3 +28,16 @@ func XCTAssertThrowsError<T, E: Error>(
...
+  _ expression: @autoclosure () throws -> T,
+  _ errorHandler: (RPCError) -> Void
+) {
+  XCTAssertThrowsError(try expression()) { error in
+    guard let error = error as? RPCError else {
+      return XCTFail("Error had unexpected type '\(type(of: error))'")
+    }
+
+    errorHandler(error)
+  }
```
glbrntt, line 42: nit: just use the helper above

glbrntt (changes requested):

Sources/GRPCHTTP2Core/GRPCMessageDeframer.swift
```diff
@@ -0,0 +1,90 @@
...
+
+  private let decompressor: Zlib.Decompressor?
+  private let maximumPayloadSize: Int?
+  private var effectiveMaximumPayloadSize: Int {
+    self.maximumPayloadSize ?? Self.defaultMaximumPayloadSize
+  }
+
+  /// Create a new ``GRPCMessageDeframer``.
+  /// - Parameters:
+  ///   - maximumPayloadSize: The maximum size a message payload can be.
```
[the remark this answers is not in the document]
gjcairo: Ah, in this implementation, I'm using it as the `limit` passed to the `Zlib/Decompressor` if the compressed flag is set, or as the size of the message payload if it's not. This means that the uncompressed payload size isn't checked if it's compressed - we only check the decompressed size.

In v1, there are two different limits yes. One for the message payload, that's checked before decompression (if any), and one for the decompressed message size. Do we want to maintain this?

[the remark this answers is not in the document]
glbrntt: It seems not, I believe that was correct at the time but after some spelunking I found these two changes to the core lib:
- https://github.com/grpc/grpc/pull/23177
- https://github.com/grpc/grpc/pull/22575

The upshot is that we should have one limit and we should check the length of the compressed message _and_ the decompressed message against that limit.

Sources/GRPCHTTP2Core/GRPCMessageDeframer.swift
```diff
@@ -0,0 +1,93 @@
...
+        )
+      }
+      return try decompressor.decompress(&message, limit: self.maximumPayloadSize)
+    } else {
+      if message.readableBytes > self.maximumPayloadSize {
+        throw RPCError(
+          code: .resourceExhausted,
+          message:
+            "Message has exceeded the configured maximum payload size (max: \(self.maximumPayloadSize), actual: \(message.readableBytes))"
+        )
```
glbrntt, line 84: We need to move this check to when we first know the length (which means we can't use `readLengthPrefixedSlice`). A malicious actor could send us a huge message which we'd happily buffer and read here before throwing.

Sources/GRPCHTTP2Core/GRPCMessageDeframer.swift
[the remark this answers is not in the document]
gjcairo: Nice find, will make that change.

Tests/GRPCHTTP2CoreTests/GRPCMessageDeframerTests.swift
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  func testReadMessageOverSizeLimitWithoutCompression() throws {
+    let deframer = GRPCMessageDeframer(maximumPayloadSize: 100)
+    let processor = NIOSingleStepByteToMessageProcessor(deframer)
+
+    var buffer = ByteBuffer()
+    buffer.writeInteger(UInt8(0))
+    buffer.writeInteger(UInt32(101))
+    buffer.writeRepeatingByte(42, count: 101)
```
glbrntt, line 58: Can we have one more test effectively which is the same as this but _doesn't_ write the 101 bytes, i.e. is just the compression byte and length? This would demonstrate we do the length check at the right time.

glbrntt (changes requested):
This looks great but we need one more test please!

glbrntt (approved):
This looks great, thanks Gus!

PR #1779, opened 2024-01-29
Make testRetriesEventuallySucceed more reliable

glbrntt:
Motivation:

`testRetriesEventuallySucceed` fails infrequently. When attempting a retry, the executor checks whether it's safe for the broadcast sequence to be replayed, it does so after removing all subscribers. However, there is a timing window between invalidating subscribers and checking whether it's safe in which a subscribtion can be registered. For retries, this can happen if the server responds before the client has subscribed. In practice this is very unlikely to happen because of network latency.

Modifications:

Have the server in the test consume the inbound stream before failing, this ensures the subscribtion is created by the time the client realises the call has failed.

Result:

Test passes more reliably


gjcairo (approved):
Nice catch

PR #1782, opened 2024-01-29
Update renderer-blank line after imports, delete blank line from docs

stefanadranca:
Motivation:

We want to have a blank line between the imports and the generated code and no blank line between the docs and the protocols/methods declarations.

Modifications:

Updated the TextBasedRenderer and the tests.

Result:

The generated code will have the right format.

### DIFF
--- a/Sources/GRPCCodeGen/Internal/Renderer/TextBasedRenderer.swift
+++ b/Sources/GRPCCodeGen/Internal/Renderer/TextBasedRenderer.swift
@@ -147,7 +147,7 @@ struct TextBasedRenderer: RendererProtocol {
   /// Renders the specified comment.
   func renderComment(_ comment: Comment) {
     let prefix: String
-    let commentString: String
+    var commentString: String
     switch comment {
     case .inline(let string):
       prefix = "//"
@@ -166,6 +166,7 @@ struct TextBasedRenderer: RendererProtocol {
       commentString = string
     }
     if prefix.isEmpty {
+      while commentString.hasSuffix("\n") { commentString.removeLast() }
       writer.writeLine(commentString)
     } else {
       let lines = commentString.transformingLines { line in
@@ -177,10 +178,18 @@ struct TextBasedRenderer: RendererProtocol {
   }
 
   /// Renders the specified import statements.
-  func renderImports(_ imports: [ImportDescription]?) { (imports ?? []).forEach(renderImport) }
+  func renderImports(_ imports: [ImportDescription]?) {
+    for (`import`, isLast) in (imports ?? []).enumeratedWithLastMarker() {
+      renderImport(`import`, isLast: isLast)
+      if isLast {
+        writer.nextLineAppendsToLastLine()
+        writer.writeLine("\n")
+      }
+    }
+  }
 
   /// Renders a single import statement.
-  func renderImport(_ description: ImportDescription) {
+  func renderImport(_ description: ImportDescription, isLast: Bool) {
     func render(preconcurrency: Bool) {
       let spiPrefix = description.spi.map { "@_spi(\($0)) " } ?? ""
       let preconcurrencyPrefix = preconcurrency ? "@preconcurrency " : ""
--- a/Sources/GRPCCodeGen/Internal/Renderer/TextBasedRenderer.swift
+++ b/Sources/GRPCCodeGen/Internal/Renderer/TextBasedRenderer.swift
@@ -102,6 +102,11 @@ final class StringCodeWriter {
   ///
   /// Safe to call repeatedly, it gets reset by `writeLine`.
   func nextLineAppendsToLastLine() { nextWriteAppendsToLastLine = true }
+
+  /// Sets a flag on the writer so that the next call to `writeLine` starts a new line.
+  ///
+  /// Safe to call repeatedly, it gets reset by `writeLine`.
+  func nextLineOnNewLine() { nextWriteAppendsToLastLine = false }
 }
 
 /// A renderer that uses string interpolation and concatenation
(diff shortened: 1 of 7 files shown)


Sources/GRPCCodeGen/Internal/Renderer/TextBasedRenderer.swift
```diff
@@ -177,10 +178,18 @@ struct TextBasedRenderer: RendererProtocol {
   }
 
   /// Renders the specified import statements.
-  func renderImports(_ imports: [ImportDescription]?) { (imports ?? []).forEach(renderImport) }
+  func renderImports(_ imports: [ImportDescription]?) {
+    for (`import`, isLast) in (imports ?? []).enumeratedWithLastMarker() {
+      renderImport(`import`, isLast: isLast)
+      if isLast {
+        writer.nextLineAppendsToLastLine()
+        writer.writeLine("\n")
```
glbrntt, line 186: I don't think this is the right approach as it special cases the rendering of imports. I think adding a `CodeBlockItem` for an empty line is probably the better way to do this because it's more flexible: you can then use it in other places like between declarations, for example.

Sources/GRPCCodeGen/Internal/Renderer/TextBasedRenderer.swift
```diff
@@ -166,6 +166,7 @@ struct TextBasedRenderer: RendererProtocol {
       commentString = string
     }
     if prefix.isEmpty {
+      while commentString.hasSuffix("\n") { commentString.removeLast() }
```
glbrntt, line 169: Where are the problematic extra new lines coming from with the preformatted comments? Modifying it here doesn't seem like the right thing to do, they are after all preformatted and we're re-formatting them...

glbrntt (changes requested):

Sources/GRPCCodeGen/Internal/Renderer/TextBasedRenderer.swift
glbrntt: So if we have a proto that's like:

```proto
// Some comments about Foo
service Foo {
}
```

and get the comments using `protoSourceComments()`, do we get `"// Some comments about Foo"`, `"// Some comments about Foo\n"` or `"// Some comments about Foo\n\n`" or something else?

It's not clear to me where the extra newline is coming from but I don't think we should reformat pre-formatted comments.

stefanadranca: yes, we get "// Some comments about Foo\n"

glbrntt: Okay, so how come when we render it we get the extra new line? Does the writer add a newline?

stefanadranca: The `writeLine()` function of the renderer that is used sets a property "nextWriteAppendsToLastLine" to false at the end, this way a new line is added after every line that is rendered.

glbrntt: I'm not sure that's the problem, I think the issue is that `writeLine` just appends a single `String` to an array of `lines` in the writer which are eventually joined by newlines. In other words the writer adds in new lines between each "line" added by `writeLine`. 

However when we render the preformatted comments we append the whole block as a single line. The code below - for rendering the other comments - is different: it splits the lines and writes each separately. This lets the _writer_ deal with adding new lines.

I think we should do the same _but_ when the last line is empty (i.e. the text ended in "\n" before it was split) we should discard that empty line.

glbrntt (changes requested):

Tests/GRPCCodeGenTests/Internal/Renderer/TextBasedRendererTests.swift
```diff
@@ -356,6 +372,16 @@ final class Test_TextBasedRenderer: XCTestCase {
     )
   }
 
+  func testEmptyLine() throws {
+    try _test(
+      .emptyLine,
+      renderedBy: TextBasedRenderer.renderCodeBlockItem(_:),
+      rendersAs: #"""
+
+        """#
```
glbrntt, line 381: I don't think the test tests what you think it tests 🙂

The last line of a multiline string literal _doesn't_ include a line break. As the line is otherwise empty it means that the multiline string in your test is really just an empty string. (You should be able to replace it with `""` and the test would be the same.)

---

From https://docs.swift.org/swift-book/documentation/the-swift-programming-language/stringsandcharacters#Multiline-String-Literals

> A multiline string literal includes all of the lines between its opening and closing quotation marks. The string begins on the first line after the opening quotation marks (""") and ends on the line before the closing quotation marks, which means that neither of the strings below start or end with a line break:
> ```swift
> let singleLineString = "These are the same."
> let multilineString = """
> These are the same.
> """
> ```

glbrntt (changes requested):

Tests/GRPCCodeGenTests/Internal/Renderer/TextBasedRendererTests.swift
glbrntt: FWIW the correct test here should be rendering as `"\n"`

Tests/GRPCCodeGenTests/Internal/Translator/IDLToStructuredSwiftTranslatorSnippetBasedTests.swift
```diff
@@ -262,7 +335,7 @@ final class IDLToStructuredSwiftTranslatorSnippetBasedTests: XCTestCase {
 
   func testSameGeneratedNameServicesSameNamespaceError() throws {
     let serviceA = ServiceDescriptor(
-      documentation: "Documentation for AService",
+      documentation: "/// Documentation for AService\n",
```
glbrntt, line 338: I don't think the service or method docs should be pre-formatted. We should only pre-format the leading trivia at the top of the proto file.

glbrntt (changes requested):
I'm happy with this fix but a couple of small things need sorting first

Tests/GRPCCodeGenTests/Internal/Translator/IDLToStructuredSwiftTranslatorSnippetBasedTests.swift
stefanadranca: They are also extracted from the `Location` objects associated with the methods and services, so they are pre-formatted. I remember we've decided to have all the docs / comments from the proto file as pre-formatted and only the ones we generate  as `.docs`

glbrntt: I thought we decided that only the copyright header would be pre-formatted because those are the only docs we should copy verbatim. I should've picked this up in the PR which added it.

The other docs are formatted by `protoSourceComments()` which inserts the leading `///`, which isn't actually necessary as the renderer can (and should) do it instead. 

Either way we can just leave it as is for now, it's not particularly important.

glbrntt (approved):

PR #1787, opened 2024-02-01
Add GRPCStreamStateMachine

gjcairo:
## Motivation

We need a state machine to drive the stream lifetime logic for both the client and the server.

## Modifications

 Added a GRPCStreamStateMachine that drives the logic for both clients and servers.

## Result

Closer to having a complete HTTP2 transport implementation.


Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,222 @@
...
+    case .clientOpenServerClosed:
+      preconditionFailure("Cannot have received anything from a closed server.")
+    case .clientClosedServerIdle, .clientClosedServerOpen, .clientClosedServerClosed:
+      preconditionFailure("Shouldn't receive anything if client's closed.")
+    }
+  }
+  
+  // - MARK: Client-opening transitions
+
+  mutating func clientSentMetadata() throws {
```
glbrntt, line 144: nit: should probably be `Send` rather than `Sent` as the client hasn't sent it yet

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,222 @@
...
+    // Client sends metadata only when opening the stream.
+    // They send grpc-timeout and method name along with it.
+    // TODO: should these things be validated in the handler or here?
```
glbrntt, line 147: For the timeout we'll insert it at a higher level (as it isn't transport specific) so no need to verify it here. For the method name (which is transport specific) we should insert it here, which I think means the state machine will need to hold the `MethodDescriptor` of the RPC we're going to call.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,222 @@
...
+    } else {
+      self.clientReceivedMetadata()
+    }
+  }
+  
+  mutating func receive(message: ByteBuffer, messageProcessor: ([UInt8]) throws -> Void) throws {
+    // This is a message received by the client, from the server.
+    switch self.state {
+    case .clientIdleServerIdle(let configuration):
+      preconditionFailure("Cannot have received anything from server if client is not yet open.")
```
glbrntt, line 130: We've changed our stance a bit on using `preconditionFailure`, anything that's network reachable should be an assertion failure (i.e. we can discover it in debug mode) but an error in release mode.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,222 @@
...
+    case .clientOpenServerClosed:
+      // Nothing to do: no point in sending a message if the server's closed.
+      ()
+    case .clientClosedServerIdle, .clientClosedServerOpen, .clientClosedServerClosed:
+      preconditionFailure("Client is closed, cannot send a message")
+    }
+  }
+  
+  mutating func send(status: String, trailingMetadata: Metadata) throws {
+    // Only server does this.
```
glbrntt, line 90: The client does need to send an end though.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,222 @@
...
+      return nil
+    case .clientClosedServerIdle, .clientClosedServerOpen, .clientClosedServerClosed:
+      preconditionFailure("Can't send request if client is closed")
+    }
+  }
+  
+  mutating func receive(metadata: Metadata, endStream: Bool) {
+    // This is metadata received by the client from the server.
+    // It can be initial, which confirms that the server is now open;
+    // or an END_STREAM trailer, meaning the response is over.
```
glbrntt, line 118: The server might send trailers without end stream set and then follow it with an empty data frame with end stream set, so we also need to be wary of that. One way to "know" they are meant as trailers is the presence of 'grpc-status' . This check could be done by the handler though, in essence "faking" the end stream, I think that'd be fine.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,222 @@
...
+    if endStream {
+      self.clientReceivedEndHeader()
+    } else {
+      self.clientReceivedMetadata()
+    }
+  }
+  
+  mutating func receive(message: ByteBuffer, messageProcessor: ([UInt8]) throws -> Void) throws {
```
glbrntt, line 126: Hmm not sure about the message processor, I think instead our processor should buffer messages then when we get channel read complete we can read them out in a similar way to how we handle messages to send.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,222 @@
...
+    let decompressor: Zlib.Decompressor?
+  }
+
+  private var state: State
+  
+  enum State {
+    case clientIdleServerIdle(Configuration)
+    case clientOpenServerIdle(ClientOpenState)
+    case clientOpenServerOpen(ClientOpenState)
+    case clientOpenServerClosed(ClientOpenState)
```
glbrntt, line 35: One thing I find really helpful to avoid bugs in state machines is having a separate state struct for each case. That way you can't accidentally switch to the wrong case by accident, e.g.:

```swift
switch state {
  case .clientOpenServerIdle(var state):
    state.doSomething()
    self.state = .clientOpenServerOpen(state) // oops, was meant to go back to idle
  default:
    // ...
}
```

You can also encode valid transitions by only having `init`s from other states, it makes the state transitions very deliberate.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,222 @@
...
+import GRPCCore
+import NIOCore
+
+// TODO: this is all done from client's perspective for now - work on server-side later.
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+struct GRPCStreamStateMachine {
+  struct Configuration {
+    let maximumPayloadSize: Int
+    let compressor: Zlib.Compressor?
+    let decompressor: Zlib.Decompressor?
```
glbrntt, line 26: We won't have these at this point: the client will only know the compression algorithms it's willing to support for compression and decompression.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
gjcairo: Right, but it's not sent in the form of a status and trailer, right? IIUC the END_STREAM flag is set at the end of the sent message(s)

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
gjcairo: Ah yeah forgot to leave a comment here. I definitely don't like this, just added it temporarily until I came up with something better. It's the downside of using the NIO processor, we need to fit things to that API. Buffering at this level feels like the only thing we can do here really, so I'll go for that.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
glbrntt: Correct, it was to highlight we still need a state transition. I saw `clientSentEnd` further but didn't think to return to and remove this comment.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+      self.deframer = previousState.deframer
+      self.decompressor = previousState.decompressor
+      self.inboundMessageBuffer = previousState.inboundMessageBuffer
+    }
+  }
+}
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+struct GRPCStreamStateMachine {
+  private var _stateMachine: GRPCStreamStateMachineProtocol
```
glbrntt, line 164: We should avoid going through a protocol here and just have an enum of the two different state machines.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+    var inboundMessageBuffer: OneOrManyQueue<[UInt8]>
+    
+    init(
+      previousState: ClientIdleServerIdleState,
+      compressionAlgorithm: Encoding?
+    ) {
+      if let zlibMethod = Zlib.Method(encoding: compressionAlgorithm) {
+        self.compressor = Zlib.Compressor(method: zlibMethod)
+        self.decompressor = Zlib.Decompressor(method: zlibMethod)
+      }
```
glbrntt, line 76: These can actually be different: the client e.g. set "message-encoding" to "deflate" and "message-accept-encoding" to "identity, gzip" meaning that they will send compressed messages with accept responses which are either not compressed or compressed with "gzip"

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+            message: "Endpoint is missing: client cannot send initial metadata without it."
+          )
+        }
+
+        self.state = .clientOpenServerIdle(.init(
+          previousState: state,
+          compressionAlgorithm: metadata.encoding
+        ))
+      case .clientOpenServerIdle, .clientOpenServerOpen, .clientOpenServerClosed:
+        throw assertionFailureAndCreateRPCErrorOnFailedPrecondition("Client is already open: shouldn't be sending metadata.")
```
glbrntt, line 238: nit: self (and elsewhere)

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+        return request
+      case .clientOpenServerOpen(var clientOpenState):
+        let request = try clientOpenState.framer.next(compressor: clientOpenState.compressor)
+        self.state = .clientOpenServerOpen(clientOpenState)
+        return request
+      case .clientOpenServerClosed:
+        // Nothing to do: no point in sending request if server is closed.
+        return nil
+      case .clientClosedServerIdle, .clientClosedServerOpen, .clientClosedServerClosed:
+        throw assertionFailureAndCreateRPCErrorOnFailedPrecondition("Can't send request if client is closed.")
```
glbrntt, line 294: It can: 

```swift
try state.send(message: [1], endStream: false)
try state.send(message: [2], endStream: false)
try state.send(message: [3], endStream: true) // client closed now
let message = try state.nextOoutboundMessage() // <- shouldn't throw

```

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+    mutating func clientReceivedEndHeader() throws {
+      switch self.state {
+      case .clientIdleServerIdle:
+        throw assertionFailureAndCreateRPCErrorOnFailedPrecondition("Client can't have received a stream end trailer if both client and server are idle.")
+      case .clientOpenServerIdle:
+        throw assertionFailureAndCreateRPCErrorOnFailedPrecondition("Server cannot have sent an end stream header if it is still idle.")
+      case .clientOpenServerOpen(let state):
+        self.state = .clientOpenServerClosed(.init(previousState: state))
+      case .clientOpenServerClosed:
+        throw assertionFailureAndCreateRPCErrorOnFailedPrecondition("Server is already closed, can't have received the end stream trailer twice.")
```
glbrntt, line 319: It _might_ make sense to ignore this: the server may send headers with 'grpc-status' but without end stream set followed by empty headers with end stream set. We should just ignore it if that happens.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+      case .clientClosedServerIdle:
+        throw assertionFailureAndCreateRPCErrorOnFailedPrecondition("Server cannot have sent end stream trailer if it is idle.")
```
glbrntt, line 321: This is valid and is usually called a "trailers-only" response (see https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#outline)

If happens if the server doesn't want to process the RPC at all (e.g. user isn't authenticated, server is about to shut down etc)

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+      case .clientClosedServerOpen(let state):
+        state.compressor?.end()
+        state.decompressor?.end()
+        self.state = .clientClosedServerClosed
+      case .clientClosedServerClosed:
+        throw assertionFailureAndCreateRPCErrorOnFailedPrecondition("Server cannot have sent end stream trailer if it is already closed.")
```
glbrntt, line 327: Similar to above, might make sense to ignore this one too.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+      }
+    }
+    
+    mutating func clientReceivedMetadata() throws {
+      switch self.state {
+      case .clientIdleServerIdle:
+        throw assertionFailureAndCreateRPCErrorOnFailedPrecondition("Server cannot have sent metadata if the client is idle.")
+      case .clientOpenServerIdle(let state):
+        self.state = .clientOpenServerOpen(.init(previousState: state))
```
glbrntt, line 336: We need to inspect the response headers because the server will indicate what compression algorithm the server is using, if any.

We also need to check the ":status" pseudo header (must be 200, although some http error codes will map directly to a gRPC status and in such a case we need to move to closed and forward an appropriate `RPCStatus` see also https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md) and content-type.

glbrntt: This function will also need to deal with converting the `HPACKHeaders` to `Metadata`

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+      case .clientOpenServerOpen:
+        // This state is valid: server can send trailing metadata without END_STREAM
+        // set, and follow it with an empty message frame where the flag *is* set.
+        ()
+        // TODO: I believe we should set some flag in the state to signal that
+        // we're expecting an empty data frame with END_STREAM set; otherwise,
+        // we could get an infinite number of metadata frames from the server -
+        // not sure this should be allowed.
```
glbrntt, line 344: I would check whether we think it should be an end stream state and then call the other function. We can be more lenient about receiving more metadata afterwards as we're about to close the stream anyway.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+      case .clientOpenServerClosed:
+        throw assertionFailureAndCreateRPCErrorOnFailedPrecondition("Server is closed, nothing could have been sent.")
+      case .clientClosedServerClosed, .clientClosedServerIdle, .clientClosedServerOpen:
+        throw assertionFailureAndCreateRPCErrorOnFailedPrecondition("Client is closed, shouldn't have received anything.")
```
glbrntt, line 348: This isn't true: if the client is closed it means it's sent end stream, it can still receive messages until the server sends end stream. `.clientClosedServerIdle` is a valid state here and `.clientClosedServerOpen` needs to be treated like `.clientOpenServerOpen`.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+        }
+        if endStream {
+          self.state = .clientOpenServerClosed(.init(previousState: state))
+        } else {
+          self.state = .clientOpenServerOpen(state)
+        }
+      case .clientOpenServerClosed:
+        throw assertionFailureAndCreateRPCErrorOnFailedPrecondition("Cannot have received anything from a closed server.")
+      case .clientClosedServerIdle, .clientClosedServerOpen, .clientClosedServerClosed:
+        throw assertionFailureAndCreateRPCErrorOnFailedPrecondition("Shouldn't receive anything if client's closed.")
```
glbrntt, line 371: We can still receive when the client is closed as long as the server is open.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+        return message
+      case .clientOpenServerClosed(var state):
+        let message = state.inboundMessageBuffer.pop()
+        self.state = .clientOpenServerClosed(state)
+        return message
+      case .clientOpenServerIdle, 
+          .clientIdleServerIdle, 
+          .clientClosedServerIdle,
+          .clientClosedServerOpen,
+          .clientClosedServerClosed:
```
glbrntt, line 389: I think both of these states need to allow for reading inbound messages

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+        return nil
+      }
+    }
+    
+    private func assertionFailureAndCreateRPCErrorOnFailedPrecondition(_ message: String) -> RPCError {
+      if !self.skipAssertions {
+        assertionFailure(message)
+      }
+      return RPCError(code: .failedPrecondition, message: message)
```
glbrntt, line 398: I would bias towards `internalError` here, it's more indicative of a gRPC implementation issue than `failedPrecondition` which is normally used for errors at the service/application level.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+      case .clientOpenServerIdle, 
+          .clientIdleServerIdle, 
+          .clientClosedServerIdle,
+          .clientClosedServerOpen,
+          .clientClosedServerClosed:
+        return nil
+      }
+    }
+    
+    private func assertionFailureAndCreateRPCErrorOnFailedPrecondition(_ message: String) -> RPCError {
```
glbrntt, line 394: Can you pass the `#line` through to `assertionFailure` too?

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+      if !self.skipAssertions {
+        assertionFailure(message)
+      }
```
glbrntt, line 397: Not that it matters much but you could do `assert(self.skipAssertions, message)`

glbrntt (commented):
Just dropping my first set of comments. Will do the server state machine in a separate batch.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+            .string(self.supportedCompressionAlgorithms
+              .map({ $0.rawValue })
+              .joined(separator: ",")
+            )
+          ))
+          return .reject(status: status, trailers: trailers)
+        }
+
+        guard self.supportedCompressionAlgorithms.contains(where: { $0 == encoding }) else {
+          if self.supportedCompressionAlgorithms.isEmpty {
```
glbrntt, line 516: We need to be careful that we always support `identity`, i.e. no compression. At the moment `encoding` could be `identity` and supported algorithms could be empty and we'd throw an error which would be wrong!

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+            throw RPCError(
+              code: .unimplemented,
+              message: "Compression is not supported"
+            )
```
glbrntt, line 520: This should result in rejection as well

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+            let trailers = Metadata(dictionaryLiteral: (
+              "grpc-accept-encoding",
+              .string(self.supportedCompressionAlgorithms
+                .map({ $0.rawValue })
+                .joined(separator: ",")
+              )
+            ))
+            return .reject(status: status, trailers: trailers)
+          }
```
glbrntt, line 534: This is repeated from above, can we refactor to avoid the repetition?

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+        ))
+        return .doNothing
+      case .clientOpenServerIdle, .clientOpenServerOpen, .clientOpenServerClosed:
+        throw assertionFailureAndCreateRPCErrorOnFailedPrecondition("Client shouldn't have sent metadata twice.")
+      case .clientClosedServerIdle, .clientClosedServerOpen, .clientClosedServerClosed:
+        throw assertionFailureAndCreateRPCErrorOnFailedPrecondition("Client can't have sent metadata if closed.")
+      }
+    }
+    
+    mutating func receive(message: ByteBuffer, endStream: Bool) throws {
```
glbrntt, line 554: nit: at this point we don't know it's a message (it may be part of a message, multiple messages, etc.). It might be clearer to call this `bytes`/`buffer`/`data` something along those lines to make it clear it hasn't yet been de-framed.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+      switch self.state {
+      case .clientIdleServerIdle, .clientOpenServerIdle, .clientClosedServerIdle:
+        throw assertionFailureAndCreateRPCErrorOnFailedPrecondition("Server is not open yet.")
+      case .clientOpenServerOpen(var state):
+        let response = try state.framer.next(compressor: state.compressor)
+        self.state = .clientOpenServerOpen(state)
+        return response
+      case .clientClosedServerOpen:
+        // No point in sending response if client is closed: do nothing.
+        return nil
```
glbrntt, line 597: We can still send messages to the client in this state

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+      case .clientOpenServerClosed, .clientClosedServerClosed:
+        throw assertionFailureAndCreateRPCErrorOnFailedPrecondition("Can't send response if server is closed.")
```
glbrntt, line 599: There's no guarantee our outbound buffer has been drained yet so we shouldn't throw here

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+        return request
+      case .clientClosedServerOpen(var state):
+        let request = state.inboundMessageBuffer.pop()
+        self.state = .clientClosedServerOpen(state)
+        return request
+      case .clientClosedServerIdle,
+          .clientIdleServerIdle,
+          .clientOpenServerClosed,
+          .clientClosedServerClosed:
+        return nil
```
glbrntt, line 621: I think we need to try when the client is closed too

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+}
+
+extension MethodDescriptor {
+  init?(fullyQualifiedMethod: String) {
+    let split = fullyQualifiedMethod.split(separator: "/")
+    guard split.count == 2 else {
+      return nil
+    }
+    self.init(service: String(split[0]), method: String(split[1]))
+  }
```
glbrntt, line 641: This is fine for now but this is likely to be something we can optimise in the future. The split, for example, allocates an array which we could avoid by looking for indexes, and we _might_ allocate creating `String`s from the substrings (depending on how long the substrings are). Fixing that one would probably require changing the internals of `MethodDescriptor`.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+}
+
+extension Metadata {
```
glbrntt, line 644: These will actually need to be on `HPACKHeaders` and we should keep them internal

glbrntt (commented):
This is taking shape. One thing that surprised me a little was having two separate state machines, as I figured they'd have more in common. Did you consider having a single state machine and switching on whether it's a client/server in the cases where the logic differs?

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,687 @@
...
+        return request
+      case .clientOpenServerClosed:
+        // Nothing to do: no point in sending request if server is closed.
+        return nil
+      case .clientClosedServerIdle, .clientClosedServerOpen, .clientClosedServerClosed:
+        throw assertionFailureAndCreateRPCErrorOnFailedPrecondition("Can't send request if client is closed.")
+      }
+    }
+    
+    mutating func receive(metadata: Metadata, endStream: Bool) throws -> OnMetadataReceived {
```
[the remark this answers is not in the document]
gjcairo: Ah, I was thinking this conversion would happen in the handler. I kind of went back and forth on this though, wasn't sure if it fell more under the responsibilities of the handler or the state machine. I decided to assume the conversion would happen in the handler and thus the state machine would already get all the "correct" GRPC-specific types instead of types coming from the H2 layer.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
gjcairo: I thought having the code completely separate instead of having each method be a big `switch` case and containing the logic for both sides was slightly more readable. But yes probably the performance cost is too high and I understand why we'd want to do it the other way. I can change it.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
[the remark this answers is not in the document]
glbrntt: Ah right. I think it needs to do the conversion because it holds information required to do so (enabled compression algorithms being one example). It's already taking care of framing/deframing which is part of the conversion.

Tests/GRPCHTTP2CoreTests/GRPCStreamStateMachineTests.swift
```diff
@@ -0,0 +1,2118 @@
...
+    
+    XCTAssertNoThrow(try stateMachine.send(message: [42, 42], endStream: false))
+    request = try XCTUnwrap(stateMachine.nextOutboundMessage())
+    
+    let expectedBytes: [UInt8] = [
+      0, // compression flag: unset
+      0, 0, 0, 2, // message length: 2 bytes
+      42, 42 // original message
+    ]
+    XCTAssertEqual(Array(buffer: request!), expectedBytes)
```
glbrntt, line 765: nit, avoid the `!`

```suggestion
    XCTAssertNil(try stateMachine.nextOutboundMessage())
    
    XCTAssertNoThrow(try stateMachine.send(message: [42, 42], endStream: false))
    let request = try XCTUnwrap(stateMachine.nextOutboundMessage())
    
    let expectedBytes: [UInt8] = [
      0, // compression flag: unset
      0, 0, 0, 2, // message length: 2 bytes
      42, 42 // original message
    ]
    XCTAssertEqual(Array(buffer: request), expectedBytes)
```

Tests/GRPCHTTP2CoreTests/GRPCStreamStateMachineTests.swift
```diff
@@ -0,0 +1,2118 @@
...
+    XCTAssertNoThrow(try stateMachine.send(message: [], endStream: true))
+    
+    // Even though the client is closed, because it received a message while open,
+    // we must get the message now.
+    let receivedMessage = try XCTUnwrap(stateMachine.nextInboundMessage())
+    XCTAssertEqual(receivedMessage, [42, 42])
+    
+    XCTAssertNil(stateMachine.nextInboundMessage())
+  }
+}
```
glbrntt, line 1090: All of these tests cover performing an action in each state (which is great). I think we need a second kind of test which exercises common paths through the state machine, e.g.:

1. client sends metadata, client sends message(s), client sends end, server sends metadata, server sends message(s), server sends end
2. client sends metadata, server sends metadata, ... etc.

This is effectively like picking parts of individual tests above and combining them but I think is still valuable as you can exercise different routes through the state machine and is closer to how the state machine will actually be used.

Tests/GRPCHTTP2CoreTests/GRPCStreamStateMachineTests.swift
```diff
@@ -0,0 +1,2118 @@
...
+  func testSendMetadataWhenClientIdleAndServerIdle() throws {
+    var stateMachine = makeClientStateMachine()
+    XCTAssertNoThrow(try stateMachine.send(metadata: self.testMetadata))
+  }
+  
+  func testSendMetadataWhenClientOpenAndServerIdle() throws {
+    var stateMachine = makeClientStateMachine()
+    
+    // Open client
+    XCTAssertNoThrow(try stateMachine.send(metadata: self.testMetadata))
```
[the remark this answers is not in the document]
glbrntt: Okay, having looked at more tests, I _strongly_ recommend doing this 😄

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1109 @@
...
+          if configuration.acceptedEncodings.isEmpty {
+            throw RPCError(
+              code: .unimplemented,
+              message: "Compression is not supported"
+            )
+          } else {
+            let status = Status(
+              code: .unimplemented,
+              message: "\(rawEncoding) compression is not supported; supported algorithms are listed in grpc-accept-encoding"
+            )
```
glbrntt: We'll need to encode these directly into the headers since that'll be what's returned to the client. It also means we need to form a trailers-only response and move into the server closed state.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1109 @@
...
+            let trailers = HPACKHeaders(httpHeaders: [
+              "grpc-accept-encoding":
+                configuration.acceptedEncodings
+                .map({ $0.name })
+                .joined(separator: ",")
```
glbrntt: Not strictly necessary, we can also repeat the key and have a different value for each key. The latter probably makes more sense when you take into account the `:status`, `grpc-status`, `grpc-message` etc.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1109 @@
...
+  ) -> HPACKHeaders {
+    // Trailers always contain the grpc-status header, and optionally,
+    // grpc-status-message, and custom metadata.
+    // If it's a trailers-only response, they will also contain :status and
+    // content-type.
+    var headers = HPACKHeaders()
+    
+    if trailersOnly {
+      // Reserve 5 for capacity: 3 for the required headers, and 1 for the
+      // optional status message.
```
glbrntt, line 749: 3 + 1 != 5

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
glbrntt: We should remove this extension now

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1109 @@
...
+        .compactMap { CompressionAlgorithm(rawValue: String($0)) }
+    }
+  }
+  
+  private func firstString(forKey key: GRPCHTTP2Keys) -> String? {
+    self[stringValues: key.rawValue].first(where: { _ in true })
+  }
+}
+
+extension HPACKHeaders {
```
glbrntt, line 1028: I'm slightly concerned about using all of these setters when constructing `HPACKHeaders` because they use `replaceOrAdd` which is O(n). Because we're creating the `HPACKHeaders` we know that there's no value to replace so a simple add would be absolutely fine.

glbrntt (changes requested):

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1109 @@
...
+  }
+}
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+extension GRPCStreamStateMachine {
+  private func makeResponseHeaders(outboundEncoding: CompressionAlgorithm?, customMetadata: Metadata) -> HPACKHeaders {
+    // Response headers always contain :status (HTTP Status 200) and content-type.
+    // They may also contain grpc-encoding, grpc-accept-encoding, and custom metadata.
+    var headers = HPACKHeaders()
+    headers.reserveCapacity(4 + customMetadata.count)
```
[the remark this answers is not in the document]
gjcairo: Forgot to add accepted encodings

Sources/GRPCHTTP2Core/Internal/GRPCStatusMessageMarshaller.swift
gjcairo, line 1: This was copied from v1

Sources/GRPCCore/Metadata.swift
```diff
@@ -85,6 +85,17 @@ public struct Metadata: Sendable, Hashable {
   public enum Value: Sendable, Hashable {
     case string(String)
     case binary([UInt8])
+
+    /// The value as a String. If it was originally stored as a binary, the base64-encoded String version
+    /// of the binary data will be returned instead.
+    public func encoded() -> String {
```
FranzBusch, line 91: To align with other existing API https://developer.apple.com/documentation/foundation/data/2142853-base64encodedstring
```suggestion
    public func base64EncodedString() -> String {
```

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1422 @@
...
+      return try self.serverSend(metadata: metadata, configuration: serverConfiguration)
+    }
+  }
+
+  mutating func send(message: [UInt8], endStream: Bool) throws {
+    switch self.configuration {
+    case .client:
+      try self.clientSend(message: message, endStream: endStream)
+    case .server:
+      assert(
```
FranzBusch, line 314: Why are we not using `self.assertionFailureAndCreateRPCErrorOnInternalError` here?

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1422 @@
...
+        self.state = .clientOpenServerIdle(state)
+      }
+    case .clientOpenServerOpen(var state):
+      state.framer.append(message)
+      if endStream {
+        self.state = .clientClosedServerOpen(.init(previousState: state))
+      } else {
+        self.state = .clientOpenServerOpen(state)
+      }
+    case .clientOpenServerClosed(let state):
```
FranzBusch, line 487: This one seems interesting to me. Shouldn't the `writer.write` method throw an error in this case and the write shouldn't even make it here or is there a potential race?

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1422 @@
...
+  private mutating func clientReceive(
+    metadata: HPACKHeaders,
+    endStream: Bool
+  ) throws -> OnMetadataReceived {
+    switch self.state {
+    case .clientOpenServerIdle(let state):
+      guard metadata.grpcStatus != nil || metadata.status == "200" else {
+        let httpStatusCode = metadata.status
+          .flatMap { Int($0) }
+          .map { HTTPResponseStatus(statusCode: $0) }
```
FranzBusch, line 560: NIT: Any reason to not write this in a single line?

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1422 @@
...
+        // This state is valid: server can send trailing metadata without grpc-status
+        // or END_STREAM set, and follow it with an empty message frame where they are set.
+        ()
+      }
+      return .receivedMetadata(Metadata(headers: metadata))
+    case .clientClosedServerIdle(let state):
+      guard metadata.grpcStatus != nil || metadata.status == "200" else {
+        let httpStatusCode = metadata.status
+          .flatMap { Int($0) }
+          .map { HTTPResponseStatus(statusCode: $0) }
```
FranzBusch, line 630: Same nit here

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1422 @@
...
+        "Server cannot have sent metadata if the client is idle."
+      )
+    case .clientOpenServerClosed:
+      throw self.assertionFailureAndCreateRPCErrorOnInternalError(
+        "Server is closed, nothing could have been sent."
+      )
+    }
+  }
+
+  private mutating func clientReceive(bytes: ByteBuffer, endStream: Bool) throws {
```
FranzBusch, line 719: Why do some methods operate on `[UInt8]` and others on `ByteBuffer`?

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1422 @@
...
+    // Trailers always contain the grpc-status header, and optionally,
+    // grpc-status-message, and custom metadata.
+    // If it's a trailers-only response, they will also contain :status and
+    // content-type.
+    var headers = HPACKHeaders()
+    let customMetadataCount = customMetadata?.count ?? 0
+    if trailersOnly {
+      // Reserve 4 for capacity: 3 for the required headers, and 1 for the
+      // optional status message.
+      headers.reserveCapacity(4 + customMetadataCount)
```
FranzBusch, line 906: The conditional reserving seems a bit too heavily optimised in my opinion. I personally would just always reserve 4.

Sources/GRPCHTTP2Core/Internal/ContentType.swift
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// See:
+// - https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md
+enum ContentType {
+  case protobuf
```
FranzBusch, line 20: Is this naming from their docs because I find it confusing that it is `protobuf` but we send `"application/grpc"`

Tests/GRPCHTTP2CoreTests/GRPCStreamStateMachineTests.swift
```diff
@@ -0,0 +1,2045 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import GRPCCore
+import NIOCore
+import NIOHPACK
+import XCTest
+
+@testable import GRPCHTTP2Core
```
FranzBusch, line 22: Out of interest, could we do this without a `@testable` import? I always try to aim for not testing internals or use the `package` modifier since that allows to run tests in release mode

FranzBusch (commented):
Overall looks great! Left some minor comments

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
gjcairo: You mean why not `.flatMap { HTTPResponseStatus(statusCode: Int($0)) }`?
`Int($0)` can return `nil` so this way I avoid having to explicitly unwrap the optional.

Tests/GRPCHTTP2CoreTests/GRPCStreamStateMachineTests.swift
gjcairo: I'm using the state enum, as well as the header key names which are all internal in the state machine - that's why the `@testable` is needed. 
As for `package`, you mean using that instead of `internal` for the types that have to be used from tests?

Sources/GRPCHTTP2Core/Internal/ContentType.swift
gjcairo: I kept the same naming from v1. `application/grpc` and `application/grpc+proto` are interchangeable, so I think that's where the name comes from.

Tests/GRPCHTTP2CoreTests/GRPCStreamStateMachineTests.swift
FranzBusch: Yes exactly. The `package` modifier should be able to replace all `@testable` usage or `@_spi(Testing)` in the long run.

Sources/GRPCCore/Metadata.swift
glbrntt: It's only a base64 encoded if the underlying value is `binary` so the `base64EncodedString` isn't accurate.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
gjcairo: When we receive data, we receive `ByteBuffer`s from the channel. However we then deal with `[UInt8]` internally, as we decode bytes into RPC messages, meaning the outbound methods operate on `[UInt8]`.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1423 @@
...
+  }
+
+  private mutating func serverNextOutboundMessage() throws -> OnNextOutboundMessage {
+    switch self.state {
+    case .clientIdleServerIdle, .clientOpenServerIdle, .clientClosedServerIdle:
+      throw self.assertionFailureAndCreateRPCErrorOnInternalError("Server is not open yet.")
+    case .clientOpenServerOpen(var state):
+      let response = try state.framer.next(compressor: state.compressor)
+      self.state = .clientOpenServerOpen(state)
+      return response != nil ? .sendMessage(response!) : .awaitMoreMessages
```
glbrntt, line 1123: nit: avoid the `!` by using `map`: `response.map { .sendMessage($0) } ?? .awaitMoreMessages`

glbrntt (changes requested):

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1423 @@
...
+
+  enum OnMetadataReceived: Equatable {
+    case receivedMetadata(Metadata)
+
+    // Client-specific actions
+    case failedRequest(Status)
+    case doNothing
+
+    // Server-specific actions
+    case rejectRPC(status: Status?, trailers: HPACKHeaders?)
```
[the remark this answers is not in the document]
gjcairo: I was thinking the handler will call `send(status:metadata:)` afterwards when receiving this return value, so we need the status to be separate. It will be added as a header anyways in the call to `send(status:metadata:)`.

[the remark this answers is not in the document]
glbrntt: I think it'd be simpler to skip the extra call into the state machine and just return the well formed `HPACKHeaders` from here instead.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1423 @@
...
+        let decoder = GRPCMessageDeframer(
+          maximumPayloadSize: previousState.maximumPayloadSize,
+          decompressor: self.decompressor
+        )
+        self.deframer = NIOSingleStepByteToMessageProcessor(decoder)
+      }
+    }
+  }
+
+  struct ClientClosedServerClosedState {
```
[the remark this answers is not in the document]
gjcairo: Hm, if nothing goes wrong, we should always reach this state. It's possible we throw an invalid state error though and in that case we won't be transitioning to this state. I'm not sure where the right place to call end would be in that scenario though.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1423 @@
...
+      self.state = .clientClosedServerClosed(state)
+      return message
+    case .clientIdleServerIdle,
+      .clientOpenServerIdle,
+      .clientClosedServerIdle:
+      return nil
+    }
+  }
+
+  private func assertionFailureAndCreateRPCErrorOnInternalError(
```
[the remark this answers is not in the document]
glbrntt: > agree it's unconventional API, I think it's unfortunate that there's no way to say "this method _will always_ throw". 

There is: `func whatever() throws -> Never`

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1391 @@
...
+              customTrailers.addString(
+                acceptedEncoding.name,
+                forKey: GRPCHTTP2Keys.acceptEncoding.rawValue
+              )
+            }
+            let trailers = self.makeTrailers(
+              status: status,
+              customMetadata: customTrailers,
+              trailersOnly: true
+            )
```
glbrntt, line 1135: Can we reduce the duplication here? The only diff is the status message and the custom metadata.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1391 @@
...
+      }
+
+      // Secondly, find a compatible encoding the server can use to compress outbound messages,
+      // based on the encodings the client has advertised.
+      let outboundEncoding: CompressionAlgorithm
+      let clientAdvertisedEncodings = metadata.values(
+        forHeader: GRPCHTTP2Keys.acceptEncoding.rawValue,
+        canonicalForm: true
+      )
+      .compactMap { CompressionAlgorithm(rawValue: String($0)) }
```
glbrntt, line 1153: It should be possible to avoid this array here by doing a plain for loop. We need to be very careful with allocations since we'll do this for every request.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1391 @@
...
+        do {
+          let decodedBinary = try header.value.base64Decoded()
+          metadata.addBinary(decodedBinary, forKey: header.name)
+        } catch {
+          metadata.addString(header.value, forKey: header.name)
+        }
+      } else {
+        if header.name == GRPCHTTP2Keys.grpcStatusMessage.rawValue {
+          let decodedStatusMessage = GRPCStatusMessageMarshaller.unmarshall(header.value)
+          metadata.addString(decodedStatusMessage, forKey: header.name)
```
glbrntt, line 1363: Hmm, didn't we have code elsewhere to remove values for this and the status code? If so there's no point special casing it here.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1391 @@
...
+      self = .gzip
+    default:
+      return nil
+    }
+  }
+}
+
+extension Metadata {
+  init(headers: HPACKHeaders) {
+    var metadata = Metadata()
```
glbrntt, line 1351: Can we reserve capacity here?

glbrntt (changes requested):

Sources/GRPCCore/Internal/Base64.swift
```diff
@@ -267,7 +267,7 @@ enum Base64 {
 
   private static let encodePaddingCharacter: UInt8 = 61
 
-  private static let encodeBase64: [UInt8] = [
+  private static let base64Encodings: [UInt8] = [
```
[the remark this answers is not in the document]
gjcairo: Formatting, there was a rule in CI complaining about having the base64 suffix when the enclosing enum is called Base64

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1391 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import GRPCCore
+import NIOCore
+import NIOHPACK
+import NIOHTTP1
```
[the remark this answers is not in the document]
gjcairo: Because I'm using `HTTPResponseStatus` which is in H1. I can `import enum NIOHTTP1.HTTPResponseStatus` if that's preferable.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1391 @@
...
+        headers.add(name: metadataPair.key, value: metadataPair.value.encoded())
+      }
+    }
+
+    return headers
+  }
+
+  private mutating func serverSend(
+    status: Status?,
+    customMetadata: Metadata?,
```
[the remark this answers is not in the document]
gjcairo: Related to an old change that I forgot to update, fixed.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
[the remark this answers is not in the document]
glbrntt: Nah, the import is fine. I was just a little surprised to see it and didn't make the connection to `HTTPResponseStatus`.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1391 @@
...
+  }
+
+  private func makeTrailers(
+    status: Status?,
+    customMetadata: Metadata?,
+    trailersOnly: Bool
+  ) -> HPACKHeaders {
+    // If status isn't present, it means we're returning a non-200 HTTP :status
+    // header. This should already be included in the custom metadata: assert
+    // this and simply return those headers.
```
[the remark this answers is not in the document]
gjcairo: Actually, I was never passing a `nil` `Status` in this method, meaning `makeNon200StatusTrailers` was never being called. This was a leftover from a previous change where I was building trailers differently.
We'll always have a grpc-status and a :status equal to 200 in all the cases we build trailers. The only scenario where we don't, is when we respond with an Unsupported Media Type HTTP status code (415) in `serverReceive(metadata:endStream:configuration:)`. I can get rid of this code altogether.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1391 @@
...
+    }
+  }
+
+  private func makeNon200StatusTrailers(_ customMetadata: Metadata?) -> HPACKHeaders {
+    assert(customMetadata != nil, "Something is very wrong")
+    var headers = HPACKHeaders()
+    for metadataPair in customMetadata! {
+      headers.add(name: metadataPair.key, value: metadataPair.value.encoded())
+    }
+    let httpStatus = headers.firstString(forKey: .status)
```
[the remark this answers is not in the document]
gjcairo: Related to the comment below, this isn't actually used anymore so I've deleted it.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1366 @@
...
+      return response.map { .sendMessage($0) } ?? .awaitMoreMessages
+    case .clientOpenServerClosed(var state):
+      let response = try state.framer.next(compressor: state.compressor)
+      if let response {
+        self.state = .clientOpenServerClosed(state)
+        return .sendMessage(response)
+      } else {
+        self.state = .clientOpenServerClosed(state)
+        return .noMoreMessages
+      }
```
glbrntt, line 1235: nit: reduce duplication

```suggestion
      let response = try state.framer.next(compressor: state.compressor)
      self.state = .clientOpenServerClosed(state)
      if let response {
        return .sendMessage(response)
      } else {
        return .noMoreMessages
      }
```

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1366 @@
...
+    case .clientClosedServerClosed(var state):
+      let response = try state.framer.next(compressor: state.compressor)
+      if let response {
+        self.state = .clientClosedServerClosed(state)
+        return .sendMessage(response)
+      } else {
+        self.state = .clientClosedServerClosed(state)
+        return .noMoreMessages
+      }
```
glbrntt, line 1244:
```suggestion
      let response = try state.framer.next(compressor: state.compressor)
      self.state = .clientClosedServerClosed(state)
      if let response {
        return .sendMessage(response)
      } else {
        return .noMoreMessages
      }
```

Tests/GRPCHTTP2CoreTests/GRPCStreamStateMachineTests.swift
```diff
@@ -0,0 +1,2149 @@
...
+  case clientOpenServerOpen
+  case clientOpenServerClosed
+  case clientClosedServerIdle
+  case clientClosedServerOpen
+  case clientClosedServerClosed
+}
+
+extension HPACKHeaders {
+  // Client
+  static let clientInitialMetadata: Self = [
```
glbrntt, line 36: can these all be `fileprivate`?

Tests/GRPCHTTP2CoreTests/GRPCStreamStateMachineTests.swift
```diff
@@ -0,0 +1,2149 @@
...
+ */
+
+import GRPCCore
+import NIOCore
+import NIOHPACK
+import XCTest
+
+@testable import GRPCHTTP2Core
+
+enum TargetStateMachineState: CaseIterable {
```
glbrntt, line 24: nit: `fileprivate` / embedded in the test class

glbrntt (changes requested):
This is getting very close to be ready to merge. I left a few nits/comments.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -0,0 +1,1366 @@
...
+        )
+      }
+      try self.serverSend(message: message)
+    }
+  }
+
+  mutating func send(
+    status: Status,
+    metadata: Metadata,
+    trailersOnly: Bool
```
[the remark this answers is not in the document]
gjcairo: Yeah you're right, this was left over from before one of the many refactors and wasn't necessary anymore; I actually found a state transition bug removing it, so good catch.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -434,11 +439,16 @@ extension GRPCStreamStateMachine {
     var headers = HPACKHeaders()
     headers.reserveCapacity(7 + customMetadata.count)
 
-    // Add required headers
+    // Add required headers.
     // See https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
-    headers.add(methodDescriptor.fullyQualifiedMethod, forKey: .path)
-    headers.add(scheme.rawValue, forKey: .scheme)
+
+    // The order is important here: reserved HTTP2 headers (those starting with `:`)
```
glbrntt, line 445: fyi they're called "pseudo-headers"

glbrntt (approved):
I think this looks good, nice one Gus!

PR #1788, opened 2024-02-02
Add service and method config

glbrntt:
Motivation:

When resoling addresses, gRPC clients can discover configuration to use
published by service maintainers. This is in the form of 'service
config' and typically encoded as JSON. However the structure of the
config is a translation of the service config protobuf message using the
standard proto to JSON mappings.

This change expands on the existing `MethodConfiguration` to allow it to
be decoded from JSON and adds a `ServiceConfiguration` which can also be
decoded from JSON.

Modifications:

- Add additional properties to `MethodConfiguration` and make it
  `Codabale`
- Add `ServiceConfiguration` and make it `Codabale`
- Add a `RuntimeError`
- Add the `service_config.proto` message and its dependencies and
  generate these messages into the test module.

Result:

Service config and method config are more fully featured


glbrntt:
This change is in two commits:
- the 'real' work in [76b5069](https://github.com/grpc/grpc-swift/pull/1788/commits/76b5069c358951b4af591a40472378e66905bf38)
- update proto script and generate in [d7aade2](https://github.com/grpc/grpc-swift/pull/1788/commits/d7aade2bd932e70e44d6454c98f2502c7216fe7f)

Sources/GRPCCore/Configuration/MethodConfiguration.swift
```diff
@@ -0,0 +1,676 @@
+/*
```
glbrntt, line 1: This isn't actually a new file, it got moved but changed quite substantially, before the configuration only included the timeout and execution policy (i.e. retry/hedging).

Sources/GRPCCore/RuntimeError.swift
```diff
@@ -0,0 +1,128 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/// An error thrown at runtime.
+///
+/// In contrast to ``RPCError``, the ``RuntimeError`` represents errors which happen at a scope
+/// wider than an individual RPC. For example, passing invalid configuration values.
+public struct RuntimeError: Error, Hashable, @unchecked Sendable {
```
glbrntt, line 21: It _might_ make sense to replace the `ClientError` and `ServerError` with this. They all have the same structure but differ only in error codes. WDYT @gjcairo?

Sources/GRPCCore/Configuration/MethodConfiguration.swift
glbrntt: These changes also have a bit of an impact on `MethodConfigurations` but I'll do those in a separate PR.

Sources/GRPCCore/RuntimeError.swift
gjcairo: Hmm my first guess would be it makes sense, but I do wonder if it would be useful from a user's perspective to be able to differentiate between client and server errors beyond the error codes. I'm thinking something like:
```swift
func wrapAndRecordMetrics(someClosure: () throws -> Void) {
  do {
    try someClosure()
  } catch is ServerError {
    // increase server error metric counter
  } catch is ClientError {
    // increase client error metric counter
  } catch {
    throw error
  }
}
```

But I'm not sure how realistic something like this would be, and I can't come up with any other obvious scenarios, so maybe it would make sense to replace them.

Tests/GRPCCoreTests/Configuration/MethodConfigurationCodingTests.swift
```diff
@@ -0,0 +1,424 @@
...
+  }
+
+  func testDecodeRetryPolicy() throws {
+    let json = """
+      {
+        "maxAttempts": 1,
+        "initialBackoff": "1s",
+        "maxBackoff": "3s",
+        "backoffMultiplier": 1.6,
+        "retryableStatusCodes": ["ABORTED"]
```
gjcairo, line 205: Bit of a nit, but maybe we could have at least one test with multiple codes here to make sure decoding works well in that scenario too?

gjcairo (approved):
LGTM overall! Left some very small nits/questions.

Sources/GRPCCore/Configuration/MethodConfiguration.swift
```diff
@@ -0,0 +1,676 @@
...
+      throw RuntimeError(
+        code: .invalidArgument,
+        message: "hedgingDelay must be greater than or equal to zero"
+      )
+    }
+  }
+}
+
+private func validateMaxAttempts(_ value: Int) throws -> Int {
+  guard value > 0 else {
```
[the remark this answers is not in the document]
glbrntt: Great catch!

gjcairo (approved):

Sources/GRPCCore/RuntimeError.swift
glbrntt: Forgot to reply to this one earlier: I don't know if that's useful because we now a third `RuntimeError` which can be thrown by either client or server so this sort of metric recording fails to be useful anyway.

PR #1789, opened 2024-02-04
[CodeGenLib] Change location of method descriptor array in generated code

stefanadranca:
Motivation:

The method descriptor array should reside inside the Methods enum (which should be called Method).

Modifications:

- added the descriptor array as a member of the Methods enum
- chenged the Methods enum name to Method
- changed the tests accordingly

Result:

Better organized generated code.

### DIFF
--- a/Sources/GRPCCodeGen/Internal/Translator/TypealiasTranslator.swift
+++ b/Sources/GRPCCodeGen/Internal/Translator/TypealiasTranslator.swift
@@ -38,10 +38,10 @@
 ///       // ...
 ///
 ///       public static let descriptors: [MethodDescriptor] = [
-///       echo.Echo.Get.descriptor,
-///       echo.Echo.Collect.descriptor,
-///       // ...
-///     ]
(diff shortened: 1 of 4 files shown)


Sources/GRPCCodeGen/Internal/Translator/TypealiasTranslator.swift
```diff
@@ -36,12 +36,13 @@
...
 ///       }
 ///       // ...
-///     }
-///     public static let methods: [MethodDescriptor] = [
+///
+///       public static let descriptors: [MethodDescriptor] = [
 ///       echo.Echo.Get.descriptor,
 ///       echo.Echo.Collect.descriptor,
 ///       // ...
 ///     ]
```
glbrntt, line 44: nit: intendation

```suggestion
///         echo.Echo.Get.descriptor,
///         echo.Echo.Collect.descriptor,
///         // ...
///       ]
```

glbrntt (approved):

PR #1792, opened 2024-02-05
[ProtobufCodeGen] Use the generated input and output type names

stefanadranca:
Motivation:

protoc-gen-swift generates message type names different from the ones in the file descriptors. The CodeGenerationRequest object should contain the generated names for the ProtobufCodeGen.

Modifications:

- the parser uses the SwiftProtobufNamer to get the generated type names and adds them in the CodeGenerationRequest objects
- propagated the change and modified tests accordingly

Result:

The inout and output type names will be the generated ones and the generated code will ocmpile.


Sources/GRPCProtobufCodeGen/ProtobufCodeGenParser.swift
```diff
@@ -23,7 +23,14 @@ import struct GRPCCodeGen.CodeGenerationRequest
 /// Parses a ``FileDescriptor`` object into a ``CodeGenerationRequest`` object.
 internal struct ProtobufCodeGenParser {
   internal init() {}
-  internal func parse(input: FileDescriptor) throws -> CodeGenerationRequest {
+  internal func parse(
+    input: FileDescriptor,
+    protoFileToModuleMappings: ProtoFileToModuleMappings
```
glbrntt, line 28: Do we actually need the mappings? It looks like we only actually need the namer. Also I think it makes more sense to configure the parser with the namer/mappings rather than passing it per call to `parse`.

glbrntt (changes requested):

Sources/GRPCProtobufCodeGen/ProtobufCodeGenParser.swift
```diff
@@ -22,9 +22,19 @@ import struct GRPCCodeGen.CodeGenerationRequest
...
+  let input: FileDescriptor
+  let namer: SwiftProtobufNamer
+
+  internal init(input: FileDescriptor) {
+    self.input = input
+    self.namer = SwiftProtobufNamer(
+      currentFile: input,
+      protoFileToModuleMappings: ProtoFileToModuleMappings()
+    )
+  }
```
glbrntt, line 34: Sorry, my last comment wasn't clear enough: I want us to pass the `SwiftProtobufNamer` in `init`. We _do_ need the mappings as it turns out, they come from the options (`options.protoToModuleMappings`).

glbrntt (approved):

PR #1797, opened 2024-02-07
Generated code fixtures: availability guards and descriptor paths

stefanadranca:
Motivation:

The generated code must contain the availability guards on top of all protocols, extensions and typealiases for protocols in order to be compiled. The full descriptor paths caused errors for namespaces and services with the same name, so we will use the relative paths to them in the descriptors array from the 'Method' enum.

Modifications:

- added the availability guards in the translators
- modified the descriptors path in the typealias translator (for the descriptors array)
- modified all tests accordingly

Result:

The generated code will be compiled.

### DIFF
--- a/Sources/GRPCCodeGen/Internal/Translator/ClientCodeTranslator.swift
+++ b/Sources/GRPCCodeGen/Internal/Translator/ClientCodeTranslator.swift
@@ -22,6 +22,7 @@
 /// a representation for the following generated code:
 ///
 /// ```swift
+/// @available(macOS 13.0, *)
 /// public protocol Foo_BarClientProtocol: Sendable {
 ///   func baz<R: Sendable>(
 ///     request: ClientRequest.Single<foo.Bar.Method.baz.Input>,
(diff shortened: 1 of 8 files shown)


Sources/GRPCCodeGen/Internal/Translator/ClientCodeTranslator.swift
```diff
@@ -22,6 +22,7 @@
 /// a representation for the following generated code:
 ///
 /// ```swift
+/// @available(macOS 13.0, *)
```
glbrntt, line 25: These need to be for other platforms too: `@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)`

glbrntt (changes requested):

Sources/GRPCCodeGen/Internal/Translator/ServerCodeTranslator.swift
```diff
@@ -22,12 +22,14 @@
 /// a representation for the following generated code:
 ///
 /// ```swift
+/// @available(macOS 13.0, *)
```
glbrntt, line 25: these need updating too

Sources/GRPCCodeGen/Internal/Translator/TypealiasTranslator.swift
```diff
@@ -38,13 +38,15 @@
...
 ///       public static let descriptors: [MethodDescriptor] = [
-///         Echo.Echo.Method.Get.descriptor,
-///         Echo.Echo.Method.Collect.descriptor,
+///         Get.descriptor,
+///         Collect.descriptor,
 ///         // ...
 ///       ]
 ///     }
 ///
+///     @available(macOS 13.0, *)
```
glbrntt, line 47: these need updating too

glbrntt (changes requested):
A few more docs need fixing but looks good otherwise

glbrntt (approved):

PR #1803, opened 2024-02-12
Add a socket address type

glbrntt:
Motivation:

Name resolvers resolve targets to an address we know how to connect to. This could be represented as an IPv4, IPv6, UDS or VSOCK address which can all be wrapped up in a broader 'SocketAddress'.

Modifications:

- Add a SocketAddress type and tests

Results:

Can represent an address to connect to

### DIFF
--- a/Tests/GRPCHTTP2CoreTests/Client/Resolver/SocketAddressTests.swift
+++ b/Tests/GRPCHTTP2CoreTests/Client/Resolver/SocketAddressTests.swift
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2024, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


Tests/GRPCHTTP2CoreTests/Client/Resolver/SocketAddressTests.swift
```diff
@@ -0,0 +1,74 @@
...
+    var address: SocketAddress = .ipv4(host: "127.0.0.1", port: 42)
+    XCTAssertDescription(address, "[ipv4]127.0.0.1:42")
+
+    address = .ipv6(host: "::1", port: 42)
+    XCTAssertDescription(address, "[ipv6]::1:42")
+
+    address = .unixDomainSocket(path: "baz")
+    XCTAssertDescription(address, "[unix]baz")
+
+    address = .vsock(contextID: 314, port: 159)
```
gjcairo, line 57: Can we add an assertion when using `.any`?

gjcairo (approved):
LGTM just one nit

Tests/GRPCHTTP2CoreTests/Client/Resolver/SocketAddressTests.swift
glbrntt: done!

PR #1804, opened 2024-02-12
Fixed .proto file imports handling in ProtobufCodeGenParser

stefanadranca:
Motivation:

The .proto file imports from a .proto file were added as file imports in the generated code. The correct behavior is to add dependencies on the modules containing generated code and types associated with the .proto files, if they are generated in a different module from the one of the .proto file we are parsing.

Modifications:

- Created a method that adds only the necessary imports associated with a .proto file into the CodeGenerationRequest object
- updated the test for the parser

Result:

Imports will be generated correctly for '.proto' files.


Tests/GRPCProtobufCodeGenTests/ProtobufCodeGenParserTests.swift
```diff
@@ -65,6 +80,11 @@ final class ProtobufCodeGenParserTests: XCTestCase {
       """
     )
 
+    XCTAssertEqual(parsedCodeGenRequest.dependencies.count, 3)
+    let expectedDependencyNames = ["DifferentModule", "ExtraModule", "GRPCProtobuf"]
+    let parsedDependencyNames = parsedCodeGenRequest.dependencies.map { $0.module }.sorted()
```
glbrntt, line 85: Is the ordering of dependencies stable? I.e. if we run the code generator multiple times will the order always be the same?

Tests/GRPCProtobufCodeGenTests/ProtobufCodeGenParserTests.swift
```diff
@@ -199,4 +216,15 @@ extension Google_Protobuf_FileDescriptorProto {
       $0.syntax = "proto3"
     }
   }
+
+  static func makeEmpty(_ name: String) -> Google_Protobuf_FileDescriptorProto {
```
glbrntt, line 220: It's not really empty, it has a name and package so this it's a bit misleading. An `init` with `name` and `package` as parameters would make this a bit clearer.

glbrntt (changes requested):

Tests/GRPCProtobufCodeGenTests/ProtobufCodeGenParserTests.swift
stefanadranca: That is a good point, I should order them alphabetically before generating. I will add this

glbrntt: That's only necessary if the order of `protoToModuleMappings.neededModules` is unstable.

stefanadranca: neededModules returns the modules imported based on the '.proto' files in alphabetical order, so the ordering is stable. My question is if we should order all of the dependencies (given that we also add "GRPCCore", "GRPCSwiftProtobuf" and the extra imports that are passed in as an option to protoc, besides the one we get from neededModules)

glbrntt: Well we insert the GRPC modules into known locations so their ordering is stable. The extra modules _could_ be passed in a different order on different invocations of `protoc-gen-grpc-swift` so we could sort those too. I don't mind whether we do that or just sort the whole lot, the result is the same.

Tests/GRPCProtobufCodeGenTests/ProtobufCodeGenParserTests.swift
```diff
@@ -23,17 +23,47 @@ import XCTest
...
     }
+    let moduleMappings = SwiftProtobuf_GenSwift_ModuleMappings.with {
+      $0.mapping = [
+        SwiftProtobuf_GenSwift_ModuleMappings.Entry.with {
+          $0.protoFilePath = ["different-module.proto"]
+          $0.moduleName = "DifferentModule"
+        },
+        SwiftProtobuf_GenSwift_ModuleMappings.Entry.with {
+          $0.protoFilePath = ["bifferent-module.proto"]
+          $0.moduleName = "BifferentModule"
```
glbrntt, line 55: Bifferent 😄

glbrntt (approved):

PR #1805, opened 2024-02-12
Update `swift-nio-http2` version

redryerye:
## Context

[A vulnerability has been found](https://github.com/advisories/GHSA-qppj-fm5r-hxr3) in `swift-nio-http2`.  
The affected versions were < 1.28.0, and the patched version is 1.28.0.

## Changes

Update the minimum version to `1.28.1`

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -36,7 +36,7 @@ let packageDependencies: [Package.Dependency] = [
   ),
   .package(
     url: "https://github.com/apple/swift-nio-http2.git",
-    from: "1.24.1"
+    from: "1.28.1"
   ),
   .package(
     url: "https://github.com/apple/swift-nio-transport-services.git",


glbrntt:
Thanks for opening this PR, however this change isn't necessary due to the way the Swift Package Manager works. The version in `Package.swift` is the lowest viable version that is usable.

If no `Package.resolved` file exists then when building SwiftPM will resolve to the latest (possible) version of each dependency and transitive dependency.

If a `Package.resolved` file _does_ exist then you'd need to run `swift package update` to update the latest version of each package anyway. At the point it doesn't matter whether the version in gRPC Swift's `Package.swift` is from `1.24.1` or `1.28.1`. 

The dependabot alert for gRPC Swift was for a `Package.resolved` file in a client-only example which has now been dismissed.

PR #1806, opened 2024-02-13
Fix namespace name for nested package names

stefanadranca:
Motivation:

The current implementation of the ProtobufParser replaces the dots with their scalar value for the CodeGenerationRequest names. We want the dots to be replaced by "_" or removed.

Modifications:

- cleaned the package name before creting the Name object associated with it
- updated the ProtobufCodeGen tests.

Result:

Tha dots from the package name will be replaced by underscores in the input and output types and will be removed form the other types that are namespace prefixed.


Sources/GRPCProtobufCodeGen/ProtobufCodeGenParser.swift
```diff
@@ -120,10 +120,14 @@ extension CodeGenerationRequest.ServiceDescriptor {
...
+    // Packages that are based on the path of the '.proto' file usually
+    // contain dots. For example: "grpc.test".
+    let cleanNamespace = package.replacingOccurrences(of: ".", with: "_")
     let namespace = CodeGenerationRequest.Name(
-      base: package,
-      generatedUpperCase: NamingUtils.toUpperCamelCase(package),
-      generatedLowerCase: NamingUtils.toLowerCamelCase(package)
+      base: cleanNamespace,
+      generatedUpperCase: NamingUtils.toUpperCamelCase(cleanNamespace),
+      generatedLowerCase: NamingUtils.toLowerCamelCase(cleanNamespace)
```
stefanadranca, line 130: The 'toUpperCamelCase()' and 'toLowerCamelCase()' methods are also transforming the snake case into camel case. This is why the generated type names (generated through CodeGenLib) don't contain underscores in the namespace prefix. On the other hand, the Input and Output types contain underscores in the namespace prefix. The same is true for the protocol names in protocol names in v1. This inconsistency is generated by the different SwiftProtobuf naming functions (some public and some internal that I can't use).

stefanadranca: I will update this PR after we merge the enum name PR then

Tests/GRPCProtobufCodeGenTests/ProtobufCodeGeneratorTests.swift
```diff
@@ -58,13 +58,13 @@ final class ProtobufCodeGeneratorTests: XCTestCase {
         import DifferentModule
         import ExtraModule
 
-        internal enum Helloworld_Greeter {
+        internal enum HelloWorld_Greeter {
```
glbrntt, line 61: Don't we want the name to be `Hello_World_Greeter` here? I.e. the `Hello_World` prefix should match that used by the messages (`Hello_World_HelloRequest` and `Hello_World_HelloReply`).

Sources/GRPCProtobufCodeGen/ProtobufCodeGenParser.swift
```diff
@@ -166,3 +171,22 @@ extension FileDescriptor {
...
   }
 }
+
+extension SwiftProtobufNamer {
+  internal func trimTrailingUnderscores(from name: String) -> String {
+    var nameCopy = name
+    while nameCopy.hasSuffix("_") {
+      nameCopy = String(nameCopy.dropLast())
+    }
+    return nameCopy
```
glbrntt, line 181: We copy the string for each `_` we drop. Instead we can look for the last index in the string which isn't an `_` (if there's no such index then the string only contains `_` or is empty in which case we can just return an empty string):

```swift
if let index = name.lastIndex({ $0 != "_" }) {
  return String(name[...index])
} else {
  return ""
}
```

This would also make more sense as an extension on `String` (which could take the `Character` to trim).

Sources/GRPCProtobufCodeGen/ProtobufCodeGenParser.swift
```diff
@@ -166,3 +171,22 @@ extension FileDescriptor {
...
+  }
+
+  internal func formattedUpperCasePackage(file: FileDescriptor) -> String {
+    let unformattedPackage = self.typePrefix(forFile: file)
+    return trimTrailingUnderscores(from: unformattedPackage)
+  }
+
+  internal func formattedLowerCasePackage(file: FileDescriptor) -> String {
+    return file.package.lowercased().replacingOccurrences(of: ".", with: "_")
```
glbrntt, line 190: This assumes `file.package` is already lowercased which I don't think is necessarily the case

stefanadranca: Doesn't `lowercased()` handle this? Or is it not the expected behaviour?

glbrntt: Sorry, I can't read 🤦‍♂️ 

Although I think we want it to be lower camel case don't we?

stefanadranca: Yep, we need lower camel case

Sources/GRPCProtobufCodeGen/ProtobufCodeGenParser.swift
```diff
@@ -173,20 +173,27 @@ extension FileDescriptor {
...
   }
 
   internal func formattedLowerCasePackage(file: FileDescriptor) -> String {
-    return file.package.lowercased().replacingOccurrences(of: ".", with: "_")
+    let upperCasePackage = self.formattedUpperCasePackage(file: file)
+    let lowerCaseFirstComponent = NamingUtils.toLowerCamelCase(
+      String(upperCasePackage.split(separator: "_")[0])
+    )
+
+    return lowerCaseFirstComponent + upperCasePackage.dropFirst(lowerCaseFirstComponent.count)
```
glbrntt, line 187: IIUC, we're only lower-camel-casing the first component here. I assume `toLowerCamelCase` lower-camel-cases each component in the name?

stefanadranca: NamingUtils.toLowerCamelCase() gets rid of the snake case, do you have a suggestion how to implement this in a nicer way (I don't really like this implementation either). Also, I've discovered that this function results in an error for empty packages, so there is still work to do

glbrntt: I think I'd go for something like the following (i.e. lower camel case each component)

```swift
let upperCasePackage = self.formattedUpperCasePackage(file: file)
let lowerCaseComponents = upperCasePackage.split(separator: "_").map { component in 
  NamingUtils.toLowerCamelCase(component)
}
return lowerCaseComponents.joined(separator: "_")
```

stefanadranca: Thanks!

glbrntt (approved):

PR #1807, opened 2024-02-13
Add name resolver protocols and registry

glbrntt:
Motivation:

When creating a client users will provide a target to connect to. This is usually a host to resolve via DNS but may be a Unix domain socket, pre-resolved IP address or something custom.

Users get a resolver from a registry by asking it to resolve a given target. Resolvers then provide a set of endpoints and a service configuration. These values can change over time.

Modifications:

- Add the `NameResolver` struct which provides an `AsyncSequence` of `NameResolutionResult`s.
- Add `NameResolutionResult` which provides an array of `Endpoint`s and a `ServiceConfiguration`.
- Add the `NameResolverFactory` protocol which returns a `NameResolver` for a given resolvable target
- Add the `ResolvableTarget` protocol
- Add the `NameResolverRegistry` which holds a set of name resolver factories and can return a `NameResolver` for a given target

Note, the actual resolver factories and targets will be added in a later PR.

Result:

All the boilerplate is in place for name resolvers.


Sources/GRPCHTTP2Core/Client/Resolver/NameResolver.swift
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  /// Create a new name resolver.
+  public init(results: RPCAsyncSequence<NameResolutionResult>, updateMode: UpdateMode) {
+    self.names = results
+    self.updateMode = updateMode
+  }
+}
+
+/// The result of name resolution, a service of endpoints to connection to and the service
```
gjcairo, line 49: [Nit] I guess you meant something like:
```suggestion
/// The result of name resolution, a list of endpoints to connect to and the service
```

Sources/GRPCHTTP2Core/Client/Resolver/NameResolver.swift
```diff
@@ -0,0 +1,117 @@
...
+  func isCompatible<Other: ResolvableTarget>(withTarget target: Other) -> Bool {
+    return target is Target
+  }
+
+  /// Returns a name resolver if the given target is compatible.
+  ///
+  /// - Parameter target: The target to make a name resolver for.
+  /// - Returns: A name resolver or `nil` if the target isn't compatible.
+  func makeResolverIfCompatible<Other: ResolvableTarget>(_ target: Other) -> NameResolver? {
+    guard let target = target as? Target else { return nil }
```
gjcairo, line 111: Maybe worth using the above func?
```suggestion
    guard self.isCompatible(withTarget: target) else { return nil }
```

Tests/GRPCHTTP2CoreTests/Client/Resolver/NameResolverRegistryTests.swift
```diff
@@ -0,0 +1,104 @@
...
+    XCTAssertEqual(resolvers.count, 0)
+  }
+
+  func testRegisterFactory() {
+    var resolvers = NameResolverRegistry()
+    resolvers.registerFactory(FailingResolver())
+    XCTAssertEqual(resolvers.count, 1)
+
+    // Adding a resolver of the same type replaces it.
+    resolvers.registerFactory(FailingResolver())
```
gjcairo, line 54: May be a bit overkill, but we're not exactly testing that the new resolver replaces the old one here: we're only asserting that there's just one left. You think we can initialise the `FailingResolver` with some form of ID, so that we can assert the new one is the one that remains?

gjcairo (changes requested):

Sources/GRPCHTTP2Core/Client/Resolver/NameResolver.swift
```diff
@@ -0,0 +1,117 @@
...
+  /// to changes in addresses by awaiting for new values in a loop.
+  ///
+  /// Resolvers with the ``UpdateMode-swift.enum/pull`` update mode shouldn't be subscribed to,
+  /// instead you should create an iterator and ask for new results as and when necessary.
+  public var names: RPCAsyncSequence<NameResolutionResult>
+
+  /// How ``names`` is updated and should be consumed.
+  public let updateMode: UpdateMode
+
+  public enum UpdateMode: Hashable, Sendable {
```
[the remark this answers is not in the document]
glbrntt: Yeah I think we should do that. I _think_ we only need to switch over the value internally so this is completely find to do, there's no downside to it (beyond more code).

Sources/GRPCHTTP2Core/Client/Resolver/NameResolver.swift
glbrntt: I'm not sure what I was thinking here... 🙈

Sources/GRPCHTTP2Core/Client/Resolver/NameResolver.swift
glbrntt: If we did use `isCompatible(withTarget:)` we'd need to call `self.resolver(for: target as! Target)` and I'd rather avoid the `!` here.

gjcairo (approved):

PR #1808, opened 2024-02-13
[CodeGenLib] Remove namespace enum generation

stefanadranca:
Motivation:

Having a namespace enum in the generated code leads to redeclaration of the package enum when we generate code for .proto files within the same package. Thus, we cannot generate the namespace enums anymore, so we need to generate service enums that contain the namespace in their names.

Modifications:

- Changed the typealias translator to generate the service enums with namespace prefixed names
- Deleted the namespace enum generation
- Modified the CodeGenerationRequest validation function for service names
- Updated tests accordingly

Result:

Redeclaration of the same enum will be avoided when generating code using CodeGenLib.


glbrntt (approved):
LGTM!

PR #1821, opened 2024-03-05
Interop Tests implementation

stefanadranca:
Motivation:

The test cases test certain functionalities of the new GRPC server and client.

Modifications:

- created the InteroperabilityTest protocol with the run method
- created the InteroperabilityTestCase enum that will be used by main to run certain tests according to the input
- implemented the interop tests for the features we support (no caching or compression tests)
- added new assertions to the test framework

Result:

We will be able to implement the main() function that runs the interop tests using in process transport.


Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -274,9 +274,18 @@ struct ServerStreaming: InteroperabilityTest {
...
+    #if swift(>=5.9)
     let asyncStream = AsyncStream.makeStream(of: Int.self)
+    let continuation = asyncStream.continuation
+    let stream = asyncStream.stream
+    #else
+    let continuation: AsyncStream<Int>.Continuation!
+    let stream = AsyncStream(Int.self) { streamContinuation in
+      continuation = streamContinuation
+    }
+    #endif
```
glbrntt: rather than doing this we can add a backported version of the API we want to this module: https://github.com/grpc/grpc-swift/blob/main/Sources/GRPCCore/Streaming/Internal/AsyncStream%2BMakeStream.swift

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,721 @@
...
+/// - call was successful
+/// - exactly zero responses
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+struct EmptyStream: InteroperabilityTest {
+  func run(client: GRPCClient) async throws {
+    let testServiceClient = Grpc_Testing_TestService.Client(client: client)
+
+    let request = ClientRequest.Stream { writer in
+      let message = Grpc_Testing_StreamingOutputCallRequest()
+      try await writer.write(message)
```
glbrntt, line 367: The test wants an _empty_ stream, this is a stream of _one_ message. In this case we just want the request to be `ClientRequest.Stream { _ in }`

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,721 @@
...
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+struct CustomMetadata: InteroperabilityTest {
+  let initialMetadataName = "x-grpc-test-echo-initial"
+  let initialMetadataValue = "test_initial_metadata_value"
+
+  let trailingMetadataName = "x-grpc-test-echo-trailing-bin"
+  let trailingMetadataValue: [UInt8] = [0xAB, 0xAB, 0xAB]
+
+  func checkInitialMetadata(_ metadata: Metadata) throws {
+    try assertEqual(metadata.count, 1)
```
glbrntt, line 429: We don't need to assert this and there's no guarantee it's true either.

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,721 @@
...
+    var metadataIterator = metadata.makeIterator()
+    guard let initialMetadataPair = metadataIterator.next() else {
+      try assertFailure("The initial metadata should contain a key value pair.")
+      return
+    }
+    try assertEqual(initialMetadataPair.key, self.initialMetadataName)
+    try assertEqual(initialMetadataPair.value, Metadata.Value.string(self.initialMetadataValue))
```
glbrntt, line 436: We should be looking up the values based on the key, something like:

```swift
let values = metadata[self.initialMetadataName]
XCTAssertEqual(Array(values), [.string(self.initialMetadataValue)])
```

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,721 @@
...
+
+  func checkTrailingMetadata(_ metadata: Metadata) throws {
+    try assertEqual(metadata.count, 1)
+    var metadataIterator = metadata.makeIterator()
+    guard let trailingMetadataPair = metadataIterator.next() else {
+      try assertFailure("The trailing metadata should contain a key value pair.")
+      return
+    }
+    try assertEqual(trailingMetadataPair.key, self.trailingMetadataName)
+    try assertEqual(trailingMetadataPair.value, Metadata.Value.binary(self.trailingMetadataValue))
```
glbrntt, line 447: Same here, we should be looking up by the key as above

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,721 @@
...
+    let unaryRequest = Grpc_Testing_SimpleRequest.with { request in
+      request.responseSize = 314_159
+      request.payload = Grpc_Testing_Payload.with {
+        $0.body = Data(count: 271_828)
+      }
+    }
+    var metadata = Metadata()
+    metadata.addString(self.initialMetadataValue, forKey: self.initialMetadataName)
+    metadata.addBinary(self.trailingMetadataValue, forKey: self.trailingMetadataName)
+    let streamingMetadata = metadata
```
glbrntt, line 462: You can use the fact that `Metadata` is `ExpressibleByDictionaryLiteral` here to avoid the mutable `metadata`:

```swift
let metadata: Metadata = [
  self.initialMetadataName: .string(self.initialMetadataValue),
  self.trailingMetadataName: .binary(self.trailingMetadataValue)
]
```

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,721 @@
...
+            try assertEqual(message.payload.body, Data(count: 314_159))
+          // Check the trailing metadata.
+          case .trailingMetadata(let receivedTrailingMetadata):
+            try self.checkTrailingMetadata(receivedTrailingMetadata)
+          default:
+            try assertFailure(
+              "The client should have received only a message and trailing metadata."
+            )
+          }
+        }
```
glbrntt, line 513: This only validates that we receive at least two parts, not exactly two.

One trick we can do is collect all of the parts: 

```swift
let parts = try await responseParts.reduce(into: []) { $0.append($1) }
try assertEqual(parts.count, 2)
// ...

```

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,721 @@
...
+
+    let message = Grpc_Testing_SimpleRequest.with {
+      $0.responseStatus = Grpc_Testing_EchoStatus.with {
+        $0.code = Int32(self.expectedCode)
+        $0.message = self.expectedMessage
+      }
+    }
+
+    try await testServiceClient.unaryCall(request: ClientRequest.Single(message: message)) {
+      response in
```
glbrntt, line 569:
```suggestion
    try await testServiceClient.unaryCall(
      request: ClientRequest.Single(message: message)
    ) { response in
```

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,721 @@
...
+        try assertEqual(error.message, self.expectedMessage)
+      case .success(_):
+        try assertFailure(
+          "The client should receive an error with the status code and message sent by the client."
+        )
+      }
+    }
+
+    let request = ClientRequest.Stream { writer in
+      let message = Grpc_Testing_StreamingOutputCallRequest()
```
glbrntt, line 582: We need to populate this message

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,721 @@
...
+    try await testServiceClient.fullDuplexCall(request: request) { response in
+      switch response.accepted {
+      case .failure(let error):
+        try assertEqual(error.code.rawValue, self.expectedCode)
+        try assertEqual(error.message, self.expectedMessage)
+      case .success(_):
+        try assertFailure(
+          "The client should receive an error with the status code and message sent by the client."
+        )
+      }
```
glbrntt, line 595: There are two places the server can fail a streaming RPC: before processing it and during processing it. Because this test relies on the service consuming the request to know it should fail the RPC we know that the RPC will be accepted. We therefore need to consume the messages in the response and catch the error there:

```swift
do {
for try await message in response.messages {
  throw AssertionFailure(...)
} catch let error as RPCError {
  // validate error code and message
} 

```

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,721 @@
...
+
+    let responseMessage = "\t\ntest with whitespace\r\nand Unicode BMP ☺ and non-BMP 😈\t\n"
+    let message = Grpc_Testing_SimpleRequest.with {
+      $0.responseStatus = Grpc_Testing_EchoStatus.with {
+        $0.code = 2
+        $0.message = responseMessage
+      }
+    }
+    try await testServiceClient.unaryCall(request: ClientRequest.Single(message: message)) {
+      response in
```
glbrntt, line 635:
```suggestion
    try await testServiceClient.unaryCall(
      request: ClientRequest.Single(message: message)
    ) { response in
```

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,721 @@
...
+/// Client asserts:
+/// - received status code is 12 (UNIMPLEMENTED)
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+struct UnimplementedMethod: InteroperabilityTest {
+  func run(client: GRPCClient) async throws {
+    let testServiceClient = Grpc_Testing_TestService.Client(client: client)
+    try await testServiceClient.unimplementedCall(
+      request: ClientRequest.Single(message: Grpc_Testing_Empty())
+    ) {
+      response in
```
glbrntt, line 669:
```suggestion
    ) { response in
```

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,721 @@
...
+      let result = response.accepted
+      switch result {
+      case .success(_):
+        try assertFailure("The result should be an error.")
+      case .failure(let error):
+        try assertEqual(error.code, RPCError.Code.unimplemented)
```
glbrntt, line 675:
```suggestion
        try assertEqual(error.code, .unimplemented)
```

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,721 @@
...
+/// Client asserts:
+/// - received status code is 12 (UNIMPLEMENTED)
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+struct UnimplementedService: InteroperabilityTest {
+  func run(client: GRPCClient) async throws {
+    let unimplementedServiceClient = Grpc_Testing_UnimplementedService.Client(client: client)
+    try await unimplementedServiceClient.unimplementedCall(
+      request: ClientRequest.Single(message: Grpc_Testing_Empty())
+    ) {
+      response in
```
glbrntt, line 702:
```suggestion
    ) { response in
```

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -0,0 +1,721 @@
...
+      let result = response.accepted
+      switch result {
+      case .success(_):
+        try assertFailure("The result should be an error.")
+      case .failure(let error):
+        try assertEqual(error.code, RPCError.Code.unimplemented)
```
glbrntt, line 708:
```suggestion
        try assertEqual(error.code, .unimplemented)
```

glbrntt (changes requested):
Looks good for the most part, but I think there are a couple of bugs which need fixing first.

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -505,23 +468,25 @@ struct CustomMetadata: InteroperabilityTest {
...
-              "The client should have received only a message and trailing metadata."
-            )
           }
         }
       case .failure(_):
-        try assertFailure("The client should have received a response from the server.")
+        throw AssertionFailure(
+          message: "The client should have received a response from the server.",
+          file: #fileID,
+          line: #line
```
glbrntt, line 488: same here re: init

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -574,33 +539,46 @@ struct StatusCodeAndMessage: InteroperabilityTest {
...
         try assertEqual(error.code.rawValue, self.expectedCode)
         try assertEqual(error.message, self.expectedMessage)
       case .success(_):
-        try assertFailure(
-          "The client should receive an error with the status code and message sent by the client."
+        throw AssertionFailure(
+          message:
+            "The client should receive an error with the status code and message sent by the client.",
+          file: #fileID,
+          line: #line
```
glbrntt, line 554: same here re: init

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -574,33 +539,46 @@ struct StatusCodeAndMessage: InteroperabilityTest {
...
     try await testServiceClient.fullDuplexCall(request: request) { response in
-      switch response.accepted {
-      case .failure(let error):
+      do {
+        for try await _ in response.messages {
+          throw AssertionFailure(
+            message:
+              "The client should receive an error with the status code and message sent by the client.",
+            file: #fileID,
+            line: #line
```
glbrntt, line 576: same here re: init

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -640,11 +618,16 @@ struct SpecialStatusMessage: InteroperabilityTest {
...
+    try await testServiceClient.unaryCall(
+      request: ClientRequest.Single(message: message)
+    ) { response in
       switch response.accepted {
       case .success(_):
-        try assertFailure("The response should be an error with the error code 2.")
+        throw AssertionFailure(
+          message: "The response should be an error with the error code 2.",
+          file: #fileID,
+          line: #line
```
glbrntt, line 629: same here re: init

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -674,14 +657,17 @@ struct UnimplementedMethod: InteroperabilityTest {
...
-      response in
+    ) { response in
       let result = response.accepted
       switch result {
       case .success(_):
-        try assertFailure("The result should be an error.")
+        throw AssertionFailure(
+          message: "The result should be an error.",
+          file: #fileID,
+          line: #line
```
glbrntt, line 667: same here re: init

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -707,14 +693,17 @@ struct UnimplementedService: InteroperabilityTest {
...
```
glbrntt, line 703: same here re: init

glbrntt (changes requested):
Couple of nits but looks good otherwise!

glbrntt (approved):

PR #1822, opened 2024-03-05
Add a minimum connections configuration to the ConnectionPool

gjcairo:
## Motivation
Sometimes, depending on the characteristics of the traffic, we may need to execute requests in spikes.
It's possible that when we get a large spike of requests, a lot of connections are opened, which are subsequently closed by the idle timer. However, when another spike comes next, because we don't have any available open connections, latency will suffer as a result of us having to open new connections.

## Modifications
Add a ConnectionPool configuration to allow a minimum number of connections to remain open, that is, make a number of connections not go idle.

## Result
A number of connections will remain open even when there are no open streams on them.


Sources/GRPC/ConnectionPool/GRPCChannelPool.swift
```diff
@@ -275,6 +275,10 @@ extension GRPCChannelPool.Configuration {
     /// Defaults to 100.
     public var maxWaitersPerEventLoop: Int = 100
 
+    /// The minimum number of connections to keep open in this pool.
+    /// This number of connections will never go idle and be closed.
+    public var minConnections: Int = 1
```
glbrntt, line 280: As implemented this is actually min connections per event loop (which is probably what we want, but we should make that more obvious.)

glbrntt (changes requested):
At the moment this doesn't solve the issue outlined in the motivation, it just stops some connections from idling.

When connections are added to the pool they are initially in the idle state and will only be connected once other connections have enough load, i.e. lazily.

We need to bring up min connections when we initialize the pool and (if necessary) when connections are closed. We'll also need some tests to validate this behaviour.

Sources/GRPC/ConnectionManager.swift
```diff
@@ -324,6 +332,9 @@ internal final class ConnectionManager: @unchecked Sendable {
   /// attempts should be made at all.
   private let connectionBackoff: ConnectionBackoff?
 
+  /// Whether this connection should be allowed to go idle (and thus be closed when the idle timer fires).
+  internal let idleBehavior: IdleBehavior
```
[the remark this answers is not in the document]
glbrntt: Why do we store it here if it's only used in the idle handler? Can't we just pass it the idle handler on `init`?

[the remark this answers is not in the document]
gjcairo: The thing is the min connections configuration lives/is passed in via the `ConnectionPool`: we need to have the context of the pool to understand how many connections are currently being kept open and decide whether we need to keep the connection from going idle or not. 
The idle handler is created by the `ConnectionManager` in `startConnecting(backoffIterator:muxPromise:connectTimeoutOverride:)` for each connection, thus the manager has to be aware of the idle behaviour to be able to pass it down to the handler.

Sources/GRPC/ConnectionPool/ConnectionPool.swift
```diff
@@ -220,6 +235,16 @@ internal final class ConnectionPool {
     let id = manager.id
     self._connections[id] = PerConnectionState(manager: manager)
     self.delegate?.connectionAdded(id: .init(id))
+
+    // If it's one of the connections that should be kept open, then connect
+    // straight away.
+    if case .neverGoIdle = idleBehavior {
```
glbrntt, line 241: nit: explicit switch please

Sources/GRPC/ConnectionPool/ConnectionPool.swift
```diff
@@ -689,8 +714,14 @@ extension ConnectionPool: ConnectionManagerConnectivityDelegate {
     // Grab the number of reserved streams (before invalidating the index by adding a connection).
     let reservedStreams = self._connections.values[index].reservedStreams
 
+    // If we have less than the minimum number of connections, don't let
+    // the new connection close when idle.
+    let idleBehavior =
+      self.sync.activeConnections < self.minConnections
+      ? ConnectionManager.IdleBehavior.neverGoIdle : .closeWhenIdleTimeout
```
glbrntt, line 721: Why don't we just get the idle behaviour from connection that closed? We're then always replacing like for like and avoid scanning the state of all connections.

Sources/GRPC/ConnectionPool/ConnectionPool.swift
```diff
@@ -716,6 +747,15 @@ extension ConnectionPool: ConnectionManagerConnectivityDelegate {
...
     }
+
+    // Now that a connection has become unavailable, we must make sure the minimum
+    // number of connections is still met.
+    if let connectionManager = self._connections[id]?.manager,
+      case .neverGoIdle = connectionManager.idleBehavior,
+      connectionManager.sync.isTransientFailure
+    {
+      connectionManager.sync.startConnecting()
+    }
```
glbrntt, line 758: I'm not so sure about this. If the connection drops and is in transient failure then it will start attempting to connect after a backoff period. We shouldn't ignore that period other we risk hammering a server with attempts to connect.

Sources/GRPC/ConnectionPool/GRPCChannelPool.swift
```diff
@@ -275,6 +275,10 @@ extension GRPCChannelPool.Configuration {
     /// Defaults to 100.
     public var maxWaitersPerEventLoop: Int = 100
 
+    /// The minimum number of connections to keep open in this pool, per EventLoop.
+    /// This number of connections per EventLoop will never go idle and be closed.
+    public var minConnections: Int = 1
```
glbrntt, line 280: Defaulting to one is a semantic change, we should default to zero which is the current behaviour.

Sources/GRPC/ConnectionPool/GRPCChannelPool.swift
glbrntt: Sorry, I meant by making it more explicit in the name (i.e. like `connectionsPerEventLoop` and `maxWaitersPerEventLoop`)

glbrntt (changes requested):

Sources/GRPC/ConnectionPool/ConnectionPool.swift
gjcairo: Good call.

Sources/GRPC/ConnectionPool/ConnectionPool.swift
```diff
@@ -881,6 +910,22 @@ extension ConnectionPool {
...
+
+    /// The number of connections in the pool in transient failure state.
+    internal var transientFailureConnections: Int {
+      self.pool.eventLoop.assertInEventLoop()
+      return self.pool._connections.values.reduce(0) {
+        $0 &+ ($1.manager.sync.isTransientFailure ? 1 : 0)
+      }
+    }
+
```
glbrntt, line 928: I think these are unused now?

Sources/GRPC/ConnectionPool/GRPCChannelPool.swift
```diff
@@ -275,6 +275,10 @@ extension GRPCChannelPool.Configuration {
     /// Defaults to 100.
     public var maxWaitersPerEventLoop: Int = 100
 
+    /// The minimum number of connections to keep open in this pool, per EventLoop.
+    /// This number of connections per EventLoop will never go idle and be closed.
+    public var minConnections: Int = 0
```
glbrntt, line 280: The other comment was missed: we should add `PerEventLoop` to the name so that it's inline with the other per-EL options.

glbrntt (changes requested):
Two nits, looks good otherwise!

Sources/GRPC/ConnectionPool/ConnectionPool.swift
gjcairo: Ah, I'm using them in tests.

glbrntt (approved):
LGTM, thanks Gus!

PR #1823, opened 2024-03-08
Unit tests for interop tests

stefanadranca:
Motivation:

We need unit tests that run the interop tests with the in process transport.

Modifications:

Created a test target and implemented the tests for each interop test, that catch AssertionFailure errors when the interop tests fail.

Result:

We will run the interop tests with the in process transport.

TODO:
The CustomMetadata test fails - I will add the metadata handling in the TestService in a different PR


Tests/InteroperabilityUnitTests/InteroperabilityInProcessTransportTests.swift
```diff
@@ -0,0 +1,149 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import GRPCCore
+import GRPCInProcessTransport
+import XCTest
+
+@testable import InteroperabilityTests
```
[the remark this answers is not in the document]
glbrntt: Hmm, no `@testable` is fine. But do make the switch to use `makeTest`.

Tests/inProcessInteroperabilityTests/InProcessInteroperabilityTests.swift
```diff
@@ -0,0 +1,109 @@
...
+        group.cancelAll()
+      }
+    } catch let error as AssertionFailure {
+      XCTFail(error.message)
+    }
+  }
+
+  func testEmtyUnary() async throws {
+    let emptyUnaryTestCase = InteroperabilityTestCase.emptyUnary.makeTest()
+    try await runInProcessTransport(interopTest: emptyUnaryTestCase.run)
```
glbrntt, line 57: We can simplify this yet further by change the param to `runInProcessTransport` to be `InteroperabilityTestCase` and calling `makeTest` and `run` in there.

glbrntt, line 57: Also: use explicit `self` (and in all the other tests)

glbrntt (changes requested):

Sources/InteroperabilityTests/InteroperabilityTestCase.swift
```diff
@@ -16,7 +16,7 @@
 import GRPCCore
 
 @available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
-public protocol InteroperabilityTest {
+public protocol InteroperabilityTest: Sendable {
```
glbrntt, line 19: Now that the tests have been refactored to construct the interop test in place this isn't a requirement anymore.

Tests/InProcessInteroperabilityTests/InProcessInteroperabilityTests.swift
```diff
@@ -0,0 +1,106 @@
...
+        try await group.next()
+        group.cancelAll()
+      }
+    } catch let error as AssertionFailure {
+      XCTFail(error.message)
+    }
+  }
+
+  func testEmtyUnary() async throws {
+    try await self.runInProcessTransport(interopTestCase: InteroperabilityTestCase.emptyUnary)
```
glbrntt, line 56: nit: prefer using type inference here 
```suggestion
    try await self.runInProcessTransport(interopTestCase: .emptyUnary)
```

glbrntt (commented):
Couple of nits, looks good otherwise though

glbrntt (approved):
LGTM!

PR #1824, opened 2024-03-08
Echo Metadata in TestService

stefanadranca:
Motivation:

We want to be able to test that the server receives metadata and can send both initial and trailing metadata.

Modifications:

- added extensions for server requests that extract the initial and trailing metadata that should be echoed
- added the initial and trailing metadata n the server responses

Result:

The interop tests will test that the server sends metadata as initial and trailing.


Sources/InteroperabilityTests/TestService.swift
```diff
@@ -189,3 +207,38 @@ internal struct TestService: Grpc_Testing_TestService.ServiceProtocol {
...
+    for value in self.metadata[stringValues: "x-grpc-test-echo-initial"] {
+      initialMetadata.addString(value, forKey: "x-grpc-test-echo-initial")
+    }
+
+    for value in self.metadata[binaryValues: "x-grpc-test-echo-trailing-bin"] {
+      trailingMetadata.addBinary(value, forKey: "x-grpc-test-echo-trailing-bin")
+    }
+
+    return (initialMetadata, trailingMetadata)
+  }
```
glbrntt, line 243: There's a lot of needless repetition here. Instead can we have a `fileprivate` `extension` on `Metadata` which makes the initial and trailing metadata?

Then in the test functions we can just call `request.metadata.makeInitialAndTrailingMetadata()`?

glbrntt (changes requested):

glbrntt (approved):

PR #1827, opened 2024-03-13
Add target and generated code for QPS

stefanadranca:
Motivation:

The first step in implementing QPS testing for V2 is to add the proto files representing the services and messages used in QPS and generating the code for them.

Modifications:

- Modified the fetch script to add the proto files we need
- Modified the generate script to generate code for them
- Created executable target for QPS worker

Result:

We can start implementing the performance worker.


Protos/fetch.sh
```diff
@@ -23,7 +23,7 @@ upstream="$here/upstream"
 checkouts="$(mktemp -d)"
 
 # Clone the grpc and google protos into the staging area.
-git clone --depth 1 https://github.com/grpc/grpc-proto "$checkouts/grpc-proto"
+git clone --depth 2 https://github.com/grpc/grpc-proto "$checkouts/grpc-proto"
```
glbrntt, line 26: Why did the depth change to 2?

Protos/fetch.sh
```diff
@@ -33,12 +33,21 @@ rm -rf "$upstream"
 # rather than source repository name.
 mkdir -p "$upstream/grpc"
 mkdir -p "$upstream/google"
+mkdir -p "$upstream/grpc/testing"
+mkdir -p "$upstream/grpc/core"
```
glbrntt, line 37: nit, we no longer need to `mkdir grpc` if we make subdirectories within it:

```suggestion
mkdir -p "$upstream/google"
mkdir -p "$upstream/grpc/testing"
mkdir -p "$upstream/grpc/core"
```

Protos/generate.sh
```diff
@@ -201,6 +201,26 @@ function generate_service_messages_interop_tests {
   done
 }
 
+function generate_qps_code {
```
glbrntt, line 204: can we call this `generate_worker_service`?  The previous naming we used was qps, but I'm not sure where that actually came from.

Protos/generate.sh
```diff
@@ -201,6 +201,26 @@ function generate_service_messages_interop_tests {
...
+    "$here/upstream/grpc/testing/control.proto"
+    "$here/upstream/grpc/testing/messages.proto"
+    "$here/upstream/grpc/testing/stats.proto"
+    "$here/upstream/grpc/testing/benchmark_service.proto"
+    "$here/upstream/grpc/testing/worker_service.proto"
+  )
+
+  local output="$root/Sources/performance-worker/Generated"
+  
+  generate_message "$here/upstream/grpc/core/stats.proto" "$here/upstream" "$output" "Visibility=Internal" "FileNaming=PathToUnderscores"
```
glbrntt, line 216: here we use `"FileNaming=PathToUnderscores"` but below we use `"FileNaming=DropPath"` – why?

Protos/generate.sh
```diff
@@ -201,6 +201,26 @@ function generate_service_messages_interop_tests {
...
+
+  for proto in "${protos[@]}"; do
+    generate_message "$proto" "$here/upstream" "$output" "Visibility=Internal" "FileNaming=DropPath"
+    generate_grpc "$proto" "$here/upstream/" "$output" "Visibility=Internal" "Server=true" "_V2=true" "FileNaming=DropPath"
```
glbrntt, line 220:
```suggestion
    generate_grpc "$proto" "$here/upstream" "$output" "Visibility=Internal" "Server=true" "_V2=true" "FileNaming=DropPath"
```

Protos/generate.sh
```diff
@@ -201,6 +201,26 @@ function generate_service_messages_interop_tests {
...
+  local protos=(
+    "$here/upstream/grpc/testing/payloads.proto"
+    "$here/upstream/grpc/testing/control.proto"
+    "$here/upstream/grpc/testing/messages.proto"
+    "$here/upstream/grpc/testing/stats.proto"
+    "$here/upstream/grpc/testing/benchmark_service.proto"
+    "$here/upstream/grpc/testing/worker_service.proto"
+  )
+
+  local output="$root/Sources/performance-worker/Generated"
```
glbrntt, line 214:
```suggestion
  )
  local output="$root/Sources/performance-worker/Generated"
```

glbrntt (changes requested):

Protos/generate.sh
stefanadranca: There are 2 stat.proto file - one in grpc/core and one in grpc/testing. As all the other .proto files are located within 'grpc/testing' I was thinking we can drop the path from their names, and keep the path only for the 'stats.proto' file from the other module, to differentiate the 2 files containing generated code for the 2 'stats.proto' files.

glbrntt: Oh, I didn't realise there were two, thanks for explaining. In that case I'd rather we stick to one naming style if possible.

Protos/fetch.sh
stefanadranca: Thanks for pointing this out, I left it by mistake in the PR. (it's from something I tried but wasn't necessary)

Protos/generate.sh
```diff
@@ -201,6 +201,25 @@ function generate_service_messages_interop_tests {
...
+    "$here/upstream/grpc/testing/benchmark_service.proto"
+    "$here/upstream/grpc/testing/worker_service.proto"
+  )
+  local output="$root/Sources/performance-worker/Generated"
+  
+  generate_message "$here/upstream/grpc/core/stats.proto" "$here/upstream" "$output" "Visibility=Internal" "FileNaming=PathToUnderscores"
+
+  for proto in "${protos[@]}"; do
+    generate_message "$proto" "$here/upstream" "$output" "Visibility=Internal" "FileNaming=PathToUnderscores"
+    generate_grpc "$proto" "$here/upstream" "$output" "Visibility=Internal" "Server=true" "_V2=true" "FileNaming=PathToUnderscores"
```
glbrntt, line 219: We're actually generating client _and_ server here even though we only set `Server=true` (true is the default for both client and server).

I _think_ we only need to generate the server for `worker_service.proto` but need both for `benchmark_service.proto`.

glbrntt: (Sorry, I didn't pick up on this in the last review!)

glbrntt (approved):

PR #1828, opened 2024-03-13
QPS Benchmark Service implementation

stefanadranca:
Motivation:

The 2 workers used in QPS testing are a Benchmark service client and server respectivelly, so we need to implement the Benchmark Service.

Modifications:

Implemented the 'BenchmarkService' struct that defines the service protocol methods, based on their documentation.

Result:

We will be able to proceed with the Wrorker Service implementation.

### DIFF
--- a/Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
+++ b/Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
@@ -110,13 +110,28 @@ private enum GRPCStreamStateMachineState {
   }
 
   struct ClientOpenServerClosedState {
-    var framer: GRPCMessageFramer
+    var framer: GRPCMessageFramer?
     var compressor: Zlib.Compressor?
 
     let deframer: NIOSingleStepByteToMessageProcessor<GRPCMessageDeframer>?
     var decompressor: Zlib.Decompressor?
 
     var inboundMessageBuffer: OneOrManyQueue<[UInt8]>
+
+    // This transition should only happen on the server-side when, upon receiving
+    // initial client metadata, some of the headers are invalid and we must reject
+    // the RPC.
+    // We will mark the client as open (because it sent initial metadata albeit
+    // invalid) but we'll close the server, meaning all future messages sent from
+    // the client will be ignored. Because of this, we won't need to frame or
+    // deframe any messages, as we won't be reading or writing any messages.
+    init(previousState: ClientIdleServerIdleState) {
+      self.framer = nil
+      self.compressor = nil
+      self.deframer = nil
+      self.decompressor = nil
+      self.inboundMessageBuffer = .init()
+    }
 
     init(previousState: ClientOpenServerOpenState) {
       self.framer = previousState.framer
@@ -240,11 +255,24 @@ private enum GRPCStreamStateMachineState {
     // We still need the framer and compressor in case the server has closed
     // but its buffer is not yet empty and still needs to send messages out to
     // the client.
-    var framer: GRPCMessageFramer
+    var framer: GRPCMessageFramer?
     var compressor: Zlib.Compressor?
 
     // These are already deframed, so we don't need the deframer anymore.
     var inboundMessageBuffer: OneOrManyQueue<[UInt8]>
+
+    // This transition should only happen on the server-side when, upon receiving
+    // initial client metadata, some of the headers are invalid and we must reject
+    // the RPC.
+    // We will mark the client as closed (because it set the EOS flag, even if
+    // the initial metadata was invalid) and we'll close the server too.
+    // Because of this, we won't need to frame any messages, as we
+    // won't be writing any messages.
+    init(previousState: ClientIdleServerIdleState) {
+      self.framer = nil
+      self.compressor = nil
+      self.inboundMessageBuffer = .init()
+    }
 
     init(previousState: ClientClosedServerOpenState) {
       self.framer = previousState.framer
@@ -1062,6 +1090,21 @@ extension GRPCStreamStateMachine {
     }
   }
 
+  mutating private func closeServerAndBuildRejectRPCAction(
+    currentState: GRPCStreamStateMachineState.ClientIdleServerIdleState,
+    endStream: Bool,
+    rejectWithStatus status: Status
+  ) -> OnMetadataReceived {
+    if endStream {
+      self.state = .clientClosedServerClosed(.init(previousState: currentState))
+    } else {
+      self.state = .clientOpenServerClosed(.init(previousState: currentState))
+    }
+
+    let trailers = self.makeTrailers(status: status, customMetadata: nil, trailersOnly: true)
+    return .rejectRPC(trailers: trailers)
+  }
+
   private mutating func serverReceive(
     metadata: HPACKHeaders,
     endStream: Bool,
@@ -1071,7 +1114,9 @@ extension GRPCStreamStateMachine {
     case .clientIdleServerIdle(let state):
       let contentType = metadata.firstString(forKey: .contentType)
         .flatMap { ContentType(value: $0) }
-      guard contentType != nil else {
+      if contentType == nil {
+        self.state = .clientOpenServerClosed(.init(previousState: state))
+
         // Respond with HTTP-level Unsupported Media Type status code.
         var trailers = HPACKHeaders()
         trailers.add("415", forKey: .status)
@@ -1080,13 +1125,50 @@ extension GRPCStreamStateMachine {
 
       let path = metadata.firstString(forKey: .path)
         .flatMap { MethodDescriptor(fullyQualifiedMethod: $0) }
-      guard path != nil else {
-        let status = Status(
-          code: .unimplemented,
-          message: "No \(GRPCHTTP2Keys.path.rawValue) header has been set."
+      if path == nil {
+        return self.closeServerAndBuildRejectRPCAction(
+          currentState: state,
+          endStream: endStream,
+          rejectWithStatus: Status(
+            code: .unimplemented,
+            message: "No \(GRPCHTTP2Keys.path.rawValue) header has been set."
+          )
         )
-        let trailers = self.makeTrailers(status: status, customMetadata: nil, trailersOnly: true)
-        return .rejectRPC(trailers: trailers)
+      }
+
+      let scheme = metadata.firstString(forKey: .scheme)
+        .flatMap { Scheme(rawValue: $0) }
+      if scheme == nil {
+        return self.closeServerAndBuildRejectRPCAction(
+          currentState: state,
+          endStream: endStream,
+          rejectWithStatus: Status(
+            code: .invalidArgument,
+            message: ":scheme header must be present and one of \"http\" or \"https\"."
+          )
+        )
+      }
+
+      guard let method = metadata.firstString(forKey: .method), method == "POST" else {
+        return self.closeServerAndBuildRejectRPCAction(
+          currentState: state,
+          endStream: endStream,
+          rejectWithStatus: Status(
+            code: .invalidArgument,
+            message: ":method header is expected to be present and have a value of \"POST\"."
+          )
+        )
+      }
+
+      guard let te = metadata.firstString(forKey: .te), te == "trailers" else {
+        return self.closeServerAndBuildRejectRPCAction(
+          currentState: state,
+          endStream: endStream,
+          rejectWithStatus: Status(
+            code: .invalidArgument,
+            message: "\"te\" header is expected to be present and have a value of \"trailers\"."
+          )
+        )
       }
 
       func isIdentityOrCompatibleEncoding(_ clientEncoding: CompressionAlgorithm) -> Bool {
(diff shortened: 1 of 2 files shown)


Sources/performance-worker/BenchmarkService.swift
```diff
@@ -0,0 +1,135 @@
...
+struct BenchmarkService: Grpc_Testing_BenchmarkService.ServiceProtocol {
+  /// One request followed by one response.
+  /// The server returns the client payload as-is.
+  func unaryCall(
+    request: GRPCCore.ServerRequest.Single<Grpc_Testing_BenchmarkService.Method.UnaryCall.Input>
+  ) async throws
+    -> GRPCCore.ServerResponse.Single<Grpc_Testing_BenchmarkService.Method.UnaryCall.Output>
+  {
+
+    // Throw an error if the status is not `ok`. Otherwise, an `ok` status is automatically sent
```
glbrntt, line 31:
```suggestion
  {
    // Throw an error if the status is not `ok`. Otherwise, an `ok` status is automatically sent
```

Sources/performance-worker/BenchmarkService.swift
```diff
@@ -0,0 +1,135 @@
...
+        )
+        try await Task.sleep(nanoseconds: 10)
+      }
+    }
+  }
+}
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+extension BenchmarkService {
+  private func echoStatus(responseStatus: Grpc_Testing_EchoStatus) throws {
```
glbrntt, line 125: The naming/usage is not all that clear because it's `try echoStatus(...)` which makes it seem like it throws if it can't echo the status.

One analogue I can think of is throwing a cancellation error if a task is cancelled and that's spelled `try Task.checkCancellation()`. 

Drawing on that I think something like `try checkOkStatus(_ responseStatus: Grpc_Testing_EchoStatus)` (note the dropped label as 'status' is included in the name) has clearer semantics: it returns if the status is okay, otherwise it throws.

Sources/performance-worker/BenchmarkService.swift
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  /// Single-sided unbounded streaming from client to server
+  /// The server returns the client payload as-is once the client does WritesDone
+  func streamingFromClient(
+    request: ServerRequest.Stream<Grpc_Testing_BenchmarkService.Method.StreamingFromClient.Input>
+  ) async throws
+    -> ServerResponse.Single<Grpc_Testing_BenchmarkService.Method.StreamingFromClient.Output>
+  {
+    throw RPCError(code: .unimplemented, message: "The RPC is not implemented.")
```
glbrntt, line 73: Why don't we implement this one?

Sources/performance-worker/BenchmarkService.swift
```diff
@@ -0,0 +1,135 @@
...
+        try self.echoStatus(responseStatus: request.message.responseStatus)
+      }
+
+      while true {
+        try await writer.write(
+          Grpc_Testing_BenchmarkService.Method.StreamingCall.Output.with {
+            $0.payload = request.message.payload
+          }
+        )
+        try await Task.sleep(nanoseconds: 10)
```
glbrntt, line 94: We shouldn't need to sleep here

Sources/performance-worker/BenchmarkService.swift
```diff
@@ -0,0 +1,135 @@
...
+    request: ServerRequest.Single<Grpc_Testing_BenchmarkService.Method.StreamingFromServer.Input>
+  ) async throws
+    -> ServerResponse.Stream<Grpc_Testing_BenchmarkService.Method.StreamingFromServer.Output>
+  {
+    return ServerResponse.Stream { writer in
+      if request.message.responseStatus.isInitialized {
+        try self.echoStatus(responseStatus: request.message.responseStatus)
+      }
+
+      while true {
```
glbrntt, line 88: We need to stop eventually, `while !Task.isCancelled` is probably a better condition to use here.

Sources/performance-worker/BenchmarkService.swift
```diff
@@ -0,0 +1,135 @@
...
+  /// Both sides send the content of their own choice to the other.
+  func streamingBothWays(
+    request: GRPCCore.ServerRequest.Stream<
+      Grpc_Testing_BenchmarkService.Method.StreamingBothWays.Input
+    >
+  ) async throws
+    -> GRPCCore.ServerResponse.Stream<Grpc_Testing_BenchmarkService.Method.StreamingBothWays.Output>
+  {
+    return ServerResponse.Stream { writer in
+      while true {
```
glbrntt, line 109: Same here, we need to stop eventually

Sources/performance-worker/BenchmarkService.swift
```diff
@@ -0,0 +1,135 @@
...
+        try await writer.write(
+          Grpc_Testing_BenchmarkService.Method.StreamingCall.Output.with {
+            $0.payload = Grpc_Testing_Payload.with {
+              $0.body = Data(count: Int.random(in: 1 ..< 10))
+            }
+          }
+        )
+        try await Task.sleep(nanoseconds: 10)
```
glbrntt, line 117: No need to sleep here

Sources/performance-worker/BenchmarkService.swift
```diff
@@ -0,0 +1,135 @@
...
+    // Throw an error if the status is not `ok`. Otherwise, an `ok` status is automatically sent
+    // if the request is successful.
+    if request.message.responseStatus.isInitialized {
+      try self.echoStatus(responseStatus: request.message.responseStatus)
+    }
+
+    return ServerResponse.Single(
+      message: Grpc_Testing_BenchmarkService.Method.UnaryCall.Output.with {
+        $0.payload = request.message.payload
+      }
```
glbrntt, line 40: This isn't quite right: the request tells us how big the response payload should be (via `responeSize`). Same goes for all the other RPCs.

Sources/performance-worker/BenchmarkService.swift
```diff
@@ -0,0 +1,135 @@
...
+    }
+  }
+}
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+extension BenchmarkService {
+  private func echoStatus(responseStatus: Grpc_Testing_EchoStatus) throws {
+    guard let code = Status.Code(rawValue: Int(responseStatus.code))
+    else {
```
glbrntt, line 127:
```suggestion
    guard let code = Status.Code(rawValue: Int(responseStatus.code)) else {
```

Sources/performance-worker/BenchmarkService.swift
```diff
@@ -0,0 +1,135 @@
...
+      throw RPCError(code: .invalidArgument, message: "The response status code is invalid.")
+    }
+    let status = Status(code: code, message: responseStatus.message)
+    if let error = RPCError(status: status) {
+      throw error
+    }
```
glbrntt, line 133: nit: we always create the `Status` but don't always use it (in fact we don't need to use it at all):

```swift
if let code = RPCError.Code(code) {
  throw RPCError(code: code, message: "...")
}
```

glbrntt (changes requested):

Sources/performance-worker/BenchmarkService.swift
stefanadranca: is this also true for `streamingBothWays()`? The documentation for it says "Both sides send the content of their own choice to the other."

glbrntt: Yes, I believe so, I think the docs are outdated because the scenario descriptions which drive the tests expect different request/response sizes.

stefanadranca: Wouldn't making this change mean that `streamingCall()` and `streamingBothWays()` become the same function?

Sources/performance-worker/BenchmarkService.swift
```diff
@@ -0,0 +1,161 @@
...
+    return ServerResponse.Stream { writer in
+      if request.message.responseStatus.isInitialized {
+        try self.checkOkStatus(request.message.responseStatus)
+      }
+
+      while working.load(ordering: .acquiring) {
+        try await writer.write(
+          Grpc_Testing_BenchmarkService.Method.StreamingCall.Output.with {
+            $0.payload = Grpc_Testing_Payload.with {
+              $0.body = Data(count: Int(request.message.responseSize))
```
glbrntt, line 113: This will create the message each time which is potentially expensive. Can we create it once before the loop?

Sources/performance-worker/BenchmarkService.swift
```diff
@@ -0,0 +1,161 @@
...
+      if message.responseStatus.isInitialized {
+        try self.checkOkStatus(message.responseStatus)
+      }
+    }
+    return ServerResponse.Stream { writer in
+      while working.load(ordering: .acquiring) {
+        try await writer.write(
+          Grpc_Testing_BenchmarkService.Method.StreamingCall.Output.with {
+            $0.payload = Grpc_Testing_Payload.with {
+              $0.body = Data(count: 100)
```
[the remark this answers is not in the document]
stefanadranca: Sorry, was there a specific reason we used exactly 100? (We are using a single value so we can create the message once - as you pointed before I will create it before the loop, but was there a reason for this size specifically?)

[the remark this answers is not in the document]
glbrntt: https://github.com/grpc/grpc-java/blob/36e9f0dfacc853b5d54f64bb14305f0b4c323589/benchmarks/src/main/java/io/grpc/benchmarks/qps/AsyncServer.java#L192-L199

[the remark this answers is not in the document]
stefanadranca: so because the java implementation does this? my understanding of the comment is that they are using the same size for all responses because the spec allows this, but not that the spec says it should be 100. Am I missing something?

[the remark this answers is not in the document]
glbrntt: You're not missing anything. We're doing it because Java (and other impls) are using 100, i.e. there is precedent elsewhere and we haven't just picked 100 out of thin air (someone else did, we're just following).

Sources/performance-worker/BenchmarkService.swift
```diff
@@ -0,0 +1,161 @@
...
+      $0.payload = Grpc_Testing_Payload.with {
+        $0.body = Data(count: 100)
+      }
+    }
+    return ServerResponse.Stream { writer in
+      while working.load(ordering: .acquiring) {
+        try await writer.write(response)
+      }
+      return [:]
+    }
```
glbrntt: We need to read and write _at the same time_ which means we need to do the work in a task group within the body of the response stream. We can have one task for reading and one for writing. You can use an atomic bool as the stopping condition for writing (we should stop writing when all requests have been consumed).

Sources/performance-worker/BenchmarkService.swift
```diff
@@ -0,0 +1,185 @@
...
+    try await withThrowingTaskGroup(
+      of: ServerResponse.Stream<Grpc_Testing_BenchmarkService.Method.StreamingBothWays.Output>?.self
+    ) { group in
+      group.addTask {
+        for try await message in request.messages {
+          if message.responseStatus.isInitialized {
+            try self.checkOkStatus(message.responseStatus)
+          }
+        }
+        _ = self.working.exchange(false, ordering: .acquiring)
```
glbrntt, line 151: I don't think this is quite right: `self.working` will be toggled by the quit method on the worker service, it's an external trigger to shutdown. Looking at it in a different way: I don't think it makes sense that the end of the request stream for this method would stop the response stream from streaming-from-server. 

Instead I think we want a separate atomic to signal between the request stream and response stream.

Sources/performance-worker/BenchmarkService.swift
```diff
@@ -0,0 +1,185 @@
...
+      for try await result in group {
+        guard let result = result else {
+          continue
+        }
+        serverResponse = result
+        group.cancelAll()
+      }
+      group.cancelAll()
+    }
+    return serverResponse
```
glbrntt, line 171: This is a bit inverted. At the moment the behaviour will be:
- RPC starts
- Task group is created
  - One task runs to consume the request messages
  - The other task returns the response stream immediately
- We then consume the response stream, the first result we'll get will be the server response because it's returned immediately so we cancel the task group (and stop consuming requests)
- We then wait for all child tasks in the task group to finish (this should be fast because one has already finished and the other has been cancelled)
- Then we return the server response so that gRPC can run it.

Instead we should be consuming the request stream within a task group _inside_ the response stream. The rough shape is something like this:

```swift
return ServerResponse.Stream { writer in 
    group.addTask { 
      // consume requests 
    } 
    group.addTask { 
      // write responses 
    }
  }
  return [:]
}
```

glbrntt (changes requested):

Sources/performance-worker/BenchmarkService.swift
stefanadranca: Thank you! It makes sense to use a different atomic for this. One question though: I should still check when writing the responses if 'working' is set to 'true', besides checking the new atomic, right?

glbrntt (changes requested):
one nit but looks good otherwise

glbrntt (approved):

PR #1832, opened 2024-03-15
Add GRPCServerStreamHandler

gjcairo:
## Motivation
We need a channel handler to drive request reads and response writes on the server for the HTTP2 transport.

## Modifications
Add a GRPCServerStreamHandler.

## Result
One step closer to having a working HTTP2 transport implementation.

### DIFF
--- a/Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
+++ b/Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
@@ -1,0 +1,174 @@
+/*
+ * Copyright 2024, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import GRPCCore
+import NIOCore
+import NIOHTTP2
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+final class GRPCServerStreamHandler: ChannelDuplexHandler {
+  typealias InboundIn = HTTP2Frame.FramePayload
+  typealias InboundOut = RPCRequestPart
+
+  typealias OutboundIn = RPCResponsePart
+  typealias OutboundOut = HTTP2Frame.FramePayload
+
+  private var stateMachine: GRPCStreamStateMachine
+
+  private var isReading = false
+  private var flushPending = false
+
+  init(
+    scheme: Scheme,
+    acceptedEncodings: [CompressionAlgorithm],
+    maximumPayloadSize: Int
+  ) {
+    self.stateMachine = .init(
+      configuration: .server(.init(scheme: scheme, acceptedEncodings: acceptedEncodings)),
+      maximumPayloadSize: maximumPayloadSize
+    )
+  }
+
+  func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+    self.isReading = true
+    let frame = self.unwrapInboundIn(data)
+    switch frame {
+    case .data(let frameData):
+      let endStream = frameData.endStream
+      switch frameData.data {
+      case .byteBuffer(let buffer):
+        do {
+          try self.stateMachine.receive(message: buffer, endStream: endStream)
+          switch self.stateMachine.nextInboundMessage() {
+          case .awaitMoreMessages:
+            ()
+          case .receiveMessage(let message):
+            context.fireChannelRead(self.wrapInboundOut(.message(message)))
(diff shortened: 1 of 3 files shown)


Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
```diff
@@ -0,0 +1,176 @@
...
+            code: .internalError,
+            message: "Server cannot get receivedStatusAndMetadata."
+          )
+        case .doNothing:
+          throw RPCError(code: .internalError, message: "Server cannot receive doNothing.")
+        }
+      } catch {
+        context.fireErrorCaught(error)
+      }
+    case .ping, .goAway, .priority, .rstStream, .settings, .pushPromise, .windowUpdate,
```
glbrntt, line 95: Can you add a test which verifies that these frames are ignored?

Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
```diff
@@ -0,0 +1,176 @@
...
+      case .byteBuffer(let buffer):
+        do {
+          try self.stateMachine.receive(message: buffer, endStream: endStream)
+          switch self.stateMachine.nextInboundMessage() {
+          case .awaitMoreMessages:
+            ()
+          case .receiveMessage(let message):
+            context.fireChannelRead(self.wrapInboundOut(.message(message)))
+          case .noMoreMessages:
+            context.channel.close(mode: .input, promise: nil)
```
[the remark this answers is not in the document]
gjcairo: I just checked and it looks like we ignore the mode: https://github.com/apple/swift-nio-http2/blob/a0c580be160af313d1d7d3bf5dda874b9abd025b/Sources/NIOHTTP2/HTTP2StreamChannel.swift#L583
Should I file an issue?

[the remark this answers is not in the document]
glbrntt: Please!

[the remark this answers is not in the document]
gjcairo: https://github.com/apple/swift-nio-http2/issues/437

Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
```diff
@@ -0,0 +1,192 @@
...
+      }
+
+    case .message(let message):
+      do {
+        try self.stateMachine.send(message: message, endStream: false)
+        self.flushIfNeeded(context)
+        // TODO: move the promise handling into the state machine
+        promise?.succeed()
+      } catch {
+        context.fireErrorCaught(error)
```
gjcairo, line 152: I see that this sends the error to the next inbound handler, but we're on the outbound side here. Should I throw the error when on the outbound side instead?

glbrntt: No, firing the error and failing the promise is the right thing to do here

Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
```diff
@@ -0,0 +1,218 @@
...
+        if let pendingTrailers = self.pendingTrailers {
+          context.write(self.wrapOutboundOut(pendingTrailers), promise: nil)
+          self.flushPending = true
+          self.pendingTrailers = nil
+        }
+      case .awaitMoreMessages:
+        ()
+      case .sendMessage:
+        preconditionFailure("This isn't possible: we'd still be inside the while loop.")
+      }
```
glbrntt, line 208: Similar to above, we can simplify this and avoid the `preconditionFailure`:

```swift
loop: while true {
  switch try self.stateMachine.nextOutboundMessage() {
  case .noMoreMessages:
    // ...
    break loop

  case .awaitMoreMessages:
    break loop

  case .sendMessage(let buffer):
    // ...
  }
}
```

Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
```diff
@@ -0,0 +1,218 @@
...
+      } catch {
+        context.fireErrorCaught(error)
+        // TODO: move the promise handling into the state machine
+        promise?.fail(error)
+      }
+    }
+  }
+
+  func flush(context: ChannelHandlerContext) {
+    do {
```
glbrntt, line 181: Can you scope the do-catch to a small block? The `self.isReading` can come before the do-catch.

Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
```diff
@@ -0,0 +1,218 @@
...
+        }
+      case .awaitMoreMessages:
+        ()
+      case .sendMessage:
+        preconditionFailure("This isn't possible: we'd still be inside the while loop.")
+      }
+
+      if self.flushPending {
+        context.flush()
+        self.flushPending = false
```
glbrntt, line 212: General rule of thumb: always update state _before_ making an outcall. The outcall may re-entrantly call this handler which could update state in some other way, which would then be trampled over when the stack unwinds.

```suggestion
        self.flushPending = false
        context.flush()
```

glbrntt (changes requested):
couple of nits

glbrntt (approved):

PR #1833, opened 2024-03-18
Implement 'quit' and 'core count' RPCs on the worker service

stefanadranca:
Motivation:

These are 2 of the RPCs on the worker service that we need for benchmarking.

Modifications:

- Created the WorkerService struct that has a GRPCClient and a GRPCServer property
- Implemented the 'coreCount' and 'quitWorker' RPCs

Result:

- The driver will be able to request 'quitWorker' and 'coreCount' from the workers

### DIFF
--- a/Sources/performance-worker/WorkerService.swift
+++ b/Sources/performance-worker/WorkerService.swift
@@ -15,20 +15,54 @@
  */
 
 import GRPCCore
+import NIOConcurrencyHelpers
 import NIOCore
 
 @available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
-struct WorkerService: Grpc_Testing_WorkerService.ServiceProtocol {
-  var testClient: GRPCClient? = nil
-  var testServer: GRPCServer? = nil
+final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable {
+  let state: NIOLockedValueBox<State>
+
+  init() {
+    let clientAndServer = State()
+    self.state = NIOLockedValueBox(clientAndServer)
+  }
+
+  struct State {
+    var role: Role?
+
+    enum Role {
+      case client(GRPCClient)
+      case server(GRPCServer)
+    }
+
+    init() {}
+
+    init(role: Role) {
+      self.role = role
+    }
+
+    init(server: GRPCServer) {
+      self.role = .server(server)
+    }
+
+    init(client: GRPCClient) {
+      self.role = .client(client)
+    }
+  }
 
   func quitWorker(
     request: ServerRequest.Single<Grpc_Testing_WorkerService.Method.QuitWorker.Input>
   ) async throws -> ServerResponse.Single<Grpc_Testing_WorkerService.Method.QuitWorker.Output> {
-    if let testClient = self.testClient {
-      testClient.close()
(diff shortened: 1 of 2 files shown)


Sources/performance-worker/WorkerService.swift
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+
+import GRPCCore
+import NIOCore
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+struct WorkerService: Grpc_Testing_WorkerService.ServiceProtocol {
+  var testClient: GRPCClient? = nil
+  var testServer: GRPCServer? = nil
```
glbrntt, line 23: This won't work when we add other RPCs because we need this type to be `Sendable`. This means we need to turn the `struct` into a `class` and protect this state using a lock. You can wrap the state up in a `struct State` and then use a `NIOLockedValueBox` to hold the state (you'll need to import `NIOConcurrencyHelpers`).

Sources/performance-worker/WorkerService.swift
```diff
@@ -0,0 +1,62 @@
...
+
+  func quitWorker(
+    request: ServerRequest.Single<Grpc_Testing_WorkerService.Method.QuitWorker.Input>
+  ) async throws -> ServerResponse.Single<Grpc_Testing_WorkerService.Method.QuitWorker.Output> {
+    if let testClient = self.testClient {
+      testClient.close()
+    } else if let testServer = self.testServer {
+      testServer.stopListening()
```
glbrntt, line 31: We will also need to flip the atomic that we'll set on the benchmark service here.

glbrntt (changes requested):

Sources/performance-worker/WorkerService.swift
```diff
@@ -0,0 +1,96 @@
...
+ * limitations under the License.
+ */
+
+import GRPCCore
+import NIOConcurrencyHelpers
+import NIOCore
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable {
+  let state: NIOLockedValueBox<State>
```
glbrntt, line 23: This, and the `State` struct can be private

Sources/performance-worker/WorkerService.swift
```diff
@@ -0,0 +1,96 @@
...
+    init(client: GRPCClient) {
+      self.role = .client(client)
+    }
+  }
+
+  func quitWorker(
+    request: ServerRequest.Single<Grpc_Testing_WorkerService.Method.QuitWorker.Input>
+  ) async throws -> ServerResponse.Single<Grpc_Testing_WorkerService.Method.QuitWorker.Output> {
+
+    if let role = self.state.withLockedValue({ $0.role }) {
```
glbrntt, line 57: I don't think there's any reason to delay setting the `$0.role` back to `nil`, we can just do it in one go:

```swift
let role = self.state.withLockedValue { 
  defer { $0.role = nil }
  return $0.role
}
```

glbrntt (changes requested):
two nits, looks good otherwise

glbrntt (approved):

PR #1834, opened 2024-03-20
Server Stats Collection

stefanadranca:
Motivation:

In order to implement the 'runServer' RPC on the WorkerService we need to capture a snapshot with the resorces usage for the server and compute the difference between 2 of these snapshots.

Modifications:

- Created the ServerStats struct that contains all stats needed in the QPS benchmarking from the server
- Implemented the init() that collects all these stats
- Implemented the function that computes the differences between 2 snapshots of ServerStats

Result:

We will be able to implement the `runServer` RPC.

### DIFF
--- a/Sources/performance-worker/ServerStats.swift
+++ b/Sources/performance-worker/ServerStats.swift
@@ -18,40 +18,42 @@ import Dispatch
 import NIOCore
 import NIOFileSystem
 
-#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+#if canImport(Darwin)
 import Darwin
-#elseif os(Linux) || os(FreeBSD) || os(Android)
+#elseif canImport(Musl)
+import Musl
+#elseif canImport(Glibc)
 import Glibc
 #else
 let badOS = { fatalError("unsupported OS") }()
 #endif
 
-#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+#if canImport(Darwin)
 private let OUR_RUSAGE_SELF: Int32 = RUSAGE_SELF
-#elseif os(Linux) || os(FreeBSD) || os(Android)
+#elseif canImport(Musl) || canImport(Glibc)
 private let OUR_RUSAGE_SELF: Int32 = RUSAGE_SELF.rawValue
 #endif
 
 /// Current server stats.
 internal struct ServerStats: Sendable {
-  let time: Double
-  let userTime: Double
-  let systemTime: Double
-  let totalCpuTime: UInt64
-  let idleCpuTime: UInt64
+  var time: Double
+  var userTime: Double
+  var systemTime: Double
+  var totalCPUTime: UInt64
+  var idleCPUTime: UInt64
 
   init(
     time: Double,
     userTime: Double,
     systemTime: Double,
-    totalCpuTime: UInt64,
-    idleCPuTime: UInt64
+    totalCPUTime: UInt64,
+    idleCPUTime: UInt64
   ) {
     self.time = time
     self.userTime = userTime
     self.systemTime = systemTime
-    self.totalCpuTime = totalCpuTime
-    self.idleCpuTime = idleCPuTime
+    self.totalCPUTime = totalCPUTime
+    self.idleCPUTime = idleCPUTime
   }
 
   init() async throws {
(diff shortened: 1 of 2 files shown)


Sources/performance-worker/ServerStats.swift
```diff
@@ -0,0 +1,131 @@
...
+  let systemTime: Double
+  let totalCpuTime: UInt64
+  let idleCpuTime: UInt64
+
+  init(
+    time: Double,
+    userTime: Double,
+    systemTime: Double,
+    totalCpuTime: UInt64,
+    idleCPuTime: UInt64
```
glbrntt, line 48: nit: should be `CPU`

Sources/performance-worker/ServerStats.swift
```diff
@@ -0,0 +1,131 @@
...
+
+internal func changeInStats(initialStats: ServerStats, currentStats: ServerStats) -> ServerStats {
+  return ServerStats(
+    time: currentStats.time - initialStats.time,
+    userTime: currentStats.userTime - initialStats.userTime,
+    systemTime: currentStats.systemTime - initialStats.systemTime,
+    totalCpuTime: currentStats.totalCpuTime - initialStats.totalCpuTime,
+    idleCPuTime: currentStats.idleCpuTime - initialStats.idleCpuTime
+  )
+}
```
glbrntt, line 88: I think it'd make more sense to have this as a function on `ServerStats` instead, e.g. `func difference(to stats: ServerStats) -> ServerStats`

Sources/performance-worker/ServerStats.swift
```diff
@@ -0,0 +1,131 @@
...
+/// The totalCpuTime is computed as follows:
+/// total = user + nice + system + idle
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+private func getTotalAndIdleCpuTime() async throws -> (UInt64, UInt64) {
+  #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS) || os(Linux) || os(Android)
+  do {
+    let contents = try await ByteBuffer(
+      contentsOf: "/proc/stat",
+      maximumSizeAllowed: .kilobytes(20)
+    )
```
glbrntt, line 103: This is the only part that throws so we should minimise the do-catch to this

Sources/performance-worker/ServerStats.swift
```diff
@@ -0,0 +1,131 @@
...
+
+    guard let index = contents.readableBytesView.firstIndex(where: { $0 == UInt8("\n") }) else {
```
glbrntt, line 105: You can use `firstIndex(of: UInt8(ascii: "\n"))` instead of `firstIndex(where:)` in this case.

Sources/performance-worker/ServerStats.swift
```diff
@@ -0,0 +1,131 @@
...
+      return (0, 0)
+    }
+
+    guard let firstLine = contents.getString(at: 0, length: index) else {
```
glbrntt, line 109: This isn't quite right because the buffer isn't necessarily zero-indexed. The view has a start index so you could use that and then compute the right length, however, the preferred way is to create a slice of the buffer view, then turn that back into a buffer and then convert it to a `String`:

```swift
let view = contents.readableBytesView
guard let index = view.firstIndex(of: UInt8(ascii: "\n")) else { 
  return (0, 0)
}

let line = String(buffer: ByteBuffer(view))

```

Sources/performance-worker/ServerStats.swift
```diff
@@ -0,0 +1,131 @@
...
+}
+
+/// Computes the total and idle CPU time after extracting stats from the first line of '/proc/stat'.
+///
+/// The first line in '/proc/stat' file looks as follows:
+/// cpu [user] [nice] [system] [idle] [iowait] [irq] [softirq]
+/// The totalCpuTime is computed as follows:
+/// total = user + nice + system + idle
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+private func getTotalAndIdleCpuTime() async throws -> (UInt64, UInt64) {
```
glbrntt, line 97: What is each value in the return, can you add labels to the tuple?

glbrntt, line 97: Generally avoid global functions, this can be a `private static func` on `ServerStats` instead

glbrntt (changes requested):

Sources/performance-worker/ServerStats.swift
```diff
@@ -0,0 +1,135 @@
...
+    if getrusage(OUR_RUSAGE_SELF, &usage) == 0 {
+      // Adding the seconds with the microseconds transformed into seconds to get the
+      // real number of seconds as a `Double`.
+      self.userTime = Double(usage.ru_utime.tv_sec) + Double(usage.ru_utime.tv_usec) * 1e-6
+      self.systemTime = Double(usage.ru_stime.tv_sec) + Double(usage.ru_stime.tv_usec) * 1e-6
+    } else {
+      self.userTime = 0
+      self.systemTime = 0
+    }
+    if #available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *) {
```
glbrntt, line 71: Can we bubble up this availability and just put it on the type definition? These guards are required for Swift concurrency but the whole performance worker needs that, so we can just bubble it up to the top-level.

Sources/performance-worker/ServerStats.swift
```diff
@@ -0,0 +1,135 @@
...
+    )
+  }
+
+  /// Computes the total and idle CPU time after extracting stats from the first line of '/proc/stat'.
+  ///
+  /// The first line in '/proc/stat' file looks as follows:
+  /// CPU [user] [nice] [system] [idle] [iowait] [irq] [softirq]
+  /// The totalCPUTime is computed as follows:
+  /// total = user + nice + system + idle
+  @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
```
glbrntt, line 97: We can remove this is the availability goes on the type

glbrntt (changes requested):
Two small things but looks great otherwise

glbrntt (approved):
Great, thanks Stefana!

PR #1835, opened 2024-03-21
Add reflection support to plugin

samisuteria:
Tested by running `grpcurl -plaintext localhost:1234 list`.


glbrntt (commented):
Thanks for taking the time to look at this and opening a PR. However, your PR has a number of issues. The biggest of which is related to how the data is generated needs some more thought / discussion.

As proposed, the reflection data is generated for each service and is the data for the whole file containing that service. 

The issue with this approach is that it is incomplete: messages can be declared in different proto files.

Another issue is duplication: a file may contain many services, each of which would contain the same generated data. This also creates a surprising API: if a users adds `FooService` to their list of reflected services, then they may expose `BarService` without realising.

samisuteria:
Ah I see. I was assuming one service per proto file but that is obviously not a restriction of the specification. 

Currently if `ReflectionData=true` is passed into the protoc plugin, a new files `<file>.grpc.reflection` is created. When I added an option in the swift plugin to pass this flag to the protoc plugin, the `<file>.grpc.reflection` was created in the plugins output folder (DerivedData or .build). This `.reflection` file is not accessible during runtime 

What do you think about this approach: For each invocation/file with `reflection: true` a <file>.grpc.reflection.swift` is generated with contents like: 
```
import GRPC

public struct <unique-file-name>: ReflectionDataProvider {
    public static let reflectionData: String = "<reflection-data>"
}
```. 

I wondering what the downsides of putting that much data into a string would be and if there are other approaches for reading the `.reflection` file directly when using a plugin.

**Comment by glbrntt** [comment]
> Ah I see. I was assuming one service per proto file but that is obviously not a restriction of the specification.
> 
> Currently if `ReflectionData=true` is passed into the protoc plugin, a new files `<file>.grpc.reflection` is created. When I added an option in the swift plugin to pass this flag to the protoc plugin, the `<file>.grpc.reflection` was created in the plugins output folder (DerivedData or .build). This `.reflection` file is not accessible during runtime
> 
> What do you think about this approach: For each invocation/file with `reflection: true` a .grpc.reflection.swift` is generated with contents like:
> 
> ```
> import GRPC
> 
> public struct <unique-file-name>: ReflectionDataProvider {
>     public static let reflectionData: String = "<reflection-data>"
> }
> ```

If you create one Swift file per input file then you need a way to find all of those inputs at runtime. If they are generated as part of Swift structs then users would need to find and add each provider manually (i.e. user is responsible for finding all dependencies of a service), so from an ergonomics point of view I don't think it's a good solution. This is much less of an issue with the current implementation because it's easy to list all files with a given extension or in a given directory.

**Comment by samisuteria** [comment]
I agree - the ergonomics aren't great. I explored some other solutions for this. If there was a way to have some file such as:
```
import GRPC

enum GRPCReflectionProviders: String, CaseIterable {}
```

then for every invocation and each file that reflection is enabled for another file is created:
```
extension GRPCReflectionProviders {
   case <file_name> = <reflection_data>
}
```

but Swift does not allow for extensions on enums in this way.

Another approach is to have a file thats: 
```
import GRPC

struct GRPCReflectionData {
   static let fileToSerializedProtos: [String: String] = [
        "<file_name>": "<reflection_data>"
   ]
}
```

but each invocation of the plugin could potentially generate this file. 

My next idea was to flatMap all the invocations with reflections enabled and generate the above file for the protofiles that want reflection enabled.

**Comment by glbrntt** [comment]
> I agree - the ergonomics aren't great. I explored some other solutions for this. 

In terms or ergonomics, I think your original proposal is good (i.e. registering a service directly with the reflection service). It's very clear to the user. 

The issue is that it needs to include all transitive dependencies in there too. I would probably take that as a starting point and work out how to make that possible.

**Comment by samisuteria** [comment]
I tried a different approach and would like your thoughts on it. I've just added the generated reflection file as an output file for the plugin. This way the same files are generated when you use the plugin or protoc directly. By including the `<file>.grpc.reflection` as an output file, the file is now included in the module Bundle. I feel like the ergonomics of this are pretty clean and match how a user would use this if they were calling protoc directly and committing the generating files.

As far as CI is concerned, after some research it seems like installing protoc via `uses: arduino/setup-protoc@v3` doesn't actually install it in the containers that are being used for the workflow. I'm not sure how this could be fixed other than by having protoc as part of the published swift docker images.

**Comment by glbrntt** [comment]
> I tried a different approach and would like your thoughts on it. I've just added the generated reflection file as an output file for the plugin. This way the same files are generated when you use the plugin or protoc directly. By including the `<file>.grpc.reflection` as an output file, the file is now included in the module Bundle. I feel like the ergonomics of this are pretty clean and match how a user would use this if they were calling protoc directly and committing the generating files.

Oh, this is great! I didn't realise non-Swift outputs can be included in the `Bundle`.

> As far as CI is concerned, I'm not sure if installing protoc via apt is the best approach or if that should be included in a base image from somewhere else.

This is a bit of an annoying one: we don't want to add a build target which relies on the plugin because it'll force everyone who tries to build it locally to have `protoc`, so I don't think we should put that in CI (or have a new example for that matter). Ideally we would update the docs for the reflection service to reflect how to generate the data using the plugin too (https://github.com/grpc/grpc-swift/blob/main/Sources/GRPCReflectionService/Documentation.docc/ReflectionServiceTutorial.md).

File: Plugins/GRPCSwiftPlugin/plugin.swift [hunk 1]
```diff
@@ -205,6 +211,12 @@ struct GRPCSwiftPlugin {
 
       // Add the outputPath as an output file
       outputFiles.append(protobufOutputPath)
+
+        if invocation.reflection == true {
```
glbrntt, line 215: No need to compare against `true`, just `if invocation.reflection { ... }`

Plugins/GRPCSwiftPlugin/plugin.swift
```diff
@@ -205,6 +211,12 @@ struct GRPCSwiftPlugin {
...
+            let reflectionFile = file.replacingOccurrences(of: "grpc.swift", with: "grpc.reflection")
```
glbrntt, line 216: The code above deliberately doesn't use this pattern because it's not necessarily correct. The filename could be "grpc.swift.foo.bar.grpc.swift" and this would incorrectly rename it to "grpc.reflection.foo.bar.grpc.reflection".

Plugins/GRPCSwiftPlugin/plugin.swift
```diff
@@ -205,6 +211,12 @@ struct GRPCSwiftPlugin {
...
+            let reflectionOutputPath = outputDirectory.appending(reflectionFile)
+            outputFiles.append(reflectionOutputPath)
+        }
```
glbrntt, line 219: The indentation is off for this block, it looks like our formatting checker script wasn't updated to include "Plugins" directory, it should be after https://github.com/grpc/grpc-swift/pull/1845

glbrntt (changes requested):
The overall shape of this is good but we need to remove the example and CI changes unfortunately, see https://github.com/grpc/grpc-swift/pull/1835#issuecomment-2025403341.

Plugins/GRPCSwiftPlugin/plugin.swift
```diff
@@ -65,6 +65,8 @@ struct GRPCSwiftPlugin {
       var server: Bool?
       /// Whether client code is generated.
       var client: Bool?
+      /// Whether reflection data is generated.
+      var reflection: Bool?
```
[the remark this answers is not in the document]
samisuteria: done

Plugins/GRPCSwiftPlugin/plugin.swift
samisuteria: Changed the logic here so drop the last `swift` and add `reflection` similar to how `proto` is dropped and `swift is appended.

Plugins/GRPCSwiftPlugin/plugin.swift
samisuteria: Updated

samisuteria:
Pushed some commits responding to the comments above and added some comments to existing documentation.

Sources/GRPCReflectionService/Documentation.docc/ReflectionServiceTutorial.md
```diff
@@ -70,6 +70,25 @@ You can use Swift Package Manager [resources][swiftpm-resources] to add the gene
 In our example the reflection data is written into the "Generated" directory within the target
 so we include the `.copy("Generated")` rule in our target's resource list.
 
+#### SPM Plugin
```
glbrntt, line 73:
```suggestion
#### Swift Package Manager Plugin
```

Sources/GRPCReflectionService/Documentation.docc/ReflectionServiceTutorial.md
```diff
@@ -70,6 +70,25 @@ You can use Swift Package Manager [resources][swiftpm-resources] to add the gene
...
+
+Reflection data can also be generated via the SPM plugin by including `"reflectionData": true` in `grpc-swift-config.json`. This will generate the same data as running `protoc` above. The generated data will be in your `.build` folder (or `DerivedData` if running from Xcode) and included in the build process. More about [spm-plugin][spm-plugin] can be found here.
```
glbrntt, line 75: > The generated data will be in your `.build` folder (or `DerivedData` if running from Xcode) and included in the build process.

We should avoid wording it like this: the location isn't stable and users shouldn't search out these files. Instead we should tell them how to access them via the `Bundle`.

glbrntt (changes requested):
Doc nit, looks good otherwise!

Sources/GRPCReflectionService/Documentation.docc/ReflectionServiceTutorial.md
samisuteria: Updated docs to refer to Bundle

glbrntt (approved):
Great, thanks @samisuteria!

PR #1838, opened 2024-03-22
Add GRPCClientStreamHandler

gjcairo:
This PR adds a `GRPCClientStreamHandler`, which brings us a step closer to having a working HTTP2 transport implementation.

### DIFF
--- a/Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
+++ b/Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
@@ -60,7 +60,7 @@ final class GRPCClientStreamHandler: ChannelDuplexHandler {
       switch frameData.data {
       case .byteBuffer(let buffer):
         do {
-          try self.stateMachine.receive(message: buffer, endStream: endStream)
+          try self.stateMachine.receive(buffer: buffer, endStream: endStream)
           switch self.stateMachine.nextInboundMessage() {
           case .awaitMoreMessages:
             ()
@@ -79,7 +79,7 @@ final class GRPCClientStreamHandler: ChannelDuplexHandler {
     case .headers(let headers):
       do {
         let action = try self.stateMachine.receive(
-          metadata: headers.headers,
+          headers: headers.headers,
           endStream: headers.endStream
         )
         switch action {
--- a/Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
+++ b/Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
@@ -67,7 +67,7 @@ extension GRPCClientStreamHandler {
       switch frameData.data {
       case .byteBuffer(let buffer):
         do {
-          try self.stateMachine.receive(message: buffer, endStream: endStream)
+          try self.stateMachine.receive(buffer: buffer, endStream: endStream)
           loop: while true {
             switch self.stateMachine.nextInboundMessage() {
             case .receiveMessage(let message):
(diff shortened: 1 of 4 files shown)


Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
```diff
@@ -0,0 +1,234 @@
...
+        promise?.fail(error)
+      }
+    }
+  }
+
+  func close(context: ChannelHandlerContext, mode: CloseMode, promise: EventLoopPromise<Void>?) {
+    switch mode {
+    case .output:
+      do {
+        try self.stateMachine.closeOutbound()
```
glbrntt, line 172: If we close the outbound we still need to flush out the outbound messages (and then forward the close event)

Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
```diff
@@ -0,0 +1,234 @@
...
+              HTTP2Frame.FramePayload.data(
+                .init(
+                  data: .byteBuffer(.init()),
+                  endStream: true
+                )
+              )
+            ),
+            promise: nil
+          )
+          context.close(mode: .output, promise: nil)
```
glbrntt, line 218: Not sure about this, we could end up double closing if `close(context:mode:promise:)` is called with `mode: .all` because we'd call through to `_flush` here, close the output, and then forward the close all in `close(context:mode:promise:)`. I think just closing in close is okay.

Tests/GRPCHTTP2CoreTests/Client/GRPCClientStreamHandlerTests.swift
```diff
@@ -0,0 +1,724 @@
...
+      outboundEncoding: .identity,
+      acceptedEncodings: [],
+      maximumPayloadSize: 1,
+      skipStateMachineAssertions: true
+    )
+
+    let channel = EmbeddedChannel(handler: handler)
+
+    // Send client's initial metadata
+    let request = RPCRequestPart.metadata(Metadata(headers: [:]))
```
glbrntt, line 72: nit, seems weird to use the headers init here (and elsewhere), let's just init the `Metadata` directly:

```suggestion
    let request = RPCRequestPart.metadata([:])
```

glbrntt (changes requested):
This looks great but I don't think the logic in close is quite right.

Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
gjcairo: For mode `output`, this currently happens: calling `closeOutbound` on the state machine still allows the queued messages to be drained (when calling `nextOutboundMessage()` in `_flush`. We don't actually call `context.close(mode: .output)` until we have drained all queued messages in `_flush`.
I was thinking in this case we just let the usual flow happen (of flushing when read complete is fired), but we could instead force a flush by calling `_flush` like I did for the other two close modes - it's probably more consistent.

Now, for `all`, I think I should also call `closeOutbound()` on the state machine, which I'm not currently doing. This would cause issues with the current implementation of `_flush`, but I think it can be fixed by delaying the forwarding until all messages have been flushed, just as we're doing with `output`.

Let me make and push these changes to see if I'm missing anything.

glbrntt (changes requested):

glbrntt (approved):
Looks good modulo one question/nit

Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
```diff
@@ -0,0 +1,237 @@
...
+      }
+
+    case .input:
+      context.close(mode: .input, promise: promise)
+    }
+  }
+
+  func flush(context: ChannelHandlerContext) {
+    if self.isReading {
+      // We don't want to flush yet if we're still in a read loop.
```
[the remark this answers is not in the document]
gjcairo: Yes, you're right

PR #1839, opened 2024-03-22
Add GRPCServerFlushNotificationHandler

gjcairo:
The `ServerConnectionManagementHandler` relies on knowing when flushes happen but lives after the H2 handler in the pipeline, so doesn’t get notified when a stream channel causes the parent channel to flush.

`GRPCServerFlushNotificationHandler`'s sole purpose is to notify the `ServerConnectionManagementHandler` when a flush happens via the `ServerConnectionManagementHandler/SyncView`.


glbrntt (changes requested):
one nit but looks good otherwise

glbrntt (approved):

PR #1840, opened 2024-03-22
'runServer' implementation for Performance Worker Service

stefanadranca:
Motivation:

The run server RPC starts a server and reports stats back to the client. We can’t yet start a real server, because we don’t yet have the http/2 transport, but we can stub it out and do the rest of the RPC which reports stats back to the client.

Modifications:

- implemented the methods that start the server and collects the stats
- added the initial stats property used in computing the requested server stats

Result:

We will be able to start the server from a performance worker.

### DIFF
--- a/Sources/performance-worker/WorkerService.swift
+++ b/Sources/performance-worker/WorkerService.swift
@@ -32,7 +32,7 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable 
 
     enum Role {
       case client(GRPCClient)
-      case server(GRPCServer, ServerStats)
+      case server(ServerState)
     }
 
     init() {}
@@ -41,34 +41,39 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable 
       self.role = role
     }
 
-    init(client: GRPCClient) {
-      self.role = .client(client)
-    }
-
-    public func getServer() -> GRPCServer? {
+    func getServer() -> GRPCServer? {
       switch self.role {
-      case let .server(server, _):
-        return server
+      case let .server(serverState):
+        return serverState.server
       default:
         return nil
       }
     }
 
-    public func getInitialServerStats() -> ServerStats? {
+    mutating func initialServerState(set newValue: ServerState? = nil) -> ServerState? {
       switch self.role {
-      case let .server(_, initialServerStats):
+      case let .server(initialServerStats):
+        defer {
+          if let newServerState = newValue {
+            self.role = .server(newServerState)
+          }
+        }
         return initialServerStats
       default:
         return nil
       }
     }
 
-    public mutating func setInitialServerStats(_ newStats: ServerStats) {
+    mutating func setupServer(_ initialServerState: ServerState) throws {
       switch self.role {
-      case let .server(server, _):
-        self.role = .server(server, newStats)
+      case .server(_):
+        throw RPCError(code: .alreadyExists, message: "A server has already been set up.")
+
+      case .client(_):
+        throw RPCError(code: .failedPrecondition, message: "This worker has a client setup.")
+
       default:
-        ()
+        self.role = .server(initialServerState)
       }
     }
   }
@@ -86,8 +91,8 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable 
       switch role {
       case .client(let client):
         client.close()
-      case .server(let server, _):
-        server.stopListening()
+      case .server(let serverState):
+        serverState.server.stopListening()
       }
     }
 
@@ -153,18 +158,10 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable 
 extension WorkerService {
   private func setupServer(_ config: Grpc_Testing_ServerConfig) async throws -> GRPCServer {
     let server = GRPCServer(transports: [], services: [BenchmarkService()])
-    let initialStats = try await ServerStats()
+    let initialServerState = try await ServerState(server: server)
+
     try self.state.withLockedValue { state in
-      switch state.role {
-      case .server(_, _):
-        throw RPCError(code: .alreadyExists, message: "A server has already been set up.")
-
-      case .client(_):
-        throw RPCError(code: .failedPrecondition, message: "This worker has a client setup.")
-
-      default:
-        state.role = .server(server, initialStats)
-      }
+      try state.setupServer(initialServerState)
     }
     return server
   }
@@ -172,21 +169,20 @@ extension WorkerService {
   private func makeStatsResponse(
     reset: Bool
   ) async throws -> Grpc_Testing_WorkerService.Method.RunServer.Output {
-    let currentStats = try await ServerStats()
-    let initialStats = self.state.withLockedValue { state in
-      defer {
-        if reset {
-          state.setInitialServerStats(currentStats)
-        }
-      }
-      return state.getInitialServerStats()
+    let server = self.state.withLockedValue { state in state.getServer() }
+    guard let server = server else {
+      throw RPCError(code: .failedPrecondition, message: "This worker doesn't have a server setup.")
+    }
+    let currentState = try await ServerState(server: server)
+    let initialState = self.state.withLockedValue { state in
+      return state.initialServerState(set: currentState)
     }
 
-    guard let initialStats = initialStats else {
+    guard let initialState = initialState else {
       throw RPCError(code: .notFound, message: "There are no initial server stats.")
     }
 
-    let differences = currentStats.difference(to: initialStats)
+    let differences = try currentState.difference(to: initialState)
     return Grpc_Testing_WorkerService.Method.RunServer.Output.with {
       $0.stats = Grpc_Testing_ServerStats.with {
         $0.idleCpuTime = differences.idleCPUTime
--- a/Sources/performance-worker/WorkerService.swift
+++ b/Sources/performance-worker/WorkerService.swift
@@ -41,12 +41,6 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable 
       self.role = role
     }
 
-    init(server: GRPCServer) async throws {
-      let initialState = try await ServerStats()
-      self.role = .server(server, initialState)
-
-    }
-
     init(client: GRPCClient) {
       self.role = .client(client)
     }
@@ -69,7 +63,7 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable 
       }
     }
 
-    public mutating func setInitialServerStats(newStats: ServerStats) {
+    public mutating func setInitialServerStats(_ newStats: ServerStats) {
       switch self.role {
       case let .server(server, _):
         self.role = .server(server, newStats)
@@ -137,14 +131,11 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable 
       }
 
       let server = self.state.withLockedValue { state in
-        state.getServer()
+        defer { state.role = nil }
+        return state.getServer()
       }
 
-      guard let server = server else {
-        return [:]
-      }
-
-      server.stopListening()
+      server?.stopListening()
       return [:]
     }
   }
--- a/Sources/performance-worker/WorkerService.swift
+++ b/Sources/performance-worker/WorkerService.swift
@@ -29,26 +29,53 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable 
 
   private struct State {
     var role: Role?
-    var initialStats: ServerStats?
 
     enum Role {
       case client(GRPCClient)
-      case server(GRPCServer)
+      case server(GRPCServer, ServerStats)
     }
 
     init() {}
 
     init(role: Role) async throws {
       self.role = role
-      self.initialStats = try await ServerStats()
     }
 
     init(server: GRPCServer) async throws {
-      self.role = .server(server)
+      let initialState = try await ServerStats()
+      self.role = .server(server, initialState)
+
     }
 
     init(client: GRPCClient) {
       self.role = .client(client)
+    }
+
+    public func getServer() -> GRPCServer? {
+      switch self.role {
+      case let .server(server, _):
+        return server
+      default:
+        return nil
+      }
+    }
+
+    public func getInitialServerStats() -> ServerStats? {
+      switch self.role {
+      case let .server(_, initialServerStats):
+        return initialServerStats
+      default:
+        return nil
+      }
+    }
+
+    public mutating func setInitialServerStats(newStats: ServerStats) {
+      switch self.role {
+      case let .server(server, _):
+        self.role = .server(server, newStats)
+      default:
+        ()
+      }
     }
   }
 
@@ -65,7 +92,7 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable 
       switch role {
       case .client(let client):
         client.close()
-      case .server(let server):
+      case .server(let server, _):
         server.stopListening()
       }
     }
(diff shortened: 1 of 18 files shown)


Sources/performance-worker/WorkerService.swift
```diff
@@ -87,7 +90,29 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+        switch message.argtype {
+        case let .some(.setup(serverConfig)):
+          try await self.serverSetup(serverConfig)
+        case let .some(.mark(mark)):
+          let response = try await self.makeStatsResponse(for: mark)
+          try await writer.write(response)
+        case .none:
+          ()
+        }
+        throw RPCError(code: .unavailable, message: "This RPC has not been implemented yet.")
```
glbrntt, line 104: Why are we still throwing an error here?

Sources/performance-worker/WorkerService.swift
```diff
@@ -98,3 +123,52 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
     throw RPCError(code: .unimplemented, message: "This RPC has not been implemented yet.")
   }
 }
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+extension WorkerService {
+  private func serverSetup(_ config: Grpc_Testing_ServerConfig) async throws {
```
glbrntt, line 129: nit: `setupServer`

Sources/performance-worker/WorkerService.swift
```diff
@@ -98,3 +123,52 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+    try self.state.withLockedValue { state in
+      switch state.role {
+      case .server(_):
+        throw RPCError(code: .alreadyExists, message: "A server has been previously set up.")
+      case .client(_):
+        throw RPCError(code: .failedPrecondition, message: "This worker has a client set up.")
+      default:
+        ()
+      }
+    }
```
glbrntt, line 139: We're just checking the state here, we need to update it too.

Sources/performance-worker/WorkerService.swift
```diff
@@ -98,3 +123,52 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+    // The asynchronous function 'run()' can't be called inside the synchronous
+    // closure passed to 'withLockedValue'.
+    let server = GRPCServer(transports: [], services: [BenchmarkService()])
+    try await server.run()
```
glbrntt, line 143: This isn't quite right because this will suspend until the server has finished running, which means that in `runServer` we won't continue reading the request messages.

Sources/performance-worker/WorkerService.swift
```diff
@@ -37,12 +38,14 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
-    init(role: Role) {
+    init(role: Role) async throws {
       self.role = role
+      self.initialStats = try await ServerStats()
     }
 
-    init(server: GRPCServer) {
+    init(server: GRPCServer) async throws {
       self.role = .server(server)
+      self.initialStats = try await ServerStats()
```
glbrntt, line 48: I think the initial stats should be created when the server is created, not when the service is created (because the service might never run a server).

Sources/performance-worker/WorkerService.swift
```diff
@@ -87,7 +90,29 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+          ()
+        }
+        throw RPCError(code: .unavailable, message: "This RPC has not been implemented yet.")
+      }
+      try self.state.withLockedValue {
+        switch $0.role {
+        case .server(let server):
+          server.stopListening()
+        default:
+          throw RPCError(code: .unavailable, message: "There is no benchmark server.")
```
glbrntt, line 111: We don't need to throw in this case, we can just ignore it. It'd be possible (but unusual) for a client to open the `runServer` RPC and then immediately close it. If that happened we'd throw this error unnecessarily.

glbrntt (changes requested):

Sources/performance-worker/WorkerService.swift
stefanadranca: Forgot to delete it, thanks!

Sources/performance-worker/WorkerService.swift
stefanadranca: Should I add a Task group in 'runServer' that has a Task that calls runServer()?

Sources/performance-worker/WorkerService.swift
```diff
@@ -37,11 +38,12 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
     init() {}
 
-    init(role: Role) {
+    init(role: Role) async throws {
       self.role = role
+      self.initialStats = try await ServerStats()
     }
 
-    init(server: GRPCServer) {
+    init(server: GRPCServer) async throws {
```
glbrntt, line 46: This `init` isn't `async` or `throw`ing

Sources/performance-worker/WorkerService.swift
```diff
@@ -37,11 +38,12 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
 
     init() {}
 
-    init(role: Role) {
+    init(role: Role) async throws {
       self.role = role
+      self.initialStats = try await ServerStats()
```
glbrntt, line 43: We shouldn't be getting status unconditionally here

Sources/performance-worker/WorkerService.swift
```diff
@@ -87,7 +89,41 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+          try await self.setupServer(serverConfig)
+          try await withThrowingTaskGroup(of: Void.self) { group in
+            group.addTask {
+              let role = self.state.withLockedValue({ $0.role })
+              switch role {
+              case let .server(server):
+                try await server.run()
+              default:
+                ()
+              }
```
glbrntt, line 105: Rather than reaching back into the state to get the server it would make more sense for `setupServer` to update the state _and_ return the server to run.

Sources/performance-worker/WorkerService.swift
```diff
@@ -87,7 +89,41 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
   ) async throws
     -> GRPCCore.ServerResponse.Stream<Grpc_Testing_WorkerService.Method.RunServer.Output>
   {
-    throw RPCError(code: .unimplemented, message: "This RPC has not been implemented yet.")
+    return ServerResponse.Stream { writer in
+      for try await message in request.messages {
+        switch message.argtype {
+        case let .some(.setup(serverConfig)):
+          try await self.setupServer(serverConfig)
+          try await withThrowingTaskGroup(of: Void.self) { group in
```
glbrntt, line 97: This has the same problem as before: now the for loop won't continue until the task group has finished. The for loop needs to be inside the task group.

Sources/performance-worker/WorkerService.swift
```diff
@@ -87,7 +89,41 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+              let role = self.state.withLockedValue({ $0.role })
+              switch role {
+              case let .server(server):
+                try await server.run()
+              default:
+                ()
+              }
+            }
+            try await group.next()
+            group.cancelAll()
```
glbrntt, line 108: No point cancelling all: there's only one task and we just waited for it to return on the previous line. At this point there's no work to cancel.

Sources/performance-worker/WorkerService.swift
```diff
@@ -87,7 +89,41 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+        }
+      }
+      self.state.withLockedValue {
+        switch $0.role {
+        case .server(let server):
+          server.stopListening()
+        default:
+          ()
+        }
+      }
```
glbrntt, line 124: Generally speaking avoid calling out to objects while holding a lock as you don't always know what they may do (and it simply might not be necessary to hold the lock for that long). As an example, the server could have a mechanism which notifies us when the server has stopped listening, and we may want to acquire the lock in that case to update our state. If we're still holding the lock (and it isn't a re-entrant lock) then we're stuffed.

In this case we only need the lock to grab the reference to the server so `stopListening` can be called after releasing the lock.

Sources/performance-worker/WorkerService.swift
```diff
@@ -98,3 +134,53 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+    }
+    let initialStats = try await ServerStats()
+    self.state.withLockedValue {
+      $0.initialStats = initialStats
+    }
+  }
+
+  private func makeStatsResponse(
+    for mark: Grpc_Testing_Mark
```
glbrntt, line 159: `for mark: Grpc_Testing_Mark` isn't particularly helpful here, `reset: Bool` is clearer.

Sources/performance-worker/WorkerService.swift
```diff
@@ -98,3 +134,53 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+      case .server(_):
+        throw RPCError(code: .alreadyExists, message: "A server has been previously set up.")
+      case .client(_):
+        throw RPCError(code: .failedPrecondition, message: "This worker has a client set up.")
+      default:
+        let server = GRPCServer(transports: [], services: [BenchmarkService()])
+        state.role = .server(server)
+      }
+    }
+    let initialStats = try await ServerStats()
```
glbrntt, line 152: We have an implicit invariant at this point: when there is a server there must also be stats.

At the moment this is codified via `Role?` and `ServerStats?` which gives us many more options in code than is actually possible, e.g.:

```swift
switch (state.role, state.initialStats) {
// These are valid combinations:
case (.some(.client), .none),
  (.some(.server), .some),
  (.none, .none):
  // ...

// These are reachable but shouldn't be possible:
case (.some(.client), .some),
  (.some(.server), .none),
  (.none, .some):
  // ...
}
```

We can codify this more clearly by extending the role to include server stats, e.g.:

```swift
switch state.role {
case .some(.client):
  ()
case .some(.server, let stats):
  ()
}
```

glbrntt (changes requested):

Sources/performance-worker/WorkerService.swift
stefanadranca: so, should there be a task containing the for loop and a child task for 'server.run()'?

Sources/performance-worker/WorkerService.swift
```diff
@@ -87,7 +116,37 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+        state.getServer()
+      }
+
+      guard let server = server else {
+        return [:]
+      }
+
+      server.stopListening()
+      return [:]
+    }
```
glbrntt, line 149: Let's simplify this a bit:

```suggestion
      server?.stopListening()
      return [:]
    }
```

Sources/performance-worker/WorkerService.swift
```diff
@@ -41,13 +41,42 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
       self.role = role
     }
 
-    init(server: GRPCServer) {
-      self.role = .server(server)
+    init(server: GRPCServer) async throws {
+      let initialState = try await ServerStats()
+      self.role = .server(server, initialState)
+
     }
```
glbrntt, line 48: This `init` is never used, might as well remove it

Sources/performance-worker/WorkerService.swift
```diff
@@ -41,13 +41,42 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+    public func getInitialServerStats() -> ServerStats? {
+      switch self.role {
+      case let .server(_, initialServerStats):
+        return initialServerStats
+      default:
+        return nil
+      }
+    }
+
+    public mutating func setInitialServerStats(newStats: ServerStats) {
```
glbrntt, line 72: I wouldn't bother with the label here, the name of the function already indicates what the function does and what the expected parameter is. 

```suggestion
    public mutating func setInitialServerStats(_ newStats: ServerStats) {
```

Sources/performance-worker/WorkerService.swift
```diff
@@ -87,7 +116,37 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+          case .none:
+            ()
+          }
+        }
+
+        try await group.next()
+      }
+
+      let server = self.state.withLockedValue { state in
+        state.getServer()
```
glbrntt, line 140: We may want to reset the state here as well so that this worker service can be reused for a new client/server.

glbrntt (changes requested):

glbrntt (changes requested):

Sources/performance-worker/WorkerService.swift
```diff
@@ -98,3 +169,44 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
     throw RPCError(code: .unimplemented, message: "This RPC has not been implemented yet.")
   }
 }
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+extension WorkerService {
+  private func setupServer(_ config: Grpc_Testing_ServerConfig) async throws -> GRPCServer {
+    let server = GRPCServer(transports: [], services: [BenchmarkService()])
+    let initialServerState = try await State.ServerState(server: server)
```
glbrntt, line 177: It's weird to pass the state into the state machine. The `ServerState` is an implementation detail of the `State` so this is a bit of layering violation. It's clearer to pass in whatever is needed to construct the server state, i.e. the server and the stats.

Sources/performance-worker/WorkerService.swift
```diff
@@ -41,12 +55,42 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+      case let .server(serverState):
+        return serverState.server
+      default:
+        return nil
+      }
     }
 
-    init(client: GRPCClient) {
-      self.role = .client(client)
+    mutating func initialServerStats(set newValue: ServerStats? = nil) -> ServerStats? {
```
glbrntt, line 67: These aren't always the initial (i.e. when the server started) stats. 

I think `serverStats(replaceWith newStats: ServerStats? = nil) -> ServerStats?` is a better spelling here.

glbrntt (changes requested):

Sources/performance-worker/WorkerService.swift
```diff
@@ -98,3 +164,45 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+    return server
+  }
+
+  private func makeServerStatsResponse(
+    reset: Bool
+  ) async throws -> Grpc_Testing_WorkerService.Method.RunServer.Output {
+    let server = self.state.withLockedValue { state in state.server }
+    guard server != nil else {
+      throw RPCError(code: .failedPrecondition, message: "This worker doesn't have a server setup.")
+    }
```
glbrntt, line 187: This is redundant because `serverStats` returns `nil` if the role isn't `server`. However, if we should update the message of the other error to indicate that the server must be setup before calling mark.

```suggestion
```

Sources/performance-worker/WorkerService.swift
```diff
@@ -98,3 +164,45 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+    let currentStats = try await ServerStats()
+    let initialStats = self.state.withLockedValue { state in
+      return state.serverStats(replaceWith: currentStats)
```
glbrntt, line 190: We're ignoring `reset` here: 

```suggestion
      return state.serverStats(replaceWith: reset ? currentStats : nil)
```

Sources/performance-worker/WorkerService.swift
```diff
@@ -41,12 +51,41 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+    mutating func serverStats(replaceWith newStats: ServerStats? = nil) -> ServerStats? {
+      switch self.role {
+      case var .server(serverState):
+        let stats = serverState.stats
+        if let newStats = newStats {
+          serverState.stats = newStats
+          self.role = .server(serverState)
+        }
+        return stats
+      default:
```
glbrntt, line 72: It's best to switch explicitly over the different options, particularly with state machines, because if you add a new case later you might not correctly handle it (because the `default` case will catch it).

Sources/performance-worker/WorkerService.swift
```diff
@@ -41,12 +51,41 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+    mutating func setupServer(server: GRPCServer, stats: ServerStats) throws {
+      let serverState = State.ServerState(server: server, stats: stats)
+      switch self.role {
+      case .server(_):
+        throw RPCError(code: .alreadyExists, message: "A server has already been set up.")
+
+      case .client(_):
+        throw RPCError(code: .failedPrecondition, message: "This worker has a client setup.")
+
+      default:
```
glbrntt, line 86: Same here re: explicitly switching
```suggestion
      case .none:
```

Sources/performance-worker/WorkerService.swift
```diff
@@ -41,12 +51,41 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
       self.role = role
     }
 
-    init(server: GRPCServer) {
-      self.role = .server(server)
+    var server: GRPCServer? {
+      switch self.role {
+      case let .server(serverState):
+        return serverState.server
+      default:
```
glbrntt, line 58:
```suggestion
      case .client, .none:
```

glbrntt (changes requested):

glbrntt (approved):
LGTM

PR #1843, opened 2024-03-27
Handle write promises correctly

gjcairo:
We're not currently handling write promises correctly in the server and client stream handlers. 

This PR adds proper handling by storing promises linked to each write, and chaining them when they're framed together in a single frame, which can be passed onto the `context/write` call on `flush()`.

### DIFF
--- a/Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
+++ b/Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
@@ -143,23 +143,16 @@ extension GRPCClientStreamHandler {
       do {
         self.flushPending = true
         let headers = try self.stateMachine.send(metadata: metadata)
-        context.write(self.wrapOutboundOut(.headers(.init(headers: headers))), promise: nil)
-        // TODO: move the promise handling into the state machine
-        promise?.succeed()
+        context.write(self.wrapOutboundOut(.headers(.init(headers: headers))), promise: promise)
       } catch {
         context.fireErrorCaught(error)
-        // TODO: move the promise handling into the state machine
-        promise?.fail(error)
       }
 
     case .message(let message):
       do {
-        try self.stateMachine.send(message: message)
-        // TODO: move the promise handling into the state machine
-        promise?.succeed()
+        try self.stateMachine.send(message: message, promise: promise)
       } catch {
         context.fireErrorCaught(error)
-        // TODO: move the promise handling into the state machine
         promise?.fail(error)
       }
     }
@@ -198,11 +191,11 @@ extension GRPCClientStreamHandler {
     do {
       loop: while true {
         switch try self.stateMachine.nextOutboundMessage() {
-        case .sendMessage(let byteBuffer):
+        case .sendMessage(let byteBuffer, let promise):
           self.flushPending = true
           context.write(
             self.wrapOutboundOut(.data(.init(data: .byteBuffer(byteBuffer)))),
-            promise: nil
+            promise: promise
           )
 
         case .noMoreMessages:
--- a/Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
+++ b/Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
@@ -145,6 +145,7 @@ extension GRPCClientStreamHandler {
         let headers = try self.stateMachine.send(metadata: metadata)
         context.write(self.wrapOutboundOut(.headers(.init(headers: headers))), promise: promise)
       } catch {
+        promise?.fail(error)
         context.fireErrorCaught(error)
       }
 
@@ -152,8 +153,8 @@ extension GRPCClientStreamHandler {
       do {
         try self.stateMachine.send(message: message, promise: promise)
       } catch {
-        context.fireErrorCaught(error)
         promise?.fail(error)
+        context.fireErrorCaught(error)
       }
     }
   }
--- a/Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
+++ b/Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
@@ -190,8 +190,8 @@ extension GRPCClientStreamHandler {
   private func _flush(context: ChannelHandlerContext) {
     do {
       loop: while true {
-        switch try self.stateMachine.nextOutboundMessage() {
-        case .sendMessage(let byteBuffer, let promise):
+        switch try self.stateMachine.nextOutboundFrame() {
+        case .sendFrame(let byteBuffer, let promise):
           self.flushPending = true
           context.write(
             self.wrapOutboundOut(.data(.init(data: .byteBuffer(byteBuffer)))),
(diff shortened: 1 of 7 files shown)


Sources/GRPCHTTP2Core/GRPCMessageFramer.swift
```diff
@@ -69,15 +74,21 @@ struct GRPCMessageFramer {
...
 
+    var pendingWritePromise: EventLoopPromise<Void>?
     while let message = self.pendingMessages.pop() {
-      try self.encode(message, compressor: compressor)
+      try self.encode(message.bytes, compressor: compressor)
+      if pendingWritePromise == nil {
+        pendingWritePromise = message.promise
+      } else {
+        pendingWritePromise?.futureResult.cascade(to: message.promise)
+      }
```
glbrntt, line 88: nit: prefer `if let`:

```swift
if let pendingWritePromise = pendingWritePromise {
  // ...
} else {
  // ...
}
```

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -398,14 +398,17 @@ struct GRPCStreamStateMachine {
...
+  enum OnNextOutboundMessage {
     /// Either the receiving party is closed, so we shouldn't send any more messages; or the sender is done
     /// writing messages (i.e. we are now closed).
     case noMoreMessages
     /// There isn't a message ready to be sent, but we could still receive more, so keep trying.
     case awaitMoreMessages
     /// A message is ready to be sent.
-    case sendMessage(ByteBuffer)
+    case sendMessage(
+      message: ByteBuffer,
```
glbrntt, line 409: at this point it's no longer a message, it could be multiple framed messages (though I realise now that all of the naming here is message based...)

Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
```diff
@@ -35,6 +35,7 @@ final class GRPCServerStreamHandler: ChannelDuplexHandler {
   // if there are messages still not written into the channel because flush has
   // not been called, but the server sends back trailers).
   private var pendingTrailers: HTTP2Frame.FramePayload?
+  private var pendingTrailersPromise: EventLoopPromise<Void>?
```
glbrntt, line 38: These always come together, can we store them as `(HTTP2Frame.FramePayload, EventLoopPromise<Void>?)?`?

Tests/GRPCHTTP2CoreTests/GRPCMessageFramerTests.swift
```diff
@@ -22,9 +22,9 @@ import XCTest
 final class GRPCMessageFramerTests: XCTestCase {
```
glbrntt, line 22: Can you add some tests which check that promises are cascaded as expected?

glbrntt (changes requested):

Sources/GRPCHTTP2Core/GRPCMessageFramer.swift
gjcairo: I don't think I can do this. First of all, I'm overriding the value of `pendingWritePromise`, so it should be `if var ...`.
But also, since I've defined `pendingWritePromise` a couple of lines above this, wouldn't this declaration shadow that one?

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
gjcairo: Hm, you're right. I've renamed these to be more consistent although I did keep `awaitMoreMessages` and `noMoreMessages`, since I believe it's not wrong to refer to messages in those situations (as we're either waiting for more messages to arrive, regardless of how we frame them, or we know that no more messages will arrive).

Sources/GRPCHTTP2Core/GRPCMessageFramer.swift
glbrntt: > I don't think I can do this. First of all, I'm overriding the value of `pendingWritePromise`, so it should be `if var ...`. 

You only set `pendingWritePromise` if it's nil, so that'd be done in the `else` branch.

> But also, since I've defined `pendingWritePromise` a couple of lines above this, wouldn't this declaration shadow that one?

Yeah, but you can use a different name, e.g.:

```swift
if let promise = pendingWritePromise {
  promise.futureResult.cascade(to: message.promise)
} else {
  pendingWritePromise = message.promise
}
```

---

The aim is to remove the unnecessary `?` in the original `else` branch. You can reason that the value must be present (because it's `nil` in the other branch) which means you could safely replace it with `!`, at that point you can just flip the check and use `if let`.

Tests/GRPCHTTP2CoreTests/GRPCMessageFramerTests.swift
gjcairo: I added a test checking this to the `GRPCStreamStateMachine` tests, but I also just added (or rather modified an existing one) in the framer tests.

Sources/GRPCHTTP2Core/GRPCMessageFramer.swift
gjcairo: Ah I see what you mean. Thanks. I've changed this.

Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
```diff
@@ -143,23 +143,16 @@ extension GRPCClientStreamHandler {
...
         self.flushPending = true
         let headers = try self.stateMachine.send(metadata: metadata)
-        context.write(self.wrapOutboundOut(.headers(.init(headers: headers))), promise: nil)
-        // TODO: move the promise handling into the state machine
-        promise?.succeed()
+        context.write(self.wrapOutboundOut(.headers(.init(headers: headers))), promise: promise)
       } catch {
         context.fireErrorCaught(error)
-        // TODO: move the promise handling into the state machine
-        promise?.fail(error)
```
glbrntt, line 152: I think we _do_ need this one: we only reach here if we're in an invalid state and the state machine doesn't take care of the promise in that situation.

Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
```diff
@@ -142,36 +143,26 @@ extension GRPCServerStreamHandler {
...
     case .message(let message):
       do {
-        try self.stateMachine.send(message: message)
-        // TODO: move the promise handling into the state machine
-        promise?.succeed()
+        try self.stateMachine.send(message: message, promise: promise)
       } catch {
         context.fireErrorCaught(error)
-        // TODO: move the promise handling into the state machine
-        promise?.fail(error)
```
glbrntt, line 162: We need this one too, I think

glbrntt (changes requested):
Couple of places where we've incorrectly stopped failing a promise but otherwise looks good!

Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
gjcairo: Yes you're right, I fixed this in a couple of other places before but I missed these two. Good catch!

glbrntt (approved):

PR #1844, opened 2024-03-27
Fix several bugs in GRPC stream handlers and state machine

gjcairo:
This PR fixes three bugs introduced in recent changes:
- We weren't correctly handling the (invalid) scenario in which a server may send a data frame with EOS (instead of closing by sending status + trailers). This is now handled correctly by forwarding an `internalError` status code down the pipeline, and closing both client and server.
- We were forwarding channel close when the close mode was `output`. However, because the H2 stream handler ignores the close mode and just closes down the whole channel upon receiving `close`, this was inadvertently closing the whole stream. We now only forward the close if the mode's `all`.
- We were retrieving the `grpc-status-message` header using the canonical form, which caused us to only retrieve part of the original message if it included a comma. This has been fixed.

### DIFF
--- a/Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
+++ b/Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
@@ -69,10 +69,8 @@ extension GRPCClientStreamHandler {
         do {
           switch try self.stateMachine.receive(buffer: buffer, endStream: endStream) {
           case .endRPCAndForwardErrorStatus(let status):
-            if let rpcError = RPCError(status: status) {
-              context.fireErrorCaught(rpcError)
-            }
-            return
+            context.fireChannelRead(self.wrapInboundOut(.status(status, [:])))
+            context.close(promise: nil)
 
           case .readInbound:
             loop: while true {
--- a/Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
+++ b/Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
@@ -186,6 +186,7 @@ extension GRPCClientStreamHandler {
         // Force a flush by calling _flush instead of flush
         // (otherwise, we'd skip flushing if we're in a read loop)
         self._flush(context: context)
+        promise?.succeed()
       } catch {
         promise?.fail(error)
         context.fireErrorCaught(error)
--- a/Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
+++ b/Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
@@ -173,7 +173,27 @@ extension GRPCClientStreamHandler {
 
   func close(context: ChannelHandlerContext, mode: CloseMode, promise: EventLoopPromise<Void>?) {
     switch mode {
-    case .output, .all:
+    case .input:
+      context.close(mode: .input, promise: promise)
+
+    case .output:
+      // We flush all pending messages and update the internal state machine's
+      // state, but we don't close the outbound end of the channel, because
+      // forwarding the close in this case would cause the HTTP2 stream handler
+      // to close the whole channel (as the mode is ignored in its implementation).
+      do {
+        try self.stateMachine.closeOutbound()
+        // Force a flush by calling _flush instead of flush
+        // (otherwise, we'd skip flushing if we're in a read loop)
+        self._flush(context: context)
+      } catch {
+        promise?.fail(error)
+        context.fireErrorCaught(error)
+      }
+
+    case .all:
+      // Since we're closing the whole channel here, we *do* forward the close
+      // down the pipeline.
       do {
         try self.stateMachine.closeOutbound()
         // Force a flush by calling _flush
@@ -184,9 +204,6 @@ extension GRPCClientStreamHandler {
         promise?.fail(error)
         context.fireErrorCaught(error)
       }
-
-    case .input:
-      context.close(mode: .input, promise: promise)
     }
   }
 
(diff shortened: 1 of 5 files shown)


Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
```diff
@@ -67,16 +67,27 @@ extension GRPCClientStreamHandler {
...
-            case .awaitMoreMessages:
-              break loop
-            case .noMoreMessages:
-              context.fireUserInboundEventTriggered(ChannelEvent.inputClosed)
-              break loop
+          switch try self.stateMachine.receive(buffer: buffer, endStream: endStream) {
+          case .endRPCAndForwardErrorStatus(let status):
+            if let rpcError = RPCError(status: status) {
+              context.fireErrorCaught(rpcError)
+            }
```
glbrntt, line 74: This isn't quite right: we should be forwarding this as a `Status` and (empty) `Metadata`.

Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
```diff
@@ -167,7 +179,27 @@ extension GRPCClientStreamHandler {
 
   func close(context: ChannelHandlerContext, mode: CloseMode, promise: EventLoopPromise<Void>?) {
     switch mode {
-    case .output, .all:
+    case .input:
+      context.close(mode: .input, promise: promise)
+
+    case .output:
```
glbrntt, line 185: Can we avoid the duplication here? If we switch on the mode we can set two booleans, whether to flush and whether to close, then we can conditionally do that work after the switch.

Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
```diff
@@ -167,7 +179,27 @@ extension GRPCClientStreamHandler {
 
   func close(context: ChannelHandlerContext, mode: CloseMode, promise: EventLoopPromise<Void>?) {
     switch mode {
-    case .output, .all:
+    case .input:
+      context.close(mode: .input, promise: promise)
```
glbrntt, line 183: Do we definitely want to forward this one? It'll close the whole stream.

Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
```diff
@@ -167,7 +179,27 @@ extension GRPCClientStreamHandler {
...
+    case .output:
+      // We flush all pending messages and update the internal state machine's
+      // state, but we don't close the outbound end of the channel, because
+      // forwarding the close in this case would cause the HTTP2 stream handler
+      // to close the whole channel (as the mode is ignored in its implementation).
+      do {
+        try self.stateMachine.closeOutbound()
+        // Force a flush by calling _flush instead of flush
+        // (otherwise, we'd skip flushing if we're in a read loop)
+        self._flush(context: context)
```
glbrntt, line 194: In this case, we'll leak the `promise`. We can just succeed it because we're effectively half closing here.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -724,7 +742,7 @@ extension GRPCStreamStateMachine {
     }
 
     let statusMessage =
-      metadata.firstString(forKey: .grpcStatusMessage)
+      metadata.first(name: GRPCHTTP2Keys.grpcStatusMessage.rawValue)
```
glbrntt, line 745: Why don't we allow `firstString(forKey:)` to take a `canonicalForm` param and just pass that down instead? Seems a shame to drop back to using the verbose names.

glbrntt (changes requested):

Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
gjcairo: Ah - should we also forward a close then?

glbrntt: yup

Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
glbrntt: Good question. I'm not sure if we'll actually ever hit this path, but that's probably the right thing to do, yeah.

Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
gjcairo: Maybe take another look now I've made some other changes? I get the cases are quite similar, but we're doing different things on each of them:
- succeeding a promise in `output` but forwarding it on `close` on `input` and `all`;
- "closing" on `input` means forwarding the close, on `output` it means only closing in the state machine, and on `all` means doing both

I'm not sure having several different booleans (because I don't think 2 would be enough) would make this easier to read.

glbrntt: Yeah, now that we handle `input` differently it doesn't make sense to do this

Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
```diff
@@ -161,7 +171,28 @@ extension GRPCClientStreamHandler {
 
   func close(context: ChannelHandlerContext, mode: CloseMode, promise: EventLoopPromise<Void>?) {
     switch mode {
-    case .output, .all:
+    case .input:
+      context.fireUserInboundEventTriggered(ChannelEvent.inputClosed)
```
glbrntt, line 175: We need to handle the promise here

glbrntt (changes requested):
We're leaking a promise but looks good otherwise

glbrntt (approved):

PR #1847, opened 2024-04-02
Update `release` GitHub workflow action

gjcairo:
The `release` GH workflow action was still using the `plugins` Makefile target, but we deleted the Makefile some time ago so it was failing.

This PR updates the `release.yaml` GH workflow to use the SPM directly instead, and also removes some doc references to the makefile.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,7 +15,9 @@ jobs:
         run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
 
       - name: Build the plugins
-        run: make plugins
+        run: |
+          swift build --build-path=./.build --configuration=release --product protoc-gen-swift
+          swift build --build-path=./.build --configuration=release --product protoc-gen-grpc-swift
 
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -15,7 +15,9 @@ jobs:
         run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
 
       - name: Build the plugins
-        run: make plugins
+        run: |
+          swift build --build-path=./.build --configuration=release --product protoc-gen-swift
+          swift build --build-path=./.build --configuration=release --product protoc-gen-grpc-swift
```
glbrntt, line 20: I don't think this'll work, the zipped plugins are looking in the current directory, I think you need to copy them out of the build directory into the current dir. Also I don't think you need to specify the `--built-path` here.

glbrntt (changes requested):

.github/workflows/release.yml
gjcairo: Ah, good catch. Fixed.

PR #1848, opened 2024-04-03
Allow manually running `release` GH workflow

gjcairo:


glbrntt (commented):
Does this allow us to set the release version (the workflow depends on it)? This is set by the release event which usually triggers the job.

glbrntt (approved):

PR #1851, opened 2024-04-08
Performance worker - client stats for runClient RPC

stefanadranca:
Motivation:

The `runClient()` RPC receives a stream of messages that can either request to set up clients based on some parameters sent in the message or send back stats on the currently running clients.

Modifications:

- implemented the resource usage class for clients similar to the server stats one
- implemented the RPCStats struct that contains a LatencyHistogram and a dictionary that keeps track of the request result codes
- added a new abstraction layer over the GRPCClient - BenchmarkClient. It stores parameters for the client extracted from the input messages and it implements the `run()` method that starts the client and makes the given number of RPCs, while registering the latency and result code for each one.

Result:

Configuring the clients for the performance tests is now possible and the latency histogram for each client gets created.


Sources/performance-worker/BenchmarkClient.swift
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+import GRPCCore
+import NIOConcurrencyHelpers
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+struct BenchmarkClient {
+  var client: GRPCClient
+  var rpcNumber: Int32
+  var rpcType: Grpc_Testing_RpcType
```
glbrntt, line 24: Do these need to be `internal`? Or can they be `private`?

Sources/performance-worker/BenchmarkClient.swift
```diff
@@ -0,0 +1,91 @@
...
+    client: GRPCClient,
+    rpcNumber: Int32,
+    rpcType: Grpc_Testing_RpcType,
+    histogramParams: Grpc_Testing_HistogramParams?
+  ) {
+    self.client = client
+    self.rpcNumber = rpcNumber
+    self.rpcType = rpcType
+
+    var histogram: RPCStats.LatencyHistogram
```
glbrntt, line 37: This can be a `let`

Sources/performance-worker/BenchmarkClient.swift
```diff
@@ -0,0 +1,91 @@
...
+      // Make the requests to the server and register the latency for each one.
+      try await withThrowingTaskGroup(of: Void.self) { rpcsGroup in
+        for _ in 0 ..< self.rpcNumber {
+          rpcsGroup.addTask {
+            let (latency, errorCode) = self.makeRPC(client: benchmarkClient, rpcType: self.rpcType)
+            self.rpcStats.withLockedValue {
+              $0.latencyHistogram.add(value: latency)
+              if errorCode != nil {
+                $0.requestResultCount[Int32(errorCode!.rawValue), default: 1] += 1
+              }
```
glbrntt, line 65: Don't we need to record non-error status codes too?

Sources/performance-worker/ClientStats.swift
```diff
@@ -0,0 +1,73 @@
...
+import Glibc
+#else
+let badOS = { fatalError("unsupported OS") }()
+#endif
+
+#if canImport(Darwin)
+private let OUR_RUSAGE_SELF: Int32 = RUSAGE_SELF
+#elseif canImport(Musl) || canImport(Glibc)
+private let OUR_RUSAGE_SELF: Int32 = RUSAGE_SELF.rawValue
+#endif
```
glbrntt, line 33: Given we do the same for the server, can we deduplicate all of this? 

We can move the actual `getrusage` call to a static function somewhere (maybe an extension on `System` from `NIOCore`). While we're at it we can make the function a bit easier to use to so that callers don't have to pass in `rusage` and `OUR_RUSAGE_SELF`.

Sources/performance-worker/ClientStats.swift
```diff
@@ -0,0 +1,73 @@
...
+    time: Double,
+    userTime: Double,
+    systemTime: Double
+  ) {
+    self.time = time
+    self.userTime = userTime
+    self.systemTime = systemTime
+  }
+
+  init() async throws {
```
glbrntt, line 52: This isn't async or throwing

Sources/performance-worker/RPCStats.swift
```diff
@@ -0,0 +1,162 @@
...
+
+import Foundation
+import NIOConcurrencyHelpers
+
+/// Stores the real time latency histogram and error code count dictionary,
+/// for the RPCs made by a particular GRPCClient. It gets updated after
+/// each finished RPC.
+struct RPCStats {
+  var latencyHistogram: LatencyHistogram
+  var requestResultCount: [Int32: Int64]
```
glbrntt, line 25: Any reason we can't use `Status.Code` as the key?

Sources/performance-worker/RPCStats.swift
```diff
@@ -0,0 +1,162 @@
...
+    return Double(self.value) * 1e-9
+  }
+}
+
+extension Nanoseconds: CustomStringConvertible {
+  /// Description to aid debugging.
+  var description: String {
+    return "\(self.value) ns"
+  }
+}
```
glbrntt, line 162: I think we should just use `DispatchTime` to be honest.

Sources/performance-worker/RPCStats.swift
```diff
@@ -0,0 +1,162 @@
...
+        ) + 1
+      precondition(numBuckets > 1)
+      precondition(numBuckets < 100_000_000)
+      self.buckets = .init(repeating: 0, count: numBuckets)
+    }
+
+    struct HistorgramShapeMismatch: Error {}
+
+    /// Determine a bucket index given a value - does no bounds checking
+    private static func bucketForUnchecked(value: Double, oneOnLogMultiplier: Double) -> Int {
```
glbrntt, line 75:
```suggestion
    private static func uncheckedBucket(forValue value: Double, oneOnLogMultiplier: Double) -> Int {
```

Sources/performance-worker/RPCStats.swift
```diff
@@ -0,0 +1,162 @@
...
+      return Int(log(value) * oneOnLogMultiplier)
+    }
+
+    private func bucketFor(value: Double) -> Int {
```
glbrntt, line 79:
```suggestion
    private func bucket(forValue value: Double) -> Int {
```

Sources/performance-worker/RPCStats.swift
```diff
@@ -0,0 +1,162 @@
...
+      )
+      assert(bucket < self.buckets.count)
+      assert(bucket >= 0)
+      return bucket
+    }
+
+    /// Add a value to this histogram, updating buckets and stats
+    /// - parameters:
+    ///     - value: The value to add.
+    public mutating func add(value: Double) {
```
glbrntt, line 92:
```suggestion
    public mutating func record(_ value: Double) {
```

Sources/performance-worker/RPCStats.swift
```diff
@@ -0,0 +1,162 @@
...
+      if value > self.maxSeen {
+        self.maxSeen = value
+      }
+      self.buckets[self.bucketFor(value: value)] += 1
+    }
+
+    /// Merge two histograms together updating `self`
+    /// - parameters:
+    ///    - source: the other histogram to merge into this.
+    public mutating func merge(source: LatencyHistogram) throws {
```
glbrntt, line 108:
```suggestion
    public mutating func merge(_ other: LatencyHistogram) throws {
```

glbrntt (changes requested):

Sources/performance-worker/ClientStats.swift
stefanadranca: Want to make sure I understood right, do you mean the function to be like this?
`extension System {
  static func getUserAndSystemTime() -> (userTime: Double, systemTime: Double) {
    let userTime, systemTime: Double
  
    #if canImport(Darwin)
    let OUR_RUSAGE_SELF: Int32 = RUSAGE_SELF
    #elseif canImport(Musl) || canImport(Glibc)
    let OUR_RUSAGE_SELF: Int32 = RUSAGE_SELF.rawValue
    #endif
    
    var usage = rusage()
    if getrusage(OUR_RUSAGE_SELF, &usage) == 0 {
      // Adding the seconds with the microseconds transformed into seconds to get the
      // real number of seconds as a `Double`.
      userTime = Double(usage.ru_utime.tv_sec) + Double(usage.ru_utime.tv_usec) * 1e-6
      systemTime = Double(usage.ru_stime.tv_sec) + Double(usage.ru_stime.tv_usec) * 1e-6
    } else {
      userTime = 0
      systemTime = 0
    }
    
    return (userTime, systemTime)
  }
}`

Sources/performance-worker/RPCStats.swift
stefanadranca: not necessarily, the output message representing this counts uses an Int32 for the statusCode, but this doesn't mean the transformation can't be done exactly when creating the response, so I will transform the keys of this dictionary to be of Status.Code type

Sources/performance-worker/BenchmarkClient.swift
glbrntt: Oh okay, you're right then, just error codes.

Sources/performance-worker/ClientStats.swift
glbrntt: I was actually thinking we can just return `rusage`:

```
extension System {
  static func resourceUsage() -> rusage {
    // ...
  }
}
```

Sources/performance-worker/RPCStats.swift
glbrntt: Given the comments above about error vs. status, we should actually use `RPCError.Code` instead. We should use the "right" types from an API perspective and then convert to whatever the proto wants when we populate it later.

Sources/performance-worker/ResourceUsage.swift
```diff
@@ -128,3 +164,17 @@ internal struct ServerStats: Sendable {
...
+  fileprivate static func resourceUsage() throws -> rusage {
+    var usage = rusage()
+
+    if getrusage(OUR_RUSAGE_SELF, &usage) == 0 {
+      return usage
+    } else {
+      throw SystemError()
+    }
+  }
+}
```
glbrntt, line 180: Rather than throwing an error we can just return `nil` instead.

Sources/performance-worker/RPCStats.swift
```diff
@@ -0,0 +1,130 @@
...
+
+import Foundation
+import GRPCCore
+import NIOConcurrencyHelpers
+
+/// Stores the real time latency histogram and error code count dictionary,
+/// for the RPCs made by a particular GRPCClient. It gets updated after
+/// each finished RPC.
+struct RPCStats {
+  var latencyHistogram: LatencyHistogram
```
glbrntt, line 25: Can you add a comment saying that this is in nanoseconds?

glbrntt (changes requested):

glbrntt (approved):

PR #1852, opened 2024-04-09
Add subpool stats

glbrntt:
Modification:

It can be helpful to track the state of the connection pool as a whole over time to understand how heavily utilised it is. This is awkward to do at the moment because delegate functions are called on connection state changes. There's also no insight into how many RPCs are queued waiting for a stream.

Modifications:

- Add a 'GRPCSubPoolStats' type capturing stats from each subpool. This includes:
  - counts of connections in each state
  - streams in use
  - streams which are free to use
  - number of rpcs waiting for a stream
- Add a delegate method which is passed an array of subpool stats (one per subpool) and add a no-op default implementation.
- Add configuration to determine how frequently stats are collected

Result:

More insight into pool stats

### DIFF
--- a/Sources/GRPC/ConnectionPool/GRPCChannelPool.swift
+++ b/Sources/GRPC/ConnectionPool/GRPCChannelPool.swift
@@ -179,6 +179,11 @@ extension GRPCChannelPool {
     /// pool.
     public var delegate: GRPCConnectionPoolDelegate?
 
+    /// The period at which connection pool stats are published to the ``delegate``.
+    ///
+    /// Ignored if either this value or ``delegate`` are `nil`.
+    public var statsPeriod: TimeAmount?
+
     /// A logger used for background activity, such as connection state changes.
     public var backgroundActivityLogger = Logger(
       label: "io.grpc",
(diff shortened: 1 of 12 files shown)


Sources/GRPC/ConnectionPool/GRPCChannelPool.swift
```diff
@@ -365,4 +370,66 @@ public protocol GRPCConnectionPoolDelegate: Sendable {
...
+  public func connectionPoolStats(_ stats: [GRPCSubPoolStats], id: GRPCConnectionPoolID) {
+    // Default conformance to avoid breaking changes.
+  }
+}
+
+public struct GRPCSubPoolStats: Sendable, Hashable {
+  public struct ConnectionStates: Sendable, Hashable {
+    /// The number of idle connections.
+    public var idle: Int
+    /// The number of connections try to establish a connection.
```
gjcairo, line 394: Nit
```suggestion
    /// The number of connections trying to establish a connection.
```

Sources/GRPC/ConnectionPool/PoolManager.swift
```diff
@@ -353,6 +377,18 @@ internal final class PoolManager {
...
     }
   }
+
+  // MARK: - Stats
+
+  private func emitStats(delegate: GRPCConnectionPoolDelegate) {
+    let pools = self.lock.withLock { self._pools }
+    if pools.isEmpty { return }
+
+    let statsFutures = pools.map { $0.stats() }
```
gjcairo, line 387: Should we capture the stats while we're holding the lock for the pools? We could for example shut down some of the pools in between the capture and doing this map, which isn't a snapshot of an instant in time of all pools. I don't know if we care about this or not though.

gjcairo (commented):
LGTM overall, just left a question and a nit

Sources/GRPC/ConnectionPool/PoolManager.swift
glbrntt: I don't think it matters to be honest; each pool needs to exec onto its event loop to get stats so they won't be from the same instant anyway.

gjcairo (approved):

PR #1856, opened 2024-04-11
Rename a few things

glbrntt:
Motivation:

"Service config" is the name of a gRPC feature, yet we've named the corresponding type "ServiceConfiguration". We should follow the gRPC naming here.

In a similar vein, "keepAlive" should be "keepalive".

Modifications:

- Rename "ServiceConfiguration" to "ServiceConfig"
- Rename "MethodConfiguration" to "MethodConfig"
- Rename various "keepAlive"s to "keepalive"

Result:

Naming is more consistent with gRPC

### DIFF
--- a/Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionManagementHandler+StateMachine.swift
+++ b/Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionManagementHandler+StateMachine.swift
@@ -33,19 +33,19 @@ extension ServerConnectionManagementHandler {
     /// Create a new state machine.
     ///
     /// - Parameters:
-    ///   - allowKeepAliveWithoutCalls: Whether the client is permitted to send keep alive pings
+    ///   - allowKeepaliveWithoutCalls: Whether the client is permitted to send keep alive pings
     ///       when there are no active calls.
     ///   - minPingReceiveIntervalWithoutCalls: The minimum time interval required between keep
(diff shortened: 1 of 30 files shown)


Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionManagementHandler+StateMachine.swift
```diff
@@ -44,12 +44,12 @@ extension ServerConnectionManagementHandler {
       minPingReceiveIntervalWithoutCalls: TimeAmount,
       goAwayPingData: HTTP2PingData = HTTP2PingData(withInteger: .random(in: .min ... .max))
     ) {
-      let keepAlive = KeepAlive(
+      let keepalive = KeepAlive(
```
gjcairo, line 47: Do we want to rename `KeepAlive` to `Keepalive`? I really dislike it but I wonder if it may be slightly more consistent :D

gjcairo (approved):
LGTM, just a small question

Sources/GRPCHTTP2Core/Server/Connection/ServerConnectionManagementHandler+StateMachine.swift
glbrntt: Good catch, yes we should rename it. It might look weird, but it is correct: https://en.wikipedia.org/wiki/Keepalive

PR #1857, opened 2024-04-12
runClient RPC response in performance worker

stefanadranca:
Motivation:

We need the `runClient` RPC to monitor the performance of the benchmark clients.

Modifications:

- implemented the methods that set up the clients 
- implemented the methods that merge the histograms and responses count
- implemented the method that creates the response for `runClient`
- implemented the `runClient` method

Result:

We can construct a latency histogram that represents the data from all clients and the `runClient` method returns the stats avout the clients started by the worker.

### DIFF
--- a/Sources/performance-worker/RPCStats.swift
+++ b/Sources/performance-worker/RPCStats.swift
@@ -131,14 +131,12 @@ struct RPCStats {
   }
 
   @available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
-  mutating func update(withStatsFrom clients: [BenchmarkClient]) throws {
-    for benchmarkClient in clients {
-      try self.latencyHistogram.merge(
-        benchmarkClient.currentStats.latencyHistogram
-      )
-      self.requestResultCount.merge(self.requestResultCount) { (current, new) in
-        current + new
-      }
+  mutating func merge(_ other: RPCStats) throws {
+    try self.latencyHistogram.merge(
+      other.latencyHistogram
+    )
+    self.requestResultCount.merge(other.requestResultCount) { (current, new) in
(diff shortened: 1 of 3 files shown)


Sources/performance-worker/RPCStats.swift
```diff
@@ -129,4 +129,16 @@ struct RPCStats {
       }
     }
   }
+
+  @available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+  mutating func update(withStatsFrom clients: [BenchmarkClient]) throws {
```
glbrntt, line 134: This unnecessarily couples the `BenchmarkClient` to `RPCStats`; the only way to update it is using a `BenchmarkClient` which is unnecessary and makes testing it more difficult.

I think the API you want here is the simplest one: `merge(_ other: RPCStats)`, the caller can then iterate over benchmark clients and call this function each time.

Sources/performance-worker/WorkerService.swift
```diff
@@ -237,4 +320,84 @@ extension WorkerService {
...
+    reset: Bool
+  ) async throws -> Grpc_Testing_WorkerService.Method.RunClient.Output {
+    let currentStats = ClientStats()
+    let initialStats = self.state.withLockedValue { state in
+      return state.clientStats(replaceWith: reset ? currentStats : nil)
+    }
+    let rpcStats = try self.state.withLockedValue { state in
+      try state.updateRPCStats()
+      return state.clientRPCStats
+    }
```
glbrntt, line 368: Can we acquire the lock just once here?

glbrntt (changes requested):

glbrntt (approved):

PR #1861, opened 2024-04-17
BenchmarkClient RPCs implementation

stefanadranca:
Motivation:

When the WorkerService holds and monitors BenchmarkClients, they need to make the requests to the server, as configured in the config input the WorkerService receives from the Test Driver.

Modifications:

- implemented a helper function that computes the latency and extracts the error code for a RPC
- implemented the body of the makeRPC function that makes one of the 5 possible RPCs (implemented the RPCs)

Result:

The BenchmarkClient implementation for performance testing will be completed.

### DIFF
--- a/Sources/performance-worker/BenchmarkClient.swift
+++ b/Sources/performance-worker/BenchmarkClient.swift
@@ -22,7 +22,7 @@ import NIOConcurrencyHelpers
 struct BenchmarkClient {
   private var client: GRPCClient
   private var rpcNumber: Int32
-  private var rpcType: RpcType
+  private var rpcType: RPCType
   private var messagesPerStream: Int32
   private var protoParams: Grpc_Testing_SimpleProtoParams
   private let rpcStats: NIOLockedValueBox<RPCStats>
@@ -30,11 +30,11 @@ struct BenchmarkClient {
   init(
     client: GRPCClient,
     rpcNumber: Int32,
-    rpcType: RpcType,
+    rpcType: RPCType,
     messagesPerStream: Int32,
     protoParams: Grpc_Testing_SimpleProtoParams,
     histogramParams: Grpc_Testing_HistogramParams?
-  ) throws {
+  ) {
     self.client = client
     self.rpcNumber = rpcNumber
     self.messagesPerStream = messagesPerStream
@@ -54,7 +54,7 @@ struct BenchmarkClient {
     self.rpcStats = NIOLockedValueBox(RPCStats(latencyHistogram: histogram))
   }
 
-  enum RpcType {
+  enum RPCType {
     case unary
     case streaming
     case streamingFromClient
@@ -122,8 +122,7 @@ struct BenchmarkClient {
         do {
           try await benchmarkClient.unaryCall(
             request: ClientRequest.Single(message: message)
-          ) {
-            response in
+          ) { response in
             _ = try response.message
           }
           return nil
@@ -178,7 +177,7 @@ struct BenchmarkClient {
             }
           }
 
-          _ = try await benchmarkClient.streamingFromClient(
+          try await benchmarkClient.streamingFromClient(
             request: streamingRequest
           ) { response in
             _ = try response.message
--- a/Sources/performance-worker/BenchmarkClient.swift
+++ b/Sources/performance-worker/BenchmarkClient.swift
@@ -22,8 +22,9 @@ import NIOConcurrencyHelpers
 struct BenchmarkClient {
   private var client: GRPCClient
   private var rpcNumber: Int32
-  private var rpcType: Grpc_Testing_RpcType
+  private var rpcType: RpcType
   private var messagesPerStream: Int32
+  private var protoParams: Grpc_Testing_SimpleProtoParams
   private let rpcStats: NIOLockedValueBox<RPCStats>
 
   init(
@@ -31,12 +32,20 @@ struct BenchmarkClient {
     rpcNumber: Int32,
     rpcType: Grpc_Testing_RpcType,
     messagesPerStream: Int32,
+    protoParams: Grpc_Testing_SimpleProtoParams,
     histogramParams: Grpc_Testing_HistogramParams?
-  ) {
+  ) throws {
     self.client = client
     self.rpcNumber = rpcNumber
-    self.rpcType = rpcType
     self.messagesPerStream = messagesPerStream
+    self.protoParams = protoParams
+
+    switch rpcType {
+    case .unary, .streaming, .streamingFromClient, .streamingFromServer, .streamingBothWays:
+      self.rpcType = RpcType(rawValue: rpcType.rawValue)!
+    case .UNRECOGNIZED:
+      throw RPCError(code: .unknown, message: "The RPC type is UNRECOGNIZED.")
+    }
 
     let histogram: RPCStats.LatencyHistogram
     if let histogramParams = histogramParams {
@@ -49,6 +58,14 @@ struct BenchmarkClient {
     }
 
     self.rpcStats = NIOLockedValueBox(RPCStats(latencyHistogram: histogram))
+  }
+
+  enum RpcType: Int {
+    case unary
+    case streaming
+    case streamingFromClient
+    case streamingFromServer
+    case streamingBothWays
   }
 
   internal var currentStats: RPCStats {
@@ -71,9 +88,6 @@ struct BenchmarkClient {
               benchmarkClient: benchmarkClient,
               rpcType: self.rpcType
             )
-            guard errorCode != RPCError.Code.unknown else {
-              throw RPCError(code: .unknown, message: "The RPC type is UNRECOGNIZED.")
-            }
             self.rpcStats.withLockedValue {
               $0.latencyHistogram.record(latency)
               if let errorCode = errorCode {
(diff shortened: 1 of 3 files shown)


Sources/performance-worker/BenchmarkClient.swift
```diff
@@ -80,18 +89,123 @@ struct BenchmarkClient {
     }
   }
 
+  private func computeTimeAndErrorCode<Contents>(
```
glbrntt, line 92: It's a bit strange to link the timing with the capturing of error codes. Aim for simple: have a function do one thing; time the execution of the body and return the result:

```swift
func timeIt<R>(_ body: () async throws -> R) async rethrows -> (R, seconds: Double) {
  // ...
}
```

This means at the call site you can return through the error code if applicable:

```swift
let (errorCode, seconds) = await timeIt { 
  do {
    try await client.whatever() 
    return nil
  } catch let error as RPCError {
    return error.code
  } catch {
    return .unknown
  }
}
```

Sources/performance-worker/BenchmarkClient.swift
```diff
@@ -80,18 +89,123 @@ struct BenchmarkClient {
...
+        ids.continuation.yield(1)
+
+        let responseStatus = try await benchmarkClient.streamingCall(request: streamingRequest) {
+          response in
+          var id = 1
+          for try await _ in response.messages {
+            id += 1
+            ids.continuation.yield(id)
+          }
+          return response.accepted
```
glbrntt, line 156: No need to return `accepted` here, we'll throw any `RPCError` from `response.message`, so if you get to the `return` here you know there's been no error.

Effectively we know that if we don't throw that the error code should be `nil`.

Sources/performance-worker/BenchmarkClient.swift
```diff
@@ -80,18 +89,123 @@ struct BenchmarkClient {
...
+        return responseStatus
+      }
+
+    case .streamingFromServer:
+      return try await self.computeTimeAndErrorCode { request in
+        let responseStatus = try await benchmarkClient.streamingFromServer(
+          request: ClientRequest.Single(message: request)
+        ) { response in
+          return response.accepted
+        }
```
glbrntt, line 185: We need to consume the messages on the response: if we don't consume them we'll exert backpressure and the server will stop writing messages.

Sources/performance-worker/BenchmarkClient.swift
```diff
@@ -80,18 +89,123 @@ struct BenchmarkClient {
...
+      return try await self.computeTimeAndErrorCode { request in
+        let streamingRequest = ClientRequest.Stream { writer in
+          for _ in 1 ... self.messagesPerStream {
+            try await writer.write(request)
+          }
+        }
+
+        let responseStatus = try await benchmarkClient.streamingBothWays(request: streamingRequest)
+        { response in
+          return response.accepted
```
glbrntt, line 200: Same note here re: consuming the messages

glbrntt (changes requested):

Sources/performance-worker/BenchmarkClient.swift
```diff
@@ -80,18 +89,123 @@ struct BenchmarkClient {
     }
   }
 
+  private func computeTimeAndErrorCode<Contents>(
+    _ body: (Grpc_Testing_SimpleRequest) async throws -> Result<Contents, RPCError>
+  ) async throws -> (latency: Double, errorCode: RPCError.Code?) {
+    let request = Grpc_Testing_SimpleRequest.with {
+      $0.responseSize = 10
```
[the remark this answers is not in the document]
stefanadranca: no particular reason, I was thinking we can create it a single time and set a random size. Is there a specific size I should set?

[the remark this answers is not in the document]
glbrntt: Hmm, does the client config not tell us what size to use?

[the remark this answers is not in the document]
stefanadranca: you're right, I forgot about the SimpleProtoParams - they specify the request and response sizes

Sources/performance-worker/BenchmarkClient.swift
```diff
@@ -48,6 +60,14 @@ struct BenchmarkClient {
     self.rpcStats = NIOLockedValueBox(RPCStats(latencyHistogram: histogram))
   }
 
+  enum RpcType: Int {
```
glbrntt, line 63: nit: `RPCType`, also no need for it to be raw representable as an `Int` here

Sources/performance-worker/BenchmarkClient.swift
```diff
@@ -22,18 +22,30 @@ import NIOConcurrencyHelpers
...
-    self.rpcType = rpcType
+    self.messagesPerStream = messagesPerStream
+    self.protoParams = protoParams
+
+    switch rpcType {
+    case .unary, .streaming, .streamingFromClient, .streamingFromServer, .streamingBothWays:
+      self.rpcType = RpcType(rawValue: rpcType.rawValue)!
+    case .UNRECOGNIZED:
+      throw RPCError(code: .unknown, message: "The RPC type is UNRECOGNIZED.")
+    }
```
glbrntt, line 48: Can we do this _before_ creating the client?

Sources/performance-worker/BenchmarkClient.swift
```diff
@@ -64,7 +84,10 @@ struct BenchmarkClient {
       try await withThrowingTaskGroup(of: Void.self) { rpcsGroup in
         for _ in 0 ..< self.rpcNumber {
           rpcsGroup.addTask {
-            let (latency, errorCode) = self.makeRPC(client: benchmarkClient, rpcType: self.rpcType)
+            let (latency, errorCode) = try await self.makeRPC(
+              benchmarkClient: benchmarkClient,
+              rpcType: self.rpcType
```
glbrntt, line 89: We don't need to pass this in because it's stored on `self`

Sources/performance-worker/BenchmarkClient.swift
```diff
@@ -80,19 +103,138 @@ struct BenchmarkClient {
...
+
   // The result is the number of nanoseconds for processing the RPC.
   private func makeRPC(
-    client: Grpc_Testing_BenchmarkServiceClient,
-    rpcType: Grpc_Testing_RpcType
-  ) -> (latency: Double, errorCode: RPCError.Code?) {
+    benchmarkClient: Grpc_Testing_BenchmarkServiceClient,
+    rpcType: RpcType
+  ) async throws -> (latency: Double, errorCode: RPCError.Code?) {
+    let request = Grpc_Testing_SimpleRequest.with {
```
glbrntt, line 120: Can we call this `message`? I think of the "request" as the overall input to the RPC, i.e. it includes the metadata as well. A "message" or "messages" are a part of the request.

Sources/performance-worker/BenchmarkClient.swift
```diff
@@ -80,19 +103,138 @@ struct BenchmarkClient {
...
-      let endTime = DispatchTime.now().uptimeNanoseconds
-      return (
-        latency: Double(endTime - startTime), errorCode: RPCError.Code(.unimplemented)
-      )
+    case .unary:
+      let (errorCode, nanoseconds): (RPCError.Code?, Double) = await self.timeIt {
+        do {
+          try await benchmarkClient.unaryCall(request: ClientRequest.Single(message: request)) {
+            response in
+            _ = try response.message
```
glbrntt, line 133:
```suggestion
          try await benchmarkClient.unaryCall(
            request: ClientRequest.Single(message: request)
          ) { response in
            _ = try response.message
```

glbrntt (changes requested):
A few nits but looks good otherwise!

Sources/performance-worker/BenchmarkClient.swift
```diff
@@ -48,6 +54,14 @@ struct BenchmarkClient {
     self.rpcStats = NIOLockedValueBox(RPCStats(latencyHistogram: histogram))
   }
 
+  enum RpcType {
```
glbrntt, line 57: nit: should be `RPCType`

Sources/performance-worker/BenchmarkClient.swift
```diff
@@ -22,17 +22,23 @@ import NIOConcurrencyHelpers
...
   init(
     client: GRPCClient,
     rpcNumber: Int32,
-    rpcType: Grpc_Testing_RpcType,
+    rpcType: RpcType,
+    messagesPerStream: Int32,
+    protoParams: Grpc_Testing_SimpleProtoParams,
     histogramParams: Grpc_Testing_HistogramParams?
-  ) {
+  ) throws {
```
glbrntt, line 37: This isn't throwing

Sources/performance-worker/BenchmarkClient.swift
```diff
@@ -80,19 +96,139 @@ struct BenchmarkClient {
...
+    }
+
+    switch self.rpcType {
+    case .unary:
+      let (errorCode, nanoseconds): (RPCError.Code?, Double) = await self.timeIt {
+        do {
+          try await benchmarkClient.unaryCall(
+            request: ClientRequest.Single(message: message)
+          ) {
+            response in
```
glbrntt, line 126:
```suggestion
          ) { response in
```

Sources/performance-worker/BenchmarkClient.swift
```diff
@@ -80,19 +96,139 @@ struct BenchmarkClient {
...
+    case .streamingFromClient:
+      let (errorCode, nanoseconds): (RPCError.Code?, Double) = await self.timeIt {
+        do {
+          let streamingRequest = ClientRequest.Stream { writer in
+            for _ in 1 ... self.messagesPerStream {
+              try await writer.write(message)
+            }
+          }
+
+          _ = try await benchmarkClient.streamingFromClient(
```
glbrntt, line 181: This one shouldn't be necessary as the body of the closure doesn't return anything
```suggestion
          try await benchmarkClient.streamingFromClient(
```

glbrntt (changes requested):
A few things got missed from last time but basically looks good!

glbrntt (approved):

PR #1864, opened 2024-04-22
Delay the point at which a connect 'succeeds'

glbrntt:
Motivation:

One issue with the connection merged in #1859 is that it considers connect succeeded to be the point at which the TCP connection succeeds. However, when considered as an HTTP/2 connection, this is too early: the connection may fail a TLS handshake or the server might not be an HTTP/2 server. The connect should only be considered a success when the initial SETTINGS frame is received from the server.

Modifications:

- Add a new connection event 'ready' and react to this appropriately in the connection
- Add tests

Result:

The `Connection` is only considered ready when the first SETTINGS frame is received.

### DIFF
--- a/Sources/GRPCHTTP2Core/Client/Connection/ClientConnectionHandler.swift
+++ b/Sources/GRPCHTTP2Core/Client/Connection/ClientConnectionHandler.swift
@@ -31,6 +31,9 @@ public enum ClientConnectionEvent: Sendable, Hashable {
     case initiatedLocally
   }
 
+  /// The connection is now ready.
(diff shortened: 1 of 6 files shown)


gjcairo (approved):

glbrntt:
This has somehow already been merged https://github.com/grpc/grpc-swift/commit/2d56a49303eae16889a5707e7369eba727701ddd ... and yet the PR says it hasn't and CI is still running.

Given the changes are now on `main` I'm closing this despite it not having been merged yet... even though it has.

PR #1865, opened 2024-04-22
Make ServerTransport protocol possible to use with structured concurrency

gjcairo:
## Motivation
The existing `ServerProtocol` has some design issues: its `listen` method returns the async sequence of streams to be handled, while at the same time it's `async` and must run for as long as the transport is up and running. This means we'll never get the sequence of streams back in most transport implementations (such as H2).

## Modifications
This PR tries to solve this by changing the protocol so that the `listen` method now takes a closure which allows handling an individual stream. This forces the transport implementations to maintain the right structured task hierarchy, as tasks to run this stream handler will have to be added as children of the long-running "listen" task.

## Result
Server Transports can now be used in structured concurrency.


Sources/GRPCCore/Transport/ServerTransport.swift
```diff
@@ -19,17 +19,31 @@ public protocol ServerTransport: Sendable {
...
+  /// A sequence of accepted streams to handle.
+  ///
+  /// This property is `async` because implementations will typically require ``listen()`` to be called
+  /// (and for the startup to be succesful) before a stream sequence can be returned.
+  ///
+  /// If the call to ``listen()`` throws, meaning the transport failed to start, implementations of this transport
+  /// will typically throw upon getting this property.
+  ///
+  /// Once ``listen()`` stops running, the sequence will be finished.
+  var acceptedStreams: RPCAsyncSequence<RPCStream<Inbound, Outbound>> { get async throws }
```
glbrntt, line 31: I like the semantics here (i.e. throw if listen failed) but I'm not sure it's that easy to program against if you're implementing a transport. (I'm less concerned about using it because we only do that once.)

I wonder if we should change this to be a non-throwing `AsyncSequence` of events and have two events:
- `startedListening(RPCAsyncSequence<...>)`
- `failedToStartListening(any Error)` (or a better name...)

That way we can remove the `async` and `throws` from here. It also means that implementations of the transport protocol don't have to worry about e.g. multiple consumers of this property (the async sequence implementation would take care of this) or cancellation when getting the sequence (because it wouldn't be `async`). WDYT?

Sources/GRPCCore/Transport/ServerTransport.swift
```diff
@@ -15,21 +15,37 @@
...
 
-  /// Starts the transport and returns a sequence of accepted streams to handle.
+  /// A sequence of ``ListenEvent``s, describing whether the transport was started successfully or not.
+  ///
+  /// This sequence should only return a single event. If multiple events are yielded into it, they may be ignored.
+  /// Not yielding any events or finishing the sequence without yielding any events is considered a broken
+  /// implementation of this protocol.
+  ///
+  /// Once ``listen()`` stops running, the sequence should be finished if it hasn't been finished already.
+  var listenEventStream: RPCAsyncSequence<ListenEvent> { get }
```
glbrntt, line 32: I think we should broaden this slightly to be "transport events". That gives us a bit of scope in the future to extend what a transport can do. We can probably just call the property "events" and the type "TransportEvent".

glbrntt, line 32: Can we add a non-throwing version of `RPCAsyncSequnce`?  `NoThrowRPCAsyncSequnce`? That way we tidy up the semantics a bit: failing to listen always results in an event rather than an error.

glbrntt (changes requested):
I think the shape is right but we can eliminate some edge cases by removing the throws and make it less restrictive by using a more general event type.

Sources/GRPCCore/Transport/ServerTransport.swift
gjcairo: Okay, this makes sense. Should I turn the `TransportEvent` into its own type instead of being a `typealias` then?
That's what I had originally done but I turned it into an alias of `Result` because it was basically the same and felt unnecessary. But if we want it to be extendable maybe we want it to be an actual type?

Sources/GRPCCore/Transport/ServerTransport.swift
```diff
@@ -15,21 +15,37 @@
  */
 
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+/// A protocol server transport implementations must conform to.
 public protocol ServerTransport: Sendable {
   typealias Inbound = RPCAsyncSequence<RPCRequestPart>
   typealias Outbound = RPCWriter<RPCResponsePart>.Closable
+  typealias AcceptedStreams = RPCAsyncSequence<RPCStream<Inbound, Outbound>>
+  typealias ListenEvent = Result<AcceptedStreams, any Error>
```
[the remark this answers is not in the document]
gjcairo: Ah, got my answer to the previous question :D

Sources/GRPCCore/GRPCServer.swift
```diff
@@ -207,40 +207,59 @@ public struct GRPCServer: Sendable {
...
-        throw RuntimeError(
-          code: .failedToStartTransport,
-          message: """
-            Server didn't start because the '\(type(of: transport))' transport threw an error \
-            while starting.
-            """,
-          cause: cause
-        )
+        // Get the first event only: we don't care about any further events
+        // (we shouldn't even receive more than one)
```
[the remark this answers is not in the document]
gjcairo: You think it's worth adding an assertion in the `GRPCServer` that checks only one accepted streams sequence is published?

Sources/GRPCInProcessTransport/InProcessServerTransport.swift
```diff
@@ -55,12 +62,9 @@ public struct InProcessServerTransport: ServerTransport, Sendable {
...
 
-  /// Return a new ``RPCAsyncSequence`` that will contain all published ``RPCStream``s published
-  /// to this transport using the ``acceptStream(_:)`` method.
-  ///
-  /// - Returns: An ``RPCAsyncSequence`` of all published ``RPCStream``s.
-  public func listen() async throws -> RPCAsyncSequence<RPCStream<Inbound, Outbound>> {
-    RPCAsyncSequence(wrapping: self.newStreams)
+  // Signal that the stream is up and running.
+  public func listen() async {
+    self.eventStreamContinuation.yield(.success(RPCAsyncSequence(wrapping: self.newStreams)))
```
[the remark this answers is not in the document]
gjcairo: This transport doesn't really have any long-running work to do here. Is it really necessary to not return until `stopListening()` has been called? The requests will be handled anyways because the server will still be holding the accepted streams sequence.

Sources/GRPCCore/GRPCServer.swift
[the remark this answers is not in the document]
glbrntt: Yeah, that's a good idea

Sources/GRPCInProcessTransport/InProcessServerTransport.swift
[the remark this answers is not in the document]
glbrntt: It doesn't here, but in general they will. Maybe the distinction doesn't matter so much, need to think about it some more.

gjcairo:
Had to update the `swift-nio-http2` version because for some reason my local package resolution was bringing a way too old version (1.27.0) that didn't have some types we're using.

Sources/GRPCCore/GRPCServer.swift
```diff
@@ -207,40 +207,49 @@ public struct GRPCServer: Sendable {
...
 
-        throw RuntimeError(
-          code: .failedToStartTransport,
-          message: """
-            Server didn't start because the '\(type(of: transport))' transport threw an error \
-            while starting.
-            """,
-          cause: cause
-        )
+        for try await event in transport.events {
```
glbrntt, line 217: We'll need to do each of these in a separate task, otherwise we'll block subsequent transports from starting

Sources/GRPCCore/Transport/ServerTransport.swift
```diff
@@ -15,21 +15,78 @@
  */
 
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+/// A type representing the possible outcomes of calling ``ServerTransport/listen()``.
+public struct TransportEvent: Sendable {
+  public typealias AcceptedStreams = RPCAsyncSequence<
+    RPCStream<ServerTransport.Inbound, ServerTransport.Outbound>
+  >
+
+  private enum Event {
```
glbrntt, line 24: Can you make this explicitly `Sendable`?

Sources/GRPCInProcessTransport/InProcessServerTransport.swift
```diff
@@ -55,12 +62,11 @@ public struct InProcessServerTransport: ServerTransport, Sendable {
...
 
-  /// Return a new ``RPCAsyncSequence`` that will contain all published ``RPCStream``s published
-  /// to this transport using the ``acceptStream(_:)`` method.
-  ///
-  /// - Returns: An ``RPCAsyncSequence`` of all published ``RPCStream``s.
-  public func listen() async throws -> RPCAsyncSequence<RPCStream<Inbound, Outbound>> {
-    RPCAsyncSequence(wrapping: self.newStreams)
+  // Signal that the stream is up and running.
+  public func listen() async {
+    self.eventStreamContinuation.yield(
```
glbrntt, line 67: We need to finish this stream at some point, probably in `stopListening()`

Tests/GRPCCoreTests/Test Utilities/Transport/ThrowingTransport.swift
```diff
@@ -53,13 +53,23 @@ struct ThrowOnStreamCreationTransport: ClientTransport {
...
+  var events: NoThrowRPCAsyncSequence<TransportEvent> {
+    NoThrowRPCAsyncSequence(
+      wrapping: AsyncStream {
+        .failedToStartListening(
+          RPCError(
+            code: .unavailable,
+            message: "The '\(type(of: self))' transport is never available."
+          )
+        )
+      }
```
glbrntt, line 65: Does this `init` finish the continuation for you?

Sources/GRPCCore/Transport/ServerTransport.swift
```diff
@@ -15,21 +15,78 @@
  */
 
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+/// A type representing the possible outcomes of calling ``ServerTransport/listen()``.
```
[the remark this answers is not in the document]
gjcairo: Ah yes, this was outdated from before we decided to make it extendable - updated it

Sources/GRPCCore/Transport/ServerTransport.swift
```diff
@@ -15,21 +15,78 @@
...
+    case .startedListening(let acceptedStreams):
+      return .success(acceptedStreams)
+    case .failedToStartListening(let error):
+      return .failure(error)
+    }
+  }
+}
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+/// A protocol server transport implementations must conform to.
```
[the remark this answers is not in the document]
gjcairo: I'd assume it does but I'll swap it just in case

Sources/GRPCCore/GRPCServer.swift
gjcairo: Ah, this wasn't being done in a separate task because when I first implemented, the events were only listen-related.

You're right that if we keep this a generic "transport event" stream then we need to have it run as long as listen runs. However this makes things trickier because we are appending the published accepted streams to the `listeners` array, and we don't want to use that `listeners` array until all of the transports have started and they've all published their accepted streams. If the `events` stream is long running, as well as the call to `listen()`, we can't simply wait on the group's tasks to be finished before we call `handleRequests(listeners)`.

I'm not sure how to wait for all the published accepted streams to be present - we basically need to wait for all transports to be ready, which is what we added this even stream for.

Tests/GRPCCoreTests/Test Utilities/Transport/ThrowingTransport.swift
gjcairo: Sorry, not sure I follow - which continuation are we talking about?

Sources/GRPCCore/GRPCServer.swift
```diff
@@ -204,140 +204,42 @@ public struct GRPCServer: Sendable {
...
-                let taskCount = tasks.wrappingIncrementThenLoad(ordering: .sequentiallyConsistent)
-                if taskCount >= maxConcurrentSubTasks {
-                  _ = await subGroup.next()
-                  tasks.wrappingDecrement(ordering: .sequentiallyConsistent)
-                }
+      while !group.isEmpty {
+        do {
+          try await group.next()
+        } catch {
+          // Failed to start some transport, so start stopping.
```
FranzBusch, line 233: Same here this might happen while the transport is already running

Sources/GRPCCore/GRPCServer.swift
```diff
@@ -204,140 +204,42 @@ public struct GRPCServer: Sendable {
...
-              }
-            } catch {
-              // If the listener threw then the connection must be broken, cancel all work.
-              subGroup.cancelAll()
-            }
+          // Some transports may have started and have streams which need closing:
+          // tell them to stop listening.
+          for transport in self.transports {
+            transport.stopListening()
           }
```
FranzBusch, line 240: Is stopping the right thing or should we just cancel? It could be really problematic if somebody starts a server with multiple transports and then one fails but the service doesn't exit because the graceful shutdown is still ongoing. IMO a hard cancellation here is better.

Sources/GRPCInProcessTransport/InProcessServerTransport.swift
```diff
@@ -55,12 +55,17 @@ public struct InProcessServerTransport: ServerTransport, Sendable {
...
-  /// Return a new ``RPCAsyncSequence`` that will contain all published ``RPCStream``s published
-  /// to this transport using the ``acceptStream(_:)`` method.
-  ///
-  /// - Returns: An ``RPCAsyncSequence`` of all published ``RPCStream``s.
-  public func listen() async throws -> RPCAsyncSequence<RPCStream<Inbound, Outbound>> {
-    RPCAsyncSequence(wrapping: self.newStreams)
+  public func listen(
+    _ streamHandler: @escaping (RPCStream<Inbound, Outbound>) async throws -> Void
+  ) async throws {
+    try await withThrowingTaskGroup(of: Void.self) { group in
```
FranzBusch, line 61: Can we use a discarding task group here? Also we probably shouldn't fail the whole transport when we fail to handle a single stream

Sources/GRPCCore/GRPCServer.swift
```diff
@@ -204,140 +204,42 @@ public struct GRPCServer: Sendable {
...
-            } catch {
-              // If the listener threw then the connection must be broken, cancel all work.
-              subGroup.cancelAll()
+          do {
+            try await transport.listen { stream in
+              await self.router.handle(stream: stream, interceptors: self.interceptors)
             }
+          } catch {
+            throw RuntimeError(
+              code: .failedToStartTransport,
```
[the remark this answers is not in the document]
gjcairo: `listen` will only throw when starting up the transport fails. Additionally, we consider the server to be "started" once all of the transports have been started - from the docs on this type:
> A ``RuntimeError`` is thrown if any of the transports can't be started.

I think it makes sense to keep failed to start?

Sources/GRPCCore/Transport/ServerTransport.swift
```diff
@@ -14,22 +14,27 @@
...
-  /// Starts the transport and returns a sequence of accepted streams to handle.
+  /// Starts the transport.
   ///
   /// Implementations will typically bind to a listening port when this function is called
-  /// and start accepting new connections. Each accepted inbound RPC stream should be published
-  /// to the async sequence returned by the function.
+  /// and start accepting new connections. Each accepted inbound RPC stream will be handed over to
+  /// the provided `streamHandler` to handle accordingly.
+  ///
+  /// Implementations should throw if the transport fails to start.
```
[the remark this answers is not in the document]
gjcairo: I'm not convinced we need to make the closure throwing, we only call this method in `GRPCServer/run`, and in that case we don't need it to be throwing: we give the stream to the router to handle, and that method's non-throwing.

Sources/GRPCInProcessTransport/InProcessServerTransport.swift
gjcairo: Actually, this being throwing was left over from one of the many changes I went through :D I know it's related to the comment above so we may want to discuss this a bit more, but for now, I just made it non-throwing.

Sources/GRPCCore/GRPCServer.swift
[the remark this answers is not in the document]
FranzBusch: I saw that the protocol documents it like that but it feels wrong to me. Why aren transports able to throw after they started? What if a runtime error happened? How would we observe the transport failing if they can't throw?

Sources/GRPCCore/GRPCServer.swift
FranzBusch: I am doubtful that this can ever be the case. If a transport dies and there is no way it self-heals your service seems pretty doomed. We can make this configurable but I would suggest we start with a forceful cancellation and then in the future can add a config property `shouldGracefullyShutdownOnTransportFailure`. (We are getting pretty close to re-invent service lifecycle with that one though ;) )

Sources/GRPCCore/Transport/ServerTransport.swift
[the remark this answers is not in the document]
FranzBusch: I am confused about your response here since I didn't say we should make the closure throwing. This comment here is in line with what I commented above. I don't think just allowing throws to happen during start is sensible for transports.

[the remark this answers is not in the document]
gjcairo: Ah you're right, I misunderstood your request. Okay, that actually makes sense to me. I'd like to double check with @glbrntt in case he had something else in mind here.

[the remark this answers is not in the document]
glbrntt: > I think this is too limiting. They should throw when they encounter an error at any point.

I agree, but the docs don't say it can throw _only_ if the transport fails to start. I.e. the docs don't say that you can't throw errors in other situations.

Sources/GRPCCore/GRPCServer.swift
FranzBusch: Sure. My reasoning is that if a single transport fails it is often detrimental to the health of the service if this results in us gracefully shutting down the whole service we have a potentially long period where our service is in a degraded state. That's why I would argue that if a transport fail it should lead to us closing the whole server down as quickly as possible so that it can potentially restart and recover.

I can see arguments where people would want to gracefully shutdown but in my personal opinion this should not be the default.

glbrntt: Thanks for elaborating. I understand your point but I'm not yet convinced that cancellation is the right behaviour and I don't really like the idea of having configuration for the behaviour either (as it introduces more edges to test).

I've not really followed the thinking through yet, but I wonder if the server should have some kind of async sequence for its state. That way it could report, for example, when all transports have started, and when a transport throws (i.e. in some degraded state). This gives the caller control over what to do (e.g. let it shutdown gracefully, or cancel). Suspect we'd need to think a little bit more about what state/events could happen though.

gjcairo:
We discussed this offline and decided to support configuring just one transport. I made these changes, and they simplify things considerably.

Sources/GRPCCore/GRPCServer.swift
```diff
@@ -194,152 +193,18 @@ public struct GRPCServer: Sendable {
...
-    for transport in self.transports {
-      do {
-        let listener = try await transport.listen()
-        listeners.append(listener)
-      } catch let cause {
-        // Failed to start, so start stopping.
-        self.state.store(.stopping, ordering: .sequentiallyConsistent)
-        // Some listeners may have started and have streams which need closing.
-        await self.rejectRequests(listeners)
+    self.state.store(.running, ordering: .sequentiallyConsistent)
```
glbrntt, line 196: We effectively go from `starting` to `running` immediately which we can simplify by removing the `starting` state.

Sources/performance-worker/WorkerService.swift
```diff
@@ -421,3 +421,11 @@ extension WorkerService {
     }
   }
 }
+
+@available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+struct NoOpServerTransport: ServerTransport {
+  func listen(_ streamHandler: @escaping (RPCStream<Inbound, Outbound>) async -> Void) async throws
+  {}
```
glbrntt, line 428: nit: formatting
```suggestion
  func listen(
    _ streamHandler: @escaping (RPCStream<Inbound, Outbound>) async -> Void
  ) async throws {
  }
```

glbrntt (commented):
Two tiny nits but looks good otherwise!

PR #1867, opened 2024-04-23
A few helpers to pave the way for subchannel

glbrntt:
Motivation:

To keep the subchannel PR smaller, a few related but standalone bits can be added separately.

Modifications:

- Bump the NIO version to pickup a NIOAsyncChannel fix and a new cascade function
- Add the `ConnectivityState` enum
- Add `ProcessUniqueID` and `SubchannelID`

Result:

Subchannel PR will be smaller.

### DIFF
--- a/Sources/GRPCHTTP2Core/Internal/ProcessUniqueID.swift
+++ b/Sources/GRPCHTTP2Core/Internal/ProcessUniqueID.swift
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2024, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


Sources/GRPCHTTP2Core/Internal/ProcessUniqueID.swift
```diff
@@ -0,0 +1,39 @@
...
+  var description: String {
+    String(describing: self.rawValue)
+  }
+}
+
+/// A process-unique ID for a subchannel.
+struct SubchannelID: Hashable, Sendable, CustomStringConvertible {
+  private let id = ProcessUniqueID()
+  var description: String {
+    "subchan_\(self.id)"
```
gjcairo, line 37: A nit, but is there a reason why we don't just spell out `subchannel` (instead of `subchan`)?

glbrntt: it's a bit shorter.

PR #1871, opened 2024-04-29
Enable `Metadata` to be initializable from a `Sequence` of `Elements`

clintonpi:
Motivation:

This would be a convenient API to have.

Modifications:

Add a new `public init` to `Metadata` which takes `some Sequence<Element>` and initializes `Metadata` collection from the `Sequence` of `Elements`.

Result:

Users can now initialize `Metadata` from a `Sequence` of `Elements`.

### DIFF
--- a/Sources/GRPCCore/Metadata.swift
+++ b/Sources/GRPCCore/Metadata.swift
@@ -130,7 +130,7 @@ public struct Metadata: Sendable, Hashable {
     self.elements = []
   }
 
-  /// Initialize Metadata from a Sequence of Elements
+  /// Initialize `Metadata` from a `Sequence` of `Element`s.
   public init(from elements: some Sequence<Element>) {
     self.elements = elements.map { key, value in
       KeyValuePair(key: key, value: value)
--- a/Sources/GRPCCore/Metadata.swift
+++ b/Sources/GRPCCore/Metadata.swift
@@ -131,7 +131,7 @@ public struct Metadata: Sendable, Hashable {
   }
 
   /// Initialize `Metadata` from a `Sequence` of `Element`s.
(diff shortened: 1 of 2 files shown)


Sources/GRPCCore/Metadata.swift
```diff
@@ -130,6 +130,13 @@ public struct Metadata: Sendable, Hashable {
     self.elements = []
   }
 
+  /// Initialize Metadata from a Sequence of Elements
```
glbrntt, line 133: Nit: add the appropriate backticks here and end with a full-stop

```suggestion
  /// Initialize `Metadata` from a `Sequence` of `Element`s.
```

Sources/GRPCCore/Metadata.swift
```diff
@@ -130,6 +130,13 @@ public struct Metadata: Sendable, Hashable {
...
+  public init(from elements: some Sequence<Element>) {
```
glbrntt, line 134: Because this is a "value preserving" init you should omit the argument label: `init(_ elements: some Sequence<Element>)`.

See the "Argument Labels" section of https://www.swift.org/documentation/api-design-guidelines/

Tests/GRPCCoreTests/MetadataTests.swift
```diff
@@ -17,6 +17,19 @@ import GRPCCore
 import XCTest
 
 final class MetadataTests: XCTestCase {
+  func testInitFrom() {
```
glbrntt, line 20: Let's be a touch more specific: 
```suggestion
  func testInitFromSequence() {
```

glbrntt (changes requested):

PR #1873, opened 2024-05-01
Round robin load balancer

glbrntt:
Motivation:

To build a grpc channel we need to support different load-balancers. One of these is the round-robin load balancer which is added by this patch.

Modifications:

Add a `RoundRobinLoadBalancer` which maintains a set of subchannels and picks between ready subchannels for RPCs using the round-robin algorithm.

Result:

Can maintain a set of connections to a service using and pick between them using round-robin

### DIFF
--- a/Sources/GRPCHTTP2Core/Client/Connection/LoadBalancers/RoundRobinLoadBalancer.swift
+++ b/Sources/GRPCHTTP2Core/Client/Connection/LoadBalancers/RoundRobinLoadBalancer.swift
@@ -544,8 +544,8 @@ extension RoundRobinLoadBalancer {
           return (run: [], close: [], newState: nil)
         }
 
-        // The load balancer should keep subchannels to remove in service until they a new subchannel
-        // can replace it so that requests can continue to be served.
+        // The load balancer should keep subchannels to remove in service until new subchannels
+        // can replace each of them so that requests can continue to be served.
         //
         // If there are more keys to remove than to add, remove some now.
         let numberToRemoveNow = max(keysToRemove.count - keysToAdd.count, 0)
--- a/Sources/GRPCHTTP2Core/Client/Connection/LoadBalancers/RoundRobinLoadBalancer.swift
+++ b/Sources/GRPCHTTP2Core/Client/Connection/LoadBalancers/RoundRobinLoadBalancer.swift
@@ -380,8 +380,7 @@ extension RoundRobinLoadBalancer {
             subchannelToClose = nil
           }
 
-          self.refreshPicker()
-          let aggregateState = self.recomputeAggregateState()
+          let aggregateState = self.refreshPickerAndAggregateState()
 
           switch (subchannelToClose, aggregateState) {
           case (.some(let subchannel), .some(let state)):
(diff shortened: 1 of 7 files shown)


Sources/GRPCHTTP2Core/Client/Connection/LoadBalancers/RoundRobinLoadBalancer.swift
```diff
@@ -0,0 +1,766 @@
...
+        let keysToAdd = newEndpoints.subtracting(existingEndpoints)
+        let keysToRemove = existingEndpoints.subtracting(newEndpoints)
+
+        if keysToRemove.isEmpty && keysToAdd.isEmpty {
+          // Nothing to do.
+          return (run: [], close: [], newState: nil)
+        }
+
+        // The load balancer should keep subchannels to remove in service until they a new subchannel
+        // can replace it so that requests can continue to be served.
```
gjcairo, line 556: Nit, something like:
```suggestion
        // The load balancer should keep subchannels to remove in service until a new subchannel
        // can replace each of them, so that requests can continue to be served.
```

Sources/GRPCHTTP2Core/Client/Connection/LoadBalancers/RoundRobinLoadBalancer.swift
```diff
@@ -0,0 +1,766 @@
...
+            } else {
+              subchannelToClose = nil
+            }
+
+          case .idle, .connecting, .transientFailure, .shutdown:
+            subchannelToClose = nil
+          }
+
+          self.refreshPicker()
+          let aggregateState = self.recomputeAggregateState()
```
gjcairo, line 385: Given we're always calling `refreshPicker()` and `recomputeAggregateState()` as a pair, would it make sense to do both things inside the same function? I don't think we'd ever want to do one independently of the other, right?

gjcairo: Actually, I see you've added `refreshPickerAndAggregateState()` but we are never calling it - I suppose it was just an oversight. Can we make the change?

Sources/GRPCHTTP2Core/Client/Connection/LoadBalancers/RoundRobinLoadBalancer.swift
```diff
@@ -0,0 +1,766 @@
...
+        let aggregate = state.recomputeAggregateState()
+        self = .active(state)
+        return aggregate
+
+      case .closing, .closed:
+        return nil
+      }
+    }
+
+    mutating func parkSoon(_ keys: [EndpointKey]) {
```
gjcairo, line 635: We're never using this

Tests/GRPCHTTP2CoreTests/Client/Connection/LoadBalancers/RoundRobinLoadBalancerTests.swift
```diff
@@ -0,0 +1,440 @@
...
+        .connectivityStateChanged(.idle),
+        .connectivityStateChanged(.connecting),
+        .connectivityStateChanged(.ready),
+        .connectivityStateChanged(.shutdown),
+      ]
+      XCTAssertEqual(events, expected)
+    }
+  }
+
+  func testAddressUpdatesAreHandledGracefully() async throws {
```
gjcairo, line 112: I think we're missing a test case where the updated endpoints exactly match existing endpoints. I suppose in this case we don't want to see a state update event.

gjcairo (approved):

PR #1880, opened 2024-05-21
Add http2 transport namespace and config

glbrntt:
Motivation:

The grpc channel will require some config. This can be dervied from config which we'll make public for the posix and niots transports. As we need the config to create the channel, we need all the bits that make up that config and a namespace to hold them in.

Modifications:

- Add an 'HTTP2ClientTransport' namespace and a nested 'Config' namespace
- Add config and defaults

Result:

Config is in place for the grpc channel

### DIFF
--- a/Sources/GRPCHTTP2Core/Client/HTTP2ClientTransport.swift
+++ b/Sources/GRPCHTTP2Core/Client/HTTP2ClientTransport.swift
@@ -1,0 +1,147 @@
+/*
+ * Copyright 2024, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import GRPCCore
+
+/// A namespace for the HTTP/2 client transport.
+public enum HTTP2ClientTransport {}
+
+extension HTTP2ClientTransport {
+  /// A namespace for HTTP/2 client transport configuration.
+  public enum Config {}
+}
+
+extension HTTP2ClientTransport.Config {
+  public struct Compression: Sendable {
+    /// The default algorithm used for compressing outbound messages.
+    ///
+    /// This can be overridden on a per-call basis via ``CallOptions``.
+    public var algorithm: CompressionAlgorithm
+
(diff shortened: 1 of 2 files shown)


Sources/GRPCHTTP2Core/Client/HTTP2ClientTransport.swift
```diff
@@ -0,0 +1,147 @@
...
+    public var algorithm: CompressionAlgorithm
+
+    /// Compression algorithms enabled for inbound messages.
+    ///
+    /// Note that ``CompressionAlgorithm/none`` is always supported, even if it isn't set here.
+    public var enabledAlgorithms: CompressionAlgorithmSet
+
+    /// Creates a new compression configuration.
+    ///
+    /// See also ``defaults``.
```
gjcairo, line 41: Similar to the above comment, we can use `SeeAlso` here:
```suggestion
    /// - SeeAlso: ``defaults``.
```

Sources/GRPCHTTP2Core/Client/HTTP2ClientTransport.swift
```diff
@@ -0,0 +1,147 @@
...
+    public static var defaults: Self {
+      Self(algorithm: .none, enabledAlgorithms: .none)
+    }
+  }
+
+  @available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+  public struct Keepalive: Sendable {
+    /// The amount of time to wait after reading data before sending a keepalive ping.
+    ///
+    /// Note that the transport may choose to increase this value if it is less than 10 seconds.
```
gjcairo, line 57: Same nit about `- Note:`

Sources/GRPCHTTP2Core/Client/HTTP2ClientTransport.swift
```diff
@@ -0,0 +1,147 @@
...
+    public var time: Duration
+
+    /// The amount of time the server has to respond to a keepalive ping before the connection
+    /// is closed.
+    public var timeout: Duration
+
+    /// Whether the client sends keepalive pings when if there are no calls in progress.
```
gjcairo, line 64:
```suggestion
    /// Whether the client sends keepalive pings when there are no calls in progress.
```

Sources/GRPCHTTP2Core/Client/HTTP2ClientTransport.swift
```diff
@@ -0,0 +1,147 @@
...
+      self.permitWithoutCalls = permitWithoutCalls
+    }
+  }
+
+  @available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+  public struct Idle: Sendable {
+    /// The maximum amount of time a connection may be idle before it's closed.
+    public var maxTime: Duration
+
+    /// Creates a idle configuration.
```
gjcairo, line 80:
```suggestion
    /// Creates an idle configuration.
```

gjcairo (changes requested):
LGTM overall, just some small doc nits

gjcairo (approved):

PR #1881, opened 2024-05-21
Add a request queue for the grpc channel

glbrntt:
Motivation:

The grpc channel needs to enqueue requests when the channel isn't ready to handle RPCs. When the channel becomes ready, it can attempt to execute the RPCs on a load balancer.

Modifications:

- Add a request queue. The queue stores continuations for a `LoadBalancer`. Elements can be removed in-order (popped) or by ID (in case of cancellation).
- Add a queue ID

Result:

Requests can be queued

### DIFF
--- a/Tests/GRPCHTTP2CoreTests/Client/Connection/RequestQueueTests.swift
+++ b/Tests/GRPCHTTP2CoreTests/Client/Connection/RequestQueueTests.swift
@@ -66,7 +66,7 @@ final class RequestQueueTests: XCTestCase {
       var waitForReadyIDs = [QueueEntryID]()
       var failFastIDs = [QueueEntryID]()
 
-      for i in 0 ..< 50 {
+      for _ in 0 ..< 50 {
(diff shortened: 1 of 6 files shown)


Tests/GRPCHTTP2CoreTests/Client/Connection/RequestQueueTests.swift
```diff
@@ -0,0 +1,159 @@
...
+      guard let popped = queue.popFirst() else {
+        return XCTFail("Missing continuation")
+      }
+      XCTAssertNil(queue.popFirst())
+
+      popped.resume(throwing: AnErrorToAvoidALeak())
+    }
+  }
+
+  func testRemoveEntryByID() async {
```
gjcairo, line 46: Can we add/change this test where we have more than a single entry and we get one of them by ID?

gjcairo (changes requested):

gjcairo (approved):

PR #1883, opened 2024-05-22
Add NIOPosix server transport implementation

gjcairo:
## Motivation
We want to provide some commonly used server transport implementations out of the box.

## Modifications
This PR adds a `ServerTransport` implementation based on `NIOPosix`.

## Result
A `NIOPosix` implementation of a server transport.


Sources/GRPCHTTP2TransportNIOPosix/GRPCHTTP2ServerTransportBootstrapper.swift
```diff
@@ -0,0 +1,73 @@
...
+    )
+    let flushNotificationHandler = GRPCServerFlushNotificationHandler(
+      serverConnectionManagementHandler: serverConnectionHandler
+    )
+    try self.addHandler(flushNotificationHandler)
+
+    let streamMultiplexer = try self.configureAsyncHTTP2Pipeline(mode: .server) { streamChannel in
+      return streamChannel.eventLoop.makeCompletedFuture {
+        let streamHandler = GRPCServerStreamHandler(
+          scheme: configuration.tls != nil ? .https : .http,
```
gjcairo, line 51: There's definitely more to setting up TLS than just this, but I'm not sure how to set up the pipeline to support it, so this is an open question

gjcairo:
@glbrntt I'm not really sure how to unit test this, since all this code does is set up the pipeline. I feel like if I write tests they'll be more like integration tests, and we'll already have the interop tests for this, right?

Sources/GRPCHTTP2TransportNIOPosix/GRPCHTTP2ServerTransportBootstrapper.swift
```diff
@@ -0,0 +1,73 @@
...
+import NIOHTTP2
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+extension ChannelPipeline.SynchronousOperations {
+  public typealias HTTP2ConnectionChannel = NIOAsyncChannel<HTTP2Frame, HTTP2Frame>
+  public typealias HTTP2StreamMultiplexer = NIOHTTP2Handler.AsyncStreamMultiplexer<
+    (NIOAsyncChannel<RPCRequestPart, RPCResponsePart>, EventLoopFuture<MethodDescriptor>)
+  >
+
+  public func configureGRPCHTTP2ServerTransportPipeline(
```
glbrntt, line 29: We want a pipeline helper that we can use for NIOPosix _and_ NIOTS, which means it should be in the grpc http2 core module. It also means that the config should be common across the two. This also doesn't need to be public, it can use `@_spi(Package)` (until we can use `package` access).

The approach I've taken for the client is to have "structured" configuration rather than a giant config blob. The NIOPosix/NIOTS transports will each have their own configs which will compose the common bits and add their own bits of specific config on the side. See also https://github.com/grpc/grpc-swift/pull/1880/files

Sources/GRPCHTTP2TransportNIOPosix/GRPCHTTP2ServerTransportBootstrapper.swift
glbrntt: Because we want this helper to be common to both NIOPosix and NIOTS we don't want it to set up TLS (as it's done in different ways between the two).

Sources/GRPCHTTP2TransportNIOPosix/GRPCHTTP2TransportNIOPosix.swift
```diff
@@ -15,7 +15,88 @@
...
 import GRPCCore
+import GRPCHTTP2Core
+import NIOCore
+import NIOExtras
+import NIOPosix
 
-@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
-public struct GRPCHTTP2TransportNIOPosix {
+@available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+public struct GRPCHTTP2TransportNIOPosix: ServerTransport {
```
glbrntt, line 24: We don't need a GRPC prefix here. 

To help give some structure I've used some namespacing for the client, see https://github.com/grpc/grpc-swift/pull/1880/files.

For the server we should do similar:

```swift
public enum HTTPServerTransport {}

extension HTTPServerTransport {
  public struct Posix: ServerTransport { ... }
}

extension HTTPServerTransport.Posix {
  public struct Config {
    // ...
  }
}
```

I think it makes sense gathering them like this as they are _so_ closely related.

Sources/GRPCHTTP2TransportNIOPosix/GRPCHTTP2TransportNIOPosix.swift
```diff
@@ -15,7 +15,88 @@
...
+    _ streamHandler: @escaping (RPCStream<Inbound, Outbound>) async -> Void
+  ) async throws {
+    let serverChannel = try await ServerBootstrap(group: self.eventLoopGroup)
+      .serverChannelInitializer { channel in
+        let quiescingHandler = self.serverQuiescingHelper.makeServerChannelHandler(channel: channel)
+        return channel.pipeline.addHandler(quiescingHandler)
+      }
+      .bind(
+        host: self.configuration.host,
+        port: self.configuration.port,
```
glbrntt, line 47: We should support more options than just host/port. The right API is probably `GRPCHTTP2Core.SocketAddress`

Sources/GRPCHTTP2Core/Server/Connection/ServerConnection.swift
```diff
@@ -0,0 +1,51 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import GRPCCore
+import NIOCore
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+public enum ServerConnection {
```
glbrntt, line 21: I don't think there's a good reason for this to be public

Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
```diff
@@ -99,6 +103,16 @@ extension GRPCServerStreamHandler {
...
+          if let methodDescriptor = metadata.methodDescriptor {
+            self.methodDescriptor.succeed(methodDescriptor)
+          } else {
+            self.methodDescriptor.fail(
+              RPCError(
+                code: .internalError,
+                message: "State machine should have thrown if a method descriptor was not present."
+              )
+            )
+          }
```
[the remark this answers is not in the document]
glbrntt: Can we have these changes (the state machine + channel handler) ones in a separate PR?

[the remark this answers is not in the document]
gjcairo: >Parsing the method descriptor isn't free and we already do it when receiving the metadata, can we just return it with the receivedMetadata case instead?

I didn't want to do this originally because it makes no sense for the action to include the method descriptor when the client receives metadata. But I agree that we want to avoid the slowness of parsing so I'll make this change.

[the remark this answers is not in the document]
gjcairo: https://github.com/grpc/grpc-swift/pull/1891

Sources/GRPCHTTP2TransportNIOPosix/GRPCHTTP2TransportNIOPosix.swift
```diff
@@ -15,7 +15,119 @@
...
+  @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+  public struct Posix: ServerTransport {
+    private let configuration: Config
+    private let eventLoopGroup: MultiThreadedEventLoopGroup
+    private let serverQuiescingHelper: ServerQuiescingHelper
+
+    init(configuration: Config) {
+      self.configuration = configuration
+      self.eventLoopGroup = MultiThreadedEventLoopGroup(
+        numberOfThreads: self.configuration.http2.maxConcurrentStreams // TODO: not sure this is the right value to use, or if we should maybe just get the EL from the caller in the init
```
gjcairo, line 34: Open question here

glbrntt (changes requested):

Sources/GRPCHTTP2Core/Server/HTTP2ServerTransport.swift
```diff
@@ -0,0 +1,194 @@
...
+    /// - Note: This will also be set as the initial window size for the connection.
+    public var targetWindowSize: Int
+
+    /// The number of concurrent streams on the HTTP/2 connection.
+    public var maxConcurrentStreams: Int
+  }
+}
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+extension ChannelPipeline.SynchronousOperations {
```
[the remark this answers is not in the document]
gjcairo: That file doesn't exist yet, it's added in your open PR. I can create the file too and we'll just have a conflict if you want? Otherwise I can move this when I tackle the TLS config in a follow-up.

Sources/GRPCHTTP2Core/Internal/NIOChannelPipeline+GRPC.swift
```diff
@@ -0,0 +1,102 @@
...
+
+import GRPCCore
+import NIOCore
+import NIOHPACK
+import NIOHTTP2
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+extension ChannelPipeline.SynchronousOperations {
+  public typealias HTTP2ConnectionChannel = NIOAsyncChannel<HTTP2Frame, HTTP2Frame>
+  public typealias HTTP2StreamMultiplexer = NIOHTTP2Handler.AsyncStreamMultiplexer<
```
glbrntt, line 25: These need to be `@_spi(Package)` too, at the moment I think they are public for real!

Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
```diff
@@ -37,7 +37,7 @@ final class GRPCServerStreamHandler: ChannelDuplexHandler {
   private var pendingTrailers:
     (trailers: HTTP2Frame.FramePayload, promise: EventLoopPromise<Void>?)?
 
-  private let methodDescriptorPromise: EventLoopPromise<MethodDescriptor>
+  internal let methodDescriptorPromise: EventLoopPromise<MethodDescriptor>
```
glbrntt, line 40: Can we keep this private? We pass it in to the handler so we already have a handle on it.

Sources/GRPCHTTP2Core/Server/HTTP2ServerTransport.swift
```diff
@@ -0,0 +1,174 @@
...
+        permitWithoutCalls: false,
+        minPingIntervalWithoutCalls: .seconds(5 * 60)  // 5 minutes
+      )
+    }
+  }
+
+  @available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+  public struct Idle: Sendable {
+    /// The maximum amount of time a connection may be idle before it's closed.
+    public var maxTime: Duration?
```
glbrntt, line 93: Should we roll this up into the `Connection` config? `connection.maxIdleTime = <whatever>` seems quite natural, `idle.maxTime = <whatever>` feels a little less so. WDYT?

Sources/GRPCHTTP2Core/Server/HTTP2ServerTransport.swift
```diff
@@ -0,0 +1,174 @@
...
+      self.targetWindowSize = targetWindowSize
+      self.maxConcurrentStreams = maxConcurrentStreams
+    }
+
+    /// Default values. The max frame size defaults to 2^14, the target window size defaults to 2^16-1, and
+    /// the max concurrent streams default to infinite.
+    public static var defaults: Self {
+      Self(
+        maxFrameSize: 2 ^ 14,
+        targetWindowSize: (2 ^ 16) - 1,
```
glbrntt, line 155: This doesn't do what you think it does, this is `2 xor 14`. You want `1 << 14`

Sources/GRPCHTTP2Core/Server/HTTP2ServerTransport.swift
```diff
@@ -0,0 +1,174 @@
...
+    }
+  }
+
+  public struct RPC: Sendable {
+    /// The maximum request payload size.
+    public var maximumRequestPayloadSize: Int
+
+    public init(maximumRequestPayloadSize: Int) {
+      self.maximumRequestPayloadSize = maximumRequestPayloadSize
+    }
```
glbrntt, line 167: nit: s/maximum/max

glbrntt (changes requested):
It's coming together, mostly just nits now.

Sources/GRPCHTTP2TransportNIOPosix/GRPCHTTP2TransportNIOPosix.swift
```diff
@@ -15,7 +15,164 @@
...
+          }
+        }
+
+      try await serverChannel.executeThenClose { inbound in
+        try await withThrowingDiscardingTaskGroup { serverTaskGroup in
+          for try await (connectionChannel, streamMultiplexer) in inbound {
+            serverTaskGroup.addTask {
+              try await connectionChannel.executeThenClose {
+                connectionInbound,
+                connectionOutbound in
```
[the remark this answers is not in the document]
gjcairo: I sometimes hate the formatter :D

Sources/GRPCHTTP2TransportNIOPosix/GRPCHTTP2TransportNIOPosix.swift
```diff
@@ -15,7 +15,188 @@
...
+                  connectionTaskGroup.addTask {
+                    for try await _ in connectionInbound {}
+                  }
+
+                  connectionTaskGroup.addTask {
+                    try await withThrowingDiscardingTaskGroup { streamTaskGroup in
+                      for try await (http2Stream, methodDescriptor) in streamMultiplexer.inbound {
+                        streamTaskGroup.addTask {
+                          try await http2Stream.executeThenClose { inbound, outbound in
+                            let descriptor = try await methodDescriptor.get()
```
glbrntt: What happens if a child task in a throwing-discarding-task-group throws? We should check this because if the stream throws it _shouldn't_ result in the connection being closed.

Same goes for the connection task group: if a connection throws we shouldn't cancel the listening channel either.

Sources/GRPCHTTP2TransportNIOPosix/GRPCHTTP2TransportNIOPosix.swift
```diff
@@ -15,7 +15,189 @@
...
+        message:
+          "Unsupported mapping to NIOCore/SocketAddress for GRPCHTTP2Core/SocketAddress: \(socketAddress)."
+      )
+    }
+  }
+}
+
+extension ServerBootstrap {
+  @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+  public func bind<Output: Sendable>(
```
glbrntt, line 184: This one doesn't need to be public

glbrntt (approved):
Looks great, thanks Gus!

PR #1887, opened 2024-05-23
Cache swift-format

glbrntt:
Motivation:

The CI takes ~7 minutes to run, the longest job is the formatting check which spends most of its time compiling swift-format.

Modifications:

- Cache the swift-format build directory

Result:

Faster CI

### DIFF
--- a/scripts/license-check.sh
+++ b/scripts/license-check.sh
@@ -102,6 +102,10 @@ check_copyright_headers() {
     ! -name '*.grpc.swift' \
     ! -name 'LinuxMain.swift' \
     ! -name 'XCTestManifests.swift' \
+    ! -path './FuzzTesting/.build/*' \
+    ! -path './Performance/QPSBenchmark/.build/*' \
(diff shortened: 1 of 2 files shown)


scripts/license-check.sh
```diff
@@ -102,6 +102,10 @@ check_copyright_headers() {
     ! -name '*.grpc.swift' \
     ! -name 'LinuxMain.swift' \
     ! -name 'XCTestManifests.swift' \
+    ! -path './FuzzTesting/.build/*' \
+    ! -path './Performance/QPSBenchmark/.build/*' \
+    ! -path './Performance/Benchmarks/.build/*' \
+    ! -path './scripts/.swift-format-source/*' \
```
glbrntt, line 108: This is now necessary because the soundness check first runs the license check and _then_ would check out and build swift-format, meaning that it'd check the headers of swift-format. This wasn't an issue because of the ordering.

gjcairo (approved):

PR #1889, opened 2024-05-23
Update tutorial docs

clintonpi:


Sources/Examples/RouteGuide/README.md
```diff
@@ -22,11 +22,12 @@ $ swift run RouteGuideClient
 
 ## Regenerating client and server code
 
-For simplicity, a Makefile is provided in the root of this package with a target
-to generate client and server code:
+For simplicity, a shell script ([grpc-swift/Protos/generate.sh][run-protoc]) is provided,
+with a target to generate client and server code:
```
glbrntt, line 26: "target" is a Makefile term and in this case would've referred to "generate-route-guide". There's no target here so this should read:

```suggestion
For simplicity, a shell script ([grpc-swift/Protos/generate.sh][run-protoc]) is provided
to generate client and server code:
```

glbrntt (changes requested):
One small nit but looks good otherwise

glbrntt (approved):
Great, thanks Clinton!

PR #1891, opened 2024-05-24
Server transport changes

gjcairo:
## Motivation
In order to implement the H2 server transports, we need a couple of changes.

## Modifications
- Return MethodDescriptor from stream state machine
- Support quiescing event in connection management handler

## Result
Closer to having H2 server transports

### DIFF
--- a/Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
+++ b/Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
@@ -19,7 +19,7 @@ import NIOCore
 import NIOHPACK
 import NIOHTTP1
 
-enum Scheme: String {
+enum GRPCHTTP2StreamScheme: String {
   case http
   case https
 }
@@ -30,13 +30,13 @@ enum GRPCStreamStateMachineConfiguration {
 
   struct ClientConfiguration {
     var methodDescriptor: MethodDescriptor
-    var scheme: Scheme
+    var scheme: GRPCHTTP2StreamScheme
     var outboundEncoding: CompressionAlgorithm
     var acceptedEncodings: CompressionAlgorithmSet
 
     init(
       methodDescriptor: MethodDescriptor,
-      scheme: Scheme,
+      scheme: GRPCHTTP2StreamScheme,
       outboundEncoding: CompressionAlgorithm,
       acceptedEncodings: CompressionAlgorithmSet
     ) {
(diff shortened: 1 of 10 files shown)


Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -19,7 +19,7 @@ import NIOCore
 import NIOHPACK
 import NIOHTTP1
 
-enum Scheme: String {
+enum GRPCHTTP2StreamScheme: String {
```
glbrntt, line 22: Can we revert this naming change?

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -1211,9 +1211,11 @@ extension GRPCStreamStateMachine {
...
       }
 
-      let path = headers.firstString(forKey: .path)
-        .flatMap { MethodDescriptor(fullyQualifiedMethod: $0) }
-      if path == nil {
+      guard
+        let path = headers.firstString(forKey: .path).flatMap({
+          MethodDescriptor(fullyQualifiedMethod: $0)
+        })
+      else {
```
glbrntt, line 1218: Can we split this into two checks?
1. We get a ":path", and 
2. The ":path" is a valid `MethodDescriptor`

Otherwise 1.) can be fine and 2.) not-fine and the error would indicate there was ":path" which is misleading

Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
```diff
@@ -37,17 +37,21 @@ final class GRPCServerStreamHandler: ChannelDuplexHandler {
   private var pendingTrailers:
     (trailers: HTTP2Frame.FramePayload, promise: EventLoopPromise<Void>?)?
 
+  var methodDescriptorPromise: EventLoopPromise<MethodDescriptor>
```
glbrntt, line 40: I think this can be a `private let`

Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
```diff
@@ -97,11 +101,27 @@ extension GRPCServerStreamHandler {
...
+          if let methodDescriptor = methodDescriptor {
+            self.methodDescriptorPromise.succeed(methodDescriptor)
+            context.fireChannelRead(self.wrapInboundOut(.metadata(metadata)))
+          } else {
+            self.methodDescriptorPromise.fail(
+              RPCError(
+                code: .internalError,
+                message: "Method descriptor should have been present if we received metadata."
+              )
+            )
```
glbrntt, line 114: IIUC we should never hit this path on the server, so can we `assertionFailure` here instead?

Tests/GRPCHTTP2CoreTests/Server/GRPCServerStreamHandlerTests.swift
```diff
@@ -26,13 +26,14 @@ import XCTest
 @available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
 final class GRPCServerStreamHandlerTests: XCTestCase {
   func testH2FramesAreIgnored() throws {
+    let channel = EmbeddedChannel()
```
glbrntt, line 29: Can we have a test which validates that the promise is succeeded when we expect? And another which checks that its failed if the handler is immediately removed from the pipeline?

glbrntt (changes requested):

PR #1893, opened 2024-05-28
Add grpc channel

glbrntt:
Motivation:

The grpc channel provides a long-lived connection to some remote backend(s). This allows us to build up a client transport.

Modifications:

- Add the `GRPCChannel`, a long-lives way of providing connections to a remote backend

Result:

Can make RPCs to a remote backend

### DIFF
--- a/Sources/GRPCHTTP2Core/Client/Connection/GRPCChannel.swift
+++ b/Sources/GRPCHTTP2Core/Client/Connection/GRPCChannel.swift
@@ -1,0 +1,912 @@
+/*
+ * Copyright 2024, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Atomics
+import DequeModule
+@_spi(Package) import GRPCCore
+
+@available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+@_spi(Package)
+public struct GRPCChannel: ClientTransport {
+  private enum Input: Sendable {
+    /// Close the channel, if possible.
+    case close
+    /// Handle the result of a name resolution.
+    case handleResolutionResult(NameResolutionResult)
+    /// Handle the event from the underlying connection object.
+    case handleLoadBalancerEvent(LoadBalancerEvent, LoadBalancerID)
+  }
+
+  /// Events which can happen to the channel.
+  private let _connectivityState:
+    (
+      stream: AsyncStream<ConnectivityState>,
+      continuation: AsyncStream<ConnectivityState>.Continuation
+    )
+
+  /// Inputs which this channel should react to.
+  private let input: (stream: AsyncStream<Input>, continuation: AsyncStream<Input>.Continuation)
+
+  /// A resolver providing resolved names to the channel.
+  private let resolver: NameResolver
+
+  /// The state of the channel.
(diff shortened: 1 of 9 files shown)


Sources/GRPCHTTP2Core/Client/Connection/GRPCChannel.swift
```diff
@@ -0,0 +1,912 @@
...
+          // of the next kind.
+          let nextNext = makeLoadBalancer()
+          let previous = state.next
+          state.next = nextNext
+          state.past[next.id] = next
+          onChangeLoadBalancer = .runLoadBalancer(nextNext, stop: previous)
+        }
+      } else {
+        if newLoadBalancerKind.matches(loadBalancer: state.current) {
+          onChangeLoadBalancer = .updateLoadBalancer(state.current)
```
gjcairo, line 661: Isn't it possible that, even though the existing LB matches the new load balancer kind, the `makeLoadBalancer` closure builds an LB that's different in some other way (i.e. same kind but different configuration)? Wouldn't we want to update the LB here to the new one instead of using the current?

Sources/GRPCHTTP2Core/Client/Connection/GRPCChannel.swift
```diff
@@ -0,0 +1,912 @@
...
+    var close: LoadBalancer? = nil
+    var publishState: ConnectivityState? = nil
+    var resumeContinuations: ResumableContinuations? = nil
+    var finish: Bool = false
+
+    struct ResumableContinuations {
+      var continuations: [CheckedContinuation<LoadBalancer, Error>]
+      var result: Result<LoadBalancer, Error>
+    }
+  }
```
gjcairo, line 692: Is there a reason (thinking allocations or other performance) why we want this to be a struct vs having an enum of actions and returning an array of these actions from `loadBalancerStateChanged`? 

I feel like returning an array of states is safer in terms of handling all of the actions (i.e., if we ever add an action here, we will forced to update callers to handle it) and also the state machine itself will enforce the order in which we should handle each action, instead of leaving that up to the caller, which may make more sense?

gjcairo (changes requested):

Sources/GRPCHTTP2Core/Client/Connection/GRPCChannel.swift
```diff
@@ -0,0 +1,912 @@
...
+    self._methodConfig.withLockedValue { $0[descriptor] }
+  }
+
+  /// Establishes and maintains a connection to the remote destination.
+  public func connect() async {
+    self.state.withLockedValue { $0.start() }
+    self._connectivityState.continuation.yield(.idle)
+
+    await withDiscardingTaskGroup { group in
+      var iterator: Optional<RPCAsyncSequence<NameResolutionResult>.AsyncIterator>
```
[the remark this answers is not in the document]
glbrntt: Yup, we use it beyond the scope of the `pull` case on L172 so we need to declare it here.

Sources/GRPCHTTP2Core/Client/Connection/GRPCChannel.swift
glbrntt: The LB can only changed via the config from the resolver which will happen via `updateLoadBalancer`. The `makeLoadBalancer` closure mustn't change over time, all of the state it captures is immutable.

Sources/GRPCHTTP2Core/Client/Connection/GRPCChannel.swift
glbrntt: An array of actions would cause an allocation each time the state changed. 

I'm not sure the ordering matters much, we either have to get it right in the state machine (when building the array of actions) or when we act on the actions as we do now. This would be a bit different if the caller didn't know what the state machine did but here the objects are so tightly coupled that enforcing the order in the state machine doesn't really add anything.

Accidentally not handling a state is a valid concern. However, if we add an action we're doing it purposefully so I'd be surprised if that got missed while writing the code or in code review. IMO it's not worth adding an allocation for it.

Sources/GRPCHTTP2Core/Client/Connection/GRPCChannel.swift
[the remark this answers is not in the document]
gjcairo: Oops missed that

gjcairo (approved):

PR #1897, opened 2024-05-30
Add server executable for interop tests

gjcairo:
## Motivation

Now we've got an actual server transport implementation using NIOPosix, we want to run the interop tests using the real stack.

## Modifications

Add a server executable to run interop tests against. Until the client transport's ready, we'll use v1's client against v2's server transport.

## Result

Interop tests can be run against v2 server.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -335,13 +335,6 @@ extension Target {
     ]
   )
 
-  static let grpcHTTP2TransportNIOPosixTests: Target = .testTarget(
-    name: "GRPCHTTP2TransportNIOPosixTests",
-    dependencies: [
-      .grpcHTTP2TransportNIOPosix
-    ]
-  )
-
   static let grpcHTTP2TransportNIOTransportServicesTests: Target = .testTarget(
     name: "GRPCHTTP2TransportNIOTransportServicesTests",
     dependencies: [
@@ -407,6 +400,17 @@ extension Target {
     dependencies: [
       .grpcCore,
       .grpcProtobuf
+    ]
+  )
+
+  static let interoperabilityTestsExecutable: Target = .executableTarget(
+    name: "InteroperabilityTestsExecutable",
+    dependencies: [
+      .grpcCore,
+      .grpcHTTP2Core,
+      .grpcHTTP2TransportNIOPosix,
+      .interoperabilityTests,
+      .argumentParser
     ]
   )
 
@@ -736,6 +740,7 @@ let package = Package(
     .grpcProtobuf,
     .grpcProtobufCodeGen,
     .interoperabilityTestImplementation,
+    .interoperabilityTestsExecutable,
     .performanceWorker,
 
     // v2 tests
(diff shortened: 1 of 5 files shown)


Package.swift
```diff
@@ -410,6 +403,17 @@ extension Target {
     ]
   )
 
+  static let interoperabilityTestsExecutable: Target = .executableTarget(
+    name: "InteroperabilityTestsExecutable",
```
glbrntt, line 407: Can we call this `interop-tests`/`interoperability-tests`? Executables are more commonly snake/kebab case yet we never seem to follow that pattern so I'd like us to move in that direction

Sources/InteroperabilityTestsExecutable/InteroperabilityTestsExecutable.swift
```diff
@@ -0,0 +1,64 @@
+import ArgumentParser
+import GRPCCore
+import GRPCHTTP2Core
+import GRPCHTTP2TransportNIOPosix
+import NIOPosix
+
+@testable import InteroperabilityTests
```
glbrntt, line 23: Hmm we shouldn't need `@testable` here, the `InteroperabilityTests` module won't ever be a product so any API you need from it can be `public`

Sources/InteroperabilityTestsExecutable/InteroperabilityTestsExecutable.swift
```diff
@@ -0,0 +1,80 @@
...
+      var command = try parseAsRoot()
+      if var asyncCommand = command as? AsyncParsableCommand {
+        try await asyncCommand.run()
+      } else {
+        try command.run()
+      }
+    } catch {
+      exit(withError: error)
+    }
+  }
```
glbrntt, line 39: This shouldn't be necessary?

Sources/InteroperabilityTestsExecutable/InteroperabilityTestsExecutable.swift
```diff
@@ -0,0 +1,80 @@
...
+
+  struct ListTests: ParsableCommand {
+    static var configuration = CommandConfiguration(
+      abstract: "List all interoperability test names."
+    )
+
+    func run() throws {
+      InteroperabilityTestCase.allCases.forEach {
```
glbrntt, line 75: nit: can we use `for testCase in ...` rather than `forEach`? It's just more idiomatic

glbrntt (changes requested):

Sources/InteroperabilityTestsExecutable/InteroperabilityTestsExecutable.swift
gjcairo: Oh, TIL. We were manually doing this in the v1 implementation but I guess it's not necessary anymore with the `@main` annotation. Nice.

Sources/InteroperabilityTestsExecutable/InteroperabilityTestsExecutable.swift
gjcairo: Oh okay. I forgot about this, I was going to suggest having an SPI, but you're right it can simply be public.

glbrntt (approved):
Great, thanks Gus!

PR #1898, opened 2024-05-30
Always store compression algorithm in stream state machine

gjcairo:
## Motivation
The `grpc-encoding` header is currently missing when the server sends a compressed response.

## Modifications
This PR makes sure that the compression algorithm is always available to the state machine so the server can send the `grpc-encoding` header when necessary.

## Result
The `grpc-encoding` header is present when the server sends a compressed response.

### DIFF
--- a/Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
+++ b/Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
@@ -89,12 +89,14 @@ private enum GRPCStreamStateMachineState {
     init(
       previousState: ClientIdleServerIdleState,
       compressor: Zlib.Compressor?,
+      outboundCompression: CompressionAlgorithm?,
       framer: GRPCMessageFramer,
       decompressor: Zlib.Decompressor?,
       deframer: NIOSingleStepByteToMessageProcessor<GRPCMessageDeframer>?
     ) {
       self.maximumPayloadSize = previousState.maximumPayloadSize
       self.compressor = compressor
+      self.outboundCompression = outboundCompression
       self.framer = framer
       self.decompressor = decompressor
       self.deframer = deframer
@@ -105,6 +107,7 @@ private enum GRPCStreamStateMachineState {
   struct ClientOpenServerOpenState {
     var framer: GRPCMessageFramer
     var compressor: Zlib.Compressor?
+    var outboundCompression: CompressionAlgorithm?
 
     let deframer: NIOSingleStepByteToMessageProcessor<GRPCMessageDeframer>
     var decompressor: Zlib.Decompressor?
@@ -118,6 +121,7 @@ private enum GRPCStreamStateMachineState {
     ) {
       self.framer = previousState.framer
       self.compressor = previousState.compressor
+      self.outboundCompression = previousState.outboundCompression
 
       self.deframer = deframer
       self.decompressor = decompressor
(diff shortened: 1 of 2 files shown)


Tests/GRPCHTTP2CoreTests/GRPCStreamStateMachineTests.swift
```diff
@@ -1343,8 +1343,36 @@ final class GRPCStreamServerStateMachineTests: XCTestCase {
...
+    )
+    XCTAssertEqual(
+      try stateMachine.send(metadata: .init()),
+      [
+        ":status": "200",
+        "content-type": "application/grpc",
+        "grpc-accept-encoding": "deflate",
+      ]
+    )
+  }
```
glbrntt, line 1358: This is the server sending initial metadata back to the client, right? If so why do we get 'grpc-accept-encoding' when `compressionEnabled: false`?

gjcairo: Ah, nice catch

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -1019,10 +1042,6 @@ extension GRPCStreamStateMachine {
       headers.add(outboundEncoding.name, forKey: .encoding)
     }
 
-    for acceptedEncoding in configuration.acceptedEncodings.elements.filter({ $0 != .none }) {
-      headers.add(acceptedEncoding.name, forKey: .acceptEncoding)
-    }
```
glbrntt, line 1024: I don't think this is the right fix: there are times when we _should_ include the accept encoding headers, usually when the client tries to use an encoding we don't support (i.e. client uses compression and server isn't configured for compression).

It looks like there's a test bug: in `makeServerStateMachine`, the compression is unconditionally used, even if `compressionEnabled: false`.

gjcairo: I went back to the RFC and this header appears as optional in all responses: `Response-Headers → HTTP-Status [Message-Encoding] [Message-Accept-Encoding] Content-Type *Custom-Metadata`. I think I had decided to always include it to make the logic simpler, as the RFC does not specify when and when it shouldn't be present. It's okay if we want to save some bytes and not send it unless the client tries to use an encoding we don't support.

WRT a bug being present in the tests: I don't think there is - it's just that I was always adding the accept-encoding header to the server's initial metadata. This doesn't mean the server will use compression, it's just advertising to the client that it can understand it. If I make the change to only send this header when we get an unsupported algorithm from the client though, I'll make the change to not always send it back.

gjcairo: Actually, we _are_ including the accept-encoding header in the case you mentioned - it just happens in a different codepath because we are returning trailers instead of response headers: see line 1314 to 1333 in the state machine. We also have a test for this, `testReceiveMetadataWhenClientIdleAndServerIdle_ServerUnsupportedEncoding`.
I have fixed the initial server metadata in the tests though to not include the accept-encoding header unless compression is enabled, and to not accept deflate encoding unless the compressionEnabled flag is passed - both for consistency.

glbrntt: > I think I had decided to always include it to make the logic simpler, as the RFC does not specify when and when it shouldn't be present. It's okay if we want to save some bytes and not send it unless the client tries to use an encoding we don't support.

Yeah, that's fine with me.

> WRT a bug being present in the tests: I don't think there is - it's just that I was always adding the accept-encoding header to the server's initial metadata. This doesn't mean the server will use compression, it's just advertising to the client that it can understand it. 

The code I originally commented on setup a server state machine which _didn't_ using compression but still advertised compression in the accept encoding header it sent to the client. If it doesn't support compression then it mustn't advertise to the client that it does.

glbrntt (approved):
LGTM, thanks Gus!

PR #1901, opened 2024-05-31
Better distinguish between connection drops and regular closes

glbrntt:
Motivation:

If a connection is closed because the tcp connection is dropped then the `Connection` doesn't receive any indication as to why it was closed which ends up being classified as a locally initiated close. In turns the subchannel will be shutdown and can't be reused.

This isn't correct: we know the connection was dropped unexpectedly and we whether there were any active RPCs at the time. This means the subchannel can transition to idle if there were no active RPCs or transient failure if there were.

Modifications:

- Add a new 'unexpected' close reason to the connection handler which optionally includes an error. This can provide further diagnostic information as to why a connection has been dropped.
- Alter the semantics in subchannel such that go-aways and unexpected closes when there are no RPCs transition cause a transition to the idle state, and unexpected closes when there are RPCs lead to a transition to the transient failure state.

Result:

More sensible state updates when connections are closed

### DIFF
--- a/Sources/GRPCHTTP2Core/Client/Connection/LoadBalancers/Subchannel.swift
+++ b/Sources/GRPCHTTP2Core/Client/Connection/LoadBalancers/Subchannel.swift
@@ -329,11 +329,12 @@ extension Subchannel {
     guard isClosed else { return }
 
     switch reason {
-    case .idleTimeout:
-      // Connection closed due to an idle timeout; notify the load balancer about this.
+    case .idleTimeout, .remote, .error(_, wasIdle: true):
+      // Connection closed due to an idle timeout or the remote telling it to GOAWAY; notify the
+      // load balancer about this.
       self.event.continuation.yield(.connectivityStateChanged(.idle))
 
-    case .keepaliveTimeout, .error:
+    case .keepaliveTimeout, .error(_, wasIdle: false):
       // Unclean closes trigger a transient failure state change and a name resolution.
       self.event.continuation.yield(.connectivityStateChanged(.transientFailure))
       self.event.continuation.yield(.requiresNameResolution)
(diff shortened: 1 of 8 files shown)


Tests/GRPCHTTP2CoreTests/Client/Connection/Connection+Equatable.swift
```diff
@@ -48,15 +47,45 @@ extension Connection.CloseReason: Equatable {
...
+  }
+}
+
+extension ClientConnectionEvent.CloseReason: Equatable {
+  public static func == (lhs: Self, rhs: Self) -> Bool {
+    switch (lhs, rhs) {
+    case (.goAway(let lhsCode, let lhsMessage), .goAway(let rhsCode, let rhsMessage)):
+      return lhsCode == rhsCode && lhsMessage == rhsMessage
+    case (.unexpected(_, let lhsIsIdle), .unexpected(_, let rhsIsIdle)):
+      return lhsIsIdle == rhsIsIdle
```
gjcairo, line 82: Do we want to do the same as above for `Connection.CloseReason` and compare the errors if they're `RPCError`s (or `Equatable`)?

Sources/GRPCHTTP2Core/Client/Connection/LoadBalancers/Subchannel.swift
```diff
@@ -329,22 +329,21 @@ extension Subchannel {
     guard isClosed else { return }
 
     switch reason {
-    case .idleTimeout:
-      // Connection closed due to an idle timeout; notify the load balancer about this.
+    case .idleTimeout, .remote, .error(_, wasIdle: true):
+      // Connection closed due to an idle timeout or the remote telling it to GOAWAY; notify the
+      // load balancer about this.
```
gjcairo, line 334: For the `.error(_, true)` case: I understand that if there were no open streams, then the connection was idle. However, is it correct to say that the connection was closed due to an idle timeout? Couldn't we have closed for some other unrelated error and it just so it happened to be that there were no open streams?

Tests/GRPCHTTP2CoreTests/Client/Connection/Connection+Equatable.swift
glbrntt: We can't compare them as `Equatable`s, but we should do the `RPCError` comparison as above. I'll do that.

Sources/GRPCHTTP2Core/Client/Connection/LoadBalancers/Subchannel.swift
glbrntt: > However, is it correct to say that the connection was closed due to an idle timeout? 

It isn't correct, no. 

However, we aren't saying that. The underlying connection is telling us why it closed and we're saying that as a result of that the state of the subchannel has changed to idle (i.e. there's no connection and we aren't trying to create one).

gjcairo (approved):

PR #1903, opened 2024-06-04
Add request compression check to interop's TestService

gjcairo:
## Motivation
`client_compressed_unary` and `client_compressed_streaming` interop tests are failing against v2's `TestService` because we don't check that the request's `expectCompressed` property matches the presence of `grpc-encoding` header.

## Modifications
Add the check, same as in v1's TestServiceProvider.

## Result
`client_compressed_unary` and `client_compressed_streaming` interop tests now pass.

### DIFF
--- a/Sources/InteroperabilityTests/TestService.swift
+++ b/Sources/InteroperabilityTests/TestService.swift
@@ -54,9 +54,9 @@ public struct TestService: Grpc_Testing_TestService.ServiceProtocol {
     // We can't validate messages at the wire-encoding layer (i.e. where the compression byte is
     // set), so we have to check via the encoding header. Note that it is possible for the header
     // to be set and for the message to not be compressed.
-    if request.message.expectCompressed.value,
-      !request.metadata.contains(where: { $0.key == "grpc-encoding" })
-    {
+    let isRequestCompressed =
+      request.metadata["grpc-encoding"].filter({ $0 != "identity" }).count > 0
+    if request.message.expectCompressed.value, !isRequestCompressed {
       throw RPCError(
         code: .invalidArgument,
         message: "Expected compressed request, but 'grpc-encoding' was missing"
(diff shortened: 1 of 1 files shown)


Sources/InteroperabilityTests/TestService.swift
```diff
@@ -51,6 +51,18 @@ public struct TestService: Grpc_Testing_TestService.ServiceProtocol {
   public func unaryCall(
     request: ServerRequest.Single<Grpc_Testing_TestService.Method.UnaryCall.Input>
   ) async throws -> ServerResponse.Single<Grpc_Testing_TestService.Method.UnaryCall.Output> {
+    // We can't validate messages at the wire-encoding layer (i.e. where the compression byte is
+    // set), so we have to check via the encoding header. Note that it is possible for the header
+    // to be set and for the message to not be compressed.
+    if request.message.expectCompressed.value,
+      !request.metadata.contains(where: { $0.key == "grpc-encoding" })
```
glbrntt, line 58: We need to be careful here: "grpc-encoding" can have the value "identity" meaning not compressed.

If `expectCompressed` is `true` then I think it's probably sufficient to get all string values for "grpc-encoding" and then check that "identity" isn't the only value.

Sources/InteroperabilityTests/TestService.swift
```diff
@@ -142,6 +154,18 @@ public struct TestService: Grpc_Testing_TestService.ServiceProtocol {
     var aggregatedPayloadSize = 0
 
     for try await message in request.messages {
+      // We can't validate messages at the wire-encoding layer (i.e. where the compression byte is
+      // set), so we have to check via the encoding header. Note that it is possible for the header
+      // to be set and for the message to not be compressed.
+      if message.expectCompressed.value,
+        !request.metadata.contains(where: { $0.key == "grpc-encoding" })
```
glbrntt, line 161: nit: can we pull this scan out of the loop so that we don't do it for each message?

glbrntt (changes requested):

glbrntt (approved):

PR #1907, opened 2024-06-05
Add the pick-first load balancer

glbrntt:


Sources/GRPCHTTP2Core/Client/Connection/LoadBalancers/PickFirstLoadBalancer.swift
```diff
@@ -0,0 +1,530 @@
...
+      await subchannel.run()
+    }
+
+    group.addTask {
+      for await event in subchannel.events {
+        switch event {
+        case .connectivityStateChanged(let state):
+          self.handleSubchannelConnectivityStateChange(state, id: subchannel.id)
+        case .goingAway:
+          ()  // self.handleSubchannelGoingAway(key: key)
```
gjcairo, line 211: Is this comment supposed to be a to-do?

Sources/GRPCHTTP2Core/Client/Connection/LoadBalancers/PickFirstLoadBalancer.swift
```diff
@@ -0,0 +1,530 @@
...
+      ()
+    }
+  }
+
+  private func handleCloseInput() {
+    let onClose = self.state.withLockedValue { $0.close() }
+    switch onClose {
+    case .closeSubchannels(let subchannel1, let subchannel2):
+      // TODO: when does this finish?
+      self.event.continuation.yield(.connectivityStateChanged(.shutdown))
```
gjcairo, line 248: Should this be finished after this yield? Are we supposed to get more events after we've emitted shutdown?

glbrntt: Nope, this was a leftover TODO before I added `.closed` which is where it should be done. The reason we wait is that that `.shutdown` is emitted when the LB begins shutting down. We still need to wait for the existing subchannels to shutdown.

gjcairo (approved):

PR #1911, opened 2024-06-06
Pass `CallOptions` through generated stub

clintonpi:
Motivation:

`CallOptions` is passed through the RPC methods in `GRPCClient` and set to the default if none (`CallOptions`) is provided. In order to make `CallOptions` easier to access, it will be passed through the RPC methods in the generated stub.

Modifications:

- Unset the default `CallOptions` for the RPC methods in `GRPCClient`.
- Add a new `options: CallOptions` parameter to the RPC methods in the generated stub, with a default value of `CallOptions.defaults`.

Result:

`CallOptions` will be passed through the generated stub rather than through `GRPCClient`.

### DIFF
--- a/Sources/GRPCCodeGen/Internal/Translator/ClientCodeTranslator.swift
+++ b/Sources/GRPCCodeGen/Internal/Translator/ClientCodeTranslator.swift
@@ -111,8 +111,8 @@ extension ClientCodeTranslator {
         for: $0,
         in: service,
         from: codeGenerationRequest,
-        shouldBeImplemented: false,
-        callOptionsShouldDefault: false
+        includeBody: false,
+        includeDefaultCallOptions: false
       )
     }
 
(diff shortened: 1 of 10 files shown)


Sources/GRPCCodeGen/Internal/Translator/ClientCodeTranslator.swift
```diff
@@ -264,6 +275,18 @@ extension ClientCodeTranslator {
     )
   }
 
+  private func callOptionsParameter(
```
glbrntt, line 278: All this function does is call the `init` of `ParameterDescription` and this function is only ever used in one place. Having a function for this makes the code _less_ readable as there is more indirection.

It'd be better to just create the `ParameterDescription` inline in `makeParameters`.

glbrntt (changes requested):
There are a couple of small things we need to change but overall this looks good

glbrntt (approved):
Looks great, thanks Clinton!

PR #1914, opened 2024-06-07
Fix a few invalid state transitions

glbrntt:
Motivation:

The server stream state machine has a couple of rough edges. Broadly
speaking there are three categories of issue:

1. Mixing protocol errors with invalid state.
2. Remaining open when the server closes.
3. Not making a state transition.

In (1) an incorrectly implemented client could hit an assertion failure
because the state was believed to be unreachable. This isn't right, we
should close the stream instead.

For (2) the RPC is considered done when the server sends the final
status. In such situations any input from the client should be dropped
on the floor, it isn't necessarily invalid, but the server has no use
for it as the processing of the RPC has been completed.

For (3) this was missed a number of times when the server receives the
initial headers. To work around this the processing of headers was moved
to be a member function of the substate which also returns the next
state the state machine should transition to. This forces the
implementer to think about state transitions as well as side effects.

Modifications:

- Avoid invalid state in a few places
- Move to client and server closed when the server sends a status

Results:

Fewer bugs

### DIFF
--- a/Tests/GRPCHTTP2CoreTests/GRPCStreamStateMachineTests.swift
+++ b/Tests/GRPCHTTP2CoreTests/GRPCStreamStateMachineTests.swift
@@ -382,7 +382,7 @@ final class GRPCStreamClientStateMachineTests: XCTestCase {
     XCTAssertNoThrow(try stateMachine.receive(buffer: receivedUncompressedBytes, endStream: false))
     var receivedAction = stateMachine.nextInboundMessage()
     switch receivedAction {
-    case .noMoreMessages, .awaitMoreMessages:
+    case .noMoreMessages, .awaitMoreMessages, .doNothing:
       XCTFail("Should have received message")
     case .receiveMessage(let receivedMessaged):
       XCTAssertEqual(originalMessage, receivedMessaged)
@@ -398,7 +398,7 @@ final class GRPCStreamClientStateMachineTests: XCTestCase {
     )
     receivedAction = stateMachine.nextInboundMessage()
     switch receivedAction {
-    case .noMoreMessages, .awaitMoreMessages:
+    case .noMoreMessages, .awaitMoreMessages, .doNothing:
       XCTFail("Should have received message")
     case .receiveMessage(let receivedMessaged):
       XCTAssertEqual(originalMessage, receivedMessaged)
@@ -417,7 +417,7 @@ final class GRPCStreamClientStateMachineTests: XCTestCase {
     switch receivedAction {
     case .awaitMoreMessages:
       ()
-    case .noMoreMessages:
+    case .noMoreMessages, .doNothing:
       XCTFail("Should be awaiting for more messages")
     case .receiveMessage:
       XCTFail("Should not have received message")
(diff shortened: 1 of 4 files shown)


Tests/GRPCHTTP2CoreTests/GRPCStreamStateMachineTests.swift
```diff
@@ -1921,39 +1921,26 @@ final class GRPCStreamServerStateMachineTests: XCTestCase {
...
-      ofType: RPCError.self,
-      try stateMachine.receive(headers: .clientInitialMetadata, endStream: false)
-    ) { error in
-      XCTAssertEqual(error.code, .internalError)
-      XCTAssertEqual(error.message, "Client shouldn't have sent metadata twice.")
-    }
+    XCTAssertEqual(
+      try stateMachine.receive(headers: .clientInitialMetadata, endStream: false),
+      .doNothing
+    )
```
gjcairo, line 1927: I think we shouldn't just ignore these: if the client sends metadata more than once we should close the stream instead. My main concern is that otherwise we are vulnerable to bad actors / buggy clients spamming the server indefinitely with metadata packets.

Tests/GRPCHTTP2CoreTests/GRPCStreamStateMachineTests.swift
```diff
@@ -1982,14 +1969,10 @@ final class GRPCStreamServerStateMachineTests: XCTestCase {
...
-    XCTAssertThrowsError(
-      ofType: RPCError.self,
-      try stateMachine.receive(headers: .clientInitialMetadata, endStream: false)
-    ) { error in
-      XCTAssertEqual(error.code, .internalError)
-      XCTAssertEqual(error.message, "Client can't have sent metadata if closed.")
-    }
+    XCTAssertEqual(
+      try stateMachine.receive(headers: .clientInitialMetadata, endStream: false),
+      .doNothing
```
gjcairo, line 1974: If the client is closed, should we be receiving data from them?

Tests/GRPCHTTP2CoreTests/GRPCStreamStateMachineTests.swift
```diff
@@ -2686,7 +2659,145 @@ final class GRPCStreamServerStateMachineTests: XCTestCase {
...
+    XCTAssertEqual(action1, .rejectRPC(trailers: trailers))
+
+    // Extra input should be ignored.
+    for endStream in [false, true] {
+      let action = try state.receive(buffer: ByteBuffer(), endStream: endStream)
+      XCTAssertEqual(action, .doNothing)
+    }
+  }
+
+  func testServerReceivesUnsupportedCompression() throws {
```
gjcairo, line 2768: I think we already have this test: `testReceiveMetadataWhenClientIdleAndServerIdle_ServerUnsupportedEncoding`.
We're not asserting the last bit about extra input being ignored though.

gjcairo (changes requested):

Tests/GRPCHTTP2CoreTests/GRPCStreamStateMachineTests.swift
glbrntt: It depends when you consider the "close" to happen. When the server sends a status it indicates that the RPC is done, so anything from the client at that point can be ignored (because the server doesn't need to process it). 

When the server sends the status I think it makes sense for the state machine to close _both_ sides, but I can also see the argument for the client remaining open until the end stream is received.

Tests/GRPCHTTP2CoreTests/GRPCStreamStateMachineTests.swift
glbrntt: Ah cool, I'll remove this test.

Tests/GRPCHTTP2CoreTests/GRPCStreamStateMachineTests.swift
glbrntt: WDYT?

glbrntt:
Closing in favour of #1924

PR #1916, opened 2024-06-11
Add a service for testing

glbrntt:
Motivation:

The interop tests cover core functionality. However, we'd like to test
more scenarios in more depth. To do so we can use a service which
responds however the client asks it to. This allows, for example, the
client to ask the server to send a status-only response, or to respond
with N messages before failing with a particular status.

Modifications:

- Add a definition for a 'control' service
- Implement the control service

Result:

We have a service which the client can control the behaviour of allowing
us to exercise different code paths.


glbrntt:
This is split into two commits:
1. The "real" changes [0d6f0b2](https://github.com/grpc/grpc-swift/pull/1916/commits/0d6f0b25efa253d3375e783a71159b5817481f74)
2. Code gen [3906a7e](https://github.com/grpc/grpc-swift/pull/1916/commits/3906a7e96dbc6018a9909e686c1a0aca82a8a704)

Tests/GRPCHTTP2TransportTests/ControlService.swift
```diff
@@ -0,0 +1,194 @@
...
+      var iterator = transfer.wrappedValue
+      // Process the rest of the messages.
+      while let message = try await iterator.next() {
+        switch try await self.processMessage(message, metadata: request.metadata, writer: writer) {
+        case .return(let metadata):
+          return metadata
+        case .continue:
+          ()
+        }
+      }
```
gjcairo, line 99: Nit, but we could replace this with a `repeat-while` loop to avoid some duplication

glbrntt: Hmm, not sure this works because we need `message` to be non-nil inside the loop, with a repeat-while we'd have to `!`-it.

gjcairo: Yeah you're right, it's also not possible to do optional binding here so it's okay, ignore this

gjcairo (approved):

PR #1920, opened 2024-06-11
Add editorconfig

glbrntt:
Motivation:

Different projects use different indentation settings. We can configure project specific settings in '.editorconfig' which is supported in Xcode 16 beta.

Modifications:

- Add '.editorconfig' with settings for our project

Result:

Less friction when developing multiple projects with different editor settings.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,24 @@
+# Copyright 2024, gRPC Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


gjcairo (approved):

.editorconfig
```diff
@@ -0,0 +1,24 @@
+# Copyright 2024, gRPC Authors All rights reserved.
```
FranzBusch, line 1: I don't think this needs a copy right notice :D

FranzBusch: Similar to our formatting configs which don't have one either

PR #1921, opened 2024-06-12
Add client to v2 interop tests

gjcairo:
## Motivation
The counterpart to https://github.com/grpc/grpc-swift/pull/1897

## Modifications
We now have the GRPCClient integrated with the v2 interop tests. 

## Result
We can now run v2 client against v2 server interop tests.

### DIFF
--- a/Sources/InteroperabilityTests/InteroperabilityTestCase.swift
+++ b/Sources/InteroperabilityTests/InteroperabilityTestCase.swift
@@ -66,6 +66,27 @@ public enum InteroperabilityTestCase: String, CaseIterable {
   }
 }
 
+public struct InteroperabilityTestError: Error {
+  private enum _Error {
+    case testNotFound(name: String)
+    case testFailed(cause: any Error)
+  }
+
+  private let _error: _Error
+
+  private init(_error: _Error) {
+    self._error = _error
+  }
+
+  public static func testNotFound(name: String) -> Self {
+    Self(_error: .testNotFound(name: name))
+  }
+
+  public static func testFailed(cause error: any Error) -> Self {
+    Self(_error: .testFailed(cause: error))
+  }
+}
+
 @available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
 extension InteroperabilityTestCase {
   /// Return a new instance of the test case.
--- a/Sources/InteroperabilityTests/InteroperabilityTestCase.swift
+++ b/Sources/InteroperabilityTests/InteroperabilityTestCase.swift
@@ -66,27 +66,6 @@ public enum InteroperabilityTestCase: String, CaseIterable {
   }
 }
 
-public struct InteroperabilityTestError: Error {
-  private enum _Error {
-    case testNotFound(name: String)
-    case testFailed(cause: any Error)
-  }
-
-  private let _error: _Error
-
-  private init(_error: _Error) {
-    self._error = _error
-  }
-
-  public static func testNotFound(name: String) -> Self {
-    Self(_error: .testNotFound(name: name))
-  }
-
-  public static func testFailed(cause error: any Error) -> Self {
-    Self(_error: .testFailed(cause: error))
-  }
-}
-
 @available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
 extension InteroperabilityTestCase {
   /// Return a new instance of the test case.
--- a/Sources/interoperability-tests/InteroperabilityTestsExecutable.swift
+++ b/Sources/interoperability-tests/InteroperabilityTestsExecutable.swift
@@ -26,7 +26,7 @@ import NIOPosix
 struct InteroperabilityTestsExecutable: AsyncParsableCommand {
   static var configuration = CommandConfiguration(
     abstract: "gRPC Swift Interoperability Runner",
-    subcommands: [StartServer.self, ListTests.self]
+    subcommands: [StartServer.self, ListTests.self, RunTest.self, RunAllTests.self]
   )
 
   struct StartServer: AsyncParsableCommand {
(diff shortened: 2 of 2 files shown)


Sources/InteroperabilityTests/InteroperabilityTestCase.swift
```diff
@@ -66,6 +66,27 @@ public enum InteroperabilityTestCase: String, CaseIterable {
   }
 }
 
+public struct InteroperabilityTestError: Error {
```
glbrntt, line 69: This isn't public API (because the interop tests aren't a product of the package), so this _can_ just be a `public enum`.

glbrntt: Although, we don't use it in this module, so can't we just define it in `interoperability-tests`?

glbrntt: Also, can we conform it to `CustomStringConvertible`?

Sources/interoperability-tests/InteroperabilityTestsExecutable.swift
```diff
@@ -60,4 +60,95 @@ struct InteroperabilityTestsExecutable: AsyncParsableCommand {
...
+
+    internal static func runTest(
+      testCase: InteroperabilityTestCase,
+      host: String,
+      port: Int
+    ) async throws {
+      let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+      defer {
+        try! group.syncShutdownGracefully()
+      }
```
glbrntt, line 98: No need to do this, just use the default param in the `init` which is the singleton MTELG

Sources/interoperability-tests/InteroperabilityTestsExecutable.swift
```diff
@@ -26,7 +26,7 @@ import NIOPosix
 struct InteroperabilityTestsExecutable: AsyncParsableCommand {
   static var configuration = CommandConfiguration(
     abstract: "gRPC Swift Interoperability Runner",
-    subcommands: [StartServer.self, ListTests.self]
+    subcommands: [StartServer.self, ListTests.self, RunTest.self, RunAllTests.self]
```
glbrntt, line 29: Rather than having two subcommands, can we just one "RunTests"? If we pass no arguments then it should run all tests, otherwise it should run all specified tests.

This way we have a single subcommand for running tests and more flexibility as we can run whichever tests we like in one go.

Sources/interoperability-tests/InteroperabilityTestsExecutable.swift
```diff
@@ -60,4 +60,95 @@ struct InteroperabilityTestsExecutable: AsyncParsableCommand {
...
+        }
+
+        group.addTask {
+          print("Running '\(testCase.name)' ... ", terminator: "")
+          do {
+            try await testCase.makeTest().run(client: client)
+            print("PASSED")
+          } catch {
+            print("FAILED")
+            throw InteroperabilityTestError.testFailed(cause: error)
```
glbrntt, line 123: If we run more than one test we shouldn't throw an error at this point, we should wait for all tests to run first.

Sources/interoperability-tests/InteroperabilityTestsExecutable.swift
```diff
@@ -60,4 +60,95 @@ struct InteroperabilityTestsExecutable: AsyncParsableCommand {
...
+        eventLoopGroup: group
+      )
+      let client = GRPCClient(transport: transport)
+
+      try await withThrowingTaskGroup(of: Void.self) { group in
+        group.addTask {
+          try await client.run()
+        }
+
+        group.addTask {
```
glbrntt, line 116: We don't need to add a subtask for this

Sources/interoperability-tests/InteroperabilityTestsExecutable.swift
```diff
@@ -60,4 +60,95 @@ struct InteroperabilityTestsExecutable: AsyncParsableCommand {
...
+            try await testCase.makeTest().run(client: client)
+            print("PASSED")
+          } catch {
+            print("FAILED")
+            throw InteroperabilityTestError.testFailed(cause: error)
+          }
+        }
+
+        try await group.next()
+        group.cancelAll()
```
glbrntt, line 128: We can just close the client here and let it finish gracefully.

glbrntt (changes requested):

Sources/interoperability-tests/InteroperabilityTestsExecutable.swift
```diff
@@ -60,4 +60,98 @@ struct InteroperabilityTestsExecutable: AsyncParsableCommand {
...
+    )
+
+    @Option(help: "The host the server is running on")
+    var host: String
+
+    @Option(help: "The port to connect to")
+    var port: Int
+
+    @Argument(help: "The name of the test to run. If absent, all tests will be run.")
+    var testName: String?
```
glbrntt, line 80: Sorry, I meant that we can specify _multiple_ arguments, i.e. `var testNames: [String]`

Sources/interoperability-tests/InteroperabilityTestsExecutable.swift
```diff
@@ -60,4 +60,98 @@ struct InteroperabilityTestsExecutable: AsyncParsableCommand {
...
+    internal static func runTest(
+      testCase: InteroperabilityTestCase,
+      host: String,
+      port: Int
+    ) async throws {
+      var transportConfig = HTTP2ClientTransport.Posix.Config.defaults
+      transportConfig.compression.enabledAlgorithms = .all
+      let serviceConfig = ServiceConfig(loadBalancingConfig: [.roundRobin])
+      let transport = try HTTP2ClientTransport.Posix(
+        target: .ipv4(host: "0.0.0.0", port: port),
```
glbrntt, line 119:
```suggestion
        target: .ipv4(host: host, port: port),
```

Sources/interoperability-tests/InteroperabilityTestsExecutable.swift
```diff
@@ -60,4 +60,98 @@ struct InteroperabilityTestsExecutable: AsyncParsableCommand {
...
+        config: transportConfig,
+        serviceConfig: serviceConfig,
+        eventLoopGroup: .singletonMultiThreadedEventLoopGroup
```
glbrntt, line 122: This should be the default arg so you shouldn't need to pass it at all

Sources/interoperability-tests/InteroperabilityTestsExecutable.swift
```diff
@@ -60,4 +60,98 @@ struct InteroperabilityTestsExecutable: AsyncParsableCommand {
...
+
+    internal static func runTest(
+      testCase: InteroperabilityTestCase,
+      host: String,
+      port: Int
+    ) async throws {
+      var transportConfig = HTTP2ClientTransport.Posix.Config.defaults
+      transportConfig.compression.enabledAlgorithms = .all
+      let serviceConfig = ServiceConfig(loadBalancingConfig: [.roundRobin])
+      let transport = try HTTP2ClientTransport.Posix(
```
glbrntt, line 118: Can we re-use the client across all tests?

glbrntt (changes requested):

Sources/interoperability-tests/InteroperabilityTestsExecutable.swift
glbrntt: Yes, if it's empty then run everything. If you don't want to run any tests then don't call the CLI 🙂.

glbrntt (approved):

PR #1923, opened 2024-06-12
Add pick first to grpc-channel

glbrntt:
Motivation:

The pick first load balancer was added in b170fb88. However, `GRPCChannel` wasn't updated to support it.

Modifications:

Add support to `GRPCChannel`

Result:

`GRPCChannel` can use the pick first LB


Tests/GRPCHTTP2CoreTests/Client/Connection/GRPCChannelTests.swift
```diff
@@ -547,6 +547,159 @@ final class GRPCChannelTests: XCTestCase {
       }
     }
   }
+
+  func testLoadBalancerChangingFromRoundRobinToPickFirst() async throws {
```
gjcairo, line 551: Should we add a test to assert that if no valid load balancing config was provided we fall back to pick first?

Tests/GRPCHTTP2CoreTests/Client/Connection/GRPCChannelTests.swift
```diff
@@ -547,6 +547,159 @@ final class GRPCChannelTests: XCTestCase {
...
+        default:
+          ()
+        }
+      }
+    }
+  }
+
+  func testPickFirstShufflingAddressList() async throws {
+    // This test checks that the pick first load-balancer has its address list shuffled. We can't
+    // this deterministically, so instead we'll run an experiment a number of times. Each round
```
gjcairo, line 633: Nit:
```suggestion
    // This test checks that the pick first load-balancer has its address list shuffled. We can't assert
    // this deterministically, so instead we'll run an experiment a number of times. Each round
```

gjcairo (approved):

PR #1928, opened 2024-06-13
Add missing v2 interop tests

gjcairo:
This PR adds some missing implementations for interoperability tests in V2.

Note: `TimeoutOnSleepingServer` currently fails until we properly propagate timeouts.

### DIFF
--- a/Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
+++ b/Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
@@ -192,6 +192,24 @@ private enum GRPCStreamStateMachineState {
 
     var inboundMessageBuffer: OneOrManyQueue<[UInt8]>
 
+    /// This transition should only happen on the client-side.
+    /// It can happen if the request times out before the client outbound can be opened, or if the stream is
+    /// unexpectedly closed for some other reason on the client before it can transition to open.
+    init(previousState: ClientIdleServerIdleState) {
+      self.maximumPayloadSize = previousState.maximumPayloadSize
+      // We don't need a compressor since we won't be sending any messages.
+      self.framer = GRPCMessageFramer()
+      self.compressor = nil
+      self.outboundCompression = .none
+
+      // We haven't received anything from the server.
+      self.deframer = nil
+      self.decompressor = nil
+
+      self.inboundMessageBuffer = .init()
+    }
+
+    /// This transition should only happen on the server-side.
     /// We are closing the client as soon as it opens (i.e., endStream was set when receiving the client's
     /// initial metadata). We don't need to know a decompression algorithm, since we won't receive
     /// any more messages from the client anyways, as it's closed.
@@ -625,8 +643,8 @@ extension GRPCStreamStateMachine {
 
   private mutating func clientCloseOutbound() throws {
     switch self.state {
-    case .clientIdleServerIdle:
-      try self.invalidState("Client not yet open.")
+    case .clientIdleServerIdle(let state):
+      self.state = .clientClosedServerIdle(.init(previousState: state))
     case .clientOpenServerIdle(let state):
       self.state = .clientClosedServerIdle(.init(previousState: state))
     case .clientOpenServerOpen(let state):
@@ -645,16 +663,19 @@ extension GRPCStreamStateMachine {
     switch self.state {
     case .clientIdleServerIdle:
       try self.invalidState("Client is not open yet.")
+
     case .clientOpenServerIdle(var state):
       let request = try state.framer.next(compressor: state.compressor)
       self.state = .clientOpenServerIdle(state)
       return request.map { .sendFrame(frame: $0.bytes, promise: $0.promise) }
         ?? .awaitMoreMessages
+
     case .clientOpenServerOpen(var state):
       let request = try state.framer.next(compressor: state.compressor)
       self.state = .clientOpenServerOpen(state)
       return request.map { .sendFrame(frame: $0.bytes, promise: $0.promise) }
         ?? .awaitMoreMessages
+
     case .clientClosedServerIdle(var state):
       let request = try state.framer.next(compressor: state.compressor)
       self.state = .clientClosedServerIdle(state)
@@ -663,6 +684,7 @@ extension GRPCStreamStateMachine {
       } else {
         return .noMoreMessages
       }
+
     case .clientClosedServerOpen(var state):
       let request = try state.framer.next(compressor: state.compressor)
       self.state = .clientClosedServerOpen(state)
@@ -671,6 +693,7 @@ extension GRPCStreamStateMachine {
       } else {
         return .noMoreMessages
       }
+
     case .clientOpenServerClosed, .clientClosedServerClosed:
       // No point in sending any more requests if the server is closed.
       return .noMoreMessages
--- a/Sources/InteroperabilityTests/InteroperabilityTestCases.swift
+++ b/Sources/InteroperabilityTests/InteroperabilityTestCases.swift
@@ -88,6 +88,214 @@ struct LargeUnary: InteroperabilityTest {
   }
 }
 
+/// This test verifies the client can compress unary messages by sending two unary calls, for
+/// compressed and uncompressed payloads. It also sends an initial probing request to verify
+/// whether the server supports the CompressedRequest feature by checking if the probing call
+/// fails with an `INVALID_ARGUMENT` status.
+///
+/// Server features:
+/// - UnaryCall
+/// - CompressedRequest
+///
+/// Procedure:
+/// 1. Client calls UnaryCall with the feature probe, an *uncompressed* message:
+///    ```
+///    {
+///      expect_compressed:{
+///        value: true
+///      }
+///      response_size: 314159
+///      payload:{
+///        body: 271828 bytes of zeros
+///      }
+///    }
+///    ```
+/// 2. Client calls UnaryCall with the *compressed* message:
+///    ```
+///    {
+///      expect_compressed:{
+///        value: true
+///      }
+///      response_size: 314159
+///      payload:{
+///        body: 271828 bytes of zeros
+///      }
+///    }
+///    ```
+/// 3. Client calls UnaryCall with the *uncompressed* message:
+///    ```
+///    {
+///      expect_compressed:{
+///        value: false
+///      }
+///      response_size: 314159
+///      payload:{
+///        body: 271828 bytes of zeros
+///      }
+///    }
+///    ```
+///
+/// Client asserts:
+/// - First call failed with `INVALID_ARGUMENT` status.
+/// - Subsequent calls were successful.
+/// - Response payload body is 314159 bytes in size.
+/// - Clients are free to assert that the response payload body contents are zeros and comparing the
+///   entire response message against a golden response.
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+class ClientCompressedUnary: InteroperabilityTest {
+  func run(client: GRPCClient) async throws {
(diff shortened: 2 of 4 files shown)


Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -88,6 +88,214 @@ struct LargeUnary: InteroperabilityTest {
...
+      request.payload = .with { $0.body = Data(repeating: 0, count: 271_828) }
+    }
+
+    try await testServiceClient.unaryCall(
+      request: ClientRequest.Single(message: compressedRequest)
+    ) { response in
+      // We can't verify that the compression bit was set, instead we verify that the encoding header
+      // was sent by the server. This isn't quite the same since as it can still be set but the
+      // compression may _not_ be set.
+      try assertTrue(response.metadata.firstIndex(where: { $0.key == "grpc-encoding" }) != nil)
```
glbrntt, line 270: "grpc-encoding" can be "identity", instead can we test that there's a value which isn't "identity"?

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -217,6 +425,216 @@ struct ServerStreaming: InteroperabilityTest {
...
+      request.expectCompressed = .with { $0.value = false }
+      request.payload = .with { $0.body = Data(repeating: 0, count: 45_904) }
+    }
+
+    options.compression = .gzip
+
+    try await testServiceClient.streamingInputCall(
+      request: ClientRequest.Stream { writer in
+        try await writer.write(probeRequest)
+        try await writer.write(secondMessage)
```
glbrntt, line 527: We can't actually implement this test as we don't have control over the flags for individual messages, which means that in this case _both_ messages are compressed but the server is being told to have different expectations for each.

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -217,6 +425,216 @@ struct ServerStreaming: InteroperabilityTest {
...
+    let responseSizes = [31415, 92653]
+    let payloads = _LockedValueBox([Grpc_Testing_Payload]())
+
+    try await testServiceClient.streamingOutputCall(
+      request: ClientRequest.Single(message: request)
+    ) { response in
+      // We can't verify that the compression bit was set, instead we verify that the encoding header
+      // was sent by the server. This isn't quite the same since as it can still be set but the
+      // compression may be not set.
+      try assertTrue(response.metadata.firstIndex(where: { $0.key == "grpc-encoding" }) != nil)
```
glbrntt, line 608: Can we check that there is a value which isn't "identity"?

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -217,6 +425,216 @@ struct ServerStreaming: InteroperabilityTest {
...
+
+      switch response.accepted {
+      case .success(let success):
+        try assertTrue(success.metadata.firstIndex(where: { $0.key == "grpc-encoding" }) != nil)
```
glbrntt, line 612: We've already done this assertion above

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -217,6 +425,216 @@ struct ServerStreaming: InteroperabilityTest {
...
+          $0.size = 31415
+        },
+        .with {
+          $0.compressed = .with { $0.value = false }
+          $0.size = 92653
+        },
+      ]
+    }
+    let responseSizes = [31415, 92653]
+    let payloads = _LockedValueBox([Grpc_Testing_Payload]())
```
glbrntt, line 600: This locked value box isn't necessary: we should accumulate the payloads inside the response handler closure and then return them as the result (or alternatively check them in the body of the response handler)

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -681,15 +1098,53 @@ struct UnimplementedService: InteroperabilityTest {
...
+
+    try await testServiceClient.fullDuplexCall(
+      request: ClientRequest.Stream { _ in },
+      options: callOptions
+    ) { response in
+      switch response.accepted {
+      case .success:
+        throw AssertionFailure(message: "The result should be an error.")
+      case .failure(let error):
+        try assertEqual(error.code, .deadlineExceeded)
```
glbrntt, line 1146: Much like cancellation we can't rely on the client producing this error code because the mechanism for enforcing the timeout is cancellation, i.e. we'd expect a `CancellationError` here.

Also this, this test _should_ be working at the moment because we enforce the timeout on the client side, it's just that we don't yet propagate it to the server.

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -681,15 +1098,53 @@ struct UnimplementedService: InteroperabilityTest {
...
+///
+/// Client asserts:
+/// - Call completed with status DEADLINE_EXCEEDED.
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+class TimeoutOnSleepingServer: InteroperabilityTest {
+  func run(client: GRPCClient) async throws {
+    let testServiceClient = Grpc_Testing_TestService.Client(client: client)
+
+    var callOptions = CallOptions.defaults
+    callOptions.timeout = .nanoseconds(1)
```
glbrntt, line 1136: The procedure above says the timeout should be 1 millisecond

glbrntt (changes requested):
Mostly looks good, left comments on a few small things which need fixing

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -88,6 +88,214 @@ struct LargeUnary: InteroperabilityTest {
...
+    }
+
+    var uncompressedRequest = compressedRequest
+    uncompressedRequest.expectCompressed = .with { $0.value = false }
+
+    // For unary RPCs we disable compression at the call level.
+    var options = CallOptions.defaults
+
+    // With compression expected but *disabled*.
+    options.compression = CompressionAlgorithm.none
```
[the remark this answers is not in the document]
glbrntt: Oh, that's unfortunate. `nil` in this case means "do whatever the client is configured with". 

We could consider an alternative name for "none" to avoid this.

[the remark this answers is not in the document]
glbrntt: I don't really like the repetition of `compression = .notCompressed`.

Maybe the `compression` property on `CallOptions` should be a distinct type with cases which mirror the compression algorithm, e.g. something like:

```
options.compression = .useClientDefault
options.compression = .gzip
options.compression = .deflate
options.compression = .none
```

This at least makes the `nil` case more explicit.

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
gjcairo: You're right. I'll get rid of this test for now: if we end up finding a way to implement timeouts in such a way that we don't expose the CancellationError, we can add this test back. Otherwise, we can't implement it and should remain gone. I'll create a follow-up PR to fix the propagation of CancellationError though, since this is not actually working (and the test failed/evidenced this)

Sources/GRPCCore/Call/Client/CallOptions.swift
```diff
@@ -77,6 +77,42 @@ public struct CallOptions: Sendable {
...
 
+  /// Message compression algorithms.
+  public struct CompressionAlgorithm: Sendable, Equatable {
+    private enum Value {
+      case useClientTransportDefault
+      case gzip
+      case deflate
+      case none
+    }
+
```
glbrntt, line 88: Oh sorry, I didn't mean we should do this in this PR (or at all, I was just throwing out ideas for how we could improve it). For this PR we should just use `CompressionAlgorithm.none`.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -192,6 +192,23 @@ private enum GRPCStreamStateMachineState {
 
     var inboundMessageBuffer: OneOrManyQueue<[UInt8]>
 
+    /// This transition should only happen on the client-side.
+    /// It can happen if the request times out before the client outbound can be opened.
```
glbrntt, line 196: It's not just that: it can happen if the stream closes unexpectedly on the client (i.e. an error happens).

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -192,6 +192,23 @@ private enum GRPCStreamStateMachineState {
...
+    init(previousState: ClientIdleServerIdleState) {
+      self.maximumPayloadSize = previousState.maximumPayloadSize
+      // We don't need a framer or compressor since we won't be sending any messages.
+      self.framer = nil
```
glbrntt, line 200: It's fine to just create one and not use it. In fact I'd rather we did that to avoid all the `!`s.

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -623,8 +640,8 @@ extension GRPCStreamStateMachine {
 
   private mutating func clientCloseOutbound() throws {
     switch self.state {
-    case .clientIdleServerIdle:
-      try self.invalidState("Client not yet open.")
+    case .clientIdleServerIdle(let state):
+      self.state = .clientClosedServerIdle(.init(previousState: state))
```
glbrntt, line 644: Can you add a test verifying that this now _doesn't_ throw?

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -88,6 +88,217 @@ struct LargeUnary: InteroperabilityTest {
...
+    try await testServiceClient.unaryCall(
+      request: ClientRequest.Single(message: compressedRequest)
+    ) { response in
+      // We can't verify that the compression bit was set, instead we verify that the encoding header
+      // was sent by the server. This isn't quite the same since as it can still be set but the
+      // compression may _not_ be set.
+      let nonIdentityEncodingHeaderPredicate = { (metadata: Metadata.Element) in
+        metadata.key == "grpc-encoding" && metadata.value != "identity"
+      }
+      try assertTrue(response.metadata.firstIndex(where: nonIdentityEncodingHeaderPredicate) != nil)
```
glbrntt, line 273: You wrote `Metadata` so you should know the better API! 

```swift
let containsEncoding = response.metadata["grpc-encoding"].contains { $0 != "identity" }
```

Sources/InteroperabilityTests/InteroperabilityTestCases.swift
```diff
@@ -217,6 +428,108 @@ struct ServerStreaming: InteroperabilityTest {
...
+      case .success(let success):
+        // We can't verify that the compression bit was set, instead we verify that the encoding header
+        // was sent by the server. This isn't quite the same since as it can still be set but the
+        // compression may be not set.
+        let nonIdentityEncodingHeaderPredicate = { (metadata: Metadata.Element) in
+          metadata.key == "grpc-encoding" && metadata.value != "identity"
+        }
+        try assertTrue(
+          success.metadata.firstIndex(where: nonIdentityEncodingHeaderPredicate) != nil
+        )
```
glbrntt, line 506: same here re: better API

glbrntt (changes requested):

glbrntt (approved):
Nice one, thanks Gus!

PR #1933, opened 2024-06-14
Expose listening address on HTTP2 server transport

gjcairo:
## Motivation
We should expose the listening address on the H2 server transports so that our client can connect to it.

## Modifications
Exposed a `listeningAddress` async property. It will block until we bind or error, return the listening address while it's bound and listening, and throw an error when it's not listening anymore.

## Result
We can now know what address our server transport's listening on for requests.


Sources/GRPCHTTP2TransportNIOPosix/GRPCHTTP2TransportNIOPosix.swift
```diff
@@ -21,13 +21,35 @@ import NIOExtras
...
+  /// A NIOPosix-backed implementation of a server transport.
   @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
   public struct Posix: ServerTransport {
     private let address: GRPCHTTP2Core.SocketAddress
     private let config: Config
     private let eventLoopGroup: MultiThreadedEventLoopGroup
     private let serverQuiescingHelper: ServerQuiescingHelper
 
+    private let listeningAddressPromise:
+      _LockedValueBox<EventLoopPromise<GRPCHTTP2Core.SocketAddress>>
```
gjcairo, line 33: This is slightly annoying, but I backed this by a box containing the promise because I didn't want to make `listen` a mutating func, and I need to mutate the promise when we exit `listen`'s scope.

glbrntt (changes requested):

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ServerTransport+Posix.swift
```diff
@@ -21,13 +21,73 @@ import NIOExtras
...
 extension HTTP2ServerTransport {
+  /// A NIOPosix-backed implementation of a server transport.
   @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
   public struct Posix: ServerTransport {
     private let address: GRPCHTTP2Core.SocketAddress
     private let config: Config
     private let eventLoopGroup: MultiThreadedEventLoopGroup
     private let serverQuiescingHelper: ServerQuiescingHelper
 
+    private enum State {
```
glbrntt, line 32: No need to undo this but for such a tiny state machine it's totally fine storing the futures/promises inline in the enum cases!

It's also fine to omit the "-State" suffix on the state types because they're already nested within `State` (so are e.g. `State.IdleState`)

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ServerTransport+Posix.swift
```diff
@@ -37,11 +97,47 @@ extension HTTP2ServerTransport {
...
     }
 
     public func listen(
       _ streamHandler: @escaping (RPCStream<Inbound, Outbound>) async -> Void
     ) async throws {
+      defer {
+        let failedPromiseError = RuntimeError(
+          code: .serverIsStopped,
+          message: """
+            There is no listening address bound for this server: there may have been
```
glbrntt, line 119:
```suggestion
            There is no listening address bound for this server: there may have been \
```

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ServerTransport+Posix.swift
```diff
@@ -21,13 +21,73 @@ import NIOExtras
...
+    }
+    private let listeningAddressState: _LockedValueBox<State>
+
+    /// The listening address for this server transport.
+    ///
+    /// It is an `async` property because it will only return once the address has been successfully bound.
+    ///
+    /// - Throws: A runtime error will be thrown if the address could not be bound or is not bound any
+    /// longer, because the transport isn't listening anymore.
+    public var listeningAddress: NIOCore.SocketAddress {
```
glbrntt, line 72: We still want this to be a gRPC socket address, we need to convert from NIOs socket address where applicable

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ServerTransport+Posix.swift
```diff
@@ -62,6 +158,25 @@ extension HTTP2ServerTransport {
...
+        switch state {
+        case .idle(let idleState):
+          if let address = serverChannel.channel.localAddress {
+            idleState.listeningAddressPromise.succeed(address)
+          } else {
+            let failedToGetAddressError = RuntimeError(
+              code: .transportError,
+              message: "An address is not available."
+            )
+            idleState.listeningAddressPromise.fail(failedToGetAddressError)
```
glbrntt, line 171: I _think_ this can be `nil` for `vsock`, in which case we should return `self.address`

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ServerTransport+Posix.swift
```diff
@@ -62,6 +158,25 @@ extension HTTP2ServerTransport {
           }
         }
 
+      self.listeningAddressState.withLockedValue { state in
+        switch state {
+        case .idle(let idleState):
+          if let address = serverChannel.channel.localAddress {
+            idleState.listeningAddressPromise.succeed(address)
```
glbrntt, line 165: Don't make outcalls while holding the lock. Generally speaking it's better to have functions on the `State` `enum` which return something to do. This makes it easier to separate state changes from side effects and it neatly encapsulates the logic of the state machine within the state type:

```swift
extension State { 
  mutating func bound(to address: GRPCHTTP2Core.SocketAddress) -> OnBind { ... }
  mutating func closed(error: ...) -> OnClosed { ... }

  var listeningAddress: EventLoopFuture<GRPCHTTP2Core.SocketAddress> { ... }
}

```

glbrntt (changes requested):

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ServerTransport+Posix.swift
glbrntt: For a bigger state machine I'd agree, but this has so few states and functions I don't think it's worth it.

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ServerTransport+Posix.swift
gjcairo: Interesting - should we always fall back to returning `self.address` if this is `nil` (i.e. even if it's not a `vsock`)?

glbrntt: Pfft, good question. 

I think we should only succeed the promise if it's a `vsock` and fail the promise otherwise. In the failure case we should also add an assertion failure (because we might add a new address type in the future, and that'll help us surface where we should be handling it)

Sources/GRPCHTTP2Core/Internal/NIOSocketAddress+GRPCSocketAddress.swift
```diff
@@ -0,0 +1,42 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import NIOCore
+
+@_spi(Package)
+extension GRPCHTTP2Core.SocketAddress {
+  public init?(_ nioSocketAddress: NIOCore.SocketAddress) {
```
glbrntt, line 21: Rather than returning `nil`, I think for the port we can default to zero (I don't think it can actually return `nil` for the IP addresses) and for the pathname to an empty string (I don't think it should ever return `nil` anyway).

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ServerTransport+Posix.swift
```diff
@@ -21,13 +21,138 @@ import NIOExtras
...
+                message: "Invalid address returned by transport."
+              )
+              self = .closedOrInvalidAddress(invalidAddressError)
+              return .failPromise(listeningAddressPromise, error: invalidAddressError)
+            }
+
+            self = .listening(listeningAddressPromise)
+            return .succeedPromise(listeningAddressPromise, address: validAddress)
+
+          } else if userProvidedAddress.unixDomainSocket != nil {
```
glbrntt, line 80: Why UDS? Is this meant to be vsock?

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ServerTransport+Posix.swift
```diff
@@ -21,13 +21,138 @@ import NIOExtras
...
+            self = .listening(listeningAddressPromise)
+            return .succeedPromise(listeningAddressPromise, address: userProvidedAddress)
+
+          } else {
+            assertionFailure("Unknown address type")
+            let invalidAddressError = RuntimeError(
+              code: .transportError,
+              message: "Invalid address returned by transport."
```
glbrntt, line 88: Hmm I don't think it's invalid: the transport successfully bound to an address if we reached this point. "Unknown" maybe?

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ServerTransport+Posix.swift
```diff
@@ -21,13 +21,138 @@ import NIOExtras
...
+            guard let validAddress = GRPCHTTP2Core.SocketAddress(address) else {
+              let invalidAddressError = RuntimeError(
+                code: .transportError,
+                message: "Invalid address returned by transport."
+              )
+              self = .closedOrInvalidAddress(invalidAddressError)
+              return .failPromise(listeningAddressPromise, error: invalidAddressError)
+            }
+
+            self = .listening(listeningAddressPromise)
```
glbrntt, line 77: Can we store the future here instead? Once we're in the listening state the promise must've been completed so there's no need to hold the promise, completing it again won't do anything.

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ServerTransport+Posix.swift
```diff
@@ -21,13 +21,138 @@ import NIOExtras
...
+        )
+
+        switch self {
+        case .idle(let listeningAddressPromise):
+          self = .closedOrInvalidAddress(serverStoppedError)
+          return .failPromise(listeningAddressPromise, error: serverStoppedError)
+
+        case .listening(let listeningAddressPromise):
+          self = .closedOrInvalidAddress(serverStoppedError)
+          return .failPromise(listeningAddressPromise, error: serverStoppedError)
```
glbrntt, line 125: Failing this promise won't do anything, it's already been completed

Tests/GRPCHTTP2TransportNIOPosixTests/HTTP2TransportNIOPosixTests.swift
```diff
@@ -0,0 +1,144 @@
...
+    )
+
+    try await withThrowingDiscardingTaskGroup { group in
+      group.addTask {
+        try await transport.listen { _ in }
+      }
+
+      group.addTask {
+        let address = try await transport.listeningAddress
+        XCTAssertNotNil(address.ipv4)
```
glbrntt, line 36: Can we assert that the port isn't zero?

Tests/GRPCHTTP2TransportNIOPosixTests/HTTP2TransportNIOPosixTests.swift
```diff
@@ -0,0 +1,144 @@
...
+    let transport = GRPCHTTP2Core.HTTP2ServerTransport.Posix(address: .ipv6(host: "::1", port: 0))
+
+    try await withThrowingDiscardingTaskGroup { group in
+      group.addTask {
+        try await transport.listen { _ in }
+      }
+
+      group.addTask {
+        let address = try await transport.listeningAddress
+        XCTAssertNotNil(address.ipv6)
```
glbrntt, line 52: Can we assert that the port isn't zero?

Tests/GRPCHTTP2TransportNIOPosixTests/HTTP2TransportNIOPosixTests.swift
```diff
@@ -0,0 +1,144 @@
...
+      group.addTask {
+        let address = try await transport.listeningAddress
+        XCTAssertEqual(
+          address.unixDomainSocket,
+          GRPCHTTP2Core.SocketAddress.UnixDomainSocket(path: "/tmp/test")
+        )
+        transport.stopListening()
+      }
+    }
+  }
```
glbrntt, line 77: Can we have a unix-only vsock test? There's a function called `vsockAvailable()` somewhere in the repo which you can copy/use to check whether the test should be run or not

Tests/GRPCHTTP2TransportNIOPosixTests/HTTP2TransportNIOPosixTests.swift
```diff
@@ -0,0 +1,144 @@
...
+
+  func testGetListeningAddress_InvalidAddress() async {
+    let transport = GRPCHTTP2Core.HTTP2ServerTransport.Posix(
+      address: .unixDomainSocket(path: "/this/is/an/invalid/path")
```
glbrntt, line 81: Probably 😄

glbrntt (changes requested):

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ServerTransport+Posix.swift
gjcairo: Oops, yes, good catch

Tests/GRPCHTTP2TransportNIOPosixTests/HTTP2TransportNIOPosixTests.swift
gjcairo: Heh I agree - is there a better way to get a surely nonexistent path?

glbrntt: Maybe the empty path? If not this should be fine

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ServerTransport+Posix.swift
```diff
@@ -21,13 +21,132 @@ import NIOExtras
...
+            an error which caused the transport to close, or it may have shut down.
+            """
+        )
+
+        switch self {
+        case .idle(let listeningAddressPromise):
+          self = .closedOrInvalidAddress(serverStoppedError)
+          return .failPromise(listeningAddressPromise, error: serverStoppedError)
+
+        case .listening(let listeningAddressPromise):
```
glbrntt, line 117: It's a future, but we also don't use it here so we can ignore it.

```suggestion
        case .listening:
```

Tests/GRPCHTTP2TransportTests/HTTP2TransportNIOPosixTests.swift
```diff
@@ -0,0 +1,166 @@
...
+
+    try await withThrowingDiscardingTaskGroup { group in
+      group.addTask {
+        try await transport.listen { _ in }
+      }
+
+      group.addTask {
+        let address = try await transport.listeningAddress
+        XCTAssertNotNil(address.ipv4)
+        XCTAssertNotEqual(address.ipv4!.port, 0)
```
glbrntt, line 37: nit: prefer using `XCTUnwrap`

```suggestion
        let ipv4 = try XCTUnwrap(address?.ipv4)
        XCTAssertNotEqual(ipv4.port, 0)
```

Tests/GRPCHTTP2TransportTests/HTTP2TransportNIOPosixTests.swift
```diff
@@ -0,0 +1,166 @@
...
+
+    try await withThrowingDiscardingTaskGroup { group in
+      group.addTask {
+        try await transport.listen { _ in }
+      }
+
+      group.addTask {
+        let address = try await transport.listeningAddress
+        XCTAssertNotNil(address.ipv6)
+        XCTAssertNotEqual(address.ipv6!.port, 0)
```
glbrntt, line 54: nit: prefer using `XCTUnwrap`

```suggestion
        let ipv6 = try XCTUnwrap(address?.ipv6)
        XCTAssertNotEqual(ipv6.port, 0)
```

Tests/GRPCHTTP2TransportTests/HTTP2TransportNIOPosixTests.swift
```diff
@@ -0,0 +1,166 @@
...
+    try? await withThrowingDiscardingTaskGroup { group in
+      group.addTask {
+        try await transport.listen { _ in }
+      }
+
+      group.addTask {
+        do {
+          _ = try await transport.listeningAddress
+        } catch let error as RuntimeError {
+          errorThrown.withLockedValue { $0 = true }
```
glbrntt, line 116: We don't need a locked value box here, if we `XCTFail` on the happy-path then we know each branch is covered. We also know that all must run and complete before the task group returns.

Tests/GRPCHTTP2TransportTests/HTTP2TransportNIOPosixTests.swift
```diff
@@ -0,0 +1,166 @@
...
+    let errorThrown = _LockedValueBox(false)
+
+    try? await withThrowingDiscardingTaskGroup { group in
+      group.addTask {
+        try await transport.listen { _ in }
+
+        do {
+          _ = try await transport.listeningAddress
+        } catch let error as RuntimeError {
+          errorThrown.withLockedValue { $0 = true }
```
glbrntt, line 145: Same note here about the locked value box

glbrntt (changes requested):

glbrntt (approved):
LGTM, thanks Gus!

PR #1937, opened 2024-06-14
Remove input/output typealiases for RPC methods

clintonpi:
Motivation:

The generated code for v2 includes typealiases for the input/output types for each RPC method. These typealiases end up making things less clear to the end user as they must follow the typealias to know what type is being used.

Modifications:

- Adjust `GRPCCodeGen` (both for the client and server) to not generate typealiases for the input/output types.
- Adjust `TestService` to use the actual message types.
- Update the generated code examples in `ClientCodeTranslator` and `ServerCodeTranslator`.

Result:

The generated code will use the actual message types for the input/output types.


glbrntt (approved):
Great, thanks @clintonpi!

PR #1938, opened 2024-06-17
Enable SwiftPM build plugin to use the new code generator plugin

clintonpi:
Motivation:

The SwiftPM build plugin could not be configured to use the new code generator plugin.

Modifications:

- Add a `_V2` boolean option to the SwiftPM build plugin configuration. If this option is set, it is passed to protoc-gen-grpc-swift which then uses the new code generator plugin.

Result:

The SwiftPM build plugin can now be configured use the new code generator plugin.

### DIFF
--- a/Plugins/GRPCSwiftPlugin/plugin.swift
+++ b/Plugins/GRPCSwiftPlugin/plugin.swift
@@ -202,6 +202,10 @@ struct GRPCSwiftPlugin {
       protocArgs.append("--grpc-swift_opt=KeepMethodCasing=\(keepMethodCasingOption)")
     }
 
+    if let v2 {
+      protocArgs.append("--grpc-swift_opt=_V2=\(v2)")
+    }
+
     var inputFiles = [Path]()
     var outputFiles = [Path]()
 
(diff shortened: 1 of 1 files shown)


Plugins/GRPCSwiftPlugin/plugin.swift
```diff
@@ -223,6 +229,10 @@ struct GRPCSwiftPlugin {
       }
     }
 
+    if let v2 {
+      protocArgs.append("--grpc-swift_opt=_V2=\(v2)")
+    }
+
```
glbrntt, line 235: Can you move this up to where the other `protoc` args are prepared?

glbrntt (changes requested):

Plugins/GRPCSwiftPlugin/plugin.swift
```diff
@@ -85,6 +85,9 @@ struct GRPCSwiftPlugin {
 
     /// A list of invocations of `protoc` with the `GRPCSwiftPlugin`.
     var invocations: [Invocation]
+
+    /// Whether this configuration is for `grpc-swift` v2.
+    var _V2: Bool?
```
glbrntt, line 90: Sorry, I didn't missed this the first time around but this should be part of the `Invocation` configuration `struct` above

glbrntt (changes requested):

glbrntt (approved):

PR #1941, opened 2024-06-19
Don't double ack pings

glbrntt:
Motivation:

The `ServerConnectionManagementHandler` acks pings, however NIOs HTTP/2 handler does this already. This results in pings received by the server being acked twice.

Modifications:

- Don't ack pings in the `ServerConnectionManagementHandler`
- Update tests

Result:

Pings aren't ack'd twice


glbrntt:
FWIW: we've already made the equivalent change for the client.

gjcairo (approved):

PR #1942, opened 2024-06-20
Avoid dropping continuations in grpc channel

glbrntt:
Motivation:

If the request queue was non-empty and close was called on the grpc channel, the request queue would be dropped along with any continuations. The request queue can be non-empty if the active load balancer isn't in the ready state (i.e. connecting).

Moreover, if close is called while a subchannel is connecting it can result in the shutdown event being fired twice.

Modifications:

- Fail any continuations in the request queue when closing the grpc channel
- Alter when shutdown events are fired by the subchannel. At the moment they're typically fired when close is called and on some paths, when the connection is closed as well. Shutdown events are now fired when entering the closing state or when transitioning directly to closed (i.e. from idle/connected).

Result:

Continuations aren't dropped

### DIFF
--- a/Tests/GRPCHTTP2CoreTests/Client/Connection/GRPCChannelTests.swift
+++ b/Tests/GRPCHTTP2CoreTests/Client/Connection/GRPCChannelTests.swift
@@ -747,6 +747,65 @@ final class GRPCChannelTests: XCTestCase {
       group.cancelAll()
     }
   }
+
+  func testQueueRequestsThenClose() async throws {
+    // Start a server.
+    let server = TestServer(eventLoopGroup: .singletonMultiThreadedEventLoopGroup)
+    let address = try await server.bind()
+    let (resolver, continuation) = NameResolver.dynamic(updateMode: .push)
+    continuation.yield(.init(endpoints: [Endpoint(address)], serviceConfig: nil))
+
+    let channel = GRPCChannel(
+      resolver: resolver,
+      connector: .posix(),
+      config: .defaults,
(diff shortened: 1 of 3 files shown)


Tests/GRPCHTTP2CoreTests/Client/Connection/GRPCChannelTests.swift
```diff
@@ -747,6 +747,65 @@ final class GRPCChannelTests: XCTestCase {
...
+            _ = try await iterator.next()  // initial metadata
+
+            // Drop the connection, this will trigger a state change to transient failure.
+            server.clients.first?.close(promise: nil)
+          }
+
+        case .transientFailure:
+          group.addTask {
+            channel.close()
+          }
```
gjcairo, line 790: Does this have to be inside a task?

glbrntt: This test is inherently racy: we have to close after the continuation for `withStream` has been queued, but we can't order that in code (because `withStream` waits for the continuation to be resumed). 

While the channel doesn't pick up the close event until it consumes it from the async sequence, the task adds a little bit more of a delay meaning that the continuation gets queued first. However, I think we can lean into this more by adding a small sleep before calling close. I'll add that.

glbrntt: I ended up reworking the test to make it a bit more reliable.

PR #1944, opened 2024-06-21
Various fixes to the benchmark client and server

glbrntt:
Motivation:

The benchmark client and server were implemented before we had a working transport making them impossible to test. As it turns out there were a few issues.

Modifications:

- BenchmarkClient:
  - Some RPC types have been removed. All other implementations only support unary and streaming and we never run scenarios using the other types.
  - The implementation of the streaming RPC has been altered to record the latency per request/response message pair. This aligns with other implementations.
  - The implementation of the streaming RPC has also been changed to send the response message sequence into the request writer. This yields a fairly substantial performance improvement (~3.5x) over the existing implementation.
  - The streaming RPC now respects the messages per stream config being zero (meaning no limit).
  - An 'is shutting down' atomic is used to stop the client from initiating new RPCs before closing.
- BenchmarkService:
  - No semantic changes; the typealiases have been desugared following from changes in fea1b722
- WorkerService:
  - The state machine has been tightened up a bit to more clearly separate state from side effects and to avoid leaking the implementation of the state machine into the service.
  - Added logic for creating clients and servers with an HTTP/2 transport.

Result:

Can run perf tests

### DIFF
--- a/Sources/performance-worker/WorkerService.swift
+++ b/Sources/performance-worker/WorkerService.swift
@@ -15,92 +15,46 @@
  */
 
 import GRPCCore
+import GRPCHTTP2Core
+import GRPCHTTP2TransportNIOPosix
 import NIOConcurrencyHelpers
 import NIOCore
+import NIOPosix
 
 @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
-final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable {
+final class WorkerService: Sendable {
   private let state: NIOLockedValueBox<State>
 
   init() {
-    let clientAndServer = State()
-    self.state = NIOLockedValueBox(clientAndServer)
+    self.state = NIOLockedValueBox(State())
   }
 
   private struct State {
-    var role: Role?
+    private var role: Role
 
     enum Role {
-      case client(ClientState)
-      case server(ServerState)
+      case none
+      case client(Client)
+      case server(Server)
     }
 
-    struct ServerState {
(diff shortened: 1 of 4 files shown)


Sources/performance-worker/WorkerService.swift
```diff
@@ -114,98 +68,176 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+        )
+        action = .invalidState(error)
+      }
 
-      case .client(_):
-        throw RPCError(code: .failedPrecondition, message: "Clients have already been set up.")
+      return action
+    }
 
+    mutating func serverShutdown() -> EventLoopGroup? {
```
gjcairo, line 154: Bit of a nit, but should we have an action with cases `shutdownEventLoopGroup(EventLoopGroup)` and `nothing` instead of this optional ELG? It would be more consistent with other funcs and also clearer about what we're supposed to do with the ELG).

Sources/performance-worker/WorkerService.swift
```diff
@@ -114,98 +68,176 @@ final class WorkerService: Grpc_Testing_WorkerService.ServiceProtocol, Sendable
...
+        self.role = .none
+        return state.eventLoopGroup
       case .none:
-        self.role = .client(clientState)
+        return nil
       }
     }
 
-    mutating func updateRPCStats() throws {
+    func stopListening() -> GRPCServer? {
```
gjcairo, line 166: Similar to the above, maybe we can have an action here instead of the optional server?

gjcairo (approved):

PR #1947, opened 2024-06-21
Propagate timeout to server

glbrntt:
Motivation:

gRPC timeouts can be propagated to the server via the 'grpc-timeout' metadata field. The server can then also choose to enforce the timeout. At the moment this value isn't propagated, but it should be.

Modifications:

- Convert the timeout to a deadline when it's passed to the rpc executor and use this internally. Compute a timeout from the deadline and current time when executing an RPC.

Result:

Timeout is propagated to server

### DIFF
--- a/Tests/GRPCCoreTests/Call/Client/Internal/ClientRPCExecutorTests.swift
+++ b/Tests/GRPCCoreTests/Call/Client/Internal/ClientRPCExecutorTests.swift
@@ -227,6 +227,7 @@ final class ClientRPCExecutorTests: XCTestCase {
   }
 
   func testTimeoutIsPropagated() async throws {
+    // 'nil' means no retires or hedging, just try to execute the RPC once.
     var policies: [RPCExecutionPolicy?] = [nil]
 
     let retryPolicy = RetryPolicy(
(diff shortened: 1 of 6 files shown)


Tests/GRPCCoreTests/Call/Client/Internal/ClientRPCExecutorTests.swift
```diff
@@ -223,4 +225,47 @@ final class ClientRPCExecutorTests: XCTestCase {
     XCTAssertEqual(tester.clientStreamOpenFailures, 1)
     XCTAssertEqual(tester.serverStreamsAccepted, 0)
   }
+
+  func testTimeoutIsPropagated() async throws {
+    var policies: [RPCExecutionPolicy?] = [nil]
```
gjcairo, line 230: This `nil` tests the one-shot executor, right? Maybe it's a bit annoying but do you mind adding a comment explaining this? It's not immediately obvious from looking at the test.

gjcairo (approved):
LGTM, just left a small comment

Tests/GRPCCoreTests/Call/Client/Internal/ClientRPCExecutorTests.swift
glbrntt, line 230:
```suggestion
    // 'nil' means no retires or hedging, just try to execute the RPC once.
    var policies: [RPCExecutionPolicy?] = [nil]
```

PR #1949, opened 2024-06-24
Make Subchannel internal state handling clearer

glbrntt:
Motivation:

The subchannel conflates closing (recoverable) with shutting down (terminal). Shutting down is initiated by a higher level (load-balancer, user) while closing happens when the subchannel closes unexpectedtly or is no longer required (i.e. becomes idle). A closed subchannel can be re-opeend, a shutdown subchannel can't. This distinction isn't clear enough in the state handling.

Modifications:

- Rename 'close' to 'shutDown' where applicable
- Add a new 'shutting down' state and renaming the 'closing' state to 'going away'.
- Add a state machine diagram
- Fix up a few state transitions

Result:

More robust state handling

### DIFF
--- a/Sources/GRPCHTTP2Core/Client/Connection/LoadBalancers/Subchannel.swift
+++ b/Sources/GRPCHTTP2Core/Client/Connection/LoadBalancers/Subchannel.swift
@@ -23,7 +23,7 @@ import NIOConcurrencyHelpers
 /// endpoint. You can tell it to start connecting by calling ``connect()`` and you can listen
 /// to connectivity state changes by consuming the ``events`` sequence.
 ///
-/// You must call ``close()`` on the ``Subchannel`` when it's no longer required. This will move
+/// You must call ``shutDown()`` on the ``Subchannel`` when it's no longer required. This will move
 /// it to the ``ConnectivityState/shutdown`` state: existing RPCs may continue but all subsequent
 /// calls to ``makeStream(descriptor:options:)`` will fail.
 ///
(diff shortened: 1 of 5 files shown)


Sources/GRPCHTTP2Core/Client/Connection/LoadBalancers/Subchannel.swift
```diff
@@ -363,24 +370,58 @@ extension Subchannel {
...
+  ///   │        ┌───────────────┐      │      ┌───────────────┐  │
+  ///   │        │   CONNECTED   │──shutDown──▶│ SHUTTING DOWN │  │
+  ///   │        └───────────────┘      │      └───────────────┘  │
+  ///   │                │              │              │          │
+  ///   │             goAway            │            closed       │
+  ///   │                │              │              │          │
+  ///   │                ▼              │              ▼          │
+  ///   │        ┌───────────────┐      │      ┌───────────────┐  │
+  ///   └────────│  GOING AWAY   │──────┘      │   SHUT DOWN   │◀─┘
+  ///            └───────────────┘             └───────────────┘
```
gjcairo, line 396: Thanks for this :D

gjcairo (approved):
LGTM!

Sources/GRPCHTTP2Core/Client/Connection/LoadBalancers/Subchannel.swift
glbrntt: np, it was helpful for me too! 😅

PR #1950, opened 2024-06-24
Fix NIOTS transport test

gjcairo:
There's currently a bug in NIOTS where we don't unlink UDS paths when closing the server. A test fails if we run it more than once on the same host because the address is already bound.
This PR adds a UUID to the path to prevent reusing the same address.


gjcairo:
Closing in favour of https://github.com/grpc/grpc-swift/pull/1951

PR #1953, opened 2024-06-28
Synthesise status when stream closes unexpectedly

gjcairo:
## Motivation
When the stream closes unexpectedly (the H2 stream channel becomes inactive, an error is fired through the pipeline, or we receive a RST_STREAM frame), the client-side inbound sequence either finishes or throws an error. 
To make it easier for higher layers to retry, we want to synthesise a `Status` instead.

## Modifications
- On the client side, write a `Status` when one of the aforementioned unexpected closures happens.
- On the server side, make sure we're always firing an error when closing unexpectedly.
- On both, make sure we transition the stream state machine to closed.

## Result
Inbound sequence now returns a `Status` instead of finishing/throwing


Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
gjcairo, line 1: I'm unsure if we want all these changes here. Particularly, the behaviour change on `GRPCServerStreamHandlerTests/testReceiveMultipleHeaders` (where we now throw when writing the duplicate headers on line 647; this is caused by `channelInactive` being called, because of a protocol violation error being thrown) makes me wonder if we're okay with it or not.

Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
```diff
@@ -130,7 +130,17 @@ extension GRPCClientStreamHandler {
         context.fireErrorCaught(error)
       }
 
-    case .ping, .goAway, .priority, .rstStream, .settings, .pushPromise, .windowUpdate,
+    case .rstStream:
+      switch self.stateMachine.unexpectedInboundClose(reason: .streamReset) {
+      case .forwardStatus(let status):
+        context.fireChannelRead(self.wrapInboundOut(.status(status, [:])))
+      case .doNothing:
+        context.fireChannelInactive()
```
glbrntt, line 138: We shouldn't do this when the stream is reset, only in channel inactive.

Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
```diff
@@ -148,6 +158,28 @@ extension GRPCClientStreamHandler {
   func handlerRemoved(context: ChannelHandlerContext) {
     self.stateMachine.tearDown()
   }
+
+  func channelInactive(context: ChannelHandlerContext) {
+    switch self.stateMachine.unexpectedInboundClose(reason: .channelInactive) {
+    case .forwardStatus(let status):
+      context.fireChannelRead(self.wrapInboundOut(.status(status, [:])))
```
glbrntt, line 165: We should do this _then_ `fireChannelInactive`

Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
```diff
@@ -148,6 +158,28 @@ extension GRPCClientStreamHandler {
...
+      assertionFailure("`fireError` should only happen on the server side, never on the client.")
+    }
+  }
+
+  func errorCaught(context: ChannelHandlerContext, error: any Error) {
+    switch self.stateMachine.unexpectedInboundClose(reason: .errorThrown(error)) {
+    case .forwardStatus(let status):
+      context.fireChannelRead(self.wrapInboundOut(.status(status, [:])))
+    case .doNothing:
+      context.fireErrorCaught(error)
```
glbrntt, line 178: Hmm, I'm not sure this is right because we want to avoid the extra errors. Presumably this we `doNothing` when the inbound stream is already finished?

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -1469,6 +1521,43 @@ extension GRPCStreamStateMachine {
       return .awaitMoreMessages
     }
   }
+
+  private func getServerUnexpectedInboundCloseAction(
+    for reason: UnexpectedInboundCloseReason
+  ) -> OnUnexpectedInboundClose {
```
glbrntt, line 1527: Any reason this shouldn't be an `init` on `OnUnexpectedInboundClose`?

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -1469,6 +1521,43 @@ extension GRPCStreamStateMachine {
...
+    switch reason {
+    case .streamReset, .channelInactive:
+      // We can force-unwrap here: the status code is `.unavailable`, so RPCError won't be nil
+      return .fireError(RPCError(status: Status(reason))!)
```
glbrntt, line 1531: Does this hint at the wrong construction? I can't imagine an unexpected close would ever result in an `ok` status, so it seems like the `RPCError` should be created from the reason and then, where necessary, a `Status` from the `RPCError`. WDYT?

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -1612,3 +1701,20 @@ extension MethodDescriptor {
...
+    switch reason {
+    case .streamReset:
+      self = Status(
+        code: .unavailable,
+        message: "Stream unexpectedly closed: a RST_STREAM frame was received."
+      )
+    case .channelInactive:
+      self = Status(code: .unavailable, message: "Stream unexpectedly closed.")
+    case .errorThrown(let error):
+      self = Status(code: .unavailable, message: "Stream unexpectedly closed with error: \(error).")
```
glbrntt, line 1717: I don't think we should include the error here: it's an internal error to the server. In most cases the client won't care but it could also leak internal details about the server that it shouldn't.

Tests/GRPCHTTP2CoreTests/Client/GRPCClientStreamHandlerTests.swift
```diff
@@ -19,10 +19,10 @@ import NIOCore
 import NIOEmbedded
 import NIOHPACK
 import NIOHTTP1
-import NIOHTTP2
 import XCTest
 
 @testable import GRPCHTTP2Core
+@testable import NIOHTTP2
```
glbrntt, line 25: We really shouldn't do this because we're taking a dependency on the internal API of a module we don't own.

Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
```diff
@@ -163,6 +175,32 @@ extension GRPCServerStreamHandler {
...
+  func channelInactive(context: ChannelHandlerContext) {
+    switch self.stateMachine.unexpectedInboundClose(reason: .channelInactive) {
+    case .forwardStatus:
+      assertionFailure(
+        "`forwardStatus` should only happen on the client side, never on the server."
+      )
+    case .doNothing:
+      context.fireChannelInactive()
+    case .fireError(let error):
+      context.fireErrorCaught(error)
```
glbrntt, line 188: We should fire channel inactive afterwards too

Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
```diff
@@ -163,6 +175,32 @@ extension GRPCServerStreamHandler {
...
+  }
+
+  func errorCaught(context: ChannelHandlerContext, error: any Error) {
+    switch self.stateMachine.unexpectedInboundClose(reason: .errorThrown(error)) {
+    case .forwardStatus:
+      assertionFailure(
+        "`forwardStatus` should only happen on the client side, never on the server."
+      )
+    case .doNothing:
+      context.fireErrorCaught(error)
```
glbrntt, line 199: Hmm, is `doNothing` the right name here? It's a bit confusing that do nothing fires an error here.

Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
glbrntt: What did we do before?

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
gjcairo: Hmm I ended up doing this here because we're using this on the server side only (whereas the type itself is also used on the client side logic). But that's probably a not good enough reason, so I can move it.

Tests/GRPCHTTP2CoreTests/Client/GRPCClientStreamHandlerTests.swift
gjcairo: Hm yes true - we should probably make the constructor public on NIO though, otherwise this API is unusable.

Tests/GRPCHTTP2CoreTests/Server/GRPCServerStreamHandlerTests.swift
```diff
@@ -646,7 +646,15 @@ final class GRPCServerStreamHandlerTests: XCTestCase {
...
-    try channel.writeInbound(HTTP2Frame.FramePayload.headers(.init(headers: clientInitialMetadata)))
+    XCTAssertThrowsError(
+      ofType: RPCError.self,
+      try channel.writeInbound(
+        HTTP2Frame.FramePayload.headers(.init(headers: clientInitialMetadata))
+      )
+    ) { error in
+      XCTAssertEqual(error.code, .unavailable)
+      XCTAssertEqual(error.message, "Stream unexpectedly closed.")
+    }
```
gjcairo, line 657: @glbrntt this is what I was referring to in https://github.com/grpc/grpc-swift/pull/1953#discussion_r1658780342

We were not throwing when doing this write, we were only getting a reset stream frame.

glbrntt: Gotcha. So previously the server just ignored the extra metadata? I think throwing the error here is appropriate: the server handler can't continue, the client messed up.

Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
```diff
@@ -130,7 +130,17 @@ extension GRPCClientStreamHandler {
...
 
-    case .ping, .goAway, .priority, .rstStream, .settings, .pushPromise, .windowUpdate,
+    case .rstStream:
+      switch self.stateMachine.unexpectedInboundClose(reason: .streamReset) {
+      case .forwardStatus(let status):
+        context.fireChannelRead(self.wrapInboundOut(.status(status, [:])))
+      case .doNothing:
+        ()
+      case .fireError:
+        assertionFailure("`fireError` should only happen on the server side, never on the client.")
```
glbrntt, line 140: Not a big deal, but I wonder if we should make this more explicit, e.g. `fireError_clientOnly`

Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
```diff
@@ -148,6 +158,30 @@ extension GRPCClientStreamHandler {
   func handlerRemoved(context: ChannelHandlerContext) {
     self.stateMachine.tearDown()
   }
+
+  func channelInactive(context: ChannelHandlerContext) {
+    switch self.stateMachine.unexpectedInboundClose(reason: .channelInactive) {
```
glbrntt, line 163: Can we avoid the duplication here? e.g. `func unexpectedClose(context: ChannelHandlerContext, reason: Reason)`

Sources/GRPCHTTP2Core/Server/GRPCServerStreamHandler.swift
```diff
@@ -139,7 +139,19 @@ extension GRPCServerStreamHandler {
         context.fireErrorCaught(error)
       }
 
-    case .ping, .goAway, .priority, .rstStream, .settings, .pushPromise, .windowUpdate,
+    case .rstStream:
+      switch self.stateMachine.unexpectedInboundClose(reason: .streamReset) {
```
glbrntt, line 143: Same thing re: duplication here

glbrntt (changes requested):
Couple of nits re: duplication but otherwise looks great

Sources/GRPCHTTP2Core/Client/GRPCClientStreamHandler.swift
gjcairo: Yeah that's a good idea

glbrntt (approved):
Fab, thanks Gus

PR #1955, opened 2024-07-01
Swift 6-ify the package manifest

glbrntt:
Motivation:

v2 will require the Swift 6 language mode. In order to continue developing v1 and v2 on main we'll need a Swift 6 specific manifest version in addition to the 5.x version.

Moving forward we'll have two package manifests, `Package.swift` for 5.x containing only v1, and `Package@swift-6.swift` using tools version 6.0 targetting v1 and v2.

Modifications:

- Remove v2 targets from Package.swift
- Add a second manifest and make it warning free with Swift 6, this requires turning `static let`s to `static var`s.
- Conditionalise some imports in `protoc-gen-grpc-swift` (as it is shared by v1 and v2)
- Fix an error in the plugin for Swift 6

Result:

v2 now only compiles with a Swift 6 compiler


Package@swift-6.swift
```diff
@@ -0,0 +1,927 @@
...
+
+  static var grpcCore: Target {
+    .target(
+      name: "GRPCCore",
+      dependencies: [
+        .dequeModule,
+        .atomics
+      ],
+      path: "Sources/GRPCCore",
+      swiftSettings: [.swiftLanguageVersion(.v5)]
```
gjcairo, line 196: Do we want to adopt Swift 6 language version for the v2 packages? Or not yet?

Sources/protoc-gen-grpc-swift/main.swift
```diff
@@ -169,6 +172,7 @@ func main(args: [String]) throws {
           generatedFiles: &generatedFiles
         )
         if options.v2 {
+          #if compiler(>=6.0)
```
gjcairo, line 175: Should we `fatalError` (or error in some other way) if the compiler is not Swift 6+ but the v2 option is set?

Package@swift-6.swift
glbrntt: I'm going to make the changes in a separate PR to keep the diff small(er). This way we can do it module by module (or at least do the larger ones separately)

Sources/protoc-gen-grpc-swift/main.swift
glbrntt: Hmm, good question. I think we should actually put the v2 option behind the compiler guard too tbh.

gjcairo (approved):

PR #1958, opened 2024-07-02
Fix warnings in SwiftPM plugin

glbrntt:
Motivation:

SwiftPMs 'Path' was all but deprecated in Swift 6 resulting in numerous warnings for the SwiftPM build plugin. It has largely been replaced by URL.

Modifications:

- Add a wrapper type to the plugin which wraps Path or URL depending on the compiler version.

Result:

No build plugin warnings for Swift 6 compilers

### DIFF
--- a/Plugins/GRPCSwiftPlugin/plugin.swift
+++ b/Plugins/GRPCSwiftPlugin/plugin.swift
@@ -99,43 +99,42 @@ struct GRPCSwiftPlugin {
   ///   - tool: The tool method from the context.
   /// - Returns: The build commands configured based on the arguments.
   func createBuildCommands(
-    pluginWorkDirectory: PackagePlugin.Path,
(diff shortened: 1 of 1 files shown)


Plugins/GRPCSwiftPlugin/plugin.swift
```diff
@@ -261,14 +260,128 @@ extension GRPCSwiftPlugin: BuildToolPlugin {
...
+
+  init(_ value: Value) {
+    self.value = value
+  }
+
+  init(_ path: String) {
+    #if compiler(<6.0)
+    self.value = Path(path)
+    #else
+    self.value = URL(string: path)!
```
gjcairo, line 298: Are we okay with this force-unwrap?

gjcairo (approved):

PR #1960, opened 2024-07-02
Implement initial set-up for the health checking service

clintonpi:
Motivation:

This setup is needed as part of the implementation of the health checking service.

Modifications:

- Pull in updates for the upstream protos and re-generate stubs.
- Update the `Protos/fetch.sh` script to copy over `grpc-proto/grpc/health/v1/health.proto` and generate the health checking stubs.
- Update the `Protos/generate.sh` script to generate the health checking service and messages for v2 and only the server.
- Add a target to the package manifest called `GRPCHealth`.

Result:

The stubs for the health checking service are now available.


Package@swift-6.swift
```diff
@@ -798,6 +798,17 @@ extension Target {
...
+
+  static var grpcHealth: Target {
+    .target(
+      name: "GRPCHealth",
+      dependencies: [
+        .grpcCore,
+        .grpcProtobuf
+      ],
+      path: "Sources/Services/Health"
+    )
```
glbrntt, line 810: Can you add: `swiftSettings: [.swiftLanguageVersion(.v6)]` here? That will enable the Swift 6 language mode for this module.

Package@swift-6.swift
```diff
@@ -912,7 +923,10 @@ let package = Package(
     .grpcHTTP2TransportTests,
     .grpcProtobufTests,
     .grpcProtobufCodeGenTests,
-    .inProcessInteroperabilityTests
+    .inProcessInteroperabilityTests,
+
+    // Health v1
```
glbrntt, line 928: This should come under the big "v2" section above.

Package@swift-6.swift
```diff
@@ -798,6 +798,17 @@ extension Target {
...
   }
+
+  static var grpcHealth: Target {
+    .target(
+      name: "GRPCHealth",
+      dependencies: [
+        .grpcCore,
+        .grpcProtobuf
+      ],
+      path: "Sources/Services/Health"
```
gjcairo, line 809: This will be changed later with the work George is doing, but for now we should probably add the `swiftSettings` param: `swiftSettings: [.swiftLanguageVersion(.v5)]`

glbrntt (changes requested):
Couple of tiny things but looks good otherwise

glbrntt (approved):
Great, thanks @clintonpi!

PR #1965, opened 2024-07-04
Eliminate unnecessary allocations by using single-element writes

gjcairo:
## Motivation
Using the collection-based write was wrapping single elements into a collection and `write(contentsOf:)` was being called. This always results in the element being wrapped in a `Deque` in the `NIOAsyncWriter`. 

## Modifications
This commit moves the write(element:) method to be required for the RPCWriter protocol so that we can call the single element write and avoid this boxing.

## Result
Fewer allocations.

### DIFF
--- a/Sources/GRPCCore/Streaming/RPCWriter+Closable.swift
+++ b/Sources/GRPCCore/Streaming/RPCWriter+Closable.swift
@@ -30,7 +30,7 @@ extension RPCWriter {
 
     /// Writes a single element.
     ///
-    /// This function suspends until the element has been accepted. Implements can use this
+    /// This function suspends until the element has been accepted. Implementers can use this
     /// to exert backpressure on callers.
     ///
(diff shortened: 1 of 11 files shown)


glbrntt:
Tests should be fixed by https://github.com/grpc/grpc-swift/pull/1966

Sources/GRPCCore/Streaming/RPCWriter+Closable.swift
```diff
@@ -28,6 +28,17 @@ extension RPCWriter {
       self.writer = other
     }
 
+    /// Writes a single element.
+    ///
+    /// This function suspends until the element has been accepted. Implements can use this
```
glbrntt, line 33: Here and elsewhere: s/Implements/Implementers (sorry, looks like this is copy+pasta from my own typo 😅)

glbrntt (approved):
Looks good modulo the typos :)

PR #1971, opened 2024-07-08
Reuse inbound request headers storage for server streams

glbrntt:
Motivation:

Each stream starts with headers. When the server accepts a stream it will respond with headers and eventually, once the RPC has been completed, respond with trailers. For each RPC on the server, three sets of headers are allocated.

However, once the response headers are received we drop any ref to them (as they are copied into metadata), rather then dropping it on the floor we can hold onto the headers and reuse their already allocated storage for both the initial and trailing metadata. Assuming there is enough capacity this saves two allocations per RPC.

Modifications:

- Hold on to headers in the stream state machine, clearning and re-populating them when necessary for the initial and trailing resposne headers.
- Add a '_modifying' state.

Result:

Fewer allocations

### DIFF
--- a/Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
+++ b/Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
@@ -66,6 +66,7 @@ private enum GRPCStreamStateMachineState {
   case clientClosedServerIdle(ClientClosedServerIdleState)
   case clientClosedServerOpen(ClientClosedServerOpenState)
   case clientClosedServerClosed(ClientClosedServerClosedState)
+  case _modifying
 
   struct ClientIdleServerIdleState {
     let maximumPayloadSize: Int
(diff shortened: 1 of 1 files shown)


Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
```diff
@@ -1587,35 +1719,43 @@ internal enum GRPCHTTP2Keys: String {
...
 
-  static func trailersOnly(code: Status.Code, message: String, metadata: Metadata = [:]) -> Self {
-    return .trailersOnly(status: Status(code: code, message: message), metadata: metadata)
+  fileprivate static func trailersOnly(code: Status.Code, message: String) -> Self {
+    var trailers = HPACKHeaders()
+    HPACKHeaders.formTrailers(
+      &trailers,
+      isTrailersOnly: true,
+      status: Status(code: code, message: message),
+      metadata: [:]
```
gjcairo, line 1739: Does this allocate or does the compiler optimise it out? Otherwise we could make `metadata` optional? Or do you think that it's not worth optimising the trailers-only path?

gjcairo (commented):
Good one! LGTM overall, just left a question

Sources/GRPCHTTP2Core/GRPCStreamStateMachine.swift
glbrntt: It doesn't allocate. This is true for _most_ collections but isn't a requirement (an empty `CircularBuffer`, for example, allocates).

gjcairo (approved):

PR #1979, opened 2024-07-12
Simplfy top-level streamin handling in server transports

glbrntt:
Modification:

The top-level connection handling code in the two NIO H2 transport is a little hard to follow because the nesting gets quite deep. So much so that it took me some time to notice an unnecessary task group in there.

Modifications:

- Split the top-level server code into a couple of functions
- Remove unnecessary task group

Result:

Easier code to follow

### DIFF
--- a/Sources/GRPCHTTP2TransportNIOPosix/HTTP2ServerTransport+Posix.swift
+++ b/Sources/GRPCHTTP2TransportNIOPosix/HTTP2ServerTransport+Posix.swift
@@ -18,6 +18,7 @@ import GRPCCore
 import GRPCHTTP2Core
 import NIOCore
 import NIOExtras
+import NIOHTTP2
 import NIOPosix
 
 extension HTTP2ServerTransport {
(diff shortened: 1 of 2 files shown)


Sources/GRPCHTTP2TransportNIOPosix/HTTP2ServerTransport+Posix.swift
```diff
@@ -211,64 +212,79 @@ extension HTTP2ServerTransport {
...
-                      do {
-                        for try await _ in connectionInbound {}
-                      } catch {
-                        // We don't want to close the channel if one connection throws.
-                        return
-                      }
-                    }
-
-                    connectionTaskGroup.addTask {
-                      await withDiscardingTaskGroup { streamTaskGroup in
```
glbrntt, line 230: FWIW this is the unnecessary task group: it only ever has one task so we can just use the connection task group

gjcairo (approved):

PR #1980, opened 2024-07-15
Propagate generics from AsyncSequence to RPCAsyncSequence

glbrntt:
Motivation:

`RPCAsyncSequence` was added before we raised the minimum 
swift version to 6.0. As a result it used a fully erased sequence where
the iterator had to upcast from `Any` to the expected type. This 
causes an allocation per message.

Modification:

- Make `RPCAsyncSequence` generic over its `Element` and its `Failure` type
- This required bumping its availability to the latest SDKs which is 
  rather viral.
- Modify the code generator to allow visionOS to be specified
- Modify the code gen translator to bump the availability of generated code
- Regenerate

Result:

Fewer allocations


glbrntt:
This PR is broken into three commits:
- [8692408](https://github.com/grpc/grpc-swift/pull/1980/commits/86924081522b43106942099b74a1d37a3ddff446) Updates the code generator
- [74b5295](https://github.com/grpc/grpc-swift/pull/1980/commits/74b5295328002703c51a137576e20fa903838103) Regenerates the code
- [1f19154](https://github.com/grpc/grpc-swift/pull/1980/commits/1f191549053773f09e139c3e732d75f4f96f062a) makes the actual changes to `RPCAsyncSequence` and deals with the rest of the fallout from bumping availability requirements

gjcairo (approved):

PR #1982, opened 2024-07-16
Add `ServiceDescriptor` to GRPCCore

clintonpi:
Motivation:
`ServiceDescriptor` describes a service. It is needed for services that interact with the Health service.

Modifications:
- Add a `public struct ServiceDescriptor` to GRPCCore.

Result:
Services will be describable using `ServiceDescriptor`.

### DIFF
--- a/Sources/GRPCCore/ServiceDescriptor.swift
+++ b/Sources/GRPCCore/ServiceDescriptor.swift
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2024, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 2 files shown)


Sources/GRPCCore/ServiceDescriptor.swift
```diff
@@ -0,0 +1,45 @@
...
+
+  /// The fully qualified service name in the format:
+  /// - "package.service": if a package name is specified. For example, "helloworld.Greeter".
+  /// - "service": if a package name is not specified. For example, "Greeter".
+  public var fullyQualifiedService: String {
+    if self.package.count > 0 {
+      return "\(self.package).\(self.service)"
+    }
+
+    return self.service
```
glbrntt, line 34: nit: prefer `isEmpty`:

```swift
if self.package.isEmpty {
  return self.service
} else {
  return "\(self.package).\(self.service)"
}
```

glbrntt (changes requested):
one nit, looks good otherwise

glbrntt (approved):

PR #1983, opened 2024-07-16
Save two allocations by making `MapRPCWriter` and `SerializingRPCWriter` generic over wrapped writer

gjcairo:
## Motivation
We currently wrap a base writer into an `RPCWriter` inside the `MapRPCWriter` and `SerializingRPCWriter`.
This results in an unnecessary allocation in both cases. In the case of unary requests, this means two extra allocations per requests on the client side.

## Modifications
Make the `MapRPCWriter` and `SerializingRPCWriter` generic over the base writer, to avoid the extra allocations caused by wrapping them in an `RPCWriter`.

## Result
Two fewer allocations per unary request on the client side.


glbrntt (approved):
Nice one

PR #1985, opened 2024-07-22
Enable Swift 6 language mode for more modules

glbrntt:
Motivation:

v2 will support Swift 6 only. We should enable the Swift 6 language mode
across all v2 modules. This has mostly been done already for non-test 
modules. This change enables it for test modules as well.

Modifications:

- Enable Swift 6 language mode and existential any for test modules

Result:

Closer to fully supporting Swift 6 language mode


Tests/GRPCCodeGenTests/Internal/Renderer/TextBasedRendererTests.swift
```diff
@@ -41,7 +41,7 @@ final class Test_TextBasedRenderer: XCTestCase {
         Also, bar
         """#
       ),
-      renderedBy: TextBasedRenderer.renderComment,
```
gjcairo, line 44: Curious: what warning were we getting here?

Sources/GRPCCore/Streaming/RPCAsyncSequence.swift
```diff
@@ -16,7 +16,10 @@
 
 /// A type-erasing `AsyncSequence`.
 @available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *)
-public struct RPCAsyncSequence<Element, Failure: Error>: AsyncSequence, @unchecked Sendable {
+public struct RPCAsyncSequence<
+  Element: Sendable,
+  Failure: Error
+>: AsyncSequence, @unchecked Sendable {
   // @unchecked Sendable is required because 'any' doesn't support composition with primary
   // associated types. (see: https://github.com/swiftlang/swift/issue/63877)
```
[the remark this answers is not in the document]
glbrntt: Unfortunately not as we then run head first into https://github.com/swiftlang/swift/issues/75242

Tests/GRPCCodeGenTests/Internal/Renderer/TextBasedRendererTests.swift
glbrntt: Implicit capture of self which requires `TextBasedRendered` to be `Sendable` (which it isn't)

Sources/GRPCCore/Streaming/RPCAsyncSequence.swift
glbrntt, line 24:
```suggestion
  // associated types. (see: https://github.com/swiftlang/swift/issues/63877)
```

gjcairo (approved):

PR #1986, opened 2024-07-22
Avoid allocating a closure for the client rpc executor

glbrntt:
Motivation:

The client RPC executor unconditionally allocates a closure when intercepting a request. This can be avoided if there are no interceptors.

Modifications:

- Special case the no-interceptors path

Result:

Fewer allocations


gjcairo (approved):
Nice one!

PR #1987, opened 2024-07-23
Implement Health service

clintonpi:
Motivation:

Implement the Health service defined in `Protos/upstream/grpc/health/v1/health.proto`.

Modifications:

- Implement the "check" and "watch" methods for the service.
- Implement providers (public APIs) to interact with the service.
- Add tests.

Result:

The Health service will be available for use in grpc-swift.


Tests/Services/HealthTests/HealthTests.swift
```diff
@@ -0,0 +1,236 @@
...
+
+  func testWatchOnKnownService() async throws {
+    try await withHealthClient { (healthClient, healthProvider) in
+      let testServiceDescriptor = ServiceDescriptor(package: "test.package", service: "TestService")
+      let ignoredServiceDescriptor = ServiceDescriptor(
+        package: "package.to.be.ignored",
+        service: "IgnoredService"
+      )
+
+      let statuses: [ServingStatus] = [.serving, .notServing, .serving, .serving, .notServing]
```
glbrntt, line 102: We shouldn't return two of the same statuses in a row

Tests/Services/HealthTests/HealthTests.swift
```diff
@@ -0,0 +1,236 @@
...
+      )
+
+      var message = Grpc_Health_V1_HealthCheckRequest()
+      message.service = testServiceDescriptor.fullyQualifiedService
+
+      try await healthClient.watch(request: ClientRequest.Single(message: message)) { response in
+        var responseStreamIterator = response.messages.makeAsyncIterator()
+
+        for i in 0 ..< statuses.count {
+          let next = try await responseStreamIterator.next()!
```
glbrntt, line 121: Avoid '!'-ing here. There's plenty of ways that this _could_ go wrong and we'd want to avoid a crash if it did.

You can use `let message = try XCTUnwrap(next)` to unwrap the optional or throw an error (and fail the test) if it's nil.

Tests/Services/HealthTests/HealthTests.swift
```diff
@@ -0,0 +1,236 @@
...
+      }
+    }
+  }
+
+  func testWatchOnUnknownService() async throws {
+    try await withHealthClient { (healthClient, healthProvider) in
+      try healthProvider.updateService(
+        descriptor: ServiceDescriptor(package: "package.to.be.ignored", service: "IgnoredService"),
+        status: .serving
+      )
```
glbrntt, line 147: No point doing this, we ignore it

Tests/Services/HealthTests/HealthTests.swift
```diff
@@ -0,0 +1,236 @@
...
+
+        try healthProvider.updateService(
+          descriptor: testServiceDescriptor,
+          status: .notServing
+        )
+
+        next = try await responseStreamIterator.next()!
+
+        XCTAssertEqual(next.status, .notServing)
+      }
```
glbrntt, line 168: I don't understand what we're testing here: it's not an "Unknown" service because we updated the status of it to `.notServing`. If anything we should check that the first and only status is `.serviceUnknown`.

Tests/Services/HealthTests/HealthTests.swift
```diff
@@ -0,0 +1,236 @@
...
+          var responseStreamIterator = response.messages.makeAsyncIterator()
+
+          // Since responseStreamIterator.next() will never be nil (as the "watch" response stream
+          // is always open), the iteration cannot be based on when responseStreamIterator.next()
+          // is nil. Else, the iteration infinitely awaits and the test never finishes. Hence, it is
+          // based on the expected number of statuses to be received.
+          for _ in 0 ..< statusesToBeSent.count {
+            try await continuation.yield(responseStreamIterator.next()!.status)
+          }
+        }
```
glbrntt, line 205: We don't need a continuation here, we can just get the first N elements from the response stream and collect them into an array and then return that from the RPC:

```swift
try await response.messages.prefix(N).reduce(into: []) { $0.append($1) }
```

Breaking that down:

```swift
// Get an async sequence which only returns the first N messages from `response.messages`
let prefix = response.messages.prefix(N) 

// Wait for the prefix sequence to end (i.e. collect all N elements) and store them in an array:
let reduced = prefix.reduce(into: []) { elements, nextElement in 
  elements.append(nextElement)
}
```

glbrntt: We can then reduce this test to:

```swift
try await withThrowingTaskGroup(of: [Status].self) { group in
  let concurrentStreams = 2
  for _ in 0 ..< concurrentStreams {
    group.addTask {
      try await watch()
    }
  }

  for status in statuses {
    // update status
  }

  for try await next in group {
    XCTAssertEqual(next, expected)
  }
}
```

Tests/Services/HealthTests/HealthTests.swift
```diff
@@ -0,0 +1,236 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import GRPCHealth
+import GRPCInProcessTransport
+import XCTest
+
+@available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *)
+final class HealthTests: XCTestCase {
```
glbrntt, line 22: We should also have tests that:

- Run check with no service name (to get the status of the overall server)
- Run watch with no service name (to get the status of the overall server)
- Run watch and update the service status to the same value multiple times, the client should only receive one update

Tests/Services/HealthTests/HealthTests.swift
```diff
@@ -0,0 +1,236 @@
...
+        XCTFail("Unexpected error: \(error)")
+      }
+
+      group.cancelAll()
+    }
+  }
+
+  func testCheckOnKnownService() async throws {
+    try await withHealthClient { (healthClient, healthProvider) in
+      let testServiceDescriptor = ServiceDescriptor(package: "test.package", service: "TestService")
```
glbrntt, line 53: Rather than spelling out the descriptor every time we can add an extension within this file to `ServiceDescriptor`, e.g.:

```swift
extension ServiceDescriptor {
  fileprivate static let echo = ServiceDescriptor(package: "echo", service: "Echo")
}
```

Sources/Services/Health/HealthProvider.swift
```diff
@@ -0,0 +1,55 @@
...
+    self.healthService = healthService
+  }
+}
+
+extension Grpc_Health_V1_HealthCheckResponse.ServingStatus {
+  /// Constructs a new ``Grpc_Health_V1_HealthCheckResponse.ServingStatus`` from ``ServingStatus``.
+  ///
+  /// - Parameters:
+  ///   - from: The base status.
+  package init(from status: ServingStatus) {
```
glbrntt, line 49: This is a 'value preserving type conversion' so should be:

```swift
  package init(_ status: ServingStatus) {
```

see: https://www.swift.org/documentation/api-design-guidelines/#type-conversion

Sources/Services/Health/HealthProvider.swift
```diff
@@ -0,0 +1,55 @@
...
+    switch status.value {
+    case .serving: self = .serving
+    case .notServing: self = .notServing
+    }
```
glbrntt, line 53: prefer each line doing one thing only:

```suggestion
    switch status.value {
    case .serving: 
      self = .serving
    case .notServing: 
      self = .notServing
    }
```

glbrntt (changes requested):

Tests/Services/HealthTests/HealthTests.swift
```diff
@@ -0,0 +1,322 @@
...
+    }
+  }
+
+  func testWatchOnKnownService() async throws {
+    try await withHealthClient { (healthClient, healthProvider) in
+      let testServiceDescriptor = ServiceDescriptor.testService
+
+      let statusesToBeSent: [ServingStatus] = [.serving, .notServing, .serving]
+
+      healthProvider.updateStatus(statusesToBeSent[0], ofService: testServiceDescriptor)
```
glbrntt, line 99: Can you add a comment as to why we send this status before starting the RPC because that's quite important.

Tests/Services/HealthTests/HealthTests.swift
```diff
@@ -0,0 +1,322 @@
...
+
+      let statusesToBeSent: [ServingStatus] = [.serving, .notServing, .serving]
+
+      try await withThrowingTaskGroup(
+        of: [Grpc_Health_V1_HealthCheckResponse.ServingStatus].self
+      ) { group in
+        var message = Grpc_Health_V1_HealthCheckRequest()
+        message.service = testServiceDescriptor.fullyQualifiedService
+
+        let immutableMessage = message
```
glbrntt, line 159: You can just do:


```swift
let message = Grpc_Health_V1_HealthCheckRequest.with { 
  $0.service = ...
}
```

Tests/Services/HealthTests/HealthTests.swift
```diff
@@ -0,0 +1,322 @@
...
+      let expectedReceivedStatuses: [Grpc_Health_V1_HealthCheckResponse.ServingStatus] = [
+        // As the service will be "watched" before being updated, the first status received should
+        // be .serviceUnknown.
+        .serviceUnknown,
+        .serving,
+        .notServing,  // The repeated `.notServing` updates should be received only once.
+        .serving,
+      ]
+
+      try await withThrowingTaskGroup(
```
glbrntt, line 234: I don't think this test needs a task group, we should be able to do everything we need within `watch`:

```swift
healthClient.watch(...) { response in 
  var iterator = response.messages.makeAsyncIterator()
  // wait for 'serviceUnknown'
  // send status updates
  // read updates
}
```

glbrntt (changes requested):

Sources/Services/Health/Health.swift
```diff
@@ -16,62 +16,71 @@
 
 import GRPCCore
 
-/// A coupled Health service and provider.
+/// ``Health`` is gRPC’s mechanism for checking whether a server is able to handle RPCs. Its semantics are documented in
+/// https://github.com/grpc/grpc/blob/master/doc/health-checking.md.
+///
```
glbrntt, line 21: We should go into more detail here, as an example:


```
/// Each ``Health`` has two objects, a ``Health/Service-swift.struct`` and a
/// ``Health/Provider-swift.struct``. The `service` implements the `Health` service from the
/// `grpc.health.v1` package. It can be registered with your server like any other service.
///
/// The ``Health/Service-swift.struct`` service doesn't know about other services running on the
/// server so you must provide it with information about the status of services via
/// the ``Health/Provider-swift.struct``.
///
/// To make this easier, generated code for services includes an extension to `ServiceDescriptor`
/// so you can easily specify which service is being updated. The following shows an example
/// of creating a health service and updating the state of the 'Foo' service in the 'bar' package.
///
/// ```
/// let health = Health()
/// let server = GRPCServer(transport: transport, services: [health, FooService()])
/// health.provider.updateStatus(.service, forService: .bar_Foo)
/// ```
```

Sources/Services/Health/Health.swift
```diff
@@ -16,62 +16,71 @@
...
+/// `Health` initializes a new `Health.Service` and a `Health.Provider`.
+/// - `Health.Service` is a registerable RPC service to probe whether a server is able to handle RPCs.
+/// - `Health.Provider` provides handlers to interact with `Health.Service`.
 @available(macOS 15.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
 public struct Health: Sendable {
-  private let internalHealthService = InternalHealthService()
-
   /// A registerable RPC service to probe whether a server is able to handle RPCs.
```
glbrntt, line 27: /// An implementation of the 'grpc.health.v1.Health' service.

Sources/Services/Health/Health.swift
```diff
@@ -16,62 +16,71 @@
...
   public let service: Health.Service
 
   /// Provides handlers to interact with the coupled Health service.
```
glbrntt, line 30: /// Provides status updates to the health service.

Sources/Services/Health/HealthService.swift
```diff
@@ -59,33 +59,31 @@ internal final class InternalHealthService: Grpc_Health_V1_HealthServiceProtocol
...
   ) {
-    self.state.updateStatus(status, ofService: service)
+    self.state.updateStatus(status, forService: service)
   }
 }
 
 @available(macOS 15.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
-extension InternalHealthService {
+extension HealthService {
+  /// The state of the Health service.
```
glbrntt, line 70: No need to document this

Sources/Services/Health/HealthService.swift
```diff
@@ -125,6 +124,7 @@ extension InternalHealthService {
       _ continuation: AsyncStream<Grpc_Health_V1_HealthCheckResponse.ServingStatus>.Continuation
     ) {
       self.continuations.append(continuation)
+      continuation.yield(self.currentStatus)
     }
 
     fileprivate init(status: Grpc_Health_V1_HealthCheckResponse.ServingStatus) {
```
glbrntt, line 130: It probably makes sense to default this to "unknown"

glbrntt (changes requested):

Sources/Services/Health/Health.swift
```diff
@@ -0,0 +1,110 @@
...
+
+    /// Updates the status of a service.
+    ///
+    /// - Parameters:
+    ///   - status: The status of the service.
+    ///   - service: The description of the service.
+    public func updateStatus(
+      _ status: ServingStatus,
+      forService service: ServiceDescriptor
+    ) {
```
glbrntt, line 88: We should probably provide an override where a `String` can be used instead of a `ServiceDescriptor`.

Sources/Services/Health/HealthService.swift
```diff
@@ -0,0 +1,132 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import GRPCCore
+
+@available(macOS 15.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+internal final class HealthService: Grpc_Health_V1_HealthServiceProtocol {
```
glbrntt, line 20: Should be possible for this to be a `struct`

Sources/Services/Health/ServingStatus.swift
```diff
@@ -0,0 +1,38 @@
...
+    case notServing
+  }
+
+  /// A status indicating that a service is healthy.
+  public static let serving = ServingStatus(.serving)
+
+  /// A status indicating that a service unhealthy.
+  public static let notServing = ServingStatus(.notServing)
+
+  package var value: Value
```
gjcairo, line 33: Do this prop and the `Value` enum really have to be `package`? Can't they be `private`?

Sources/Services/Health/HealthService.swift
```diff
@@ -0,0 +1,132 @@
...
+
+    var response = Grpc_Health_V1_HealthCheckResponse()
+    response.status = status
+
+    return ServerResponse.Single(message: response)
+  }
+
+  func watch(
+    request: ServerRequest.Single<Grpc_Health_V1_HealthCheckRequest>
+  ) async throws -> ServerResponse.Stream<Grpc_Health_V1_HealthCheckResponse> {
```
gjcairo, line 40: Nit: this func doesn't have to be throwing

Sources/Services/Health/HealthService.swift
```diff
@@ -0,0 +1,132 @@
...
+internal struct HealthService: Grpc_Health_V1_HealthServiceProtocol {
+  private let state = HealthService.State()
+
+  func check(
+    request: ServerRequest.Single<Grpc_Health_V1_HealthCheckRequest>
+  ) async throws -> ServerResponse.Single<Grpc_Health_V1_HealthCheckResponse> {
+    let service = request.message.service
+
+    guard let status = self.state.currentStatus(ofService: service) else {
+      throw RPCError(code: .notFound, message: "Requested service unknown.")
```
gjcairo, line 29: For consistency with `watch` below, should we throw here or return a `serviceUnknown` state?

Sources/Services/Health/ServingStatus.swift
clintonpi: Oh yes, they don't have to be `package`. But they should be `internal`. The `value` property is accessed [in a different file](https://github.com/grpc/grpc-swift/pull/1987/files#diff-f379eb3d9b44ebf8b98c37c7f9af8ca6499e7500c83fc07c56c61ed4023cab09R120).

Sources/Services/Health/HealthService.swift
clintonpi: According to [the specification](https://github.com/grpc/grpc-swift/blob/main/Protos/upstream/grpc/health/v1/health.proto#L46) for `Check`, "If the requested service is unknown, the call will fail with status NOT_FOUND".

gjcairo (approved):

Package@swift-6.swift
```diff
@@ -510,6 +511,19 @@ extension Target {
...
+  static var grpcHealthTests: Target {
+    .testTarget(
+      name: "GRPCHealthTests",
+      dependencies: [
+        .grpcHealth,
+        .grpcProtobuf,
+        .grpcInProcessTransport
+      ],
+      path: "Tests/Services/HealthTests",
+      swiftSettings: [._swiftLanguageMode(.v6), .enableUpcomingFeature("ExistentialAny")]
```
glbrntt, line 523: Can you update the health module and health tests to use explicit imports like in Gus' PR https://github.com/grpc/grpc-swift/pull/2003?

I'm happy for this to be done in a separate PR as well (actually it probably needs to be done separately because Gus needs to do the imports on the code-gen too).

glbrntt (approved):

PR #1992, opened 2024-07-24
Change visibility of `_makeBackpressuredStream` to `package`

gjcairo:
We were using `public` as the access modifier because we started not supporting Swift 6 only. We can now change the visibility to be `package`.

### DIFF
--- a/Sources/GRPCCore/Streaming/Internal/RPCAsyncSequence+Buffered.swift
+++ b/Sources/GRPCCore/Streaming/Internal/RPCAsyncSequence+Buffered.swift
@@ -30,7 +30,7 @@ extension RPCAsyncSequence where Failure == any Error {
   }
 
   @inlinable
-  public static func _makeBackpressuredStream(
+  package static func _makeBackpressuredStream(
     of elementType: Element.Type = Element.self,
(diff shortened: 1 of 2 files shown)


Sources/GRPCCore/Streaming/Internal/RPCAsyncSequence+Buffered.swift
```diff
@@ -30,7 +30,7 @@ extension RPCAsyncSequence where Failure == any Error {
   }
 
   @inlinable
-  public static func _makeBackpressuredStream(
+  package static func _makeBackpressuredStream(
```
glbrntt, line 33: We can remove the underscore too

glbrntt (changes requested):

Sources/GRPCCore/Streaming/Internal/RPCAsyncSequence+Buffered.swift
gjcairo: Oops, yeah we can actually get rid of the whole definition here :D

glbrntt (approved):

PR #1993, opened 2024-07-25
Simplify client rpc execution

glbrntt:
Motivation:

The existing code for executing RPCs on the client is layered, the one-shot, retry and hedging executors serialize the messages of the requests they're processing into bytes. These 'raw' requests are processed by the stream executor which is structured in such a way as requiring a task group of events it needs to handled. These events are handled in a run method (which is in turn another task for the caller to run). This requires quite a bit of machinery (and allocations) for it to work.

Modifications:

- Push the serialization/deserialization of messages down into the stream executor. This lets a typed request be transformed directly to `RPCRequestPart`s and for `RPCResponsePart`s to be transformed directly to a typed response (rather than via an intermediary request/response typed to `[UInt8]`).
- Remove the sendability requirement from the `next` function for client interceptors. This makes sense: interceptors should be straight-line code so shouldn't be shuffled off into a subtask. This unlocks the ability to remove the task group and event stream previously used by the stream executor as it can add child tasks to a provided task group (rather than needing a separate task group with an event stream).
- Apply this change to the one-shot, hedging, and retry executors.

Result:

- ~35% reduction in allocations for unary RPCs on the client

### DIFF
--- a/Sources/GRPCCore/Call/Client/ClientInterceptor.swift
+++ b/Sources/GRPCCore/Call/Client/ClientInterceptor.swift
@@ -102,7 +102,7 @@ public protocol ClientInterceptor: Sendable {
   func intercept<Input: Sendable, Output: Sendable>(
     request: ClientRequest.Stream<Input>,
     context: ClientInterceptorContext,
-    next: @Sendable (
+    next: (
       _ request: ClientRequest.Stream<Input>,
       _ context: ClientInterceptorContext
     ) async throws -> ClientResponse.Stream<Output>
--- a/Sources/GRPCCore/Call/Client/Internal/ClientStreamExecutor.swift
+++ b/Sources/GRPCCore/Call/Client/Internal/ClientStreamExecutor.swift
@@ -16,59 +16,7 @@
 
 @available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *)
 @usableFromInline
-internal struct ClientStreamExecutor<Transport: ClientTransport> {
-  /// The client transport to execute the stream on.
-  @usableFromInline
-  let _transport: Transport
-
-  /// An `AsyncStream` and continuation to send and receive processing events on.
-  @usableFromInline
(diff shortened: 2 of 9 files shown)


Sources/GRPCCore/Call/Client/ClientInterceptor.swift
```diff
@@ -102,7 +102,7 @@ public protocol ClientInterceptor: Sendable {
   func intercept<Input: Sendable, Output: Sendable>(
     request: ClientRequest.Stream<Input>,
     context: ClientInterceptorContext,
-    next: @Sendable (
+    next: (
```
FranzBusch, line 105: you might want to adopt `sending` here instead. So that you can call an interceptor on a child task if one wants to

glbrntt: Yeah I was thinking about that. I was going to go through this module separately and adopt `sending` since we're still using v5 language mode and I think we need it in a few places.

Sources/GRPCCore/Call/Client/Internal/ClientStreamExecutor.swift
```diff
@@ -77,37 +25,52 @@ internal struct ClientStreamExecutor<Transport: ClientTransport> {
...
-    stream: RPCStream<Transport.Inbound, Transport.Outbound>
-  ) async -> ClientResponse.Stream<[UInt8]> {
-    // Each execution method can add work to process in the 'run' method. They must not add
-    // new work once they return.
-    defer { self._work.continuation.finish() }
-
-    // Start processing the request.
-    self._work.continuation.yield(.request(request, stream.outbound))
+  @inlinable  // would be private
+  static func execute<Input: Sendable, Output: Sendable>(
```
gjcairo, line 29: Nit: do we want to rename to `_execute`?

glbrntt: Ah thanks. We actually don't want it to be `_execute`, it should be `internal`, it's the comment that's wrong.

gjcairo (approved):

PR #2000, opened 2024-07-31
Clamp HTTP2 config values in configureGRPCServerPipeline

gjcairo:
We currently clamp a couple of H2 settings when setting up a v2 gRPC client to keep them within the RFC-specified range - we should do the same for the server.

### DIFF
--- a/Sources/GRPCHTTP2Core/Internal/NIOChannelPipeline+GRPC.swift
+++ b/Sources/GRPCHTTP2Core/Internal/NIOChannelPipeline+GRPC.swift
@@ -51,10 +51,23 @@ extension ChannelPipeline.SynchronousOperations {
     )
     try self.addHandler(flushNotificationHandler)
 
+    // Window size which mustn't exceed 2^31 - 1 (RFC 9113 § 6.5.2).
+    let clampedTargetWindowSize = min(http2Config.targetWindowSize, (1 << 31) - 1)
+
+    // Max frame size must be in the range 2^14 ..< 2^24 (RFC 9113 § 4.2).
+    let clampedMaxFrameSize: Int
+    if http2Config.maxFrameSize >= (1 << 24) {
+      clampedMaxFrameSize = (1 << 24) - 1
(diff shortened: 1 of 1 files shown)


Sources/GRPCHTTP2Core/Internal/NIOChannelPipeline+GRPC.swift
```diff
@@ -51,10 +51,23 @@ extension ChannelPipeline.SynchronousOperations {
...
+
+    // Max frame size must be in the range 2^14 ..< 2^24 (RFC 9113 § 4.2).
+    let clampedMaxFrameSize: Int
+    if http2Config.maxFrameSize >= (1 << 24) {
+      clampedMaxFrameSize = (1 << 24) - 1
+    } else if http2Config.maxFrameSize < (1 << 14) {
+      clampedMaxFrameSize = (1 << 14)
+    } else {
+      clampedMaxFrameSize = http2Config.maxFrameSize
+    }
```
glbrntt, line 65: Presumably this is all repeated in the client? Might be worth extracting into a couple of internal `validateHTTP2<Thing>` functions

Sources/GRPCHTTP2Core/Internal/NIOChannelPipeline+GRPC.swift
```diff
@@ -181,3 +161,23 @@ extension ChannelPipeline.SynchronousOperations {
     return (connection, multiplexer)
   }
 }
+
+extension ChannelPipeline.SynchronousOperations {
+  /// Max frame size must be in the range `2^14 ..< 2^24` (RFC 9113 § 4.2).
+  internal func clampMaxFrameSize(_ maxFrameSize: Int) -> Int {
```
glbrntt, line 167:
```suggestion
  fileprivate func clampMaxFrameSize(_ maxFrameSize: Int) -> Int {
```

glbrntt (approved):
One nit but looks good otherwise

PR #2001, opened 2024-08-01
Fix tests to work with unreleased swift-nio changes

czechboy0:
## Summary

Incoming swift-nio changes slightly changed the semantics of testing event loops, making some tests fail.

This PR explicitly shuts down the ConnectionManager to force correct ordering in tests.

## Test Plan

All tests pass locally with both the latest released and "main" swift-nio.


glbrntt (approved):
Thank you!

PR #2003, opened 2024-08-02
Enable `InternalImportsByDefault` in v2 modules

gjcairo:
Starting with Swift 5.10, `InternalImportsByDefault` is an optional feature flag to make all `import` statements be `internal` by default, instead of `public`, which has been the default since the beginning in Swift.
With a future language mode, this will become the new default, and all imports will be `internal` unless otherwise specified. The main reason for this change is to minimise dependency creep.

This PR enables the feature flag on v2 modules, and adds `public`/`package` access modifiers to `import`s where required.

Note that sadly, `@usableFromInline` has not been implemented for `import` statements, so in `@inlinable`/`@usableFromInline` contexts where non-publicly-imported types are used, the only workaround for now is to import the module as `public`. I've left a comment next to these imports.

### DIFF
--- a/Package@swift-6.swift
+++ b/Package@swift-6.swift
@@ -207,7 +207,11 @@ extension Target {
         .atomics
       ],
       path: "Sources/GRPCCore",
-      swiftSettings: [._swiftLanguageMode(.v5), .enableUpcomingFeature("ExistentialAny")]
+      swiftSettings: [
+        ._swiftLanguageMode(.v5),
+        .enableUpcomingFeature("ExistentialAny"),
+        .enableUpcomingFeature("InternalImportsByDefault")
+      ]
     )
   }
 
@@ -217,7 +221,11 @@ extension Target {
       dependencies: [
         .grpcCore
       ],
-      swiftSettings: [._swiftLanguageMode(.v6), .enableUpcomingFeature("ExistentialAny")]
+      swiftSettings: [
+        ._swiftLanguageMode(.v6),
+        .enableUpcomingFeature("ExistentialAny"),
+        .enableUpcomingFeature("InternalImportsByDefault")
+      ]
     )
   }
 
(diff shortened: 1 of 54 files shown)


Package@swift-6.swift
```diff
@@ -510,7 +538,7 @@ extension Target {
         .grpcCore,
         .grpcProtobuf
       ],
-      swiftSettings: [._swiftLanguageMode(.v5), .enableUpcomingFeature("ExistentialAny")]
+      swiftSettings: [._swiftLanguageMode(.v6), .enableUpcomingFeature("ExistentialAny")]
```
glbrntt, line 541: This one doesn't have `"InternalImportsByDefault"`

Package@swift-6.swift
```diff
@@ -207,7 +207,11 @@ extension Target {
         .atomics
       ],
       path: "Sources/GRPCCore",
-      swiftSettings: [._swiftLanguageMode(.v5), .enableUpcomingFeature("ExistentialAny")]
+      swiftSettings: [
+        ._swiftLanguageMode(.v5),
+        .enableUpcomingFeature("ExistentialAny"),
+        .enableUpcomingFeature("InternalImportsByDefault")
```
glbrntt, line 213: If we have to enable this, then presumably `public` is the default for Swift 6?

Sources/GRPCHTTP2Core/Client/Connection/ConnectionFactory.swift
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-import NIOCore
-import NIOHTTP2
+package import NIOCore
+package import NIOHTTP2
 import NIOPosix
```
glbrntt, line 19: Sorry to be a pain, but can we have explicit imports for everything? (Hopefully `sed` can make this an easier fix...!)

Package@swift-6.swift
gjcairo: Yes: the default will become `internal` in a future language mode ([see proposal](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md#default-import-access-level)).

Package@swift-6.swift
gjcairo: Yes, it's because there are a bunch of generated files in this module. I will enable it for this one in a follow up PR after I've changed the code generator.

glbrntt: Cool, thanks for clarifying!

glbrntt (approved):
LGTM, thanks Gus!

PR #2004, opened 2024-08-05
Update generated code to use service descriptors

clintonpi:
Motivation:

The generated code refers to services literally using strings. With the addition of service descriptors, the generated code can now utilize them to refer to services.

Modifications:

- Update the generated code to use service descriptors.
- For each service in the generated code, extend `ServiceDescriptor` to include a `static` property that holds the service descriptor for that service. (I.e. for a service `Foo` in the `bar` package, the service descriptor for `Foo` can be accessed using `ServiceDescriptor.bar_Foo`.)

Result:

The generated code will utilize and provide easy access to service descriptors.

### DIFF
--- a/Sources/Examples/v2/Echo/Generated/echo.grpc.swift
+++ b/Sources/Examples/v2/Echo/Generated/echo.grpc.swift
@@ -25,7 +25,7 @@ import GRPCCore
 import GRPCProtobuf
 
 internal enum Echo_Echo {
-    internal static let serviceDescriptor = ServiceDescriptor(
+    internal static let descriptor = ServiceDescriptor(
         package: "echo",
         service: "Echo"
     )
@@ -34,7 +34,7 @@ internal enum Echo_Echo {
             internal typealias Input = Echo_EchoRequest
             internal typealias Output = Echo_EchoResponse
             internal static let descriptor = MethodDescriptor(
-                service: Echo_Echo.serviceDescriptor.fullyQualifiedService,
+                service: Echo_Echo.descriptor.fullyQualifiedService,
                 method: "Get"
             )
         }
@@ -42,7 +42,7 @@ internal enum Echo_Echo {
             internal typealias Input = Echo_EchoRequest
             internal typealias Output = Echo_EchoResponse
             internal static let descriptor = MethodDescriptor(
-                service: Echo_Echo.serviceDescriptor.fullyQualifiedService,
+                service: Echo_Echo.descriptor.fullyQualifiedService,
                 method: "Expand"
             )
         }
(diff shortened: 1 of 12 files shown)


Sources/Examples/v2/Echo/Generated/echo.grpc.swift
```diff
@@ -25,36 +25,40 @@ import GRPCCore
 import GRPCProtobuf
 
 internal enum Echo_Echo {
+    internal static let serviceDescriptor = ServiceDescriptor(
```
glbrntt, line 28: can we just call this `descriptor`?

Sources/GRPCCodeGen/Internal/Translator/TypealiasTranslator.swift
```diff
@@ -111,6 +128,32 @@ extension TypealiasTranslator {
...
+      )
+    )
+
+    let staticServiceDescriptorProperty = VariableDescription(
+      accessModifier: self.accessModifier,
+      isStatic: true,
+      kind: .let,
+      left: .identifierPattern("serviceDescriptor"),
+      right: serviceDescriptorInitialization
+    )
```
glbrntt, line 154: Can we extract this into a helper? Otherwise these code-gen methods get unwieldy.

Sources/GRPCCodeGen/Internal/Translator/TypealiasTranslator.swift
```diff
@@ -270,4 +323,36 @@ extension TypealiasTranslator {
...
     )
   }
+
+  private func makeServiceDescriptorExtension(
+    for service: CodeGenerationRequest.ServiceDescriptor
+  ) -> Declaration {
+    let serviceIdentifier =
+      service.namespace.generatedLowerCase.isEmpty
+      ? service.name.generatedLowerCase
+      : "\(service.namespace.generatedLowerCase)_\(service.name.generatedUpperCase)"
```
glbrntt, line 333: Use the ternary operator if the condition and branches are short (e.g. `let foo = x > 10 ? bar : baz`) otherwise they're just less readable. 

If it can't fit on one line then just use an `if`/`else` – it will be easier to read (this also _doesn't_ mean that if it fits on one line then it's always acceptable to use the ternary operator).

glbrntt: In this instance you probably want to split the identifier into two parts: a prefix and the service name:

```swift
let prefix = service.namespace.isEmpty ? "" : service.namespace + "_"
let serviceIdentifier = prefix + service.name.generatedUpperCase
```

glbrntt (changes requested):

Sources/GRPCCodeGen/Internal/Translator/TypealiasTranslator.swift
clintonpi: True!

Sources/Examples/v2/Echo/Generated/echo.grpc.swift
glbrntt: This didn't get done, looks like you just need to regenerate the code

glbrntt (changes requested):

glbrntt (commented):
One nit, but looks good otherwise.

glbrntt (approved):

PR #2007, opened 2024-08-08
Generate default response handlers

clintonpi:
Motivation:

The client stub for all RPCs accepts a response handler and can return the result of that response. For unary and client-streaming RPCs, the most common handler will just return the response message. This handler should be generated as the default for those RPC types.

Modifications:

- Default the response handler parameter (`body`) to `{ try $0.message }`.

Result:

As an example, part of the generated code for the Echo service which looks like:

```swift
extension Echo_Echo.ClientProtocol {
    internal func get<R>(
        request: ClientRequest.Single<Echo_EchoRequest>,
        options: CallOptions = .defaults,
        _ body: @Sendable @escaping (ClientResponse.Single<Echo_EchoResponse>) async throws -> R
    ) async throws -> R where R: Sendable {
        try await self.get(...)
    }
```

Becomes:

```swift
extension Echo_Echo.ClientProtocol {
    internal func get<R>(
        request: ClientRequest.Single<Echo_EchoRequest>,
        options: CallOptions = .defaults,
        _ body: @Sendable @escaping (ClientResponse.Single<Echo_EchoResponse>) async throws -> R = {
	    try $0.message
        }
    ) async throws -> R where R: Sendable {
        try await self.get(...)
    }
```

### DIFF
--- a/Sources/GRPCCodeGen/Internal/Translator/ClientCodeTranslator.swift
+++ b/Sources/GRPCCodeGen/Internal/Translator/ClientCodeTranslator.swift
@@ -37,7 +37,9 @@
 ///   public func baz<R>(
 ///     request: ClientRequest.Single<Foo_Bar_Input>,
 ///     options: CallOptions = .defaults,
-///     _ body: @Sendable @escaping (ClientResponse.Single<Foo_Bar_Output>) async throws -> R
+///     _ body: @Sendable @escaping (ClientResponse.Single<Foo_Bar_Output>) async throws -> R = {
+///       try $0.message
+///     }
 ///   ) async throws -> R where R: Sendable {
 ///     try await self.baz(
 ///       request: request,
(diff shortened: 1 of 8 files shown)


Sources/GRPCCodeGen/Internal/Translator/ClientCodeTranslator.swift
```diff
@@ -324,7 +337,30 @@ extension ClientCodeTranslator {
...
+            item: .expression(
+              .try(
+                .memberAccess(
+                  MemberAccessDescription(left: .identifierPattern("$0"), right: "message")
+                )
+              )
+            )
+          )
+        ]
+      )
```
glbrntt, line 356: This can be much simpler:

```swift
  ClosureInvocationDescription(body: [.expression(.try(.identifierPattern("$0").dot("message")))])
```

glbrntt (changes requested):
One nit, looks good otherwise.

PR #2010, opened 2024-08-08
Add explicit access-level modifier support for imports in code generator

gjcairo:
## Motivation
This is a follow-up of https://github.com/grpc/grpc-swift/pull/2003.
As of 5.9, Swift supports access-level modifiers on imports. In an upcoming Swift 6.x release, the default modifier will become `internal`.

## Modifications
This PR adds explicit access-level modifiers to the generated code and a few other modules that use this generated code.

## Result
Explicit access level imports present in more places.

### DIFF
--- a/Sources/GRPCCodeGen/CodeGenerationRequest.swift
+++ b/Sources/GRPCCodeGen/CodeGenerationRequest.swift
@@ -87,9 +87,6 @@ public struct CodeGenerationRequest {
     public var item: Item?
 
     /// The access level to be included in imports of this dependency.
-    ///
-    /// - Note: This defaults to `public`, which is the default access level for imports on compilers
-    /// up to and including Swift 6.0.
     public var accessLevel: SourceGenerator.Configuration.AccessLevel
 
     /// The name of the imported module or of the module an item is imported from.
--- a/Sources/GRPCCodeGen/CodeGenerationRequest.swift
+++ b/Sources/GRPCCodeGen/CodeGenerationRequest.swift
@@ -84,7 +84,13 @@ public struct CodeGenerationRequest {
   public struct Dependency: Equatable {
     /// If the dependency is an item, the property's value is the item representation.
     /// If the dependency is a module, this property is nil.
-    public var item: Item? = nil
+    public var item: Item?
+
+    /// The access level to be included in imports of this dependency.
+    ///
+    /// - Note: This defaults to `public`, which is the default access level for imports on compilers
+    /// up to and including Swift 6.0.
+    public var accessLevel: SourceGenerator.Configuration.AccessLevel
 
     /// The name of the imported module or of the module an item is imported from.
     public var module: String
@@ -102,12 +108,14 @@ public struct CodeGenerationRequest {
       item: Item? = nil,
       module: String,
       spi: String? = nil,
-      preconcurrency: PreconcurrencyRequirement = .notRequired
+      preconcurrency: PreconcurrencyRequirement = .notRequired,
+      accessLevel: SourceGenerator.Configuration.AccessLevel
     ) {
       self.item = item
       self.module = module
       self.spi = spi
       self.preconcurrency = preconcurrency
+      self.accessLevel = accessLevel
     }
 
     /// Represents an item imported from a module.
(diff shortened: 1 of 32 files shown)


Sources/GRPCCodeGen/CodeGenerationRequest.swift
```diff
@@ -84,7 +84,13 @@ public struct CodeGenerationRequest {
   public struct Dependency: Equatable {
     /// If the dependency is an item, the property's value is the item representation.
     /// If the dependency is a module, this property is nil.
-    public var item: Item? = nil
+    public var item: Item?
+
+    /// The access level to be included in imports of this dependency.
+    ///
+    /// - Note: This defaults to `public`, which is the default access level for imports on compilers
+    /// up to and including Swift 6.0.
```
glbrntt, line 92: Bit misleading because we don't default it. I think you mean if it isn't explicitly set then assumed implicit default is public.

Sources/GRPCCodeGen/Internal/StructuredSwiftRepresentation.swift
```diff
@@ -31,6 +31,11 @@
 ///
 /// For example: `import Foo`.
 struct ImportDescription: Equatable, Codable {
+  /// The access level of the imported module.
+  ///
+  /// For example, the `public` in `public import Foo`.
+  var accessLevel: AccessModifier
```
glbrntt, line 37: This should be optional.

You should think about building this from the other direction: you should start with the build blocks, i.e. the 'structured swift' and build from there. Starting there it's more obvious that this should be optional: because you don't have to have explicit import access levels.

It's only when you get to the grpc code gen layer that we decide to be explicit about the imports.

Doing this also means we can have less churn in the tests as all the existing tests for import description would remain as valid tests. You'd only need to add one more test with an explicit access level to make sure our bases our covered.

glbrntt (changes requested):

Sources/GRPCCodeGen/Internal/StructuredSwiftRepresentation.swift
gjcairo: I had started with this as an optional (both here and in `ImportDescription`), but because by the time I got to the higher layers and was always being explicit about it, I thought it added some unnecessary logic/unwraps elsewhere.

I'm not sure I understand the "It's only when you get to the grpc code gen layer that we decide to be explicit about the imports." point. Both this `StructuredSwiftRepresentation` and the `ImportDescription` are at the `GRPCCodeGen` layer - that's precisely why I ended up changing my mind about it. 

Having said that, I can make it optional again: I can see the argument for it from a "Swift doesn't require you to have it, so shouldn't we" perspective. I was just trying to simplify the logic if we'll always have a value for it in our codegen.

glbrntt: > I'm not sure I understand the "It's only when you get to the grpc code gen layer that we decide to be explicit about the imports." point. Both this `StructuredSwiftRepresentation` and the `ImportDescription` are at the `GRPCCodeGen` layer - that's precisely why I ended up changing my mind about it.

You're conflating the module and layer here. You can have different API layers within a module and that's what we have here. 

The surface API of the module is the code gen request and response. Within the module there are separate API layers: a translation layer which coverts the request into structured swift, and a rendering layer which coverts the structured swift into text. 

It's the translation layer that decides to be explicit about imports, the rendering layer doesn't express an opinion on whether they should have access levels, it just knows how to render them.

Making it optional is a smaller change (so reduces churn) and gives us more flexibility. If, for example, we need to add an option about whether generated code should have explicit imports then making that change would be easy: we'd express that opinion in the request and pass it to the translation layer.

Sources/GRPCCodeGen/CodeGenerationRequest.swift
gjcairo: Hm yes but with this API this isn't an optional argument, we're not defaulting it so it must always be explicit. I'll just get rid of this comment (unless we want to default it in the `init`)

glbrntt (approved):
LGTM, thanks Gus!

PR #2012, opened 2024-08-09
Use fully qualified types in generated code

clintonpi:
Motivation:

The generated code has possible namespace collisions between user-defined types and gRPC types.

Modifications:

- Adjust `GRPCCodeGen` and `GRPCProtobufCodeGen` to use the fully qualified names of gRPC types when generating code.

Result:

This will help avoid namespace collision between user-defined types and gRPC types.

### DIFF
--- a/Sources/GRPCCodeGen/Internal/Translator/ClientCodeTranslator.swift
+++ b/Sources/GRPCCodeGen/Internal/Translator/ClientCodeTranslator.swift
@@ -472,7 +472,7 @@ extension ClientCodeTranslator {
     parameters.append(
       ParameterDescription(
         label: "options",
-        type: .member("GRPCCore.CallOptions"),
+        type: .member(["GRPCCore", "CallOptions"]),
         defaultValue: includeDefaultCallOptions
           ? .memberAccess(MemberAccessDescription(right: "defaults")) : nil
       )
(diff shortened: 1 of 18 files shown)


Sources/GRPCCodeGen/Internal/Translator/ClientCodeTranslator.swift
```diff
@@ -254,7 +254,7 @@ extension ClientCodeTranslator {
     parameters.append(
       ParameterDescription(
         label: "options",
-        type: .member("CallOptions"),
+        type: .member("GRPCCore.CallOptions"),
```
glbrntt, line 257: The `.member(String)` API is an overload, in this instance prefer using the array version which will add the dots for you: `.member(["GRPCCore", "CallOptions"])`

Sources/GRPCCodeGen/Internal/Translator/ClientCodeTranslator.swift
```diff
@@ -267,7 +267,7 @@ extension ClientCodeTranslator {
     in service: CodeGenerationRequest.ServiceDescriptor
   ) -> ParameterDescription {
     let requestType = method.isInputStreaming ? "Stream" : "Single"
-    let clientRequestType = ExistingTypeDescription.member(["ClientRequest", requestType])
+    let clientRequestType = ExistingTypeDescription.member(["GRPCCore.ClientRequest", requestType])
```
glbrntt, line 270: Same here (and in a bunch of places), `GRPCCore` should be an additional string in the array

glbrntt (changes requested):
Broadly fine there are a bunch of merge conflicts and one issue that needs fixing in a bunch of places.

glbrntt (approved):
LGTM, thanks!

PR #2013, opened 2024-08-09
Add check that generated code is up-to-date

glbrntt:
Motivation:

It's easy for checked in generated code to not be up-to-date.

Modifications:

- Add a script to check it's up-to-date and run it in CI

Result:

More guarantess that checked in code is up-to-date

### DIFF
--- a/Protos/generate.sh
+++ b/Protos/generate.sh
@@ -119,7 +119,7 @@ function generate_reflection_service {
   mv "$output_v1/reflection.grpc.swift" "$output_v1/reflection-v1.grpc.swift"
 
   local proto_v1alpha="$here/upstream/grpc/reflection/v1alpha/reflection.proto"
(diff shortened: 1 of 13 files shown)


Protos/generate.sh
```diff
@@ -119,7 +119,7 @@ function generate_reflection_service {
   mv "$output_v1/reflection.grpc.swift" "$output_v1/reflection-v1.grpc.swift"
 
   local proto_v1alpha="$here/upstream/grpc/reflection/v1alpha/reflection.proto"
-  local output_v1alpha="$root/Sources/GRPCReflectionService/v1alpha"
+  local output_v1alpha="$root/Sources/GRPCReflectionService/v1Alpha"
```
glbrntt, line 122: FWIW this wasn't an issue on macOS as AFPS is case-insensitive by default.

gjcairo (approved):

PR #2014, opened 2024-08-12
Improve the names of a few commonly used APIs

glbrntt:
Motivation:

Naming is important; it should be clear and concise.

Modifications:

The follow renames all offer more precise names:
- Rename `server.run()` to `server.serve()`
- Rename `server.stopListening()` to `server.beginGracefulShutdown()`
- Rename `client.close()` to `client.beginGracefulShutdown()`

Result:

Clearer APIs

### DIFF
--- a/Sources/GRPCCore/GRPCServer.swift
+++ b/Sources/GRPCCore/GRPCServer.swift
@@ -60,11 +60,11 @@ internal import Atomics
 ///
 /// ```swift
 /// // Start running the server.
-/// try await server.run()
+/// try await server.serve()
 /// ```
 ///
 /// The ``run()`` method won't return until the server has finished handling all requests. You can
-/// signal to the server that it should stop accepting new requests by calling ``stopListening()``.
+/// signal to the server that it should stop accepting new requests by calling ``beginGracefulShutdown()``.
 /// This allows the server to drain existing requests gracefully. To stop the server more abruptly
 /// you can cancel the task running your server. If your application requires additional resources
 /// that need their lifecycles managed you should consider using [Swift Service
@@ -154,13 +154,13 @@ public struct GRPCServer: Sendable {
   ///
   /// This function returns when the configured transport has stopped listening and all requests have been
   /// handled. You can signal to the transport that it should stop listening by calling
-  /// ``stopListening()``. The server will continue to process existing requests.
+  /// ``beginGracefulShutdown()``. The server will continue to process existing requests.
   ///
   /// To stop the server more abruptly you can cancel the task that this function is running in.
   ///
   /// - Note: You can only call this function once, repeated calls will result in a
   ///   ``RuntimeError`` being thrown.
-  public func run() async throws {
+  public func serve() async throws {
     let (wasNotStarted, actualState) = self.state.compareExchange(
       expected: .notStarted,
       desired: .running,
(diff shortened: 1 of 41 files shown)


Sources/GRPCCore/GRPCServer.swift
```diff
@@ -154,13 +154,13 @@ public struct GRPCServer: Sendable {
...
   /// handled. You can signal to the transport that it should stop listening by calling
-  /// ``stopListening()``. The server will continue to process existing requests.
+  /// ``beginGracefulShutdown()``. The server will continue to process existing requests.
   ///
   /// To stop the server more abruptly you can cancel the task that this function is running in.
   ///
   /// - Note: You can only call this function once, repeated calls will result in a
   ///   ``RuntimeError`` being thrown.
-  public func run() async throws {
+  public func serve() async throws {
```
FranzBusch, line 163: I am a bit wary of renaming this from `run` to `serve`. While I understand what you are getting at. We do have many libraries that now adopted the `run` method name to align with what `Lifecycle` expects. If you rename this here and conform to `Service` in the future your type will have two methods that do the same.

Sources/GRPCCore/GRPCServer.swift
```diff
@@ -206,10 +206,10 @@ public struct GRPCServer: Sendable {
   /// Signal to the server that it should stop listening for new requests.
   ///
   /// By calling this function you indicate to clients that they mustn't start new requests
-  /// against this server. Once the server has processed all requests the ``run()`` method returns.
+  /// against this server. Once the server has processed all requests the ``serve()`` method returns.
   ///
   /// Calling this on a server which is already stopping or has stopped has no effect.
-  public func stopListening() {
+  public func beginGracefulShutdown() {
```
FranzBusch, line 212: I would suggest using `triggerGracefulShutdown` to align with https://github.com/swift-server/swift-service-lifecycle/blob/24c800fb494fbee6e42bc156dc94232dc08971af/Sources/ServiceLifecycle/ServiceGroup.swift#L167.

Sources/GRPCCore/GRPCServer.swift
glbrntt: `run` is less clear and is worse for users which don't depend on lifecycle. Users would have to provide shims to use `Lifecycle` anyway (to avoid retroactive conformance). If we provide conformance in the future (e.g. via package traits) then having a `run()` method calling through to `serve()` isn't an issue.

glbrntt: From the perspective of a user using `Lifecycle` the names of these methods don't matter at all because they won't be called directly by the user.

From the perspective of a user _not_ using `Lifecycle`, these methods matter a lot: they _will_ be called by the user so they should be clear.

FranzBusch: I still remain unconvinced. `serve` and `run` are pretty close in my mind. Even leaving `Lifecycle` aside just the proliferation of `run` methods across the ecosystem is IMO enough to justify aligning. It becomes very natural that you call `run` on your server/clients when you don't use something like `Lifecycle`.

glbrntt: Noted. I'm pretty strongly against aligning for the sake of aligning on a generic name: I think we should pick names that express their intent and provide the most clarity, the APIs should be approachable for all users including those new to Swift and SoS!

gjcairo (approved):

PR #2015, opened 2024-08-12
Fixup various doc issues

glbrntt:


Sources/GRPCHTTP2Core/Client/HTTP2ClientTransport.swift
```diff
@@ -28,12 +28,12 @@ extension HTTP2ClientTransport.Config {
...
     /// The default algorithm used for compressing outbound messages.
     ///
-    /// This can be overridden on a per-call basis via ``CallOptions``.
+    /// This can be overridden on a per-call basis via `CallOptions`.
     public var algorithm: CompressionAlgorithm
 
     /// Compression algorithms enabled for inbound messages.
     ///
-    /// - Note: ``CompressionAlgorithm/none`` is always supported, even if it isn't set here.
+    /// - Note: `CompressionAlgorithm.none` is always supported, even if it isn't set here.
```
gjcairo, line 36: Can these not be linked?

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ServerTransport+Posix.swift
```diff
@@ -294,7 +294,7 @@ extension HTTP2ServerTransport {
 
 @available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *)
 extension HTTP2ServerTransport.Posix {
-  /// Configuration for the ``GRPCHTTP2TransportNIOPosix/GRPCHTTP2Core/HTTP2ServerTransport/Posix``.
+  /// Config for the `Posix` transport.
```
gjcairo, line 297: I think having the links to the types makes navigation easier if you're going over it. If it's a matter of making it less "verbose", can we at least add a `- SeeAlso:` section (here and in the few other places we've removed links)?

Sources/GRPCHTTP2Core/Client/HTTP2ClientTransport.swift
glbrntt: Nope, because the come from a different module (I think originally they were in the right module but moved them to core)

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ServerTransport+Posix.swift
glbrntt: I agree but DocC gets trips over these and emits warnings. I think the problem stems from the namespace being defined in `GRPCHTTP2Core` and the type being defined in an extension in `GRPCHTTP2TransportNIOPosix`.

gjcairo: Ah, I see that's sad :( 
Okay then, ignore this, nothing we can do.

gjcairo (approved):

PR #2016, opened 2024-08-13
Add docc docs for grpc core

glbrntt:
Motivation:

As we move towards v2 we need to make a clearer separation between v1 and v2 docs to avoid confusion.

Modifications:

- Remove v1 specific content out of README.md (it is duplicated in the v1 DocC docs) and provide a clearer update on the state of the project. This also points users to the v1 docs on Swift Package Index where the duplicated v1 content lives.
- Add a doc catalog to gRPC core with a very high level description of the package products
- Add a "Benchmarks"  tutorial to the v2 doc catalog (it was content removed from the README)

Result:

- DocC catalog in place for v2
- README is clearer

### DIFF
--- a/Sources/GRPCCore/Documentation.docc/Documentation.md
+++ b/Sources/GRPCCore/Documentation.docc/Documentation.md
@@ -1,0 +1,28 @@
+# ``GRPCCore``
+
+A gRPC library for Swift written natively in Swift.
+
+> 🚧 This module is part of gRPC Swift v2 which is under active development and in the pre-release
+> stage.
+
+## Package structure
+
+gRPC Swift is made up of a number of modules, each of which is documented separately. However this
+module – ``GRPCCore`` – includes higher level documentation such as tutorials. The following list
+contains products of this package:
+
+- ``GRPCCore`` contains core types and abstractions and is the 'base' module for the project.
+- `GRPCInProcessTransport` contains an implementation of an in-process transport.
+- `GRPCHTTP2TransportNIOPosix` provides client and server implementations of HTTP/2 transports built
+  on top of SwiftNIO's POSIX Sockets abstractions.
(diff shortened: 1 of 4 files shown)


Sources/GRPCCore/Documentation.docc/Documentation.md
```diff
@@ -0,0 +1,28 @@
...
+
+- ``GRPCCore`` contains core types and abstractions and is the 'base' module for the project.
+- `GRPCInProcessTransport` contains an implementation of an in-process transport.
+- `GRPCHTTP2TransportNIOPosix` provides client and server implementations of HTTP/2 transports built
+  on top of SwiftNIO's POSIX Sockets abstractions.
+- `GRPCHTTP2TransportNIOTransportServices` provides client and server implementations of HTTP/2
+  transports built on top of SwiftNIO's Network.framework abstraction, `NIOTransportServices`.
+- `GRPCProtobuf` provides serialization and deserialization components for `SwiftProtobuf`.
+
+## Topics
```
gjcairo, line 22: Is this section supposed to be empty or do we want to remove it / add something to it?

gjcairo (approved):
LGTM, just a small comment

Sources/GRPCCore/Documentation.docc/Documentation.md
glbrntt: It's not empty, it has the "Getting involved" subsection. There will be more subsections (e.g. tutorials/articles) in the future too. 

`## Topics` is also magic: if you don't add it then DocC will add it and list all articles etc.
 underneath it (as well as API docs). 

If you add it manually it'll add in everything that you didn't explicitly add, so at the moment it'll look like this:

Screenshot 2024-08-14 at 13 22 47

PR #2018, opened 2024-08-14
Remove 'BufferedStream'

glbrntt:
Motivation:

We use buffered stream in a couple of places which don't require backpressure.

Modifications:

- Replace occurrences with `AsyncStream`

Result:

Less code to maintain


Sources/GRPCCore/Streaming/RPCWriterProtocol.swift
```diff
@@ -65,3 +65,27 @@ public protocol ClosableRPCWriterProtocol<Element>: RPCWriterProtocol {
...
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension AsyncThrowingStream.Continuation: ClosableRPCWriterProtocol where Failure == any Error {
+  public func finish() {
+    self.finish(throwing: nil)
+  }
+
+  public func finish(throwing error: any Error) {
+    self.finish(throwing: error as (any Error)?)
+  }
```
gjcairo, line 90: Question: do we want to consider making `ClosableRPCWriterProtocol/finish(throwing:)` take an `Optional<any Error>`? That way we could avoid having to implement these.

If we do keep this, could we rewrite `self.finish(throwing: error as (any Error)?)` as `self.finish(throwing: .some(error))`? It's a lot more readable to me.

glbrntt: I agree that `error as (any Error)?` is weird though but I think having that internally in one place within the library is preferable to having public API which can be called as `finish(throwing: nil)`. (Which is weird, especially when referenced in docs because it'll always include the `throwing:`.) WDYT?

gjcairo: Hm yes, I agree that the `throwing:` is not great - probably better to leave this as is then. But what do you think of replacing the `error as (any Error)?` with `.some(error)`?

glbrntt: Yeah, that's a much better spelling. Sorry, didn't meant to miss it beforehand!

gjcairo (approved):

gjcairo (approved):

PR #2022, opened 2024-08-19
Don't police "te: trailers" at the grpc layer

glbrntt:
Motivation:

"te: trailers" is required in the request to detect incompatible proxies. However, "te" is a hop-by-hop header so isn't meant to be forwarded, as such the only validation the server should do is to check whether the value is "trailers" iff "te" is present. This doesn't need to be done at the gRPC layer because NIOHTTP2 already does this validation.

Modifications:

- Remove the check for "te: trailers"

Result:

Fewer bugs


gjcairo:
We're also adding this trailer in `GRPCStreamStateMachine.makeClientHeaders(methodDescriptor:scheme:outboundEncoding:acceptedEncodings:customMetadata:)` - I suppose we should get rid of this?
And I believe we can get rid of the `te` case in `GRPCHTTP2Keys` if we do that, because there aren't other uses.

gjcairo (changes requested):
We're also adding this trailer in `GRPCStreamStateMachine.makeClientHeaders(methodDescriptor:scheme:outboundEncoding:acceptedEncodings:customMetadata:)` - I suppose we should get rid of this?
And I believe we can get rid of the `te` case in `GRPCHTTP2Keys` if we do that, because there aren't other uses.

glbrntt:
> We're also adding this trailer in `GRPCStreamStateMachine.makeClientHeaders(methodDescriptor:scheme:outboundEncoding:acceptedEncodings:customMetadata:)` - I suppose we should get rid of this? And I believe we can get rid of the `te` case in `GRPCHTTP2Keys` if we do that, because there aren't other uses.

No, the client must send it, but the server doesn't have to validate it.

gjcairo (approved):

PR #2026, opened 2024-08-21
Adopt `Mutex` where possible

glbrntt:
Motivation:

Swift 6 has a `Mutex` which we can use in place of `LockedValueBox` or `NIOLockedValueBox` in a bunch of places.

Modifications:

- Use `Mutex` where possible
- This causes a number of other changes too as `Mutex` doesn't allocate: the allocations move to the type holding it instead. This makes sense as the objects which were previously structs didn't have value semantics.

Result:

Fewer uses of our own lock, no uses of NIOs lock


Sources/GRPCHTTP2TransportNIOTransportServices/HTTP2ServerTransport+TransportServices.swift
```diff
@@ -23,10 +23,12 @@ internal import NIOCore
 internal import NIOExtras
 internal import NIOHTTP2
 
+internal import Synchronization
```
gjcairo, line 26: This can be `private`

Sources/GRPCInProcessTransport/InProcessClientTransport.swift
```diff
@@ -15,6 +15,7 @@
  */
 
 public import GRPCCore
+internal import Synchronization
```
gjcairo, line 18: This can be `private`

gjcairo (changes requested):

Sources/GRPCCore/Internal/Concurrency Primitives/Lock.swift
```diff
@@ -237,48 +237,48 @@ extension UnsafeMutablePointer {
 }
 
 @usableFromInline
-package struct LockedValueBox<Value> {
```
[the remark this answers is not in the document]
glbrntt: Not easily, `Mutex` doesn't have the manual `lock()`/`unlock()` methods which we rely on in `BroadcastAsyncSequence`. I want to look at that separately as that's likely to require a larger change.

gjcairo (approved):

PR #2029, opened 2024-08-21
Add helpers for creating transports

glbrntt:
Motivation:

Creating an HTTP/2 transport requires quite a bit of typing and knowing what types to use. We can make use of type inference to make this a bit easier for users.

Modifications:

- Add constrained extensions to `ClientTransport` and `ServerTransport` allowing for `http2NIO` and `http2NIOTS` transports to be created via static methods. This makes it possible for transports to be created inline in the `init` of a `GRPCClient` or `GRPCServer` without the users need the full type.
- Extend the `defaults` functions so that they can take an optional trailing closure to further modify the defaults. This allows for configs to be created and modified inline when making a transport.
- A config is now required to create client transports and the NIOTS server transport as these will soon require a transport security config.

Results:

Easier to create clients/servers


Sources/GRPCCore/GRPCServer.swift
```diff
@@ -74,7 +74,7 @@ public struct GRPCServer: Sendable {
   typealias Stream = RPCStream<ServerTransport.Inbound, ServerTransport.Outbound>
 
   /// The ``ServerTransport`` implementation that the server uses to listen for new requests.
-  private let transport: any ServerTransport
+  public let transport: any ServerTransport
```
gjcairo, line 77: Does it make sense to have this `public`? Is there any reason why a user would want to reach out to the transport directly after it's been configured? Otherwise it could just be `package` to be used in the `ListeningServerTransport` extension.

Sources/GRPCHTTP2Core/ListeningServerTransport.swift
```diff
@@ -0,0 +1,41 @@
...
+public protocol ListeningServerTransport: ServerTransport {
+  /// Returns the listening address of the server transport once it has started.
+  var listeningAddress: SocketAddress { get async throws }
+}
+
+@available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *)
+extension GRPCServer {
+  /// Returns the listening address of the server transport once it has started.
+  ///
+  /// This will `nil` if the transport doesn't conform to ``ListeningServerTransport``.
```
gjcairo, line 31:
```suggestion
  /// This will be `nil` if the transport doesn't conform to ``ListeningServerTransport``.
```

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ClientTransport+Posix.swift
```diff
@@ -69,8 +69,8 @@ extension HTTP2ClientTransport {
     /// - Throws: When no suitable resolver could be found for the `target`.
     public init(
       target: any ResolvableTarget,
+      config: Config,
```
gjcairo, line 72: Nit: can we reorder the param in the docs too?

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ClientTransport+Posix.swift
```diff
@@ -202,3 +207,34 @@ extension GRPCChannel.Config {
...
 }
+
+@available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *)
+extension ClientTransport where Self == HTTP2ClientTransport.Posix {
+  /// Creates a new Posix based HTTP/2 client transport.
+  ///
+  /// - Parameters:
+  ///   - target: A target to resolve.
+  ///   - resolverRegistry: A registry of resolver factories.
+  ///   - config: Configuration for the transport.
```
gjcairo, line 218: Can we swap these to match the func signature?

gjcairo (changes requested):
Nice one! Really like the `defaults` taking the closure to customise the config. Just a few very minor nits.

Sources/GRPCCore/GRPCServer.swift
glbrntt: The rationale for making it public was that the user passed in the transport, so providing a `public` reference to it doesn't give them anything they didn't already have.

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ClientTransport+Posix.swift
glbrntt: Absolutely, good catch!

PR #2032, opened 2024-08-23
Allow access level on imports to be disabled

glbrntt:
Motivation:

The v2 code generator defaults to generating access levels on imports. SwiftProtobuf also has but hasn't yet been released. This causes problems as explicit 'internal' imports in gRPC are ambiguous with the implicit imports from Protobuf.

Modifications:

- Allow an opt-out of explicit imports for v2 code generation

Result:

Can build against current version of Protobuf without `InternalImportsByDefault`.


Sources/GRPCCodeGen/SourceGenerator.swift
```diff
@@ -28,15 +28,24 @@ public struct SourceGenerator: Sendable {
...
+    public var accessLevelOnImports: Bool
     /// The indentation of the generated code as the number of spaces.
     public var indentation: Int
     /// Whether or not client code should be generated.
     public var client: Bool
     /// Whether or not server code should be generated.
     public var server: Bool
 
-    public init(accessLevel: AccessLevel, client: Bool, server: Bool, indentation: Int = 4) {
+    public init(
```
gjcairo, line 40: Not related to this change but this init is missing docs

gjcairo (approved):

Sources/GRPCCodeGen/SourceGenerator.swift
glbrntt: Fixed!

PR #2036, opened 2024-08-28
Add TLS support for NIOPosix H2 client

gjcairo:
## Motivation
We currently have a `NIOPosix` client transport implementation in gRPC v2, but it doesn't support TLS.

## Modifications
This PR adds support for TLS in the NIOPosix-backed HTTP/2 implementation of the client transport for gRPC v2.

## Result
We now support TLS when using the NIOPosix client transport in gRPC V2.

### DIFF
--- a/Sources/GRPCHTTP2TransportNIOPosix/HTTP2ClientTransport+Posix.swift
+++ b/Sources/GRPCHTTP2TransportNIOPosix/HTTP2ClientTransport+Posix.swift
@@ -158,7 +158,7 @@ extension HTTP2ClientTransport.Posix {
           )
         }
       }
-    #endif
+      #endif
     }
 
     func establishConnection(
--- a/Sources/GRPCHTTP2TransportNIOPosix/HTTP2ClientTransport+Posix.swift
+++ b/Sources/GRPCHTTP2TransportNIOPosix/HTTP2ClientTransport+Posix.swift
@@ -94,7 +94,7 @@ extension HTTP2ClientTransport {
       // Configure a connector.
       self.channel = GRPCChannel(
         resolver: resolver,
-        connector: Connector(eventLoopGroup: eventLoopGroup, config: config),
+        connector: try Connector(eventLoopGroup: eventLoopGroup, config: config),
         config: GRPCChannel.Config(posix: config),
         defaultServiceConfig: serviceConfig
       )
@@ -132,25 +132,24 @@ extension HTTP2ClientTransport.Posix {
     private let config: HTTP2ClientTransport.Posix.Config
     private let eventLoopGroup: any EventLoopGroup
 
-    init(eventLoopGroup: any EventLoopGroup, config: HTTP2ClientTransport.Posix.Config) {
+    #if canImport(NIOSSL)
+    private let nioSSLContext: NIOSSLContext?
+    private let serverHostname: String?
+    #endif
+
+    init(eventLoopGroup: any EventLoopGroup, config: HTTP2ClientTransport.Posix.Config) throws {
       self.eventLoopGroup = eventLoopGroup
       self.config = config
-    }
 
-    func establishConnection(
-      to address: GRPCHTTP2Core.SocketAddress
-    ) async throws -> HTTP2Connection {
       #if canImport(NIOSSL)
-      let nioSSLContext: NIOSSLContext?
-      let serverHostname: String?
       switch self.config.transportSecurity.wrapped {
       case .plaintext:
-        nioSSLContext = nil
-        serverHostname = nil
+        self.nioSSLContext = nil
+        self.serverHostname = nil
       case .tls(let tlsConfig):
         do {
-          nioSSLContext = try NIOSSLContext(configuration: TLSConfiguration(tlsConfig))
-          serverHostname = tlsConfig.serverHostname
+          self.nioSSLContext = try NIOSSLContext(configuration: TLSConfiguration(tlsConfig))
+          self.serverHostname = tlsConfig.serverHostname
         } catch {
           throw RuntimeError(
             code: .transportError,
@@ -159,18 +158,22 @@ extension HTTP2ClientTransport.Posix {
           )
         }
       }
-      #endif
+    #endif
+    }
 
+    func establishConnection(
+      to address: GRPCHTTP2Core.SocketAddress
+    ) async throws -> HTTP2Connection {
       let (channel, multiplexer) = try await ClientBootstrap(
         group: self.eventLoopGroup
       ).connect(to: address) { channel in
         channel.eventLoop.makeCompletedFuture {
           #if canImport(NIOSSL)
-          if let nioSSLContext {
+          if let nioSSLContext = self.nioSSLContext {
             try channel.pipeline.syncOperations.addHandler(
               NIOSSLClientHandler(
                 context: nioSSLContext,
-                serverHostname: serverHostname
+                serverHostname: self.serverHostname
               )
             )
           }
--- a/Sources/GRPCHTTP2TransportNIOPosix/HTTP2ClientTransport+Posix.swift
+++ b/Sources/GRPCHTTP2TransportNIOPosix/HTTP2ClientTransport+Posix.swift
@@ -19,14 +19,18 @@ public import GRPCHTTP2Core  // should be @usableFromInline
 public import NIOCore
 public import NIOPosix  // has to be public because of default argument value in init
 
+#if canImport(NIOSSL)
+import NIOSSL
+#endif
+
 @available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *)
 extension HTTP2ClientTransport {
-  /// A `ClientTransport` using HTTP/2 built on top of `NIOPosix`.
+  /// A ``GRPCCore/ClientTransport`` using HTTP/2 built on top of `NIOPosix`.
   ///
   /// This transport builds on top of SwiftNIO's Posix networking layer and is suitable for use
   /// on Linux and Darwin based platform (macOS, iOS, etc.) However, it's *strongly* recommended
   /// that if you are targeting Darwin platforms then you should use the `NIOTS` variant of
-  /// the `HTTP2ClientTransport`.
+  /// the ``GRPCHTTP2Core/HTTP2ClientTransport``.
   ///
   /// To use this transport you need to provide a 'target' to connect to which will be resolved
   /// by an appropriate resolver from the resolver registry. By default the resolver registry can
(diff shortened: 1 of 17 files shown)


Sources/GRPCHTTP2TransportNIOPosix/HTTP2ClientTransport+Posix.swift
```diff
@@ -19,31 +19,38 @@ public import GRPCHTTP2Core  // should be @usableFromInline
...
   /// Beyond creating the transport you don't need to interact with it directly, instead, pass it
   /// to a `GRPCClient`:
   ///
   /// ```swift
-  /// try await withThrowingDiscardingTaskGroup {
-  ///   let transport = try HTTP2ClientTransport.Posix(target: .dns(host: "example.com"))
+  /// try await withThrowingDiscardingTaskGroup { group in
+  ///   let transport = try HTTP2ClientTransport.Posix(
+  ///     target: .ipv4(host: "example.com"),
+  ///     config: .defaults(transportSecurity: someTransportSecurity)
```
glbrntt, line 52: Can we use `.plaintext`? IMO it's better if example are as close to compiling as possible.

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ClientTransport+Posix.swift
```diff
@@ -133,11 +140,37 @@ extension HTTP2ClientTransport.Posix {
...
+
       let (channel, multiplexer) = try await ClientBootstrap(
         group: self.eventLoopGroup
       ).connect(to: address) { channel in
         channel.eventLoop.makeCompletedFuture {
-          try channel.pipeline.syncOperations.configureGRPCClientPipeline(
+          #if canImport(NIOSSL)
+          if let nioSSLContext {
+            try channel.pipeline.syncOperations.addHandler(
+              NIOSSLServerHandler(context: nioSSLContext)
```
glbrntt, line 168: copy-pasta? Should be `NIOSSLClientHandler`

Sources/GRPCHTTP2TransportNIOPosix/NIOSSL+GRPC.swift
```diff
@@ -144,5 +144,25 @@ extension TLSConfiguration {
...
+    }
+
+    tlsConfiguration.minimumTLSVersion = .tlsv12
+    tlsConfiguration.certificateVerification = CertificateVerification(
+      tlsConfig.serverCertificateVerification
+    )
+    tlsConfiguration.trustRoots = try NIOSSLTrustRoots(tlsConfig.trustRoots)
+    tlsConfiguration.applicationProtocols = ["grpc-exp", "h2"]
+
+    self = tlsConfiguration
```
glbrntt, line 165: (nit) seems a bit indirect, why not `self = TLSConfiguration.makeClientConfiguration()` then modify self directly?

Sources/GRPCHTTP2TransportNIOPosix/TLSConfig.swift
```diff
@@ -217,3 +217,54 @@ extension HTTP2ServerTransport.Posix.Config {
...
+    public var trustRoots: TLSConfig.TrustRootsSource
+
+    /// Create a new HTTP2 NIO Posix transport TLS config, with some values defaulted:
+    /// - `certificateChain` is `[]`
+    /// - `privateKey` is `nil`
+    /// - `serverCertificateVerification` equals `fullVerification`
+    /// - `trustRoots` equals `systemDefault`
+    ///
+    /// - Returns: A new HTTP2 NIO Posix transport TLS config.
+    public static func defaults() -> Self {
```
glbrntt, line 261: Prefer doing a `static var defaults: Self { ... }` as we don't have any parameters

Sources/GRPCHTTP2TransportNIOPosix/TLSConfig.swift
```diff
@@ -217,3 +217,54 @@ extension HTTP2ServerTransport.Posix.Config {
...
+    /// This connection is plaintext: no encryption will take place.
+    public static let plaintext = Self(wrapped: .plaintext)
+
+    /// This connection will use TLS.
+    public static func tls(_ tls: TLS) -> Self {
+      Self(wrapped: .tls(tls))
+    }
+  }
+
+  public struct TLS: Sendable {
```
glbrntt, line 241: Do we have a `public init` on the server equivalent? If so we should have the same here

Sources/GRPCHTTP2TransportNIOPosix/TLSConfig.swift
```diff
@@ -217,3 +217,54 @@ extension HTTP2ServerTransport.Posix.Config {
...
+
+    /// Create a new HTTP2 NIO Posix transport TLS config, with some values defaulted:
+    /// - `certificateChain` is `[]`
+    /// - `privateKey` is `nil`
+    /// - `serverCertificateVerification` equals `fullVerification`
+    /// - `trustRoots` equals `systemDefault`
+    ///
+    /// - Returns: A new HTTP2 NIO Posix transport TLS config.
+    public static func defaults() -> Self {
+      Self.init(
```
glbrntt, line 262:
```suggestion
      Self(
```

Sources/GRPCHTTP2TransportNIOPosix/TLSConfig.swift
```diff
@@ -217,3 +217,54 @@ extension HTTP2ServerTransport.Posix.Config {
...
+        certificateChain: [],
+        privateKey: nil,
+        serverCertificateVerification: .fullVerification,
+        trustRoots: .systemDefault
+      )
+    }
```
glbrntt, line 268: Can we add an mTLS helper too?

Sources/GRPCHTTP2TransportNIOPosix/TLSConfig.swift
glbrntt, line 241: We'll also need to be able to configure an optional `serverHostname`

Tests/GRPCHTTP2TransportTests/HTTP2TransportTests.swift
```diff
@@ -199,7 +199,7 @@ final class HTTP2TransportTests: XCTestCase {
       serviceConfig.loadBalancingConfig = [.roundRobin]
       transport = try HTTP2ClientTransport.Posix(
         target: target,
-        config: .defaults {
+        config: .defaults(transportSecurity: .plaintext) {
```
glbrntt, line 202: Now that we have client and server support for TLS we should add some tests. @simonjbeaumont was going to use `swift-certificates` in NIOSSL to generate test certs in tests rather than futzing around with scripts to regenerate them every year so might be worth syncing with him.

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ClientTransport+Posix.swift
gjcairo: Fair enough - I was torn between having it compile as-is, or suggesting potentially bad practices i.e. having someone copy this and use it in their own code and defaulting to unsecured connections :D

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ClientTransport+Posix.swift
gjcairo: Ah, yes 🫠 
Now, the client requires an (optional) hostname to validate certificates against. But we only have the server's address, not its hostname. Should we get the hostname as part of the TLS config?

Sources/GRPCHTTP2TransportNIOPosix/NIOSSL+GRPC.swift
gjcairo: Yeah, not any reason specifically. I wrote the `init` above like this and I just copied this one. I'll modify both.

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ClientTransport+Posix.swift
gjcairo: Ignore that comment - just saw your comment about the TLSConfig answering my question :D

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ClientTransport+Posix.swift
glbrntt: That's fair, I think it's okay in this case because it's explicit. It'd be a different story if it was `config: .defaults` for example.

Sources/GRPCHTTP2TransportNIOPosix/TLSConfig.swift
gjcairo: We don't, no. We only have the `defaults` and `mTLS` flavours and you can then edit those.

Sources/GRPCHTTP2TransportNIOPosix/TLSConfig.swift
gjcairo: Ah, I didn't because I don't think anything changes? I may be wrong here but the client will verify server certificates in either "normal" TLS or mTLS, so the configuration is the same in both cases.

glbrntt: In mTLS the client will present its own cert to the server as well. We have some tests in v1 which cover this: https://github.com/grpc/grpc-swift/blob/6f396ca3aed37ecb4f9c7b4d6fbb0b98df883860/Tests/GRPCTests/MutualTLSTests.swift#L156

Tests/GRPCHTTP2TransportTests/HTTP2TransportTests.swift
gjcairo: We decided we'll do this in a follow-up PR

glbrntt (changes requested):
One small thing to shuffle around but looks good otherwise, thanks Gus!

Sources/GRPCHTTP2TransportNIOPosix/HTTP2ClientTransport+Posix.swift
```diff
@@ -133,11 +140,43 @@ extension HTTP2ClientTransport.Posix {
...
+      #if canImport(NIOSSL)
+      let nioSSLContext: NIOSSLContext?
+      let serverHostname: String?
+      switch self.config.transportSecurity.wrapped {
+      case .plaintext:
+        nioSSLContext = nil
+        serverHostname = nil
+      case .tls(let tlsConfig):
+        do {
+          nioSSLContext = try NIOSSLContext(configuration: TLSConfiguration(tlsConfig))
```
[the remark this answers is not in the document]
gjcairo: While making this change I realised that we're exposing the `tls` case for `TransportSecurity` even when `NIOSSL` cannot be imported. I think this is unsafe, because it means users may be able to add a `TLSConfig` to their server/client, which then simply will be ignored if the platform the code's being built on doesn't support NIOSSL. If this happens on both the server and the client, it could mean the connection is plaintext and there would be no warnings that the config wasn't used.
I've wrapped the `tls` case of `TransportSecurity` in `#if canImport(NIOSSL)` to avoid this. Let me know if you agree with it or if I'm missing something/think it should be resolved in some other way.

glbrntt (approved):
LGTM, thanks Gus!

PR #2042, opened 2024-09-03
Enable `InternalImportsByDefault` in remaining modules

gjcairo:
This PR enables `InternalImportsByDefault` on `GRPCHealth` and `InteroperabilityTests`.
It also regenerates the protos with the latest version of swift-protobuf (v1.28.1), and increases its min version to 1.28.1 to make sure that we don't get build failures when using earlier generators.


Package@swift-6.swift
```diff
@@ -56,7 +56,7 @@ let packageDependencies: [Package.Dependency] = [
   ),
   .package(
     url: "https://github.com/apple/swift-protobuf.git",
-    from: "1.27.0"
+    from: "1.28.1"
```
glbrntt, line 59: Can you bump this in `Package.swift` as well?

Protos/generate.sh
```diff
@@ -228,7 +228,7 @@ function generate_service_messages_interop_tests {
   local output="$root/Sources/InteroperabilityTests/Generated"
 
   for proto in "${protos[@]}"; do
-    generate_message "$proto" "$here/tests/interoperability" "$output" "Visibility=Public" "FileNaming=DropPath"
+    generate_message "$proto" "$here/tests/interoperability" "$output" "Visibility=Public" "FileNaming=DropPath" "UseAccessLevelOnImports=true"
```
glbrntt, line 231: Isn't this the default for protobuf?

gjcairo: It's the default only if compiling with Swift 6 (it defaults to false on <6). Since we use this only in the v2 modules I could get rid of it, but I thought it was clearer. I can get rid of it though, I don't think it matters.

glbrntt: Gotcha. I think we should be consistent with `generate_grpc`, so either explicitly set both or set neither.

glbrntt: Oh, protobuf uses Swift 6 _language mode_:

https://github.com/apple/swift-protobuf/blob/beeb4142e775f40fa4162b52af32ed29ade4db1e/Sources/protoc-gen-swift/GeneratorOptions.swift#L82-L86

We default to true but ignore it in v1: https://github.com/grpc/grpc-swift/blob/63168d007b52d8d5d7493af3f5feb4cf8cb7c801/Sources/protoc-gen-grpc-swift/options.swift#L74

This means if you use gRPC Swift v2 with a Swift 6 compiler but in Swift 5 language mode you get imports for gRPC but not Protobuf  (unless you specify otherwise). Would you mind changing our options to match Protobuf?

gjcairo: Ah, good catch. Fixed.

Protos/generate.sh
```diff
@@ -29,7 +29,7 @@ bin_path=$(swift build -c release --show-bin-path)
 protoc_gen_swift="$bin_path/protoc-gen-swift"
 protoc_generate_grpc_swift="$bin_path/protoc-gen-grpc-swift"
 
-# Genreates gRPC by invoking protoc with the gRPC Swift plugin.
+# Generates gRPC by invoking protoc with the gRPC Swift plugin.
```
glbrntt, line 32: Spelling is hard 🥲

Protos/generate.sh
glbrntt: Half fixed 🙃 

Can we have the args to `generate_message` and `generate_grpc` be consistent? Either both explicitly set `"UseAccessLevelOnImports=true"` or have neither of them set it?

glbrntt (approved):
One nit got missed but looks good otherwise

Protos/generate.sh
gjcairo: Okay, so I've been playing around with this and I think we've got a bit of a problem.
Because the compiler flag dance `swift-protobuf` did (and we copied) is on the language version, it doesn't matter what compiler we're using: we'll default this option to `false` because both have their language modes set to 5 for the plugin target.

We could get around this on gRPC and get the right default if, in the `Package@swift-6.swift` definition of the protoc target, we enable v6 language mode. However, this doesn't affect the `swift-protobuf` generation, because language mode is still v5 for `swift-protobuf`'s plugin.

I'm starting to think we should just change the option defaults to use `#if compiler` instead of `#if swift` both here and on protobuf. I've made this change on gRPC, but for now I'll also explicitly set the options for both in `generate.sh` because we'll need this to change in swift-protobuf too.

glbrntt: Good point. I think the behaviour we really want is dependent on the language mode of the module into which the code is being generated which isn't possible for us to know.

Failing that I think doing it based on the compiler version is reasonable.

glbrntt (approved):
Awesome, thanks Gus!

PR #2048, opened 2024-09-09
Start adopting swift-testing

glbrntt:
Motivation:

Modifications:

- Convert the MethodConfig coding tests
- Convert the Status tests
- Convert the Metadata tests
- Convert the Timeout
- Add a dependency on swift-testing on Linux (this can be removed when it's included as part of the toolchain)

Results:

Fewer bugs, better tests


glbrntt:
Closing in favour of separate PRs per test area.

PR #2049, opened 2024-09-09
Adopt swift-testing for status tests

glbrntt:
Motivation:

Modifications:

- Convert the Status tests
- Add a dependency on swift-testing on Linux (this can be removed when it's included as part of the toolchain)

Results:

Better tests

### DIFF
--- a/Tests/GRPCCoreTests/StatusTests.swift
+++ b/Tests/GRPCCoreTests/StatusTests.swift
@@ -13,105 +13,60 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 import GRPCCore
-import XCTest
+import Testing
 
-final class StatusTests: XCTestCase {
-  private static let statusCodeRawValue: [(Status.Code, Int)] = [
-    (.ok, 0),
-    (.cancelled, 1),
-    (.unknown, 2),
-    (.invalidArgument, 3),
-    (.deadlineExceeded, 4),
-    (.notFound, 5),
-    (.alreadyExists, 6),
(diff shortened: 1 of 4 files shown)


Tests/GRPCCoreTests/StatusTests.swift
```diff
@@ -13,105 +13,60 @@
...
+    @Test("Initialize from rawValue", arguments: zip(0 ... 16, Status.Code.all))
+    func initFromRawValue(rawValue: Int, expected: Status.Code) {
+      #expect(Status.Code(rawValue: rawValue) == expected)
     }
 
-    // API accepts `Int` so try invalid `Int` values too.
-    XCTAssertNil(Status.Code(rawValue: -1))
-    XCTAssertNil(Status.Code(rawValue: 1000))
-    XCTAssertNil(Status.Code(rawValue: .max))
+    @Test("Initialize from invalid rawValue", arguments: [-1, 17, 100, .max])
```
gjcairo, line 45: I'm not sure it's really important, but we're not testing for all values from 17 to UInt8.max anymore

gjcairo (approved):
LGTM, only left a small comment.
This is a lot nicer btw :D exciting times.

Tests/GRPCCoreTests/StatusTests.swift
glbrntt: Yeah, not sure it really matters tbh. The `UInt8` is an implementation detail so a bit odd to do all of those values given the parameter is an `Int`.

PR #2050, opened 2024-09-09
Adopt swift-testing for metadata tests

glbrntt:
Motivation:

Modifications:

- Convert the Metadata tests

Results:

Better tests


gjcairo (approved):
LGTM!

PR #2054, opened 2024-09-11
Add NIOTS client transport

gjcairo:
This PR adds a `NIOTransportServices` H2 client transport to gRPC v2.

### DIFF
--- a/Sources/GRPCHTTP2TransportNIOTransportServices/HTTP2ClientTransport+TransportServices.swift
+++ b/Sources/GRPCHTTP2TransportNIOTransportServices/HTTP2ClientTransport+TransportServices.swift
@@ -18,7 +18,7 @@
 public import GRPCCore
 public import GRPCHTTP2Core
 public import NIOTransportServices  // has to be public because of default argument value in init
-public import NIOCore // has to be public because of EventLoopGroup param in init
+public import NIOCore  // has to be public because of EventLoopGroup param in init
 
 private import Network
 
--- a/Sources/GRPCHTTP2TransportNIOTransportServices/HTTP2ClientTransport+TransportServices.swift
+++ b/Sources/GRPCHTTP2TransportNIOTransportServices/HTTP2ClientTransport+TransportServices.swift
@@ -14,29 +14,28 @@
  * limitations under the License.
  */
 
+#if canImport(Network)
 public import GRPCCore
-public import GRPCHTTP2Core  // should be @usableFromInline
+public import GRPCHTTP2Core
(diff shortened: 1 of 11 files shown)


Sources/GRPCHTTP2TransportNIOTransportServices/HTTP2ClientTransport+TransportServices.swift
```diff
@@ -0,0 +1,339 @@
...
+public import NIOCore  // has to be public because of EventLoopGroup param in init
+
+private import Network
+
+@available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *)
+extension HTTP2ClientTransport {
+  /// A ``GRPCCore/ClientTransport`` using HTTP/2 built on top of `NIOTransportServices`.
+  ///
+  /// This transport builds on top of SwiftNIO's Transport Services networking layer and is the recommended
+  /// variant for use for use on Darwin-based platforms **only** (macOS, iOS, etc.).
```
glbrntt, line 30: It would be hard to recommend it anywhere else since it won't exist on other platforms 😃

glbrntt (changes requested):
This all looks really good! There one additional thing we should do in this PR though which is allowing the `HTTP2TransportTests` to use the NIOTS client transport.

Sources/GRPCHTTP2TransportNIOTransportServices/HTTP2ClientTransport+TransportServices.swift
gjcairo: Haha yeah my point with the only was to make it clear it's only available on Darwin platforms, but maybe it doesn't clarify as I expected it to. I'll get rid of it.

Tests/GRPCHTTP2TransportTests/HTTP2TransportTests.swift
```diff
@@ -1421,7 +1434,7 @@ final class HTTP2TransportTests: XCTestCase {
 extension [HTTP2TransportTests.Transport] {
   static let supported = [
     HTTP2TransportTests.Transport(server: .posix, client: .posix),
-    HTTP2TransportTests.Transport(server: .niots, client: .posix),
+    HTTP2TransportTests.Transport(server: .niots, client: .niots),
```
glbrntt, line 1437: We shouldn't remove the other one: we should do all four combinations

glbrntt (changes requested):

PR #2055, opened 2024-09-11
Adopt swift-testing for MethodConfig tests

glbrntt:
Motivation:

Modifications:

- Convert the MethodConfig coding tests

Results:

Fewer bugs, better tests


gjcairo (commented):
Overall this looks good! I do wonder though whether having input files instead of the JSON in-line at each test case makes for tidier or harder to read test cases (as you'll have to move back and forth). Any reason in particular you moved them to input files?

glbrntt:
> Overall this looks good! I do wonder though whether having input files instead of the JSON in-line at each test case makes for tidier or harder to read test cases (as you'll have to move back and forth). Any reason in particular you moved them to input files?

Yeah, so I only moved _some_ of them. Inline is easier to follow if it's small. Once they get more than a handful of lines I found them a but unwieldy and writing them inline is a bit more a pain and error prone (you need to make sure you escape everything correctly etc.)

gjcairo (approved):
That makes sense - thanks George!

PR #2058, opened 2024-09-16
Factor out shared http2 server transport code

glbrntt:
Motivation:

The NIOPosix and NIOTS HTTP/2 server transports have quite a lot of shared code. The only meaningfull difference between them is how they create their channels.

Modification:

- Factor out the shared parts to a `CommonHTTP2ServerTransport`
- Allow each tranport to inject a factory for creating a listener

Result:

Less duplication


gjcairo (approved):
Nice one

PR #2059, opened 2024-09-16
Propagate context to server RPC handlers

glbrntt:
Motivation:

The interceptors API has a context which, at the moment, only includes
the name of the RPC. This information is generally useful so should be
propagated to the server handler too. Information on the context can
also be extended later to include things like the identity of the remote
peer, or info about the state of the RPC.

Modifications:

- Rename 'ServerInterceptorContext' to 'ServerContext'
- Make the transport the source of the context and have that provide it
  to the 'listen' method. Propagate this through the server stack to the
  generated stubs.
- Update code generator to include the context
- Update generated code
- Update services

Results:

RPC handlers have a context provided by a transport


glbrntt:
Split into two commits:
- [359c6c3](https://github.com/grpc/grpc-swift/pull/2059/commits/359c6c33263bf06981d9b1ff6445699ed3c78a33) contains the bulk of the changes
- [0e8b637](https://github.com/grpc/grpc-swift/pull/2059/commits/0e8b6374ab6d3b7285357441ac521a65ff2a6f86) is the result of re-running the code generator

Tests/GRPCInProcessTransportTests/InProcessClientTransportTests.swift
```diff
@@ -171,7 +171,7 @@ final class InProcessClientTransportTests: XCTestCase {
       }
 
       group.addTask {
-        try await server.listen { stream in
```
gjcairo, line 174: Aren't we getting a warning because we're not using `context` here? Should we `_` it, or should we assert the contained method descriptor?

Tests/GRPCInProcessTransportTests/InProcessServerTransportTests.swift
```diff
@@ -41,7 +41,7 @@ final class InProcessServerTransportTests: XCTestCase {
 
     try await withThrowingTaskGroup(of: Void.self) { group in
       group.addTask {
-        try await transport.listen { stream in
```
gjcairo, line 44: Aren't we getting a warning because we're not using `context` here? Should we `_` it, or should we assert the contained method descriptor?

Tests/GRPCInProcessTransportTests/InProcessServerTransportTests.swift
```diff
@@ -71,7 +71,7 @@ final class InProcessServerTransportTests: XCTestCase {
 
     try transport.acceptStream(firstStream)
 
-    try await transport.listen { stream in
```
gjcairo, line 74: Aren't we getting a warning because we're not using `context` here? Should we `_` it, or should we assert the contained method descriptor?

Tests/GRPCInProcessTransportTests/InProcessClientTransportTests.swift
glbrntt: Nope, no warning for not using it. I don't mind whether we change it to `_` or leave it as `context`. Let me know what you prefer.

I'll add a separate test checking we get the descriptor in the context.

gjcairo: Ah, that's fine, I'm okay with the name. I thought warnings were emitted in this case :D

gjcairo (approved):

PR #2063, opened 2024-09-18
Standardise some naming

glbrntt:
Motivation:

In v2 we've predominantly used 'max' instead of 'maximum' and 'config' instead of 'configuration'. We should apply these consistently.

Modifications:

- Replaces occurrences of 'maximum' with 'max', and 'configuration' with 'config'.

Result:

More consistent naming

### DIFF
--- a/Sources/GRPCCodeGen/Internal/Translator/SpecializedTranslator.swift
+++ b/Sources/GRPCCodeGen/Internal/Translator/SpecializedTranslator.swift
@@ -19,7 +19,7 @@
 protocol SpecializedTranslator {
 
   /// The ``SourceGenerator.Configuration.AccessLevel`` object used to represent the visibility level used in the generated code.
-  var accessLevel: SourceGenerator.Configuration.AccessLevel { get }
+  var accessLevel: SourceGenerator.Config.AccessLevel { get }
 
   /// Generates an array of ``CodeBlock`` elements that will be part of the ``StructuredSwiftRepresentation`` object
   /// created by the ``Translator``.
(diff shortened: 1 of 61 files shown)


Sources/GRPCCodeGen/Internal/Translator/SpecializedTranslator.swift
```diff
@@ -19,7 +19,7 @@
 protocol SpecializedTranslator {
 
   /// The ``SourceGenerator.Configuration.AccessLevel`` object used to represent the visibility level used in the generated code.
```
gjcairo, line 21: Docs are outdated here. Would you mind trying to build docs to see which other references may have broken?

gjcairo (changes requested):

Sources/GRPCCodeGen/Internal/Translator/SpecializedTranslator.swift
glbrntt: Yes, good catch. We should add a docc CI check as well.

glbrntt: Ah, these were missed because DocC by default only checks public symbols. You can change that to internal if you use the docc-plugin (e.g. `swift package generate-documentation --target GRPCCore --symbol-graph-minimum-access-level internal`).

This surfaced a fair few issues with internal docs. I fixed a few, and ones introduced by this change, but I think we should tackle the rest separately.

gjcairo (approved):

PR #2066, opened 2024-09-20
Filter down main to v2

glbrntt:
Motivation:

v2 will be split across a number of packages, and v1 will be maintained on the 'release/1.x' branch. In order to do the package split this package needs to be whittled down to only the v2 components which will remain within this package.

Modifications:

- Remove v1 code
- Remove v2 code destined for other packages
- Reshuffle a few scripts

Result:

- Package contains only the intended code


FranzBusch (approved):
Looked at each line and LGTM! Happy Friday :)

PR #2067, opened 2024-09-23
Make examples standalone packages

glbrntt:
Motivation:

Standalone packages for examples are an easier on-ramp for newcomers as
they only see the details they care about and it can act as a starting
point which they can develop from.

Modifications:

- Turn each example into a standalone package and add a README to each
- Update the generate and fetch scripts and re-run them

Result:

Better examples


Examples/route-guide/README.md
```diff
@@ -0,0 +1,44 @@
+# Route Guide
+
+This example demonstrates all four RPC types using a 'Route Guide' service and
+client.
+
+## Overview
+
+A "rouge-guide" command line tool that uses generated stubs for an 'Route Guide'
```
gjcairo, line 8:
```suggestion
A "route-guide" command line tool that uses generated stubs for a 'Route Guide'
```

gjcairo (approved):
LGTM! Just a couple of typos

PR #2068, opened 2024-09-23
Add server stream events

glbrntt:
Motivation:

As a service author it's useful to know if the underlying stream has been cancelled (because it's timed out, the remote peer closed it, the connection dropped etc).

For cases where the stream has already closed this can be surfaced by a read or write failing. However, for cases like server-streaming RPCs where there are no reads and writes can be infrequent it's useful to have a more explicit signal.

Modifications:

- Add a `ServerStreamState` which provides two API for this:
  - An explicit "has this RPC been cancelled?"
  - An async sequence of events which can happen the underlying stream
- Add the stream state to the server context
- Update the in-process transport to use this

Result:

Server context includes a way for users to know if their RPC has been cancelled.

### DIFF
--- a/Sources/GRPCCore/Call/Server/ServerStreamState.swift
+++ b/Sources/GRPCCore/Call/Server/ServerStreamState.swift
@@ -1,0 +1,213 @@
+/*
+ * Copyright 2024, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


Sources/GRPCCore/Call/Server/ServerStreamState.swift
```diff
@@ -0,0 +1,213 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Synchronization
+
+@available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *)
+public struct ServerStreamState: Sendable {
```
glbrntt, line 20: For discussion: I don't like this name. 

The other idea I had was `ServerStreamContext` which is potentially better and aligns with `ServerContext`. Downside is that it hints at a slightly wide use.

Another option: rather than nesting `Events` have a `ServerStreamEvents` type which also has a computed property for `isRPCCancelled` (i.e. an `AsyncSequence` with extra API).

glbrntt:
Closing in favour of #2090

PR #2071, opened 2024-09-24
Restructure in-process transport

clintonpi:
Motivation:

To create a connected pair of `ServerTransport` and `ClientTransport`, users need to call `InProcessTransport.makePair(serviceConfig:)` which returns a tuple. The spelling of this API can be more intuitive and simple.

Modifications:

- Restructure `InProcessTransport` to include two properties: `server` (the `ServerTransport`) and `client` (the `ClientTransport`), and an `init(serviceConfig:)` that pairs these properties.
- Add missing GRPCInProcessTransport dependency to GRPCCore test target.

Result:

The spelling of the `InProcessTransport` API will be better.


Sources/GRPCCore/GRPCClient.swift
```diff
@@ -65,8 +65,8 @@ private import Synchronization
 /// )
 ///
 /// // Finally create a transport and instantiate the client, adding an interceptor.
-/// let inProcessServerTransport = InProcessServerTransport()
-/// let inProcessClientTransport = InProcessClientTransport(serverTransport: inProcessServerTransport)
+/// let inProcessServerTransport = InProcessTransport.Server()
+/// let inProcessClientTransport = InProcessTransport.Client(serverTransport: inProcessServerTransport)
```
glbrntt, line 69: This one was wrong before your changes: can you update it?

Sources/GRPCCore/GRPCServer.swift
```diff
@@ -35,7 +35,7 @@ private import Synchronization
 ///
 /// ```swift
 /// // Create and an in-process transport.
-/// let inProcessTransport = InProcessServerTransport()
+/// let inProcessTransport = InProcessTransport.Server()
```
glbrntt, line 38: Same here, this is also wrong, can you update this one?

Sources/GRPCInProcessTransport/Client.swift
```diff
@@ -0,0 +1,359 @@
+/*
```
glbrntt, line 1: Can you rename this file? Normally when you extend a type you name the file `<ThingBeingExtended>+<WhatItsExtendedWith>.swift` so in this case `InProcessTransport+Client.swift`. 

This makes things much easier to find: if you're looking for the in-process client transport then the obvious thing to search for is "InProcessTransport" because that's much more specific than "Client".

Sources/GRPCInProcessTransport/InProcessTransport.swift
```diff
@@ -16,24 +16,16 @@
...
 public import GRPCCore
 
-public enum InProcessTransport {
-  /// Returns a pair containing an ``InProcessServerTransport`` and an ``InProcessClientTransport``.
-  ///
-  /// This function is purely for convenience and does no more than constructing a server transport
-  /// and a client using that server transport.
+@available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *)
+public struct InProcessTransport: Sendable {
+  public let server = Self.Server()
```
glbrntt, line 21: nit: initialize this in the `init`

```suggestion
  public let server: Self.Server
```

Sources/GRPCInProcessTransport/InProcessTransport.swift
```diff
@@ -16,24 +16,16 @@
...
+  public let client: Self.Client
+
+  /// Initializes a new ``InProcessTransport`` pairing a ``ServerTransport`` and a ``ClientTransport``.
```
glbrntt, line 24:
```suggestion
  /// Initializes a new ``InProcessTransport`` pairing a ``Client`` and a ``Server``.
```

Sources/GRPCInProcessTransport/Server.swift
```diff
@@ -0,0 +1,83 @@
+/*
```
glbrntt, line 1: Can you rename this file to be `InProcessTransport+Server.swift`?

glbrntt (changes requested):
Mostly looks fine, left a few comments which need resolving.

glbrntt:
> Can you remove [60cc0bd](https://github.com/grpc/grpc-swift/pull/2071/commits/60cc0bd8e8905bc00bee7c195c49a4d5aa05a029)? It's unrelated to the changes made in this PR so shouldn't be included here (it was fixed by #2073)

Nevermind, merging from main effectively undid this.

glbrntt (approved):

PR #2072, opened 2024-09-24
Add an underscore between generated service name and component

glbrntt:
Motivation:

A 'bar' package defining a 'FooService' service will have a number of types generateed for it, for example, `Bar_FooServiceServiceProtocol`, or `Bar_FooServiceClient`. The protocol buffers style guide reccomends that services are suffixed with "Service" (e.g. "FooService", not "Foo") which results in some strange protocol names (like `Bar_FooServiceServiceProtocol`).

There should be a clearer differentiation between the qualified service name ("bar.FooServce") and the type being generated. The code generator already replaces "." with "_" in fully qualified names, so "_" is a natural choice as a separator.

Modifications:

- Add an "_" between the generated fully qualified service name and the gRPC component being generated
- Update tests

Result:

Better names in generated code.

The `Bar_FooServiceServiceProtocol` example above becomes `Bar_FooService_ServiceProtocol`, while `Bar_FooServiceClient` becomes `Bar_FooService_Client`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -65,6 +65,7 @@ let targets: [Target] = [
     name: "GRPCCoreTests",
     dependencies: [
       .target(name: "GRPCCore"),
+      .target(name: "GRPCInProcessTransport"),
       .product(name: "SwiftProtobuf", package: "swift-protobuf")
     ],
     resources: [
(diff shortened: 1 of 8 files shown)


gjcairo (approved):
LGTM - thanks George! This reads much more clearly.

A super nit for posterity, not really about the change itself: the underscore in the PR description is being rendered as markdown/italics 😅 can you escape it so it reads better if someone comes looking in the future?

PR #2083, opened 2024-10-06
Flatten `RPCError` causes if they're also `RPCError`s with the same code

gjcairo:
## Motivation
For errors happening deep in the task tree, we'd wrap them in many layers of `RPCError`s. This isn't particularly nice.

## Modifications
This PR changes the behaviour of the `RPCError` initialiser to flatten the cause as long as it's an `RPCError` with the same status code as the wrapping error.

## Result
Friendlier errors.

### DIFF
--- a/Sources/GRPCCore/RPCError.swift
+++ b/Sources/GRPCCore/RPCError.swift
@@ -49,13 +49,13 @@ public struct RPCError: Sendable, Hashable, Error {
     metadata: Metadata = [:],
     cause: (any Error)? = nil
   ) {
-    if let rpcErrorCause = cause as? RPCError, rpcErrorCause.code == code {
-      self.code = code
-      self.message = message + " \(rpcErrorCause.message)"
-      var mergedMetadata = metadata
-      mergedMetadata.add(contentsOf: rpcErrorCause.metadata)
-      self.metadata = mergedMetadata
-      self.cause = rpcErrorCause.cause
+    if let rpcErrorCause = cause as? RPCError {
+      self = .init(
+        code: code,
+        message: message,
+        metadata: metadata,
+        cause: rpcErrorCause
+      )
     } else {
       self.code = code
       self.message = message
--- a/Sources/GRPCCore/RPCError.swift
+++ b/Sources/GRPCCore/RPCError.swift
@@ -58,7 +58,7 @@ public struct RPCError: Sendable, Hashable, Error {
       while let nextRPCErrorCause = nextCause as? RPCError {
         if code == nextRPCErrorCause.code {
           finalMessage = finalMessage + " \(nextRPCErrorCause.message)"
-          finalMetadata.merge(nextRPCErrorCause.metadata)
+          finalMetadata.add(contentsOf: nextRPCErrorCause.metadata)
           nextCause = nextRPCErrorCause.cause
         } else {
           nextCause = RPCError(
--- a/Sources/GRPCCore/RPCError.swift
+++ b/Sources/GRPCCore/RPCError.swift
@@ -35,47 +35,56 @@ public struct RPCError: Sendable, Hashable, Error {
   /// The original error which led to this error being thrown.
   public var cause: (any Error)?
 
-  /// Create a new RPC error.
+  /// Create a new RPC error. If the given `cause` is also an ``RPCError`` sharing the same `code`,
+  /// then they will be flattened into a single error, by merging the messages and metadata.
   ///
   /// - Parameters:
   ///   - code: The status code.
   ///   - message: A message providing additional context about the code.
   ///   - metadata: Any metadata to attach to the error.
   ///   - cause: An underlying error which led to this error being thrown.
-  ///   - flatteningCauses: Whether to flatten the `causes` as long as they're ``RPCError``s
-  ///   sharing the same ``Code-swift.struct``.
   public init(
     code: Code,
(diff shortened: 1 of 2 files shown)


Tests/GRPCCoreTests/RPCErrorTests.swift
gjcairo, line 1: I also migrated this to swift-testing since I was adding new tests

Tests/GRPCCoreTests/RPCErrorTests.swift
```diff
@@ -14,114 +14,144 @@
...
+    (.internalError, 13),
+    (.unavailable, 14),
+    (.dataLoss, 15),
+    (.unauthenticated, 16),
+  ])
+  func testStatusCodeRawValues(statusCode: RPCError.Code, rawValue: Int) {
+    #expect(statusCode.rawValue == rawValue, "\(statusCode) had unexpected raw value")
+  }
+
+  @Test("Flatten causes with same status code")
```
gjcairo, line 144: This is the only new test.

Tests/GRPCCoreTests/RPCErrorTests.swift
```diff
@@ -14,114 +14,144 @@
...
-    XCTAssertNil(RPCError(status: status))
+    #expect(RPCError(status: status) == nil)
 
     status.code = .invalidArgument
-    var error = try XCTUnwrap(RPCError(status: status))
-    XCTAssertEqual(error.code, .invalidArgument)
-    XCTAssertEqual(error.message, "")
-    XCTAssertEqual(error.metadata, [:])
+    var error = RPCError(status: status)
+    try #require(error != nil)
```
glbrntt, line 43: Is it possible to do `var error = try #require(RPCError(status: status))` so we can avoid the `!` below?

glbrntt (changes requested):

Sources/GRPCCore/RPCError.swift
```diff
@@ -42,11 +42,46 @@ public struct RPCError: Sendable, Hashable, Error {
...
+    message: String,
+    metadata: Metadata = [:],
+    cause: (any Error)? = nil,
+    flatteningCauses: Bool = false
+  ) {
+    if flatteningCauses {
+      var finalMessage = message
+      var finalMetadata = metadata
+      var nextCause = cause
+      while let nextRPCErrorCause = nextCause as? RPCError {
```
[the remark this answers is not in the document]
gjcairo: I can't really think of a reason why we wouldn't want to do this unconditionally other than flattening may potentially be expensive since we're creating new errors and dealing with strings.

If we do it unconditionally, then the loop would be unnecessary because all underlying causes would have been flattened already anyways.

And yeah good idea about the overload.

Tests/GRPCCoreTests/RPCErrorTests.swift
gjcairo: Oh yeah! I swear I thought I'd tried this - will fix

Sources/GRPCCore/RPCError.swift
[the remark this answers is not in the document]
glbrntt: > I can't really think of a reason why we wouldn't want to do this unconditionally other than flattening may potentially be expensive since we're creating new errors and dealing with strings.

We're already creating a new `RPCError` so that part isn't an issue. Strings a potentially expensive but that's the tradeoff for more info I guess. 

> If we do it unconditionally, then the loop would be unnecessary because all underlying causes would have been flattened already anyways.

Correct.

We can always add an opt-out later if we think it's necessary.

Sources/GRPCCore/RPCError.swift
```diff
@@ -35,18 +35,62 @@ public struct RPCError: Sendable, Hashable, Error {
...
+    cause: (any Error)? = nil
+  ) {
+    if let rpcErrorCause = cause as? RPCError, rpcErrorCause.code == code {
+      self.code = code
+      self.message = message + " \(rpcErrorCause.message)"
+      var mergedMetadata = metadata
+      mergedMetadata.add(contentsOf: rpcErrorCause.metadata)
+      self.metadata = mergedMetadata
+      self.cause = rpcErrorCause.cause
+    } else {
```
glbrntt, line 59: Can't we just use the other `init` here rather than duplicating this?
```swift
    if let rpcErrorCause = cause as? RPCError {
      self = .init(...)
    } else {
```

Tests/GRPCCoreTests/RPCErrorTests.swift
```diff
@@ -14,114 +14,151 @@
...
+      (.dataLoss, 15),
+      (.unauthenticated, 16),
+    ]
+  )
+  func testStatusCodeRawValues(statusCode: RPCError.Code, rawValue: Int) {
+    #expect(statusCode.rawValue == rawValue, "\(statusCode) had unexpected raw value")
+  }
+
+  @Test("Flatten causes with same status code")
+  func testFlattenCausesWithSameStatusCode() {
```
glbrntt, line 143: This one needs updating now that `flatteningCauses` doesn't exist.

It'd also be good to have a test which verifies a cause with a different status code is given isn't flattened

glbrntt (changes requested):

glbrntt (approved):
LGTM, thanks Gus!

PR #2084, opened 2024-10-06
Add a merge method to Metadata

gjcairo:
## Motivation
We want to be able to flatten `RPCError`s, and to do so we need to be able to merge the `Metadata` contained in each.

## Modifications
This PR adds a helper function to merge one `Metadata` instance into another.

## Result
Unblocks https://github.com/grpc/grpc-swift/pull/2083 and also provides a potentially useful API for users.

**- Note:** Because of the way `Metadata` has been implemented, we can have multiple _identical_ key-value pairs. This isn't ideal, as it's particularly feasible that we'll end up with multiple repeated identical pairs when merging two `Metadata`s. I think we should reconsider the backing data structure (using a set for example) or add a check before inserting to avoid this.

### DIFF
--- a/Sources/GRPCCore/Metadata.swift
+++ b/Sources/GRPCCore/Metadata.swift
@@ -170,6 +170,15 @@ public struct Metadata: Sendable, Hashable {
   internal mutating func addValue(_ value: Value, forKey key: String) {
     self.elements.append(.init(key: key, value: value))
   }
+  
+  /// Merge another instance of `Metadata` into this one.
+  ///
+  /// - Parameter other: the `Metadata` which key-value pairs should be merged into this one.
+  public mutating func merge(_ other: Metadata) {
+    for (key, value) in other {
+      self.addValue(value, forKey: key)
+    }
+  }
 
   /// Removes all values associated with the given key.
   ///
--- a/Sources/GRPCCore/Metadata.swift
+++ b/Sources/GRPCCore/Metadata.swift
@@ -171,13 +171,11 @@ public struct Metadata: Sendable, Hashable {
     self.elements.append(.init(key: key, value: value))
   }
   
-  /// Merge another instance of `Metadata` into this one.
+  /// Add the contents of a `Sequence` of key-value pairs to this `Metadata` instance.
   ///
-  /// - Parameter other: the `Metadata` which key-value pairs should be merged into this one.
-  public mutating func merge(_ other: Metadata) {
-    for (key, value) in other {
-      self.addValue(value, forKey: key)
-    }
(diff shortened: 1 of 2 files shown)


Tests/GRPCCoreTests/MetadataTests.swift
```diff
@@ -252,4 +252,52 @@ struct MetadataTests {
...
+      self.otherMetadata.addString("value1-1", forKey: "key1")
+      self.metadata.merge(otherMetadata)
+
+      #expect(self.metadata == [
+        "key1": "value1-1",
+        "key2": "value2",
+        "key3": "value3",
+        "key4": "value4",
+        "key5": "value5",
+        "key1": "value1-1",
```
gjcairo, line 294: I don't think this pair should actually be here, but this is a result of how `Metadata` was implemented (see the note in the PR description). I think we should consider changing this.

glbrntt: I don't think we should: there's no requirement for key/value pairs of metadata to be unique.

I think `merge` is a misleading name here (you can merge values into a dictionary because it _does_ have a requirement that keys are unique). In this case a more appropriate spelling would be `add(contentsOf metadata: Metadata)` (we should also add an `add(contentsOf other: some Sequence<Element>)` while we're at it).

Sources/GRPCCore/Metadata.swift
```diff
@@ -170,6 +170,15 @@ public struct Metadata: Sendable, Hashable {
   internal mutating func addValue(_ value: Value, forKey key: String) {
     self.elements.append(.init(key: key, value: value))
   }
+  
+  /// Merge another instance of `Metadata` into this one.
+  ///
+  /// - Parameter other: the `Metadata` which key-value pairs should be merged into this one.
+  public mutating func merge(_ other: Metadata) {
```
glbrntt, line 177: We can do `self.elements.append(contentsOf: other.elements)` which will reserve enough capacity for the new elements.

glbrntt (changes requested):

Tests/GRPCCoreTests/MetadataTests.swift
gjcairo: The `some Sequence<Element>` variant is enough to be able to pass a `Metadata`. Should we still have the `Metadata` overload for optimisation purposes or is it unnecessary?

gjcairo: Indeed it saves an allocation, I've added it.

glbrntt (approved):
Awesome, thanks Gus

PR #2096, opened 2024-10-16
Allow adding `ServerInterceptor`s to specific services and methods

gjcairo:
## Motivation
We want to allow users to customise the RPCs a registered interceptor should apply to on the server:
- Intercept all requests
- Intercept requests only meant for specific services
- Intercept requests only meant for specific methods

## Modifications
This PR adds a new `ServerInterceptorTarget` type that allows users to specify what the target of the interceptor should be.
Existing APIs accepting `[any ServerInterceptor]` have been changed to instead take `[ServerInterceptorTarget]`.

## Result
Users can have more control over to which requests interceptors are applied.

### DIFF
--- a/Sources/GRPCCore/Call/Server/RPCRouter.swift
+++ b/Sources/GRPCCore/Call/Server/RPCRouter.swift
@@ -136,18 +136,6 @@ public struct RPCRouter: Sendable {
     self.handlers[descriptor] = (handler, [])
   }
 
-  @inlinable
-  public mutating func registerInterceptors(
-    pipeline: [ServerInterceptorOperation]
-  ) {
-    for descriptor in self.handlers.keys {
-      let applicableOperations = pipeline.filter { $0.applies(to: descriptor) }
-      if !applicableOperations.isEmpty {
-        self.handlers[descriptor]?.interceptors = applicableOperations.map { $0.interceptor }
-      }
-    }
-  }
-
   /// Removes any handler registered for the specified method.
   ///
   /// - Parameter descriptor: A descriptor of the method to remove a handler for.
@@ -155,6 +143,16 @@ public struct RPCRouter: Sendable {
   @discardableResult
   public mutating func removeHandler(forMethod descriptor: MethodDescriptor) -> Bool {
     return self.handlers.removeValue(forKey: descriptor) != nil
+  }
+
+  @inlinable
+  mutating func registerInterceptors(pipeline: [ServerInterceptorOperation]) {
+    for descriptor in self.handlers.keys {
+      let applicableOperations = pipeline.filter { $0._applies(to: descriptor) }
+      if !applicableOperations.isEmpty {
+        self.handlers[descriptor]?.interceptors = applicableOperations.map { $0.interceptor }
+      }
+    }
   }
 }
 
--- a/Sources/GRPCCore/Call/Server/RPCRouter.swift
+++ b/Sources/GRPCCore/Call/Server/RPCRouter.swift
@@ -43,7 +43,7 @@ public struct RPCRouter: Sendable {
           RPCWriter<RPCResponsePart>.Closable
         >,
         _ context: ServerContext,
-        _ interceptors: [ServerInterceptorTarget]
+        _ interceptors: [any ServerInterceptor]
       ) async -> Void
 
     @inlinable
@@ -75,7 +75,7 @@ public struct RPCRouter: Sendable {
         RPCWriter<RPCResponsePart>.Closable
       >,
       context: ServerContext,
-      interceptors: [ServerInterceptorTarget]
+      interceptors: [any ServerInterceptor]
     ) async {
       await self._fn(stream, context, interceptors)
     }
@@ -151,7 +151,7 @@ extension RPCRouter {
       RPCWriter<RPCResponsePart>.Closable
     >,
     context: ServerContext,
-    interceptors: [ServerInterceptorTarget]
+    interceptors: [any ServerInterceptor]
   ) async {
     if let handler = self.handlers[stream.descriptor] {
       await handler.handle(stream: stream, context: context, interceptors: interceptors)
--- a/Sources/GRPCCore/Call/Server/RPCRouter.swift
+++ b/Sources/GRPCCore/Call/Server/RPCRouter.swift
@@ -22,6 +22,8 @@
 /// the router has a handler for a method with ``hasHandler(forMethod:)`` or get a list of all
 /// methods with handlers registered by calling ``methods``. You can also remove the handler for a
 /// given method by calling ``removeHandler(forMethod:)``.
+/// You can also register any interceptors that you want applied to registered handlers via the
+/// ``registerInterceptors(pipeline:)`` method.
 ///
 /// In most cases you won't need to interact with the router directly. Instead you should register
 /// your services with ``GRPCServer/init(transport:services:interceptors:)`` which will in turn
@@ -145,10 +147,19 @@ public struct RPCRouter: Sendable {
     return self.handlers.removeValue(forKey: descriptor) != nil
   }
 
+  /// Registers applicable interceptors to all currently-registered handlers.
+  ///
+  /// - Important: Calling this method will apply the interceptors only to existing handlers. Any handlers registered via
+  ///  ``registerHandler(forMethod:deserializer:serializer:handler:)`` _after_ calling this method will not have
+  ///    any interceptors applied to them. If you want to make sure all registered methods have any applicable interceptors applied,
+  ///    only call this method _after_ you have registered all handlers.
+  /// - Parameter pipeline: The interceptor pipeline operations to register to all currently-registered handlers. The order of the
+  ///  interceptors matters.
+  /// - SeeAlso: ``ServerInterceptorPipelineOperation``.
   @inlinable
-  mutating func registerInterceptors(pipeline: [ServerInterceptorOperation]) {
+  public mutating func registerInterceptors(pipeline: [ServerInterceptorPipelineOperation]) {
     for descriptor in self.handlers.keys {
-      let applicableOperations = pipeline.filter { $0._applies(to: descriptor) }
+      let applicableOperations = pipeline.filter { $0.applies(to: descriptor) }
       if !applicableOperations.isEmpty {
         self.handlers[descriptor]?.interceptors = applicableOperations.map { $0.interceptor }
       }
--- a/Sources/GRPCCore/Call/Server/RPCRouter.swift
+++ b/Sources/GRPCCore/Call/Server/RPCRouter.swift
@@ -82,7 +82,8 @@ public struct RPCRouter: Sendable {
   }
 
   @usableFromInline
-  private(set) var handlers: [MethodDescriptor: (handler: RPCHandler, interceptors: [any ServerInterceptor])]
+  private(set) var handlers:
+    [MethodDescriptor: (handler: RPCHandler, interceptors: [any ServerInterceptor])]
 
   /// Creates a new router with no methods registered.
   public init() {
--- a/Sources/GRPCCore/Call/Server/RPCRouter.swift
+++ b/Sources/GRPCCore/Call/Server/RPCRouter.swift
@@ -43,7 +43,7 @@ public struct RPCRouter: Sendable {
           RPCWriter<RPCResponsePart>.Closable
         >,
         _ context: ServerContext,
-        _ interceptors: [any ServerInterceptor]
+        _ interceptors: [ServerInterceptorTarget]
       ) async -> Void
 
     @inlinable
@@ -75,7 +75,7 @@ public struct RPCRouter: Sendable {
         RPCWriter<RPCResponsePart>.Closable
       >,
       context: ServerContext,
-      interceptors: [any ServerInterceptor]
+      interceptors: [ServerInterceptorTarget]
     ) async {
       await self._fn(stream, context, interceptors)
     }
@@ -151,7 +151,7 @@ extension RPCRouter {
       RPCWriter<RPCResponsePart>.Closable
     >,
     context: ServerContext,
-    interceptors: [any ServerInterceptor]
+    interceptors: [ServerInterceptorTarget]
   ) async {
     if let handler = self.handlers[stream.descriptor] {
       await handler.handle(stream: stream, context: context, interceptors: interceptors)
--- a/Sources/GRPCCore/Call/Server/RPCRouter.swift
+++ b/Sources/GRPCCore/Call/Server/RPCRouter.swift
@@ -82,7 +82,7 @@ public struct RPCRouter: Sendable {
   }
 
   @usableFromInline
-  private(set) var handlers: [MethodDescriptor: RPCHandler]
+  private(set) var handlers: [MethodDescriptor: (handler: RPCHandler, interceptors: [any ServerInterceptor])]
 
   /// Creates a new router with no methods registered.
   public init() {
@@ -126,12 +126,25 @@ public struct RPCRouter: Sendable {
       _ context: ServerContext
     ) async throws -> StreamingServerResponse<Output>
   ) {
-    self.handlers[descriptor] = RPCHandler(
+    let handler = RPCHandler(
       method: descriptor,
       deserializer: deserializer,
       serializer: serializer,
       handler: handler
     )
+    self.handlers[descriptor] = (handler, [])
+  }
+
+  @inlinable
+  public mutating func registerInterceptors(
+    pipeline: [ServerInterceptorOperation]
+  ) {
+    for descriptor in self.handlers.keys {
+      let applicableOperations = pipeline.filter { $0.applies(to: descriptor) }
+      if !applicableOperations.isEmpty {
+        self.handlers[descriptor]?.interceptors = applicableOperations.map { $0.interceptor }
+      }
+    }
   }
 
   /// Removes any handler registered for the specified method.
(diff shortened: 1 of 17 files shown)


glbrntt (changes requested):
Thanks Gus – I think this is broadly the right approach and I like how the user has full control over the ordering of global/service/method specific interceptors. I have a few pieces of higher level feedback:

- I don't think we should propagate the `ServerInterceptorTarget` right down into the server executor. It couples the surface API (i.e. configuration) with the internals more than is necessary, so rather than pushing that all the way down I think the server should pre-compute the array of interceptors for each method and pass that down to the executor (instead of the array of targets). This also means we can avoid working out which interceptors to apply when each interceptor is called (which is potentially quite expensive).
- I think there's value in making the common case (interceptors apply to all services) as easy as it currently is, so the server should likely have an overload for the `interceptors` (i.e. one should take `[any ServerInterceptor]` and apply to everything) and the other variant should all users to customise the setup a bit more.

I think it might also make sense to not think of "targets" but operations on an interceptor pipeline. It could be something like:

```swift
let server = GRPCServer(
  ...
  interceptorPipeline: [
    // Add to all
    .add(LoggingInterceptor()),
    // Add to foo service
    .add(AuthInterceptor(), services: [foo]),
    // Add to foo service _and_ the bar method 
    .add(Blah(), services: [foo], methods: [bar]),
    // Add to the bar method
    .add(BlahBlah(), methods: [bar])
  ]
)
```

This also lets us leave open the door to adding operations that _don't_ add an interceptor for a particular service/method as well. (Maybe "include" is better terminology here than "add"?)

gjcairo:
Thanks for the review George, you bring up good points. A couple of things:

> I don't think we should propagate the ServerInterceptorTarget right down into the server executor.

Yeah, I went a bit back and forth on this one. I had originally precomputed the array of interceptors, but it felt like a lot of unnecessary new arrays being passed around. However I'm not actually sure performance would be affected - I didn't instrument it before making the change. I'm happy to make the change.

> I think there's value in making the common case (interceptors apply to all services) as easy as it currently is

Agreed, good idea.

Also noted about naming. I'll give it a bit more thought.

glbrntt:
> Thanks for the review George, you bring up good points. A couple of things:
> 
> > I don't think we should propagate the ServerInterceptorTarget right down into the server executor.
> 
> Yeah, I went a bit back and forth on this one. I had originally precomputed the array of interceptors, but it felt like a lot of unnecessary new arrays being passed around. However I'm not actually sure performance would be affected - I didn't instrument it before making the change. I'm happy to make the change.

It's potentially quite expensive to do the filtering (depends on how many interceptors are registered and the rules for which services/methods they apply to), and this is done each time the RPC is called. It's especially bad if none of the interceptors apply to the RPC being handled! 

I think we should precompute and store an array of interceptors for each method registered with the router. That way we avoid repeating the work each time an RPC is executed.

If we do this carefully we can arrange for there to only be one array allocation per unique interceptor pipeline (rather than one array allocation per method).

Sources/GRPCCore/GRPCServer.swift
```diff
@@ -172,15 +190,39 @@ public final class GRPCServer: Sendable {
...
+  ///   - router: A ``RPCRouter`` used by the server to route accepted streams to method handlers.
+  ///   - interceptors: A collection of interceptors providing cross-cutting functionality to each
+  ///       accepted RPC. The order in which interceptors are added reflects the order in which they
+  ///       are called. The first interceptor added will be the first interceptor to intercept each
+  ///       request. The last interceptor added will be the final interceptor to intercept each
+  ///       request before calling the appropriate handler.
+  public init(
+    transport: any ServerTransport,
+    router: RPCRouter,
+    interceptorPipeline: [ServerInterceptorOperation]
```
glbrntt, line 218: Previously the interceptors were passed to the router when `handle` is called. This meant that at this point in the code – i.e. when the router is passed to the server – that the router was fully configured and the caller was responsible for configuring it.

In the _other_ API, where the caller passes services and interceptors to the server, the server was responsible for configuring the router.

With the router now accepting interceptors the lines are blurred: it's not clear who is responsible for configuring the router. In this case it's partly the caller and partly the server. I think this distinction should be clearer.

I see a few options:
1. Remove the ability to create the server with a router _and_ interceptors and make the caller responsible for configuring the router entirely. (If the API where they pass it into the server.)
2. Stop storing the interceptors on the router and continue storing them on the server. This would require the server to lookup the interceptors for each RPC and pass them into the router (which is similar to the existing setup).

Sources/GRPCCore/GRPCServer.swift
```diff
@@ -206,7 +248,10 @@ public final class GRPCServer: Sendable {
 
     do {
       try await transport.listen { stream, context in
-        await self.router.handle(stream: stream, context: context, interceptors: self.interceptors)
+        await self.router.handle(
+          stream: stream,
+          context: context
+        )
```
glbrntt, line 254: stylistic nit: prefer a single line if it fits

Tests/GRPCCoreTests/Call/Server/Internal/ServerRPCExecutorTests.swift
```diff
@@ -258,7 +258,9 @@ final class ServerRPCExecutorTests: XCTestCase {
     )
 
     // The interceptor skips the handler altogether.
-    let harness = ServerRPCExecutorTestHarness(interceptors: [.rejectAll(with: error)])
+    let harness = ServerRPCExecutorTestHarness(interceptors: [
```
glbrntt, line 261: nit: avoid formatting only changes

Sources/GRPCCore/Call/Server/ServerInterceptorOperation.swift
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  /// An operation specifying an interceptor that will be applied only to RPCs directed to the specified services.
+  /// - Parameters:
+  ///   - interceptor: The interceptor to register with the server.
+  ///   - services: The list of service names for which this interceptor should intercept RPCs.
+  /// - Returns: A ``ServerInterceptorOperation``.
+  public static func apply(
+    _ interceptor: any ServerInterceptor,
+    onlyToServices services: [String]
```
glbrntt, line 49: I think we can drop the "only": `apply(_:toServices:)` is clear

glbrntt, line 49: Also: we have a `ServiceDescriptor` – we should use it instead of `String` here

Sources/GRPCCore/Call/Server/ServerInterceptorOperation.swift
```diff
@@ -0,0 +1,107 @@
...
+        interceptor: interceptor,
+        services: services
+      )
+    )
+  }
+
+  /// An operation specifying an interceptor that will be applied only to RPCs directed to the specified service methods.
+  /// - Parameters:
+  ///   - interceptor: The interceptor to register with the server.
+  ///   - services: The list of method descriptors for which this interceptor should intercept RPCs.
```
glbrntt, line 62:
```suggestion
  ///   - methods: The list of method descriptors for which this interceptor should intercept RPCs.
```

Sources/GRPCCore/Call/Server/ServerInterceptorOperation.swift
```diff
@@ -0,0 +1,107 @@
...
+  /// - Returns: A ``ServerInterceptorOperation``.
+  public static func apply(
+    _ interceptor: any ServerInterceptor,
+    onlyToMethods methods: [MethodDescriptor]
```
glbrntt, line 66: same comment re: only

Sources/GRPCCore/Call/Server/ServerInterceptorOperation.swift
```diff
@@ -0,0 +1,107 @@
...
+/// - requests directed only to specific methods (of a specific service).
+///
+/// - SeeAlso: ``ServerInterceptor`` for more information on server interceptors, and
+///  ``ClientInterceptorOperation`` for the client-side version of this type.
+public struct ServerInterceptorOperation: Sendable {
+  internal enum Wrapped: Sendable {
+    case allServices(interceptor: any ServerInterceptor)
+    case serviceSpecific(interceptor: any ServerInterceptor, services: [String])
+    case methodSpecific(interceptor: any ServerInterceptor, methods: [MethodDescriptor])
+  }
```
glbrntt, line 31: The distinguishing part is what the interceptor applies to, rather than the combination so I'd probably represent this as:

```swift
public struct ServerInterceptorOperation {
  enum AppliesTo {
    case all
    case services([ServiceDescriptor])
    case methods([MethodDescriptor])
  }

  var interceptor: any ServerInterceptor
  var appliesTo: AppliesTo
}
```

This also lets you remove the computer interceptor property on your `Wrapped` type

glbrntt: Thinking about this more, it might actually make sense to promote the `AppliesTo` to the API, then you can reduce the number of static funcs on `ServerInterceptorOperation` to one:

```swift
public struct ServerInterceptorOperation {
  public static func apply(_ interceptor: any ServerInteceptor, to subject: AppliesTo) -> Self
}
```

So then we calling it reads quite naturally:

```swift
interceptorPipeline: [
  .apply(foo, to: .all),
  .apply(bar, to: .services([...])
  .apply(baz, to: .methods([...])
]
```

Sources/GRPCCore/Call/Server/ServerInterceptorOperation.swift
```diff
@@ -0,0 +1,107 @@
...
+      return interceptor
+    case .methodSpecific(let interceptor, _):
+      return interceptor
+    }
+  }
+
+  /// Returns whether this ``ServerInterceptorOperation`` applies to the given `descriptor`.
+  /// - Parameter descriptor: A ``MethodDescriptor`` for which to test whether this interceptor applies.
+  /// - Returns: `true` if this interceptor applies to the given `descriptor`, or `false` otherwise.
+  public func applies(to descriptor: MethodDescriptor) -> Bool {
```
glbrntt, line 97: Does this need to be public?

Sources/GRPCCore/Call/Server/ServerInterceptorOperation.swift
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  private let wrapped: Wrapped
+
+  private init(wrapped: Wrapped) {
+    self.wrapped = wrapped
+  }
+
+  /// Get the ``ServerInterceptor`` associated with this ``ServerInterceptorOperation``.
+  public var interceptor: any ServerInterceptor {
```
glbrntt, line 83: Does this need to be public?

Tests/GRPCCoreTests/Test Utilities/Services/HelloWorld.swift
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright 2023, gRPC Authors All rights reserved.
```
glbrntt, line 2: It's closer to 2025 than it is to 2023 ;)

Sources/GRPCCore/Call/Server/RPCRouter.swift
```diff
@@ -126,12 +127,25 @@ public struct RPCRouter: Sendable {
...
       method: descriptor,
       deserializer: deserializer,
       serializer: serializer,
       handler: handler
     )
+    self.handlers[descriptor] = (handler, [])
+  }
+
+  @inlinable
+  public mutating func registerInterceptors(
```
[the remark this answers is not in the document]
gjcairo: Not really

Sources/GRPCCore/Call/Server/ServerInterceptorOperation.swift
gjcairo: Only if we want to keep `RPCRouter/registerInterceptors` inlinable.

Sources/GRPCCore/Call/Server/ServerInterceptorOperation.swift
gjcairo: Should the `service` in `MethodDescriptor` be a `ServiceDescriptor` then, instead of a `String`?
This is the main reason why I chose a `String` for this API.

Sources/GRPCCore/Call/Server/ServerInterceptorOperation.swift
glbrntt: Okay, in that case we should prefix it with an underscore

Sources/GRPCCore/Call/Server/ServerInterceptorOperation.swift
glbrntt: Probably, yes, but I think that'd have to come as part of a wider change though.

There are a couple of things that'd have to be considered before doing that:
- generated code would need updating
- parsing the ":path" would need to decompose the fully qualified service name (i.e. package + service) or have `ServiceDescriptor` do that lazily (store the fully qualified string, and compute the package/service name separately on demand)

Either way, I think `ServiceDescriptor` is the right type here and we should make any change to `MethodDescriptor` separately.

Sources/GRPCCore/GRPCServer.swift
gjcairo: Good point. I honestly think that 1 is the right approach here, since, in the init where the router is passed, we are already assuming it's set up properly (with e.g. each service's methods registered).

Sources/GRPCCore/Call/Server/ServerInterceptorOperation.swift
gjcairo: Nice one! I agree it reads more nicely

Sources/GRPCCore/GRPCServer.swift
glbrntt: Great, I'm happy with (1)

Sources/GRPCCore/Call/Server/ServerInterceptorOperation.swift
gjcairo: Yes, I wasn't thinking we should do this as part of this PR, but we should probably make this change too for consistency.

Sources/GRPCCore/Call/Server/ServerInterceptorOperation.swift
```diff
@@ -0,0 +1,96 @@
...
+/// A `ServerInterceptorOperation` describes to which RPCs a server interceptor should be applied.
+///  
+/// You can configure a server interceptor to be applied to:
+/// - all RPCs and services;
+/// - requests directed only to specific services registered with your server; or
+/// - requests directed only to specific methods (of a specific service).
+///  
+/// - SeeAlso: ``ServerInterceptor`` for more information on server interceptors, and
+///  ``ClientInterceptorOperation`` for the client-side version of this type.
+public struct ServerInterceptorOperation: Sendable {
```
glbrntt, line 26: Can we add "pipeline" to this name? i.e. "ServerInterceptorPipelienOperation"?

Sources/GRPCCore/Call/Server/ServerInterceptorOperation.swift
```diff
@@ -0,0 +1,96 @@
...
+  /// The subject of a ``ServerInterceptorOperation``.
+  /// The subject of an interceptor can either be all services and methods, only specific services, or only specific methods.
+  public struct Subject: Sendable {
+    internal enum Wrapped: Sendable {
+      case all
+      case services([ServiceDescriptor])
+      case methods([MethodDescriptor])
```
glbrntt, line 33: The number of methods/services here could potentially be quite large. Should we use `Set` instead?

Sources/GRPCCore/Call/Server/ServerInterceptorOperation.swift
```diff
@@ -0,0 +1,96 @@
...
+  ///   - subject: The ``Subject`` to which the `interceptor` applies.
+  /// - Returns: A ``ServerInterceptorOperation``.
+  public static func apply(_ interceptor: any ServerInterceptor, to subject: Subject) -> Self {
+    Self(interceptor: interceptor, appliesTo: subject)
+  }
+
+  /// Returns whether this ``ServerInterceptorOperation`` applies to the given `descriptor`.
+  /// - Parameter descriptor: A ``MethodDescriptor`` for which to test whether this interceptor applies.
+  /// - Returns: `true` if this interceptor applies to the given `descriptor`, or `false` otherwise.
+  public func _applies(to descriptor: MethodDescriptor) -> Bool {
```
glbrntt, line 93: Realised I gave you bad advice before: this can be `internal` and marked as `@inlinable`, there's no need for it to be `public`.

glbrntt (changes requested):

Sources/GRPCCore/Call/Server/RPCRouter.swift
```diff
@@ -142,6 +144,16 @@ public struct RPCRouter: Sendable {
   public mutating func removeHandler(forMethod descriptor: MethodDescriptor) -> Bool {
     return self.handlers.removeValue(forKey: descriptor) != nil
   }
+
+  @inlinable
+  mutating func registerInterceptors(pipeline: [ServerInterceptorOperation]) {
```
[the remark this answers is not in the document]
gjcairo: Good point. I think there's value in letting users register the interceptors manually. I've added docs - let me know if you think they're not enough.

glbrntt (approved):
Thanks Gus, this looks great. The docs check and formatting checks are failing though so those need fixing up before we can merge this.

glbrntt:
API break expected:
```
1 breaking change detected in GRPCCore:
  💔 API breakage: constructor GRPCServer.init(transport:router:interceptors:) has been removed
```

PR #2110, opened 2024-11-08
Remove v2 code from v1

glbrntt:
Motivation:

There's a bug in SwiftPM which prohibits the Xcode build system from being used if a dependency explicitly declares which Swift language mode is being used, regardless of its value. A number of users are running into this problem, and while it's not a gRPC Swift bug we can work around the issue by removing the v2 code from the v1 branch and removing the Swift 6 specific manifest.

Modifications:

- Remove all v2 code
- Remove workarounds in the SwiftPM build plugin which avoid deprecation warnings. It turns out that the guard was incorrect anyway: the API isn't guarded by the compiler version, but the package tools version.

Result:

Possible to build v1 with SwiftPM using the Xcode build system.


rnro (approved):
After reading every line of this diff, I approve

PR #2111, opened 2024-11-11
Bump version number to 1.24.2

glbrntt:
Motivation:

We plan on tagging a release soon.

Modifications:

- Bump the version to 1.24.2

Result:

The version in the default user-agent string will match the released version.


gjcairo (approved):

glbrntt:
The tests are failing because the certs for the interop test (which come from https://github.com/grpc/grpc/tree/master/src/core/tsi/test_creds and are yet to be updated) have expired. The tests passed on the last change so I'm going to merge over these failures.

PR #2113, opened 2024-11-13
Allow adding `ClientInterceptor`s to specific services and methods

gjcairo:
## Motivation
We want to allow users to customise the RPCs a registered interceptor should apply to on the client:
- Intercept all requests
- Intercept requests only meant for specific services
- Intercept requests only meant for specific methods

## Modifications
This PR adds a new `ClientInterceptorPipelineOperation` type that allows users to specify what the target of the interceptor should be.
Existing APIs accepting `[any ClientInterceptor]` have been kept, but new initialisers taking `[ClientInterceptorPipelineOperation]` instead have been added.

## Result
Users can have more control over to which requests interceptors are applied.

### DIFF
--- a/Sources/GRPCCore/Call/Client/ClientInterceptorPipelineOperation.swift
+++ b/Sources/GRPCCore/Call/Client/ClientInterceptorPipelineOperation.swift
@@ -14,17 +14,17 @@
  * limitations under the License.
  */
 
-/// A `ServerInterceptorPipelineOperation` describes to which RPCs a server interceptor should be applied.
+/// A `ClientInterceptorPipelineOperation` describes to which RPCs a client interceptor should be applied.
 ///
-/// You can configure a server interceptor to be applied to:
+/// You can configure a client interceptor to be applied to:
 /// - all RPCs and services;
-/// - requests directed only to specific services registered with your server; or
+/// - requests directed only to specific services; or
 /// - requests directed only to specific methods (of a specific service).
 ///
-/// - SeeAlso: ``ServerInterceptor`` for more information on server interceptors, and
-///  ``ClientInterceptorPipelineOperation`` for the client-side version of this type.
-public struct ServerInterceptorPipelineOperation: Sendable {
-  /// The subject of a ``ServerInterceptorPipelineOperation``.
+/// - SeeAlso: ``ClientInterceptor`` for more information on client interceptors, and
+///  ``ServerInterceptorPipelineOperation`` for the server-side version of this type.
+public struct ClientInterceptorPipelineOperation: Sendable {
+  /// The subject of a ``ClientInterceptorPipelineOperation``.
   /// The subject of an interceptor can either be all services and methods, only specific services, or only specific methods.
   public struct Subject: Sendable {
     internal enum Wrapped: Sendable {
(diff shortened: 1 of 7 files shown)


Sources/GRPCCore/GRPCClient.swift
```diff
@@ -112,13 +112,18 @@ public final class GRPCClient: Sendable {
...
+  ///
+  /// The list of interceptors for each method is computed from `interceptorsPipeline` when calling a method for the first time.
+  /// This caching is done to avoid having to compute the applicable interceptors for each request made.
   ///
   /// The order in which interceptors are added reflects the order in which they are called. The
   /// first interceptor added will be the first interceptor to intercept each request. The last
   /// interceptor added will be the final interceptor to intercept each request before calling
   /// the appropriate handler.
-  private let interceptors: [any ClientInterceptor]
+  private let interceptorsPerMethod: Mutex<[MethodDescriptor: [any ClientInterceptor]]>
```
gjcairo, line 126: I thought this caching could be useful to speed things up, similar to what we do in the `RPCRouter` on the server. However I wonder if we care more about the client not taking up too much space (which could be the case if it calls _many_ methods and this dictionary grows a lot).

Sources/GRPCCore/Call/Client/ClientInterceptorPipelineOperation.swift
```diff
@@ -0,0 +1,99 @@
...
+/// A `ClientInterceptorPipelineOperation` describes to which RPCs a client interceptor should be applied.
+///
+/// You can configure a client interceptor to be applied to:
+/// - all RPCs and services;
+/// - requests directed only to specific services; or
+/// - requests directed only to specific methods (of a specific service).
+///
+/// - SeeAlso: ``ClientInterceptor`` for more information on client interceptors, and
+///  ``ServerInterceptorPipelineOperation`` for the server-side version of this type.
+public struct ClientInterceptorPipelineOperation: Sendable {
```
glbrntt, line 26: Not sure how I missed this for the server, but we should add tests for this specific type as well - `applies(to:)` is the only thing we need to test here though.

Sources/GRPCCore/GRPCClient.swift
```diff
@@ -112,34 +112,53 @@ public final class GRPCClient: Sendable {
...
+      /// A collection of interceptors providing cross-cutting functionality to each accepted RPC, keyed by the method to which they apply.
+      ///
+      /// The list of interceptors for each method is computed from `interceptorsPipeline` when calling a method for the first time.
+      /// This caching is done to avoid having to compute the applicable interceptors for each request made.
+      ///
+      /// The order in which interceptors are added reflects the order in which they are called. The
+      /// first interceptor added will be the first interceptor to intercept each request. The last
+      /// interceptor added will be the final interceptor to intercept each request before calling
+      /// the appropriate handler.
+      interceptorsPerMethod: [MethodDescriptor: [any ClientInterceptor]]
```
glbrntt, line 134: The current state is orthogonal to the cache, rather than storing the cache in the enum cases, can we wrap `State` in a `StateMachine` struct and store the cache in the `StateMachine` struct as a sibling to the `State`? It also means we don't need the `_modifying` state.

```swift
struct StateMachine {
  var state: State
  var interceptors:  [MethodDescriptor: [any ClientInterceptor]]
}
```

Sources/GRPCCore/GRPCClient.swift
```diff
@@ -361,14 +406,50 @@ public final class GRPCClient: Sendable {
     var options = options
     options.formUnion(with: methodConfig)
 
+    let applicableInterceptors = self.state.withLock {
```
glbrntt, line 409: Rather than doing all of this inline can we do it in a method on the `StateMachine`?

Also, it'd be good if we can combine this with `checkExecutable` above to avoid acquiring the lock twice per RPC.

glbrntt (changes requested):

glbrntt (approved):

PR #2117, opened 2024-11-14
Add structured swift represntation for the generated metadata

glbrntt:
Motivation:

The code gen as it stands tests too much in one go: a small change leads to many tests being updated. This stems from the translator not being very testable. To improve this, and make future code gen changes less painful, we can refactor it such that helpers build structured swift. These are significantly less coupled and can be tested more easily.

The strategy here is to add all the structured representations for the metadata, client, and server and then cut over the translators to using the new code. This first change will introduce the structured represntation for metadata.

Modifications:

- Add metadata structured swift
- Add helpers for commonly used types

Result:

Metadata is in place


rnro (approved):
Looks good!

PR #2120, opened 2024-11-15
Allow metadata to be mutated on server response types

glbrntt:
Motivation:

The server response types have a metadata computed property with only a getter. It's entirely possible to mutate the metadata manually, it's just a bit of a faff. This should be easier.

Modifications:

- Add a setter to the computed property
- Migrate server response tests to swift-testing

Result:

- Easier to use API

### DIFF
--- a/Sources/GRPCCore/Call/Server/ServerResponse.swift
+++ b/Sources/GRPCCore/Call/Server/ServerResponse.swift
@@ -244,15 +244,25 @@ extension ServerResponse {
     self.accepted = .failure(error)
   }
 
-  /// Returns the metadata to be sent to the client at the start of the response.
-  ///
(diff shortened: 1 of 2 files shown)


Sources/GRPCCore/Call/Server/ServerResponse.swift
```diff
@@ -244,15 +244,25 @@ extension ServerResponse {
...
+      }
+    }
+    set {
+      switch self.accepted {
+      case var .success(contents):
+        contents.metadata = newValue
+        self.accepted = .success(contents)
+      case var .failure(error):
+        error.metadata = newValue
+        self.accepted = .failure(error)
```
FranzBusch, line 264: Just a side note: This might be prone to CoWs and something we might want to make `@inlinable` or use a `consuming switch`

FranzBusch (approved):

PR #2124, opened 2024-11-20
Document support for v1 moving forwards

glbrntt:


README.md
```diff
@@ -12,6 +12,25 @@ major version, v2.x. Pre-release versions will be available in the near future.
 In the meantime, v1.x is available and still supported. You can read more about
 it on the [Swift Package Index][spi-grpc-swift-main].
 
+## Support
```
gjcairo, line 15: Should we explicitly mention anything about platforms?

gjcairo (approved):
LGTM but left a question

README.md
glbrntt: What did you have in mind? The support window will decrease the same for all platforms we support.

gjcairo: We added a "platforms" section to the v2 docs: https://github.com/grpc/grpc-swift/blob/main/Sources/GRPCCore/Documentation.docc/Articles/Compatibility.md
So I was wondering if it's worth mentioning which platforms v1 supports explicitly as well or not.

glbrntt: It's in the docs on SPI for v1, but at this point I don't think it really matters tbh.

PR #2126, opened 2024-11-20
Simplify typealias translator

glbrntt:
Motivation:

The typealias used to generte an enum per package name and nested enums for each service within that namespace. In order for that to work, the services had to be grouped by namespace. This is no longer the case, each service is generated within a combined package-service enum.

It's simpler to just generate the services in the order they are created.

Modifications:

- Remove the sorting
- Remove tests which test the unwanted behaviour

Result:

Less code


rnro (approved):
LGTM

PR #2127, opened 2024-11-21
Use ServiceDescriptor within MethodDescriptor

glbrntt:
Motivation:

MethodDescriptor uses a string to represent the fully qualified name of the service the method belongs to. ServiceDescriptor also represents a fully qualified name of a service. It'd make more sense for MethodDescriptor to use a ServiceDescriptor instead of a string to describe the service.

Modifications:

- ServiceDescriptor now stores the fully qualified service name (as this is generally more useful) and computes the package and unqualified service name.
- ServiceDescriptor can now be created from the fully qualified name or from a separate package and service name.
- MethodDescriptor now holds a ServiceDescriptor instead of a String for the service

Result:

More coherent API


glbrntt:
API breakage is expected and okay.

FranzBusch (approved):

PR #2131, opened 2024-11-26
Simplify generated code

glbrntt:
Motivation:

A number of things have changed since the generated code was initially designed, which means it can be simplified and improved.

Modifications:

- As services are namespaced within a namespace-service enum (rather than within a service enum in a namespace enum), they no longer need to be grouped by namespace.
- As deployment targets must be set in the package manifest, the generated code no longer needs to be annotated with availability guards so these have been removed.

Result:

Simpler code generation, better generated code.

### DIFF
--- a/Sources/GRPCCodeGen/Internal/Renderer/TextBasedRenderer.swift
+++ b/Sources/GRPCCodeGen/Internal/Renderer/TextBasedRenderer.swift
@@ -1134,8 +1134,9 @@ struct TextBasedRenderer: RendererProtocol {
   /// Renders the specified code block.
   func renderCodeBlock(_ description: CodeBlock) {
     if let comment = description.comment { renderComment(comment) }
-    let item = description.item
-    renderCodeBlockItem(item)
+    if let item = description.item {
+      renderCodeBlockItem(item)
+    }
   }
 
   /// Renders the specified code blocks.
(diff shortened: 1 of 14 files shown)


Sources/GRPCCodeGen/Internal/Renderer/TextBasedRenderer.swift
```diff
@@ -1134,8 +1134,9 @@ struct TextBasedRenderer: RendererProtocol {
   /// Renders the specified code block.
   func renderCodeBlock(_ description: CodeBlock) {
     if let comment = description.comment { renderComment(comment) }
-    let item = description.item
-    renderCodeBlockItem(item)
+    if let item = description.item {
```
rnro, line 1137: The previous style had a cleaner call site; the yet was the impetus to change?

rnro (approved):
One question otherwise it looks good

Sources/GRPCCodeGen/Internal/Renderer/TextBasedRenderer.swift
glbrntt: `description.item` is now `Optional`, so we need this change (or for `renderCodeBlockItem` to take an optional)

PR #2136, opened 2024-12-02
Add remote peer info to the server context

glbrntt:
Motivation:

It's often useful to know the identity of the remote peer when handling RPCs.

Modifications:

- Add a 'peer' to the server context
- Implement this for the in-process transport
- Make some in-process inits `package`, these should never have been `public`

Result:

Server RPCs have some idea what the address of remote peer is


glbrntt:
API breakages is expected.

rnro (approved):

PR #2142, opened 2024-12-09
Enable MemberImportVisibility check on all targets

rnro:
Enable MemberImportVisibility check on all targets. Use a standard string header and footer to bracket the new block for ease of updating in the future with scripts.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -117,3 +117,14 @@ let package = Package(
   dependencies: dependencies,
   targets: targets
 )
+
+// ---    STANDARD CROSS-REPO SETTINGS DO NOT EDIT   --- //
+for target in package.targets {
+  if target.type != .plugin {
+    var settings = target.swiftSettings ?? []
+    // https://github.com/swiftlang/swift-evolution/blob/main/proposals/0444-member-import-visibility.md
+    settings.append(.enableUpcomingFeature("MemberImportVisibility"))
+    target.swiftSettings = settings
+  }
+}
+// --- END: STANDARD CROSS-REPO SETTINGS DO NOT EDIT --- //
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -13,9 +13,9 @@ jobs:
     with:
       linux_5_9_enabled: false
       linux_5_10_enabled: false
-      linux_6_0_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable"
-      linux_nightly_6_0_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable"
-      linux_nightly_main_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable"
+      linux_6_0_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable -Xswiftc -enable-upcoming-feature -Xswiftc MemberImportVisibility"
```
glbrntt, line 16: Any reason we can't just add this to our default settings in the manifest? https://github.com/grpc/grpc-swift/blob/e390c85ad880004513221555884ae77b08714f1e/Package.swift#L48

Package.swift
```diff
@@ -117,3 +117,9 @@ let package = Package(
   dependencies: dependencies,
   targets: targets
 )
+
+for target in package.targets {
+  var settings = target.swiftSettings ?? []
+  settings.append(.enableUpcomingFeature("MemberImportVisibility"))
+  target.swiftSettings = settings
+}
```
glbrntt, line 125: Can we just add this directly to the `defaultSwiftSettings` above so that they're all in the same place.

glbrntt (changes requested):

Package.swift
rnro: The intention here is that many of the core Swift on Server repositories have a block of their package manifest which a common structure which can be easily be modified with a script. This block would be used to insert settings which we wish to apply across all repositories.

glbrntt (approved):

PR #2149, opened 2024-12-20
Add reflection server example

glbrntt:
Motivation:

We've added back the reflection server; we should have an example of how
to use it.

Modifications:

- Add a reflection server example
- Split the echo service into a separate file in the echo example so that it can be symlinked from the reflection server

Result:

More examples


Examples/reflection-server/README.md
```diff
@@ -0,0 +1,61 @@
+# Reflection Server
+
+This example demonstrates the gRPC Reflection service which is described in more
+detail in the [gRPC documentation](https://github.com/grpc/grpc/blob/6fa8043bf9befb070b846993b59a3348248e6566/doc/server-reflection.md).
+
+## Overview
+
+A "reflection-server" command line tool that uses the reflection service implementation
+from [grpc/grpc-swift-extras](https://github.com/grpc/grpc-swift-extras) and the
+Echo service (see the 'echo' example).
```
gjcairo, line 10: This is a tiny nit (and I'm sorry to be annoying), but we're using double quotes for `reflection-server`, single quotes for `echo`, and we've used backticks to refer to other services or command line tools elsewhere in this README and in other docs. Can we pick one?

gjcairo (approved):
LGTM, just one very tiny comment. Thanks George!

Examples/reflection-server/README.md
glbrntt: Don't apologise my inconsistency!

PR #2154, opened 2025-01-13
Extend structured swift to support generic structs

glbrntt:
Motivation:

The structured swift representation we're using doesn't support generic struct at the moment. We'll need this for an upcoming change.

Modifications:

- Allow for generic structs with where clauses to be represented and rendered.

Result:

Generic structs can be rendered

### DIFF
--- a/Tests/GRPCCodeGenTests/Internal/Renderer/TextBasedRendererTests.swift
+++ b/Tests/GRPCCodeGenTests/Internal/Renderer/TextBasedRendererTests.swift
@@ -853,11 +853,12 @@ final class Test_TextBasedRenderer: XCTestCase {
         whereClause: WhereClause(
           requirements: [
             .conformance("T", "Foo"),
-            .conformance("T", "Sendable")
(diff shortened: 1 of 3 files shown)


Tests/GRPCCodeGenTests/Internal/Renderer/TextBasedRendererTests.swift
```diff
@@ -820,12 +820,55 @@ final class Test_TextBasedRenderer: XCTestCase {
...
+      rendersAs: #"""
+        struct Structy<T> where T: Foo {}
+        """#
+    )
+    try _test(
+      StructDescription(
+        name: "Structy",
+        generics: [.member("T")],
+        conformances: ["Hashable"],
+        whereClause: WhereClause(requirements: [.conformance("T", "Foo")])
```
gjcairo, line 865: I know we're not testing the `WhereClause`, but should we add a second requirement to make sure both get rendered appropriately?

gjcairo (approved):

PR #2157, opened 2025-01-15
Deduplicate interceptor pipeline operation

glbrntt:
Motivation:

ClientInterceptorPipelineOperation is the same as
ServerInterceptorPipelineOperation apart from the type of interceptor being used. The duplication here is unnecessary.

Modifications:

- Add a `ConditionalInterceptor` which is roughly the same as `*InterceptorPipelineOperation` but is generic.
- The init is private and can only be initialized via static methods when the generic type is appropriate (i.e. client/server interceptor).

Result:

Less duplication, less code


Sources/GRPCCore/Call/ConditionalInterceptor.swift
```diff
@@ -97,3 +84,29 @@ public struct ClientInterceptorPipelineOperation: Sendable {
...
+    to subject: Subject
+  ) -> Self {
+    Self(interceptor: interceptor, subject: subject)
+  }
+}
+
+extension ConditionalInterceptor where Interceptor == any ServerInterceptor {
+  /// Create an operation, specifying which ``ServerInterceptor`` to apply and to which ``Subject``.
+  /// - Parameters:
+  ///   - interceptor: The ``ServerInterceptor`` to register with the client.
```
gjcairo, line 104:
```suggestion
  ///   - interceptor: The ``ServerInterceptor`` to register with the server.
```

Sources/GRPCCore/Call/Server/ServerInterceptor.swift
```diff
@@ -21,7 +21,7 @@
 /// been returned from a service. They are typically used for cross-cutting concerns like filtering
 /// requests, validating messages, logging additional data, and tracing.
 ///
-/// Interceptors can be registered with the server either directly or  via ``ServerInterceptorPipelineOperation``s.
+/// Interceptors can be registered with the server either directly or  via ``ConditionalInterceptor``s.
```
gjcairo, line 24:
```suggestion
/// Interceptors can be registered with the server either directly or via ``ConditionalInterceptor``s.
```

gjcairo (approved):

PR #2158, opened 2025-01-15
Add more properties to `ClientContext` and have the `ClientTransport` provide it

gjcairo:
The `ServerTransport` provides the `ServerContext`, as it contains information that only the transport knows about (such as the remote peer's address).

For consistency and to allow the `ClientContext` to also hold some additional information (such as remote and local peer descriptions), this PR changes the `ClientTransport` protocol so that implementations also provide the corresponding `ClientContext`. 

This PR also adds additional information to the context (which will be used by the tracing interceptor but can be useful for users in general): remote and local peer addresses, server hostname, and network transport.

### DIFF
--- a/Sources/GRPCCore/Call/Client/ClientContext.swift
+++ b/Sources/GRPCCore/Call/Client/ClientContext.swift
@@ -32,9 +32,6 @@ public struct ClientContext: Sendable {
   /// - "in-process:27182".
   public var remotePeer: String
 
-  /// The hostname of the RPC server.
-  public var serverHostname: String
-
   /// A description of the local peer.
   ///
   /// The format of the description should follow the pattern "<transport>:<address>" where
@@ -48,21 +45,14 @@ public struct ClientContext: Sendable {
   /// - "in-process:27182".
   public var localPeer: String
 
-  /// The transport in use (e.g. "tcp", "udp").
-  public var networkTransportMethod: String
-
   /// Create a new client interceptor context.
   public init(
     descriptor: MethodDescriptor,
     remotePeer: String,
-    localPeer: String,
-    serverHostname: String,
-    networkTransportMethod: String
+    localPeer: String
   ) {
     self.descriptor = descriptor
     self.remotePeer = remotePeer
     self.localPeer = localPeer
-    self.serverHostname = serverHostname
-    self.networkTransportMethod = networkTransportMethod
   }
 }
--- a/Sources/GRPCCore/Call/Client/ClientContext.swift
+++ b/Sources/GRPCCore/Call/Client/ClientContext.swift
@@ -19,8 +19,50 @@ public struct ClientContext: Sendable {
   /// A description of the method being called.
   public var descriptor: MethodDescriptor
 
+  /// A description of the remote peer.
+  ///
+  /// The format of the description should follow the pattern "<transport>:<address>" where
+  /// "<transport>" indicates the underlying network transport (such as "ipv4", "unix", or
+  /// "in-process"). This is a guideline for how descriptions should be formatted; different
+  /// implementations may not follow this format so you shouldn't make assumptions based on it.
+  ///
+  /// Some examples include:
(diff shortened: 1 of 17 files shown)


gjcairo:
The API breaks are expected, since we've added the `ClientContext` to the closure signature:

```
💔 API breakage: func ClientTransport.withStream(descriptor:options:_:) has parameter 2 type change from (GRPCCore.RPCStream<Self.Inbound, Self.Outbound>) async throws -> T to (GRPCCore.RPCStream<Self.Inbound, Self.Outbound>, GRPCCore.ClientContext) async throws -> T

💔 API breakage: constructor ClientContext.init(descriptor:) has been removed

💔 API breakage: func InProcessTransport.Client.withStream(descriptor:options:_:) has parameter 2 type change from (GRPCCore.RPCStream<GRPCInProcessTransport.InProcessTransport.Client.Inbound, GRPCInProcessTransport.InProcessTransport.Client.Outbound>) async throws -> T to (GRPCCore.RPCStream<GRPCInProcessTransport.InProcessTransport.Client.Inbound, GRPCInProcessTransport.InProcessTransport.Client.Outbound>, GRPCCore.ClientContext) async throws -> T
```

Sources/GRPCCore/Call/Client/Internal/ClientRPCExecutor+OneShotExecutor.swift
```diff
@@ -98,11 +98,13 @@ extension ClientRPCExecutor.OneShotExecutor {
   ) async -> Result<R, any Error> {
     return await withTaskGroup(of: Void.self, returning: Result<R, any Error>.self) { group in
       do {
-        return try await self.transport.withStream(descriptor: method, options: options) { stream in
+        return try await self.transport.withStream(descriptor: method, options: options) {
+          stream,
+          context in
```
glbrntt, line 103: nit: swift-format did a bad job here

```suggestion
        return try await self.transport.withStream(
          descriptor: method, 
          options: options
        ) { stream, context in
```

glbrntt (changes requested):

Sources/GRPCCore/Call/Client/ClientContext.swift
```diff
@@ -19,8 +19,50 @@ public struct ClientContext: Sendable {
...
+  /// implementations may not follow this format so you shouldn't make assumptions based on it.
+  ///
+  /// Some examples include:
+  /// - "ipv4:127.0.0.1:31415",
+  /// - "ipv6:[::1]:443",
+  /// - "in-process:27182".
+  public var remotePeer: String
+
+  /// The hostname of the RPC server.
+  public var serverHostname: String
```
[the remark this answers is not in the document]
gjcairo: Hm yes, that would make sense. I was thinking of using the authority for the hostname (which may be frequently set) but the transport method annoyed me actually because it seems a bit ad-hoc/too tied to the specific tracer. I'll move them to the tracer's config.

Sources/GRPCInProcessTransport/InProcessTransport+Client.swift
```diff
@@ -297,11 +297,17 @@ extension InProcessTransport {
         }
       }
 
+      let clientContext = ClientContext(
+        descriptor: descriptor,
+        remotePeer: "<in-process>",
+        localPeer: "<in-process>"
+      )
```
glbrntt, line 304: Sorry, forgot about this in the last review but the server adds the PID as well. We actually create the string for the server here: https://github.com/grpc/grpc-swift/blob/76073a2c1a2dcae35df4f0ab83ffd07b23e4cec9/Sources/GRPCInProcessTransport/InProcessTransport.swift#L28-L30 so we can just pass that in.

glbrntt (approved):
Fab, thanks Gus!

glbrntt (approved):

glbrntt (approved):

PR #2161, opened 2025-01-16
Add remote and local peers to `ServerContext`

gjcairo:
This PR adds a `localPeer` property to the `ServerContext`, and renames `peer` to `remotePeer`.

This is related to https://github.com/grpc/grpc-swift/pull/2158 on the client side and will be used to improve the server tracing interceptor.


glbrntt (approved):

glbrntt (approved):

gjcairo:
This API break is expected: the local peer is required so we can't not have it as part of the context's initialiser.
```
💔 API breakage: constructor ServerContext.init(descriptor:peer:cancellation:) has been removed
```

gjcairo:
Also @glbrntt you'll have to merge this to override the failing required step

PR #2162, opened 2025-01-17
Use JSON (de)serializers instead of custom ones for tests

gjcairo:
In some tests, we were using a custom serializer/deserializer to deal with the `PeerInfo` type. We can just reuse the JSON (de)serialiser we have. Had to copy it over as it was in a different target.


Tests/GRPCInProcessTransportTests/Test Utilities/JSONSerializing.swift
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+import Foundation
+import GRPCCore
+
+struct JSONSerializer<Message: Encodable>: MessageSerializer {
+  private let encoder = JSONEncoder()
+
+  func serialize(_ message: Message) throws -> [UInt8] {
```
glbrntt, line 23: You've lost the contiguous-bytes-ness from these

glbrntt (changes requested):

glbrntt (approved):

PR #2167, opened 2025-01-20
Add a dev-tool subpackage

glbrntt:
Motivation:

A number of test in this package and others rely on ad-hoc services using Codable. This is less overhead than using protobuf as you it's not always available. It also means the messages are defined in Swift so they're easy to change without needing to regenerate. However, the service glue code is hand rolled. We can avoid this by having a little adapter sit on top of the code gen lib.

Modifications:

- Add a grpc-dev-tool package to dev. We can use this as a place to add tooling and other helpers without worrying about worsening the experience for end users (because of additional dependencies, more public API and so on).
- For now this has a single executable for generating code from a JSON config file. The schema for the services is limited, but that's fine, it's not a general purpose tool.

Result:

- We have a tool which can generate grpc code from a JSON definition which uses Codable message types.

### DIFF
--- a/dev/grpc-dev-tool/Sources/grpc-dev-tool/Subcommands/GenerateJSON/GRPCDevUtils+GenerateJSON.swift
+++ b/dev/grpc-dev-tool/Sources/grpc-dev-tool/Subcommands/GenerateJSON/GRPCDevUtils+GenerateJSON.swift
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2025, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import ArgumentParser
+import Foundation
+
+struct GenerateJSON: ParsableCommand {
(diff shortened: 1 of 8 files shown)


dev/grpc-dev-tool/Sources/grpc-dev-tool/Subcommands/GenerateJSON/GRPCDevUtils+GenerateJSON.swift
```diff
@@ -0,0 +1,83 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import ArgumentParser
+import Foundation
+
+struct GenerateJSON: ParsableCommand {
```
gjcairo, line 20: Should we have both commands take an `output` param as well instead of printing to stdout? Not necessary but we could automate things further with a build plugin eventually if we did this instead.

gjcairo (commented):
Could we add some tests?

glbrntt:
> Could we add some tests?

I honestly don't think it's worth it: this is a dev tool and absolutely shouldn't be used by anyone else (which is why it's a separate package). There's also not much to test as the code generator is pretty heavily tested already.

I also don't want to set a high bar for this tools package, we should bias towards adding things which make our lives easier and having rough edges is absolutely fine. We can smooth them out as and when that becomes necessary.

dev/grpc-dev-tool/Sources/grpc-dev-tool/Subcommands/GenerateJSON/GRPCDevUtils+GenerateJSON.swift
glbrntt: Similar to the other comment, I don't think we need this now: we can just redirect the output to a file using bash etc.

PR #2171, opened 2025-01-22
Use the protoc build plugin in examples

glbrntt:
Motivation:

We now have a SwiftPM build plugin to generate code as part of the build
process, the examples should make use of it.

Modifications:

- Remove generated code for examples and update the generate script to
  stop genearting example code.
- Symlink protos from examples to the checked out location in the repo;
  this avoids some duplication and makes it easier to keep things
  up-to-date.
- Add a basic config to each example.
- Update CI to install protoc before building examples.

Result:

Better examples, less code.


gjcairo (approved):
Nice one

PR #2179, opened 2025-01-24
Update Example READMEs

glbrntt:
Motivation:

The examples now use the Swift PM build plugin but the READMEs weren't updated to include the PROTOC_PATH.

Modifications:

- update READMEs

Result:

Better examples


Examples/echo/README.md
```diff
@@ -12,27 +12,32 @@ the four RPC types.
...
+You must have the Protocol Buffers compiler (`protoc`) installed. You can find
+the instructions for doing this in the [gRPC Swift Protobuf documentation][0].
+
 ## Usage
 
 Build and run the server using the CLI:
 
 ```console
-$ swift run echo serve
+$ PROTOC_PATH=$(which protoc) swift run echo serve
```
gjcairo, line 25: Similar to the comment I left on the examples, should we mention (and/or link to the plugin docs) why setting this environment variable is required?

glbrntt: Yup, done!

gjcairo (approved):

PR #2181, opened 2025-01-24
Update dependencies for rc

glbrntt:
Motivation:

We're going to tag an RC soon so bump tutorial/example dependencies.

Modifications:

- Bump dependencies to `-rc.1`.

Result:

Docs will be correct for the tagged release.


gjcairo (approved):

glbrntt:
Examples are expected to fail, they're using a dependency which hasn't been tagged yet.

Issue #2183, opened 2025-01-28
Helper to get body parts from a streaming client response


glbrntt:
The [`StreamingClientResponse`][0] type has an `accepted` property which holds a result, the success case of which holds the `Contents` of the body of the response which in turn holds a [`bodyParts`][1] async sequence.

This isn't very ergonomic to get! 

We should add a computed property called `bodyParts` to `StreamingClientResponse` which returns it to us (or throws).

[0]: https://swiftpackageindex.com/grpc/grpc-swift/2.0.0-rc.1/documentation/grpccore/streamingclientresponse
[1]: https://swiftpackageindex.com/grpc/grpc-swift/2.0.0-rc.1/documentation/grpccore/streamingclientresponse/contents/bodyparts

PR #2184, opened 2025-01-28
Add bodyParts computed prop to StreamingClientResponse

gjcairo:
This PR resolves https://github.com/grpc/grpc-swift/issues/2183

### DIFF
--- a/Sources/GRPCCore/Call/Client/ClientResponse.swift
+++ b/Sources/GRPCCore/Call/Client/ClientResponse.swift
@@ -382,4 +382,13 @@ extension StreamingClientResponse {
       return RPCAsyncSequence.throwing(error)
     }
   }
+
+  /// Returns the body parts (i.e. `messages` and `trailingMetadata`) returned from the server.
+  ///
+  /// For rejected RPCs (in other words, where ``accepted`` is `failure`), this method throws an `RPCError`.
+  public var bodyParts: RPCAsyncSequence<Contents.BodyPart, any Error> {
+    get throws {
+      try self.accepted.get().bodyParts
+    }
+  }
 }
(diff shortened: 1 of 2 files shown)


Sources/GRPCCore/Call/Client/ClientResponse.swift
```diff
@@ -382,4 +382,13 @@ extension StreamingClientResponse {
...
   }
+
+  /// Returns the body parts (i.e. `messages` and `trailingMetadata`) returned from the server.
+  ///
+  /// For rejected RPCs (in other words, where ``accepted`` is `failure`), this method throws an `RPCError`.
+  public var bodyParts: RPCAsyncSequence<Contents.BodyPart, any Error> {
+    get throws {
+      try self.accepted.get().bodyParts
+    }
+  }
```
glbrntt, line 393: If the caller gets this and then iterates it in two expression they'll have:

```swift
let bodyParts = try response.bodyParts
for try await part in bodyParts {
  // ...
}
```

Can we do what's done for `messages` above this so that we can avoid that? (Sorry, the issue does so "or throws"; I didn't think about it enough when filing it.)

Tests/GRPCCoreTests/Call/Client/ClientResponseTests.swift
```diff
@@ -182,3 +210,21 @@ final class ClientResponseTests: XCTestCase {
...
+      return lhsMessage == rhsMessage
+
+    case (.trailingMetadata(let lhsMetadata), .trailingMetadata(let rhsMetadata)):
+      return lhsMetadata == rhsMetadata
+
+    default:
+      return false
+    }
+  }
+}
```
glbrntt, line 230: We should just make `BodyPart` and `Contents` conditionally `Hashable` when `Message` is `Hashable` (as public API).

glbrntt (changes requested):

glbrntt (changes requested):
LGTM modulo doc nits. A bit surprised the docs check didn't pick this up to be honest.

glbrntt (approved):
Thanks @gjcairo!

PR #2185, opened 2025-01-28
Fix Metadata's description for prettier printing

gjcairo:
This PR changes `Metadata`'s description to include quotes around string values.

From this:

```
[("some-key", some-value)]
```

to this:

```
[("some-key", "some-value")]
```

### DIFF
--- a/Sources/GRPCCore/Metadata.swift
+++ b/Sources/GRPCCore/Metadata.swift
@@ -494,7 +494,15 @@ extension Metadata.Value: ExpressibleByArrayLiteral {
 
 extension Metadata: CustomStringConvertible {
   public var description: String {
-    String(describing: self.map({ ($0.key.description, $0.value.description) }))
+    var description = "["
+    description = self.dropLast().reduce(into: description) { partialResult, nextPair in
+      partialResult += "\"\(nextPair.key)\": \(nextPair.value), "
+    }
+    if let lastElement = self.last {
+      description += "\"\(lastElement.key)\": \(lastElement.value)"
+    }
+    description += "]"
+    return description
   }
 }
 
@@ -502,7 +510,7 @@ extension Metadata.Value: CustomStringConvertible {
   public var description: String {
     switch self {
     case .string(let stringValue):
-      return stringValue
+      return "\"\(stringValue)\""
     case .binary(let binaryValue):
       return String(describing: binaryValue)
     }
--- a/Sources/GRPCCore/Metadata.swift
+++ b/Sources/GRPCCore/Metadata.swift
@@ -494,7 +494,7 @@ extension Metadata.Value: ExpressibleByArrayLiteral {
 
 extension Metadata: CustomStringConvertible {
   public var description: String {
-    String(describing: self.map({ ($0.key, $0.value) }))
+    String(describing: self.map({ ($0.key.description, $0.value.description) }))
   }
 }
 
@@ -502,7 +502,7 @@ extension Metadata.Value: CustomStringConvertible {
   public var description: String {
     switch self {
     case .string(let stringValue):
-      return String(describing: stringValue)
+      return stringValue
     case .binary(let binaryValue):
       return String(describing: binaryValue)
     }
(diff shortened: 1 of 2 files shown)


Sources/GRPCCore/Metadata.swift
```diff
@@ -494,15 +494,15 @@ extension Metadata.Value: ExpressibleByArrayLiteral {
 
 extension Metadata: CustomStringConvertible {
   public var description: String {
-    String(describing: self.map({ ($0.key, $0.value) }))
+    String(describing: self.map({ ($0.key.description, $0.value.description) }))
```
glbrntt, line 497: Can you add a test to verify the description is as we'd expect for a single `Metadata.Value` and for a `Metadata`?

glbrntt (changes requested):

Sources/GRPCCore/Metadata.swift
```diff
@@ -494,15 +494,23 @@ extension Metadata.Value: ExpressibleByArrayLiteral {
...
+    return description
   }
 }
 
 extension Metadata.Value: CustomStringConvertible {
   public var description: String {
     switch self {
     case .string(let stringValue):
-      return String(describing: stringValue)
+      return "\"\(stringValue)\""
```
glbrntt, line 513: This has the downside that when printing an individual value it will be quoted. When you print a sequence the `CustomDebugStringConvertible` conformance will be used. This is why `print(["foo"])` prints `["foo"]` and `print("foo")` prints `foo`. We should use `CustomDebugStringConvertible` here.

Sources/GRPCCore/Metadata.swift
```diff
@@ -494,15 +494,23 @@ extension Metadata.Value: ExpressibleByArrayLiteral {
 
 extension Metadata: CustomStringConvertible {
   public var description: String {
-    String(describing: self.map({ ($0.key, $0.value) }))
+    var description = "["
```
glbrntt, line 497: Couple of things:
1. For an empty metadata collection we should print `[:]`, right now we'll print `[]`.
2. We can use map and joined here: `self.map { ... }.joined(separator: ", ")`
3. To quote the keys (and get the correct quoting on values) we should use `String(reflecting:)` (to get the `CustomStringConvertible` value).

Sources/GRPCCore/Metadata.swift
glbrntt: I think it's obvious without quotes for a single value: `[0, 1, 2]` vs `foo`.

gjcairo: Well, it could be the string "[1, 2, 3]" but maybe I'm too cynical

Tests/GRPCCoreTests/MetadataTests.swift
```diff
@@ -314,4 +314,31 @@ struct MetadataTests {
...
+        case "key1":
+          #expect("\(value)" == "value1")
+        case "key2":
+          #expect("\(value)" == "value2")
+        case "key-bin":
+          #expect("\(value)" == "[1, 2, 3]")
+        default:
+          Issue.record("Should not have reached this point")
+        }
+      }
```
glbrntt, line 341: This should be a separate test: it's testing how `Metadata.Value` conforms to `CustomStringConvertible` vs. how `Metadata` does (which doesn't rely on `Metadata.Value`'s conformance to `CustomStringConvertible`).

glbrntt (approved):
LGTM, thanks Gus!

Tests/GRPCCoreTests/MetadataTests.swift
```diff
@@ -314,4 +314,34 @@ struct MetadataTests {
...
     }
   }
+
+  @Suite("Description")
+  struct Description {
+    let metadata = Metadata([
+      "key1": "value1",
+      "key2": "value2",
+      "key-bin": .binary([1, 2, 3]),
+    ])
```
glbrntt, line 324: This is using the `init` from a `Sequence`, so we're creating a dictionary then iterating the sequence of key-value pairs in the dictionary. Dictionary iteration isn't by insertion order meaning that the metadata may be ordered differently (which is why a test failed).

If you do: `let metadata: Metadata = [ .... ]` then you'll go via the `init` for `ExpressibleByDictionaryLiteral` which will use the correct order.

PR #2186, opened 2025-01-28
Fix grpc-swift-protobuf version in hello-world example

gjcairo:
CI is currently failing to build because `grpc-swift-protobuf`'s dependency on the `hello-world` example points to no a non-existing tag.


glbrntt (approved):
🤦‍♂️

PR #2189, opened 2025-01-31
Don't build examples twice

glbrntt:
Motivation:

The CI for examples builds them in the setup stage and the actual run stages. That's just a silly oversight.

Modifications:

- Don't build them during set.

Result:

Examples don't build twice in CI


rnro (approved):
It's easily done done

PR #2199, opened 2025-02-18
Add a migration guide

glbrntt:
Motivation:

Moving from 1.x to 2.x isn't trivial. We should provide a guide for doing this.

Modifications:

- Add a guide for migrating client code and services
- Add a script which can automate parts of this

Result:

Some guidance on migrating.


dev/v1-to-v2/v1_to_v2.sh
```diff
@@ -0,0 +1,125 @@
...
+# Clones v1 into the given directory and applies a number of patches to rename
+# the package from 'grpc-swift' to 'grpc-swift-v1' and 'protoc-gen-grpc-swift'
+# to 'protoc-gen-grpc-swift-v1'.
+function checkout_v1 {
+  # The directory to clone grpc-swift into.
+  grpc_checkout_dir="$(realpath "$1")"
+  # The path of the checkout.
+  grpc_checkout_path="${grpc_checkout_dir}/grpc-swift-v1"
+  # Version of grpc-swift to checkout.
+  grpc_v1_tag="1.24.2"
```
gjcairo, line 31: Would it be worth getting the latest tag by curling GH instead of hardcoding this here?

dev/v1-to-v2/v1_to_v2.sh
```diff
@@ -0,0 +1,125 @@
...
+  find \
+    "${grpc_checkout_path}/Sources" \
+    "${grpc_checkout_path}/Tests" \
+    "${grpc_checkout_path}/Plugins" \
+    -type f \
+    -name '*.swift' \
+    -exec sed -i '' 's/protoc-gen-grpc-swift/protoc-gen-grpc-swift-v1/g' {} +
+
+  # Update the path of the protoc plugin so it aligns with the target name.
+  log "Updating direcotry name for protoc-gen-grpc-swift-v1"
```
gjcairo, line 64:
```suggestion
  log "Updating directory name for protoc-gen-grpc-swift-v1"
```

Sources/GRPCCore/Documentation.docc/Articles/Migration-guide.md
```diff
@@ -0,0 +1,501 @@
...
+
+The intended audience for this guide is users of the `async` variants of clients
+and services from 1.x, not the versions using the older `EventLoopFuture` API.
+
+The guide takes you through a number of steps to migrate your gRPC app
+from 1.x to 2.x. You'll use the following strategy:
+
+1. Setup your package so it depends on a local copy of gRPC Swift 1.x and the
+   upstream version of 2.x.
+2. Generate code for 2.x to alongside generated 1.x code.
```
gjcairo, line 15:
```suggestion
2. Generate code for 2.x alongside generated 1.x code.
```

Sources/GRPCCore/Documentation.docc/Articles/Migration-guide.md
```diff
@@ -0,0 +1,501 @@
...
+3. Incrementally migrate targets to 2.x.
+4. Remove the code generated for, and the dependency on 1.x.
```
gjcairo, line 17:
```suggestion
4. Remove the code generated for 1.x, and the dependency on 1.x.
```

FranzBusch (commented):
Thanks for writing this up. I have a few questions here:

This seems to only outline a migration path for the simplest scenarios. What if an application author has multiple dependencies that use gRPC v1
  - Do I have to wait until all of them vendored v1 so that I can depend on v2?
  - If multiple dependencies vendor gRPC v1 then we should probably advise them rename/alias the modules otherwise we will have conflicts

Sources/GRPCCore/Documentation.docc/Articles/Migration-guide.md
```diff
@@ -0,0 +1,501 @@
...
+Repeat this until you've done all services in your package.
+
+### ServiceProtocol
+
+> If the `SimpleServiceProtocol` worked then you can skip this section.
+
+If you're reading this section then you're likely relying on metadata in your
+service. This means you need to implement the `ServiceProtocol` instead of the
+`SimpleServiceProtocol` and the transformations you need to apply are
+aren't well suited automation. The best approach is to conform your
```
gjcairo, line 287:
```suggestion
aren't well suited for automation. The best approach is to conform your
```

Sources/GRPCCore/Documentation.docc/Articles/Migration-guide.md
```diff
@@ -0,0 +1,501 @@
...
+Buffers package then this would be `Foo_Bar_Baz.ServiceProtocol`).
+
+Let Xcode generate stubs for the methods which haven't been implemented yet and
+fill each one with a `fatalError` so that you app builds. Each method
+should take a `ServerRequest` or `StreamingServerRequest` and context as input
+and return a `ServerResponse` or `StreamingServerResponse`. Request metadata is
+available on the request object. For single responses you can set initial and
+trailing metadata when you create the response. For streaming responses you can
+set initial metadata in the initializer and return trailing metadata from the
+closure you provide to the initializer.
```
gjcairo, line 300: Should we link to the docs or tutorials in case this compressed version isn't enough to understand what's going on?

Sources/GRPCCore/Documentation.docc/Articles/Migration-guide.md
```diff
@@ -0,0 +1,501 @@
...
+```swift
+var client: Foo_Bar_Baz.Client {
+  fatalError("TODO")
+}
+```
+
+If a generated client is passed into the function then duplicate the function
+and replace the body of the new version with a `fatalError()`. You can also mark
+it as deprecated to help you find usages of the function. You'll now have two
+versions of the same function.
```
gjcairo, line 398: Feels like this paragraph and the one above beginning with "If the function is passed a generated client then duplicate it" should be merged.

Sources/GRPCCore/Documentation.docc/Articles/Migration-guide.md
```diff
@@ -0,0 +1,501 @@
...
+
+Now you need to update the function to use the new client. For unary calls the API
+is very similar, so you may not have to change any code. An important change to
+highlight is that for RPCs which stream their responses you must handle the
+response stream _within_ the closure passed to the client. By way of example,
+imagine the following server streaming RPC from 1.x:
```
gjcairo, line 404: Might also be helpful to link to appropriate tutorials/examples/docs from here?

gjcairo (changes requested):

dev/v1-to-v2/v1_to_v2.sh
glbrntt: Yeah, we can just look at the tags though.

Sources/GRPCCore/Documentation.docc/Articles/Migration-guide.md
glbrntt: I'm not sure what would be useful beyond the inline examples?

glbrntt:
> Thanks for writing this up. I have a few questions here:
> 
> This seems to only outline a migration path for the simplest scenarios. What if an application author has multiple dependencies that use gRPC v1
> 
> * Do I have to wait until all of them vendored v1 so that I can depend on v2?
> * If multiple dependencies vendor gRPC v1 then we should probably advise them rename/alias the modules otherwise we will have conflicts

The guide doesn't recommend that you keep the copy of v1. Indeed step 4 of the overview is to remove the copy of v1; It's only there to minimise the number of compilation errors during migration (this is also mentioned in the overview).

Sources/GRPCCore/Documentation.docc/Articles/Migration-guide.md
```diff
@@ -384,19 +386,14 @@ Change the body of the function using the 1.x client to just `fatalError()`.
 Later you'll remove this function altogether.
 
 If the generated client is a stored type then add a new computed property
-returning an instance of it, the body can just call `fatalError()` for now:
+returning an instance of it. the body can just call `fatalError()` for now:
```
gjcairo, line 389:
```suggestion
returning an instance of it. The body can just call `fatalError()` for now:
```

Sources/GRPCCore/Documentation.docc/Articles/Migration-guide.md
gjcairo: Okay fair, I was thinking it would be nice to have a link to the rest of the docs but this will likely be read in the context of the whole documentation anyways so finding the other docs should be easy.

gjcairo (approved):
LGTM, just a nit

PR #2201, opened 2025-02-26
Add code gen config for the module name

glbrntt:
Modifications:

- Add config to the code generator allowing a different module name to be specified.
- Remove unused code

Result:

The grpc module name can be specified in generated code


glbrntt:
The API breakage checker is a false positive, this is a `package` level API:

```
1 breaking change detected in GRPCCodeGen:
  💔 API breakage: func IDLToStructuredSwiftTranslator.makeImports(dependencies:accessLevel:accessLevelOnImports:) has been renamed to func makeImports(dependencies:accessLevel:accessLevelOnImports:grpcCoreModuleName:)
```

glbrntt:
The examples are failing on nightly-main and nightly-next because of https://github.com/grpc/grpc-swift-nio-transport/pull/69 (which is yet to be released).

gjcairo:
Why do we need this exactly? Just for testing? If so, can we not make the new `grpcCoreModuleName` property `package` instead of `public`?

Sources/GRPCCodeGen/Internal/Namer.swift
```diff
@@ -0,0 +1,118 @@
...
+    } else {
+      return baseType
+    }
+  }
+
+  func literalNamespacedType(_ type: String) -> String {
+    return self.grpcCore + "." + type
+  }
+
+  func serverRequest(forType type: String?, streaming: Bool) -> ExistingTypeDescription {
```
gjcairo, line 50: Nit: I think having `isStreaming` in all these funcs instead of `streaming` makes it more consistent with `requestResponse` (and it's also more idiomatic).
```suggestion
  func serverRequest(forType type: String?, isStreaming: Bool) -> ExistingTypeDescription {
```

gjcairo (approved):
LGTM, just one nit

Sources/GRPCCodeGen/Internal/Namer.swift
glbrntt: Done in https://github.com/grpc/grpc-swift/pull/2201/commits/8fa87bf271d8b7d12e8ebca9143db6e5fe2d7803

PR #2203, opened 2025-03-03
Rename nightly_6_1 params to nightly_next

rnro:
Rename nightly_6_1 params to nightly_next


glbrntt (changes requested):
```
** ERROR: Broken symlink: IntegrationTests/Benchmarks/Thresholds/nightly-next/nightly-next
```

glbrntt (approved):

PR #2223, opened 2025-04-24
Expose APIs to customise NIOTS bootstraps

gjcairo:
`swift-nio-transport-services` added support for customising the `NWParameters` used in the underlying connections in https://github.com/apple/swift-nio-transport-services/pull/230.

This PR adds API in `grpc-swift` v1 to allow users to customise the bootstraps used when creating gRPC connections using NIOTS. This will allow them to, for example, customise their `NWParameters` using the new APIs in `swift-nio-transport-services`.

### DIFF
--- a/Sources/GRPC/ClientConnection.swift
+++ b/Sources/GRPC/ClientConnection.swift
@@ -469,6 +469,24 @@ extension ClientConnection {
     @preconcurrency
     public var debugChannelInitializer: (@Sendable (Channel) -> EventLoopFuture<Void>)?
 
+    #if canImport(Network)
+    @available(macOS 10.14, iOS 12.0, watchOS 6.0, tvOS 12.0, *)
+    public var clientBootstrapNWParametersConfigurator: (
+      @Sendable (NIOTSConnectionBootstrap) -> Void
+    )? {
+      get {
+        return self._clientBootstrapNWParametersConfigurator as! (
+          @Sendable (NIOTSConnectionBootstrap) -> Void
+        )?
+      }
+      set {
+        self._clientBootstrapNWParametersConfigurator = newValue
+      }
+    }
+
+    private var _clientBootstrapNWParametersConfigurator: (any Sendable)?
+    #endif
+
     #if canImport(NIOSSL)
     /// Create a `Configuration` with some pre-defined defaults. Prefer using
     /// `ClientConnection.secure(group:)` to build a connection secured with TLS or
--- a/Sources/GRPC/ConnectionPool/GRPCChannelPool.swift
+++ b/Sources/GRPC/ConnectionPool/GRPCChannelPool.swift
@@ -17,6 +17,10 @@ import Logging
 import NIOCore
 import NIOPosix
 
+#if canImport(Network)
+import NIOTransportServices
+#endif
+
 import struct Foundation.UUID
 
 public enum GRPCChannelPool {
@@ -191,6 +195,8 @@ extension GRPCChannelPool {
         return SwiftLogNoOpLogHandler()
       }
     )
+
+    public var transportServices: TransportServices = .defaults
   }
 }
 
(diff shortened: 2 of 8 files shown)


Sources/GRPC/ClientConnection.swift
```diff
@@ -469,6 +469,24 @@ extension ClientConnection {
     @preconcurrency
     public var debugChannelInitializer: (@Sendable (Channel) -> EventLoopFuture<Void>)?
 
+    #if canImport(Network)
+    @available(macOS 10.14, iOS 12.0, watchOS 6.0, tvOS 12.0, *)
+    public var clientBootstrapNWParametersConfigurator: (
```
glbrntt, line 474: We can drop the `clientBootstrap` prefix here, we hide the concept of a bootstrap. Can you add some docs as well please?

Sources/GRPC/ClientConnection.swift
```diff
@@ -469,6 +469,24 @@ extension ClientConnection {
...
+      @Sendable (NIOTSConnectionBootstrap) -> Void
```
glbrntt, line 475: We should be passing in `NWParameters` here (and elsewhere)

Sources/GRPC/Server.swift
```diff
@@ -384,6 +390,24 @@ extension Server {
     /// the need to recalculate this dictionary each time we receive an rpc.
     internal var serviceProvidersByName: [Substring: CallHandlerProvider]
 
+    #if canImport(Network)
+    @available(macOS 10.14, iOS 12.0, watchOS 6.0, tvOS 12.0, *)
+    public var serverBootstrapNWParametersConfigurator: (
+      @Sendable (NIOTSListenerBootstrap) -> Void
+    )? {
```
glbrntt, line 397: Similar feedback here, we can drop `serverBootstrap` and the closure should take a `NWParameters`

Tests/GRPCTests/ConnectionManagerTests.swift
```diff
@@ -1412,6 +1418,54 @@ extension ConnectionManagerTests {
       XCTAssert(error is DoomedChannelError)
     }
   }
+
+  #if canImport(Network)
+  func testDefaultChannelProvider_NWParametersConfigurator() throws {
+    let counter = NIOLockedValueBox(0)
+    let group = NIOTSEventLoopGroup(loopCount: 1)
```
glbrntt, line 1425: The shutdown should be done in a `defer` block just after this or just use the singleton.

Tests/GRPCTests/ConnectionPool/GRPCChannelPoolTests.swift
```diff
@@ -618,5 +623,24 @@ final class GRPCChannelPoolTests: GRPCTestCase {
 
     XCTAssertGreaterThan(statsEvents.count, 0)
   }
+
+  #if canImport(Network)
+  func testNWParametersConfigurator() {
+    let counter = NIOLockedValueBox(0)
+    self.group = NIOTSEventLoopGroup()
```
glbrntt, line 630: We're somewhat skirting around how these test helpers were designed which could lead to some surprises. 

I think the more appropriate solution is to add a flag to `configureEventLoopGroup` which allows you to toggle whether an MTELG is used vs. a NIOTSELG, that way the other helpers should work as expected (e.g. `setUpClientAndServer`)

Tests/GRPCTests/ServerTests.swift
```diff
@@ -0,0 +1,44 @@
...
+    var serverConfiguration = Server.Configuration.default(
+      target: .hostAndPort("localhost", 0),
+      eventLoopGroup: serverEventLoopGroup,
+      serviceProviders: []
+    )
+    serverConfiguration.serverBootstrapNWParametersConfigurator = { _ in
+      counter.withLockedValue { $0 += 1 }
+    }
+
+    _ = try Server.start(configuration: serverConfiguration).wait()
```
glbrntt, line 40: you need to shut down the server when you're done with it

glbrntt (changes requested):

Sources/GRPC/ClientConnection.swift
gjcairo: Ah yeah this was me changing my mind a couple of times about whether we should allow the bootstrap to be configured or only the parameters - that's why the name of the property is a bit confused too.
I take from your previous comment that we should hide the concept of a bootstrap, so I'll stick to just customising the NWParameters then

Sources/GRPC/ConnectionPool/GRPCChannelPool.swift
```diff
@@ -191,6 +195,10 @@ extension GRPCChannelPool {
         return SwiftLogNoOpLogHandler()
       }
     )
+
+    #if canImport(Network)
+    public var transportServices: TransportServices = .defaults
```
glbrntt, line 200: Probably worth adding a comment saying that this will be ignored unless an appropriate event loop group is used.

Sources/GRPC/Server.swift
```diff
@@ -384,6 +390,21 @@ extension Server {
     /// the need to recalculate this dictionary each time we receive an rpc.
     internal var serviceProvidersByName: [Substring: CallHandlerProvider]
 
+    #if canImport(Network)
+    /// A closure allowing to customise the `NWParameters` used when establising a connection using NIOTransportServices.
+    @available(macOS 10.14, iOS 12.0, watchOS 6.0, tvOS 12.0, *)
+    public var nwParametersConfigurator: (@Sendable (NWParameters) -> Void)? {
```
glbrntt, line 396: Is this for the listener channel or for child channels?

glbrntt (commented):
Couple of small questions otherwise this looks good

Sources/GRPC/Server.swift
gjcairo: Good point - right now it's being used for the listener channel. Should we also expose one for the child channels?

glbrntt (approved):
Thanks Gus!

PR #2225, opened 2025-04-25
Set tolerance to zero when using `Task.sleep`

gjcairo:
`Task.sleep` will by default try and coalesce multiple timers into one, mostly for client-specific reasons such as performance, power consumption, etc. 

However, this is undesirable on servers, as it can increase latency, memory usage, and (in the case of gRPC) may result in timeouts not firing when they should.

We can avoid this by setting the sleep `tolerance` to zero.


glbrntt (approved):
LGTM but the formatter is unhappy

PR #2228, opened 2025-04-30
Add a new field, transportSpecific to ServerContext

jtdavey:
Motivation:

Currently there's no way to plumb through details from the transport level to a request handler. Adding this field allows transports, such as the nio transport, to add the peer certificate to the server context when using mTLS.

From there there it's easy for an interceptor to take this data and propogate it forward to a request handler.

Modifications:

This PR adds a single field to the ServerContext that transports can use

Result:

A new field will be accessible to transports and consumers of the API

### DIFF
--- a/Sources/GRPCCore/Call/Server/ServerContext.swift
+++ b/Sources/GRPCCore/Call/Server/ServerContext.swift
@@ -45,6 +45,15 @@ public struct ServerContext: Sendable {
   /// - "in-process:27182".
   public var localPeer: String
 
+  /// An optional field for transports to store specific data
+  ///
+  /// Refer to the transport documentation to understand what type of
+  /// value this field will contain, if any.
+  ///
+  /// An example of what this field can be used for, would be to store
+  /// things like a peerCertificate from a mTLS connection
+  public var transportSpecific: (any Sendable)?
+
   /// A handle for checking the cancellation status of an RPC.
   public var cancellation: RPCCancellationHandle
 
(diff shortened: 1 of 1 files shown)


Sources/GRPCCore/Call/Server/ServerContext.swift
```diff
@@ -45,6 +45,15 @@ public struct ServerContext: Sendable {
...
   public var localPeer: String
 
+  /// An optional field for transports to store specific data
+  ///
+  /// Refer to the transport documentation to understand what type of
+  /// value this field will contain, if any.
+  ///
+  /// An example of what this field can be used for, would be to store
+  /// things like a peerCertificate from a mTLS connection
+  public var transportSpecific: (any Sendable)?
```
glbrntt, line 55: I thought about this some more, rather than `any Sendable` can you define a `public protocol TransportSpecific: Sendable {}` nested within `ServerContext` and use that instead?

It has a few advantages:
1. Compiler helps users out a little more
2. It's easier to document and easier for users to find implementations.

Sources/GRPCCore/Call/Server/ServerContext.swift
```diff
@@ -45,6 +45,15 @@ public struct ServerContext: Sendable {
   /// - "in-process:27182".
   public var localPeer: String
 
+  /// An optional field for transports to store specific data
+  ///
+  /// Refer to the transport documentation to understand what type of
+  /// value this field will contain, if any.
+  ///
+  /// An example of what this field can be used for, would be to store
+  /// things like a peerCertificate from a mTLS connection
```
glbrntt, line 54:
```suggestion
  /// things like a peer certificate from a mTLS connection
```

glbrntt (changes requested):

glbrntt (approved):
Awesome, thanks @jtdavey!

PR #2234, opened 2025-05-08
Add explicit availability annotations

glbrntt:
Motivation:

gRPC is moving from specifying the platforms in the package manifest to annotating code with availability annotations.

Modifications:

- Explicitly annotate public API
- Add '-require-explicit-availability' to CI
- Drop platforms

Result:

- gRPC can be used by packages which don't specify their platforms in the manifest

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+import CompilerPluginSupport
 import PackageDescription
 
 let products: [Product] = [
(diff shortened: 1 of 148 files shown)


Package.swift
```diff
@@ -45,12 +46,28 @@ let dependencies: [Package.Dependency] = [
...
+// This adds some build settings which allow us to map "@available(gRPCSwift 2.x, *)" to
+// the appropriate OS platforms.
+let nextMinorVersion = 2
+let availabilitySettings: [SwiftSetting] = (0 ... nextMinorVersion).map { minor in
+  let name = "gRPCSwift"
+  let version = "2.\(minor)"
+  let platforms = "macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0"
+  let setting = "AvailabilityMacro=\(name) \(version):\(platforms)"
+  return .enableExperimentalFeature(setting)
+}
```
gjcairo, line 60: Are we handling each minor separately (instead of just having a `gRPCSwift 2` macro) in case we eventually add new APIs that have higher platform requirements?

With this logic, we tie each whole minor to a set of platforms. What if we ever decide to add some new API that requires higher platform requirements, alongside other new API which could keep the previous platform requirements? Would we just bump all of the new APIs to use the new platforms?

gjcairo (approved):

PR #2235, opened 2025-05-08
Allow clients to set a max age for connections

glbrntt:
Motivation:

Some load balancers reject new streams when a client certificate has expired but allow existing ones to continue. This is problematic because there's no signal that the connection can't be used anymore so new RPCs will continue to fail.

Modifications:

Allow clients to configure a max age for connections, after which time the connection will shutdown gracefully. This allows new connections to be established.

Result:

Client connections can be configured to age out


Tests/GRPCTests/MaxAgeTests.swift
```diff
@@ -0,0 +1,241 @@
...
+    func connectionRemoved(id: GRPCConnectionID) {}
+    func startedConnecting(id: GRPCConnectionID) {}
+    func connectFailed(id: GRPCConnectionID, error: any Error) {}
+    func connectionUtilizationChanged(id: GRPCConnectionID, streamsUsed: Int, streamCapacity: Int) {
+    }
+  }
+
+  func testRPCContinuesAfterQuiescing() throws {
+    // Check that an in-flight RPC can continue to run after the connection is quiescing as a result
+    // or aging out.
```
gjcairo, line 192:
```suggestion
    // of aging out.
```

gjcairo (commented):
LGTM, tiny typo.

PR #2238, opened 2025-05-09
chore: Bump deps to compatible minimums

rasberik:
Good to have updates on Package dependencies (for reference when built outside SPM)

Both recent bumps of `swift-nio-transport-services` and `swift-nio-http2` require  lastest`swift-nio`:
* `swift-nio-transport-services` 1.24.0  -> `swift-nio` `2.81.0`
* `swift-nio-http2` 1.36.0 -> `swift-nio` `2.82.0`

And latest `swift-nio` require bump of `swift-collections`:
* `swift-nio` `2.82.0` -> `swift-collections` `1.1.0`

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -32,7 +32,7 @@ let includeNIOSSL = ProcessInfo.processInfo.environment["GRPC_NO_NIO_SSL"] == ni
 let packageDependencies: [Package.Dependency] = [
   .package(
     url: "https://github.com/apple/swift-nio.git",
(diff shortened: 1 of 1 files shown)


glbrntt:
Thanks for opening this but the change isn't necessary: the build system will satisfy the requirements from all dependencies, so while the _direct_ requirement on swift-nio is 2.65.0+, the _transitive_ requirement is 2.81.0+.

PR #2261, opened 2025-10-06
Update generated code

glbrntt:
Motivation:

Newer protobuf versions changed how they generate code, so CI now fails.

Modifications:

Update generated code, bump minimum protobuf version

Result:

Protobuf is more up-to-date


gjcairo (approved):

rnro (approved):
🚀

PR #2262, opened 2025-10-06
Drop 5.10 and 6.0

glbrntt:
Motivation:

In line with our support policy (see README.md), now that Swift 6.2 has been released we can drop 5.10 and 6.0.

Modifications:

- Update CI
- Update Package.swift to use `static` computed `var`s as it no longer builds with 6.x compilers as the various SwiftPM types aren't sendable

Result:

Less support


Lukasa (approved):
We

PR #2269, opened 2025-12-09
Remove `PathLike` and use `URL` instead

mimischi:
Removes the `PathLike` definition from `GRPCSwiftPlugin` and uses `URL` instead, as this project has a minimum supported version of Swift 6.1 now.

### DIFF
--- a/Plugins/GRPCSwiftPlugin/plugin.swift
+++ b/Plugins/GRPCSwiftPlugin/plugin.swift
@@ -168,13 +168,13 @@ struct GRPCSwiftPlugin {
   ) -> Command {
     // Construct the `protoc` arguments.
     var protocArgs = [
-      "--plugin=protoc-gen-grpc-swift=\(protocGenGRPCSwiftPath)",
-      "--grpc-swift_out=\(outputDirectory)",
+      "--plugin=protoc-gen-grpc-swift=\(protocGenGRPCSwiftPath.path())",
+      "--grpc-swift_out=\(outputDirectory.path())",
     ]
 
     importPaths.forEach { path in
       protocArgs.append("-I")
-      protocArgs.append("\(path)")
+      protocArgs.append(path.path())
     }
 
     if let visibility = invocation.visibility {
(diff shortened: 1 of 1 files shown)


glbrntt (approved):

Plugins/GRPCSwiftPlugin/plugin.swift
```diff
@@ -99,42 +99,42 @@ struct GRPCSwiftPlugin {
...
-      protocPath = PathLike(environmentPath)
+      protocPath = URL(fileURLWithPath: environmentPath)
     } else {
       // The user didn't set anything so let's try see if SPM can find a binary for us
-      protocPath = try PathLike(tool("protoc"))
+      protocPath = try tool("protoc").url
     }
 
-    let protocGenGRPCSwiftPath = try PathLike(tool("protoc-gen-grpc-swift"))
+    let protocGenGRPCSwiftPath = try tool("protoc-gen-grpc-swift").url
```
glbrntt, line 137: This isn't quite right it seems, at least when the value is plumbed through and turned into an argument:

```
file: warning: directory does not exist.
file:///tmp/tmp.V377jRaSyP/.build/x86_64-unknown-linux-gnu/debug/protoc-gen-grpc-swift-tool: program not found or is not executable
Please specify a program using absolute path or make sure the program is available in your PATH system variable
```

I don't remember the exact API but you can get the string representing of the `URL` without `file://` which should fix this.

glbrntt (changes requested):
Something is amiss here around interpolating the `URL` as a string

Plugins/GRPCSwiftPlugin/plugin.swift
mimischi: I've addressed this in [e5f7944](https://github.com/grpc/grpc-swift/pull/2269/commits/e5f7944f0cc217278fb5b81717fb185a9a4f3cfe), but ended up doing this in a different place

glbrntt (approved):

PR #2271, opened 2026-02-11
Add LocalizedError conformance to GRPCStatus

DePasqualeOrg:
`GRPCStatus` conforms to `Error` and `CustomStringConvertible`, but not `LocalizedError`. This means `localizedDescription` returns a generic message like `"The operation couldn't be completed. (GRPC.GRPCStatus error 1.)"` instead of the status description. This is a problem for applications that display errors in a UI using `errorDescription`, which is the standard pattern in Swift.

This adds `LocalizedError` conformance so that `errorDescription` returns the same string as `description`, e.g. `"unavailable (14): server not reachable"`.

### DIFF
--- a/Sources/GRPC/GRPCStatus.swift
+++ b/Sources/GRPC/GRPCStatus.swift
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import Foundation
 import NIOCore
 import NIOHTTP1
 import NIOHTTP2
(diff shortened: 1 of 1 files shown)


glbrntt:
These changes don't localize the error.

I would recommend you add a shim in your application which checks whether the error is a `LocalizedError` and returns the `localizedDescription` if it is and falls back to the `description` if it isn't.

DePasqualeOrg:
@glbrntt, `LocalizedError` is a bit of a misnomer in this case. It's about having a standardized way to show a user-friendly error message, so that no custom shim is needed. What is the harm of pulling in Foundation, which should be available on all platforms? Your proposed solution requires developers to write custom code just for this error type.

Lukasa:
I don't want to speak for @glbrntt, but this is the crux: these errors aren't intended to be user-friendly errors. They're intended to be developer friendly: they exist to be logged, not printed.

DePasqualeOrg:
They are still likely to be shown in the UI as they propagate up. And currently an opaque error description will be shown, which also isn't helpful to the developer.

Issue #2085, opened 2024-10-07
When upgrading from 1.23.0 to 1.23.1 "protoc-gen-grpc-swift=/${BUILD_DIR}/${CONFIGURATION}" becomes "protoc-gen-grpc-swift=/$%7BBUILD_DIR%7D/$%7BCONFIGURATION%7D" and breaks the build


Lutzifer:
### Describe the bug

When upgrading from 1.23.0 to 1.23.1 "protoc-gen-grpc-swift=/${BUILD_DIR}/${CONFIGURATION}" becomes "protoc-gen-grpc-swift=/$%7BBUILD_DIR%7D/$%7BCONFIGURATION%7D" and breaks the build.

With 1.23.0, the shell script that is generated during the build is

```
#!/bin/sh
/usr/bin/sandbox-exec -p "(version 1)
(deny default)
(import \"system.sb\")
(allow file-read*)
(allow process*)
(allow mach-lookup (global-name \"com.apple.lsd.mapdb\"))
(allow file-write*
    (subpath \"/private/tmp\")
    (subpath \"/private/var/folders/fl/14_g861101sg0qfb9qrtrhwr0000gn/T\")
)
(deny file-write*
    (subpath \"/Users/admin/dev/Synchronizing/clinic-app-mobile/iOS\")
)
(allow file-write*
    (subpath \"/Users/admin/Library/Developer/Xcode/DerivedData/ClinicApp-aisphxldtrzqiraqscjpzjjuettp/SourcePackages/plugins/ClinicApp.output/API/GRPCSwiftPlugin\")
    (subpath \"/private/var/folders/fl/14_g861101sg0qfb9qrtrhwr0000gn/T/TemporaryItems\")
    (subpath \"/private/var/folders/fl/14_g861101sg0qfb9qrtrhwr0000gn/T/TemporaryItems/NSIRD_Xcode_lhUx2k\")
)
[...]

PR #2273, opened 2026-02-23
fix escaping (regression of https://github.com/grpc/grpc-swift/issues/2085)

Lutzifer:
this fixes the regression I posted about in https://github.com/grpc/grpc-swift/issues/2085

### DIFF
--- a/Plugins/GRPCSwiftPlugin/plugin.swift
+++ b/Plugins/GRPCSwiftPlugin/plugin.swift
@@ -255,7 +255,7 @@ struct GRPCSwiftPlugin {
   }
 }
 
-private extension URL {
+fileprivate extension URL {
   var fileSystemPath: String {
     #if canImport(Darwin)
     return self.path(percentEncoded: false)
--- a/Plugins/GRPCSwiftPlugin/plugin.swift
+++ b/Plugins/GRPCSwiftPlugin/plugin.swift
@@ -255,8 +255,8 @@ struct GRPCSwiftPlugin {
   }
 }
 
-fileprivate extension URL {
-  var fileSystemPath: String {
+extension URL {
+  fileprivate var fileSystemPath: String {
     #if canImport(Darwin)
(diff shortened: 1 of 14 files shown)


Plugins/GRPCSwiftPlugin/plugin.swift
```diff
@@ -168,8 +168,8 @@ struct GRPCSwiftPlugin {
   ) -> Command {
     // Construct the `protoc` arguments.
     var protocArgs = [
-      "--plugin=protoc-gen-grpc-swift=\(protocGenGRPCSwiftPath.path())",
       "--grpc-swift_out=\(outputDirectory.path())",
+      "--plugin=protoc-gen-grpc-swift=\(protocGenGRPCSwiftPath.fileSystemPath)",
```
glbrntt, line 171: It looks like there are a couple of places in this file where we use `path()`, could you update those as well?

glbrntt (changes requested):
Thanks for fixing this! We should apply this to all call site though.

Plugins/GRPCSwiftPlugin/plugin.swift
Lutzifer: I think I got em all, now.

Plugins/GRPCSwiftPlugin/plugin.swift
```diff
@@ -255,6 +255,16 @@ struct GRPCSwiftPlugin {
   }
 }
 
+private extension URL {
```
emuguy1: Formatter fails with wanting fileprivate instead of private
```suggestion
fileprivate extension URL {
```

glbrntt (approved):

glbrntt (changes requested):
agh, sorry, looks like you need to run `./Protos/generate.sh`

glbrntt (approved):

PR #2278, opened 2026-07-28
Fix broken documentation

samuelmurray:
~When looking at the documentation at SPI for this package, I noticed that the latest published `1.X` release is `1.24.2`, https://swiftpackageindex.com/grpc/grpc-swift/1.24.2/documentation/grpc. I'm not sure if this is because stops tracking previous major versions, once a newer major has been released (notice that `1.24.2` was the last `1.X` release before `2.0.0`), or if there's some issue with the newer `1.X` releases? For example, `GRPCCore` is listed in `spi.yml`, but that target doesn't exist on `1.X`. See https://github.com/grpc/grpc-swift/issues/2279.~

The documentation for this package contains a few broken links, both internal and external.

I updated all broken links, except for https://github.com/grpc/grpc-swift/blob/release/1.x/docs/faqs.md?plain=1#L157 which points to a file that doesn't exist.
I also removed non-existing target `GRPCCore`  from `.spi.yml`, and added a link to SPI in the README.
Last, I removed some unused markdown links.


glbrntt (approved):
Thanks!

PR #2280, opened 2026-07-29
Stop pinning documentation generation to Swift 6.0

finestructure:


glbrntt:
Thanks @finestructure -- you'll need to sign the CLA before we can accept this change.

finestructure:
That might be a bit tricky at the moment - feel free to simply redo that change via your account!

PR #2281, opened 2026-07-29
Drop swift version from spi.yml

glbrntt:
Motivation:

This package requires Swift 6.1 to build but the spi yml chooses Swift version 6.0 so fails to build. This means that SPI can't build the docs.

Modifications:

- drop swift version from spi.yml

Result:

docs build


glbrntt:
Code gen fixed in https://github.com/grpc/grpc-swift/pull/2282

rafaelcepeda (approved):

